PR #18, opened 2015-04-09
run_standalone_monitor not working

imron:
The MonitorsManager.build_monitor function changed its interface (3 parameters instead of 2), but run_standalone_monitor was not updated to reflect this, and this was preventing monitors from being run in a standalone fashion using

```
python -m scalyr_agent.run_monitor scalyr_agent.builtin_monitors.<my_monitor>
```

This PR calls ```build_monitor``` correctly, and adds a unit test calling run_standalone_monitor, which fails using the previous code, and passes with this fix.

### DIFF
--- a/scalyr_agent/run_monitor.py
+++ b/scalyr_agent/run_monitor.py
@@ -94,9 +94,8 @@ def run_standalone_monitor(monitor_module, monitor_python_path, monitor_config, 
         signal.signal(sig, handle_shutdown_signal)
 
     try:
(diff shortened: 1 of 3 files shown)


czerwingithub:
Thanks for this fix.

imron:
No problem.  I needed it to make the develop/test cycle for the syslog plugin simpler, so figured might as well share.

czerwingithub <notifications@github.com> �� 2015/4/9 16:58 �g�D�G

PR #19, opened 2015-04-10
2.4 compatibility

imron:
Thread.is_alive should be Thread.isAlive in order to maintain compatibility with python 2.4


czerwingithub:
No need for this PR.  Sorry, I had this change yesterday in a client and hadn't pushed it.  I've already pushed my changes into master.  Thanks for finding this as well though.

imron:
No worries.  I needed them for 2.4 testing on the syslog monitor so thought I��d upload them as the change hadn��t been pushed to master when I checked.  I��ll update my branch with your latest changes.

PR #46, opened 2016-10-04
Upgrade docker base image to xenial

### LINKED ISSUE #43: Update Scalyr Docker container
Right now the scalyr docker container uses `ubuntu:14.04.2`
It would be advisable to update this to `ubuntu:14.04` that way it uses the latest security patches, etc, from 14.04.4 and beyond.

It also might be worth updating it to `ubuntu:16.04` and also creating an `alpine:3.4` version (which would be significantly lighter in weight, as the base image is only 4mb), as most major docker containers now come in an alpine flavor.

zkanda:
Fixes #43

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04.2
+FROM ubuntu:16.04
 ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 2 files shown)


czerwingithub:
Hello.  Thanks for taking care of this for us.  I'll include this change in our upcoming release as long as it passes the usual testing.

PR #53, opened 2017-08-18
Support HTTP request methods other than GET for URL Monitors

code-sauce:
Still Work in Progress, the aim of this Pull Request is to support the URL Monitors with:
* Request Methods other than the default GET [optional]
* Request HTTP Headers [optional]
* Request body [optional]

The intention of this change is to support the customer requests to support POST etc.

Part of this PR change was to support response body that is in UTF8 str (bytes) and not `unicode` so we don't try to re-decode into UTF-8 that fails. I discovered this when we started accepting HTTP Headers and I set the `Accept-Encoding: gzip` so the text was not `unicode` but `str` and it broke the `__to_escaped_string` method.

Still TODO:

- [x] Unit Tests
~- [ ] Test Locally with DELETE, PUT with and without a request body. Maybe this could be an integration test where we could spin up a local server on `0.0.0.0` and support some requests? Need to discuss this with @czerwingithub~ Will punt on integration tests for now.
- [ ] Documentation on the config that needs to drive this change. Need to know how to go about documentation. Another thing to learn.

The configuration in the `agent.json` that will be needed (and also be needed to add to the documentation) is:

```
  "monitors": [
    {
      "module": "scalyr_agent.builtin_monitors.url_monitor",
      "id": "google-url-monitor",
      "url": "http://www.google.com",
      "request_method": "GET",
      "request_headers": [{"header": "Accept-Encoding", "value": "gzip"}],
      "request_body": "foobar"
    }
  ]
```

I also changed the log emitter to include the HTTP request method and the log looks like:

```
2017-08-18 19:47:12.806Z [url_monitor(google-url-monitor)] response "<!doctype html><html itemscope=\"\" itemtype=\"http://schema.org/WebPage\" lang=\"en\"><head><meta content=\"Search the world's information, including webpages, images, videos and more. Google has many speci..." url="http://www.google.com" status=200 length=10482 request_method="GET"
```

or with gzip enabled in header:

```
2017-08-18 19:28:56.885Z [url_monitor(google-url-monitor)] response "\u001f\u008b\b\u0000\u0000\u0000\u0000\u0000\u0002��:kw۶\u0092��+h�T\u0096�\u0094DR�\u0083��l\u009aG�i�n�\u0093�MS\u001f\u0090\u0084H�|\u0099�d�����;\u0003\u0090\u0014%;�=�ioo4�`0/\f\u0006\u0003�\u0017'A�W\u009b\u0082)Q\u0095&\u0097\u0017��\u0015K�\u009f\u0017�%Dt\u0090�%QU\u0015�x��\u0088�t\u0094\u0097��\u0003�~�!#JB��%,#�\u0081���\"e\u0015U�<�XV��\u001d��\u001f)U�\u0094��L\u00823���2/SZ�y�B�OVA\u009c\u0085�\u001d�" url="http://www.google.com" status=200 length=4528 request_method="GET"
```


scalyr_agent/builtin_monitors/url_monitor.py
```diff
@@ -103,11 +107,62 @@ def _initialize(self):
         else:
             self.extractor = None
 
-    def gather_sample(self):
+    @staticmethod
```
czerwingithub: Why have this as a static method?  You wouldn't have to pass in things like `logger` or `url`, etc if it was a normal method.  I take it is for testing?  How about we just require the tests to create an instead of `UrlMonitor`?

scalyr_agent/builtin_monitors/url_monitor.py
```diff
@@ -103,11 +107,62 @@ def _initialize(self):
...
+        @type request_method: str
+        @type request_data: str | None
+        @type headers: JsonArray (of tuples with header name, header value combination)
+        @type logger: logging.Logger or subtypes
+        @return: Request object
+        """
+
+        request = urllib2.Request(url, data=request_data)
+        if headers:
+            if type(headers) != JsonArray:
```
czerwingithub: Let's move this check for `type(headers) != JsonArray` to the constructor.  If there is a configuration error, we try to throw as soon as we can (like in the constructor) because the rest of the system behaves better if it does.  (The system is designed to expect configuration errors to be thrown during monitor construction).

scalyr_agent/builtin_monitors/url_monitor.py
```diff
@@ -103,11 +107,62 @@ def _initialize(self):
...
+            else:
+                for header in headers:
+                    request.add_header(header["header"], header["value"])
+
+        # seems awkward to override the GET method, but internally it flips
+        # between GET and POST anyway based on the existence of request body
+        request.get_method = lambda: request_method
+        return request
+
+    def gather_sample(self, request_method='GET', request_data=None, headers=None):
```
czerwingithub: `gather_sample` is defined in the base class, so overriding it and changing what arguments it takes seems like it's not a good idea.  The `request_method`, `request_data`, `headers` are instance variables, why not just get them from `self`?

scalyr_agent/tests/url_monitor_test.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def setUp(self):
+        self.legit_headers = JsonArray()
+        self.legit_headers.add(JsonObject({'header': 'header_foo', 'value': 'foo'}))
+        self.legit_headers.add(JsonObject({'header': 'header_bar', 'value': 'bar'}))
+
+    def tearDown(self):
+        pass
+
+    def testGetRequestNoHeaders(self):
```
czerwingithub: style question.  We are using snake case for all of our method names (for the most part).  Do you feel that tests should be treated different because the of the influence of junit (`setUp`, etc).

dev-requirements.txt
```diff
@@ -0,0 +1,4 @@
+mock==2.0.0
```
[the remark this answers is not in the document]
code-sauce: root is generally preferred, but it's not mandatory. As far as comment is concerned, it's a great idea

scalyr_agent/builtin_monitors/url_monitor.py
code-sauce: yes, you're right - it was mainly for more testability. However, we can instantiate an object in the test, no big deal.

scalyr_agent/tests/url_monitor_test.py
code-sauce: my preference is to keep tests n the same case as the rest of the code (snake_case)- yeah the `setUp`, `tearDown` throws it off but we can live with it?

code-sauce:
@czerwingithub i _think_ I have covered all the comments in the later commits.

dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
+# This is only meant for Scalyr's internal development and dev environment. These dependencies are not needed to run the tool
```
czerwingithub: Can you specify what tool this is for?  I.e., this is formated for tools like `blah`.

scalyr_agent/builtin_monitors/url_monitor.py
```diff
@@ -87,12 +96,17 @@ def _initialize(self):
...
-        self.request_data = self._config.get("request_data") or None
-        self.request_headers = self._config.get("request_headers") or []
+        self.request_method = self._config.get("request_method")
+        self.request_data = self._config.get("request_data")
+        self.request_headers = self._config.get("request_headers")
         self.timeout = self._config.get("timeout")
         self.max_characters = self._config.get("max_characters")
         self.log_all_lines = self._config.get("log_all_lines")
+        if self.request_headers and type(self.request_headers) != JsonArray:
+            self._logger.emit_value(
```
czerwingithub: Don't log here since it is unclear if the log is even open.  Instead, `raise Exception("your reason for an error")`.  The code that is above this will log the exception in the proper place.  And we do want to kill the monitor if it looks like the configuration is not good.

code-sauce:
thank you! yes I was waiting for the final approval.

PR #56, opened 2017-08-24
[Ready For Review] Add Hashing option for the redacted data

code-sauce:
Still Work in Progress.

Adds the support to hash the redacted data. This introduces 1 more configurations in the `"redaction_rules"` dictionary
e.g.:
```
"redaction_rules": [
  			{
  				"match_expression": "password=[^& ]*",
  				"replacement": "thistextwillbeusedifnothashed",
                                "hash_salt": "fdfhgkhfgjkhfgf"
  			}
```
Currently we support only MD5 hashing algorithm. All the redacted and matched groups will be replaced by (its + salt, if any) MD5 digest. 

Currently, the hashing feature is only available for regular expressions with captured groups.
We now need to know the groups that need to be hashed. Not all groups that are captured in the regular expression may be hashed. For this case, which i presume will be the general case,
we need a little terminology.

eg:

```
redaction_rules: [
       // Delete all instances of password=...
       { match_expression: "password=[^& ]*" },

       // Replace terms like "userInfo=username password" with "userInfo=username"
       {
         match_expression: "userInfo=([^ ]+) [^ ]+",
         replacement: "userInfo=\\H1"
       }
     ]
   }
```

we will need to specify the group that needs to be hashed by using this in the replacement string prefixed by `\\H` followed by the group number. This will help us capture and parse the group before replacement so we can then later hash the actual content of the matched group. One more example is: 
```
redaction_rules: [
       {
         match_expression: "secret(.*?)=([a-z]+\s?)",
         replacement: "secret\\1=\\H2"
       }
     ]
   }
```

This will hash only group number 2 and keep group 1 the same.

I currently ran the following configuration on the local dev and found it working:

```
"logs": [
  	{
  		"path": "/var/log/baz.log",
  		"attributes": {"parser": "appLog"},
  		"redaction_rules": [
  			{
  				"match_expression": "password(.*?)=([a-z]+\\s?)",
  				"replacement": "password\\1=\\H2"
  			}
  		] 	
  	}
  ```

Where the `/var/log/baz.log` looked like: 
```
somegarbage text passwordblahtext=secretstuff some more text and one more passwordfdfdfd=moresecretstuff fdfd
somegarbage text passwordblahtext=secretstuff some more text and one more passwordfdfdfd=moresecretstuff fdfd
somegarbage text passwordblahtext=secretstuff some more text and one more passwordfdfdfd=moresecretstuff fdfd
somegarbage text passwordblahtext=secretstuff some more text and one more passwordfdfdfd=moresecretstuff fdfd
```

and i could see the hashed group of password in the terminal as:
```
somegarbage text passwordblahtext=�,���<�u/�UqL�some more text and one more passwordfdfdfd=�' ԯ6� �h�N  
17:33:03.513 Saurabhs-MBP.lan /var/log/baz.log  somegarbage text passwordblahtext=�,���<�u/�UqL�some more text and one more passwordfdfdfd=�' ԯ6� �h�N  
```

TODO:
- [x] Unit Tests. I'm sure i broke the existing ones. I need to fix those and add a few others.

### DIFF
--- a/scalyr_agent/log_processing.py
+++ b/scalyr_agent/log_processing.py
@@ -1611,16 +1611,21 @@ class LogFileProcessor(object):
         self.__num_redaction_and_sampling_rules += 1
         self.__sampler.add_rule(match_expression, sampling_rate)
 
-    def add_redacter(self, match_expression, replacement):
+    def add_redacter(self, match_expression, replacement, hash_salt=''):
         """Adds a new redaction rule that will be applied after all previously added redaction rules.
 
         @param match_expression: The regular expression that must match the portion of the log line that will be
             redacted.
         @param replacement: The text to replace the matched expression with. You may use \1, \2, etc to use sub
             expressions from the match regular expression.
+        @param hash_salt: The salt used for hashing, only if hashing group replacement is used
+
+        @type hash_salt: str
         """
         self.__num_redaction_and_sampling_rules += 1
-        self.__redacter.add_redaction_rule(match_expression, replacement)
+        self.__redacter.add_redaction_rule(
+            match_expression, replacement, hash_salt
+        )
 
     def __create_events_object(self, event_message, sampling_rate):
         """Returns the events object that can be sent to the server for this log to insert the specified message.
@@ -1817,6 +1822,10 @@ class LogLineRedacter(object):
     Redaction rules can match each line multiple times.
     """
 
+    # Indicator in the replacement text[optional] to hash the group content.
+    # eg replacement string of foo\\1=\\H2 will replace the second group with its hashed content
+    HASH_GROUP_INDICATOR = "H"
+
     def __init__(self, log_file_path):
         """Initializes an instance for a single file.
 
@@ -1848,14 +1857,22 @@ class LogLineRedacter(object):
 
         return input_line, modified_it
 
-    def add_redaction_rule(self, redaction_expression, replacement_text):
+    def add_redaction_rule(self, redaction_expression, replacement_text, hash_salt=''):
         """Appends a new redaction rule to this instance.
 
         @param redaction_expression: The regular expression that must match some portion of the line.
         @param replacement_text: The text to replace the matched text with. May include \1 etc to use a portion of the
             matched text.
+        @param hash_salt: [optional] If hashing is set, then the cryptographic salt to be used
+
+        @type hash_salt: str
         """
-        self.__redaction_rules.append(RedactionRule(redaction_expression, replacement_text))
+
+        self.__redaction_rules.append(
+            RedactionRule(
+                redaction_expression, replacement_text, hash_salt=hash_salt
+            )
+        )
 
     def __apply_redaction_rule(self, line, redaction_rule):
         """Applies the specified redaction rule on line and returns the result.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: scalyr_agent/log_processing.py [hunk 1]
```diff
@@ -1866,16 +1880,59 @@ def __apply_redaction_rule(self, line, redaction_rule):
...
             indicating if a redaction was applied.
         """
+
+        def __hash_group(match):
+            print '&&&&&\n', match.groups()
+            if not match.groups():
+                print "NO GROUPS!!!!"
+                return scalyr_util.md5_digest(match.group() + redaction_rule.hash_salt)
+
+            # TODO NEED help here! for a replacement regular expression, how can we determine which
```
code-sauce: @czerwingithub Need help here. Since i'm passing a callback into the `re.subn` i need to know which captured group should I be hashing and replacing the value for. Is there a good way to determine from the already existing configuration of `replacement` like `foo=\\1` to determine that is the group number 1 that needs to be captured and hashed? Let me know if the approach doesn't make sense.

code-sauce:
think i found another way. closing this out for now.

scalyr_agent/log_processing.py
```diff
@@ -1817,6 +1824,10 @@ class LogLineRedacter(object):
     Redaction rules can match each line multiple times.
     """
 
+    # Indicator in the replacement text[optional] to hash the group content.
+    # eg replacement string of foo\\1=$HG2 will replace the second group with its hashed content
+    HASH_GROUP_INDICATOR = "$HG"
```
czerwingithub: To be consistent with the normal replacement syntax, how about we call this `\\H`?

So, you would say something like `foo\\1=\\H2`?

scalyr_agent/log_processing.py
```diff
@@ -1848,14 +1859,26 @@ def process_line(self, input_line):
 
         return input_line, modified_it
 
-    def add_redaction_rule(self, redaction_expression, replacement_text):
+    def add_redaction_rule(self, redaction_expression, replacement_text,
+                           hash_replace=False, hash_salt=''):
```
czerwingithub: Since you are now indicating, in the replacement string, whether or not to use a hash of the value rather than the value itself... I don't believe you need `hash_replace` anymore.  Can we remove that?

scalyr_agent/log_processing.py
```diff
@@ -1866,16 +1889,98 @@ def __apply_redaction_rule(self, line, redaction_rule):
         @return: A sequence of two elements, the line with the redaction applied (if any) and True or False
             indicating if a redaction was applied.
         """
+
+        def __replace_groups_with_hashed_content(re_ex, replacement_ex, line):
```
czerwingithub: I think you can simplify this code a fair bit.  Here's the approach I would take:

Suppose you have regex string `password=(.*)&` and replacement `password=\\H1`.

First, only attempt to apply the hash substitution if there is a match for `\\H\d` in the replacement string.

If that is true, then do the following:
* Perform the search, so that you get values for all of the groups.
* for 1 to the number of groups, do the following:
* replace `\\HN` in the replacement string with the hash of group(N)
* end for loop
* Then perform the replacement on the original string, using the new replacement string that has the actual digests built into it.

I think that works as long as we aren't using "global" replacement, which I don't think we are?  You could check that.  If we are, then we just have to iterate overall of the matches in the string and do them one at a time like how I specified above.

What do you think?

scalyr_agent/log_processing.py
code-sauce: well, we still need the `hash_replace` if the replace pattern doesn't have a regular expression or the regular expression doesn't have any captured groups. However i'm not sure if we will ever need to hash a static value. Thoughts?

scalyr_agent/log_processing.py
code-sauce: it does make sense in retrospect. I'll try it out. thank you for your feedback.

code-sauce: @czerwingithub  i tried using this approach and it's elegant for single matches, but I couldn't figure out how to make it cleaner than what it already is for multiple matches in a line. If there are multiple matches 
eg: 

`replacement_string`: `secret\1=\H2`
`regular expression`: `secret(.*?)=([a-z]+\s?)`
`log line`: `sometext.... secretoption=czerwin moretextsecretbar=xxx andsecret123=saurabh`

in this case, we will have to iterate over all the matches i.e. `secretoption=czerwin ` `secretbar=xxx ` and `secret123=saurabh` and replace the replacement string inside the loop and make sure the final replaced string needs to be partially replaced for each matched group. This calls for bookeeping on - 
1) making sure the replacement string is replenished after each match occurence
2) the match count is incremented and accounted for
3) the main log line should only be partially matched and we should account for the location until where the string was replaced as part of the first match. This is because replacing the first match with its hash will convert
```sometext.... secretoption=czerwin moretextsecretbar=xxx andsecret123=saurabh``` -> ```sometext.... secretoption=<hash of czerwin> moretextsecretbar=xxx andsecret123=saurabh``` and doing it again will make the first match again susceptible to being a candidate for replacement i.e.
```sometext.... secretoption=<hash of <hash of czerwin>> moretextsecretbar=xxx andsecret123=saurabh```
unless we track the last replaced line index.

This caused my code to be more unclear than what it was before. Let me know what you think.

**Comment by code-sauce** [comment]
@czerwingithub - i have renamed the indicator to be more consistent like ```secret\\2=\\H1``` etc. Other questions for you above though.

File: scalyr_agent/log_processing.py (hunk 3 above)
Comment (czerwingithub) [reply scalyr_agent/log_processing.py]: I think we should only support hashing captured groups.  It is more in line with the feature requests we see.  Also, if they wish to hash the entire matched expression, they could just surround it with parentheses.

File: scalyr_agent/log_processing.py (hunk 4 above)
Comment (czerwingithub) [reply scalyr_agent/log_processing.py]: I think you can essentially add an outer loop to applying my suggested replacement trick one replacement at a time.

Use `re.finditer` to walk over each match in a given line.  That returns a `MatchObject` for each match.  You can use its `start` and `end` methods to determine the indices into the original line where the match appears.

You can use this and I think it works out to be fairly clean.

    myindex = 0
    output_result = ""
    for x in the_regex.finditer(the_line):
        output_result = output_result + the_line[my_index:x.start(0)]
        my_index = x.end(0)
        output_result += perform_replacement_for_single(x.group(0), replacement_string)

Does that make sense?  I'm building up the overall string manually but always first adding the portion of the line that did not match replacement pattern (since our last match).  Then, I also add in the text that did match the replacement pattern with the replacements performed by just performing the replacement on x.group(0) [which is the substring that matched the pattern].  Then I iterate from there.

**Comment by code-sauce** [comment]
@czerwingithub - i took your advice on the algorithm - and i think the function is much cleaner now. I fixed the test and even ran the agent locally and saw the right result. Updated the PR description with the result.

File: scalyr_agent/log_processing.py [hunk 5]
```diff
@@ -1848,14 +1857,22 @@ def process_line(self, input_line):
...
-    def add_redaction_rule(self, redaction_expression, replacement_text):
+    def add_redaction_rule(self, redaction_expression, replacement_text, hash_salt=''):
         """Appends a new redaction rule to this instance.
 
         @param redaction_expression: The regular expression that must match some portion of the line.
         @param replacement_text: The text to replace the matched text with. May include \1 etc to use a portion of the
             matched text.
+        @param hash_salt: If hashing is set, then the cryptographic salt to be used
+
+        @type hash_salt: str (optional)
```
czerwingithub: Don't use the (optional) string here.  This will (I believe) get in the way of PyCharm's ability to use `str` as the expected type of the parameter.

scalyr_agent/log_processing.py
```diff
@@ -1866,22 +1883,85 @@ def __apply_redaction_rule(self, line, redaction_rule):
...
+
+            _matches = re.finditer(re_ex, line)
+
+            if not _matches:
+                # if no matches, return the `line` as such
+                return line, None
+
+            replacement_matches = 0
+
+            last_match_index = 0
```
czerwingithub: Add a little documentation what `last_match_index` is used for.

scalyr_agent/log_processing.py
```diff
@@ -1866,22 +1883,85 @@ def __apply_redaction_rule(self, line, redaction_rule):
...
+            replaced_string = ""
+
+            for _match in _matches:
+                _groups = _match.groups()
+                replaced_group = replacement_ex
```
czerwingithub: Add documentation for what `replaced_group` is used for.

code-sauce:
thanks @czerwingithub - i updated the doc string and added more comments. I also squashed my commits for cleaner history. Merging.

PR #59, opened 2017-09-11
Request third party

code-sauce:


setup.py
```diff
@@ -121,15 +123,43 @@ def __init__(self, **kw):
...
+
+        if _should_use_system_requests:
+            # rename the third_party/requests directory to third_party/requests_deprecated
+            project_path = os.path.dirname(os.path.abspath(__file__))
+            shutil.move(
+                os.path.join(project_path, 'scalyr_agent', 'third_party', 'requests'),
+                os.path.join(project_path, 'scalyr_agent', 'third_party', 'requests_deprecated'),
+            )
+    except Exception as ex:
+        pass
```
code-sauce: not sure if we should print or log something?

PR #60, opened 2017-09-13
allow users to easily configure `scalyr_api_key` via env variables

code-sauce:
This is especially useful for automated server spinups using a base template like cloudformation, ECS template etc. where the API Key can be baked in those templates as AWS templates have an easy interface to set up environment variables using JSON. The agent.json file doesn't need to be custom built

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -124,9 +124,9 @@ class Configuration(object):
                 self.__add_elements_from_array('monitors', content, self.__config)
                 self.__merge_server_attributes(fp, content, self.__config)
 
-            # Check for api_key in the environment variable `api_key`. Sometimes, injecting a secret key like
+            # Check for api_key in the environment variable `scalyr_api_key`. Sometimes, injecting a secret key like
             # this is preferable via environment variables
-            api_key = os.environ.get('api_key') if os.environ.get('api_key') else api_key
+            api_key = os.environ.get('scalyr_api_key') if os.environ.get('scalyr_api_key') else api_key
 
             self.__set_api_key(self.__config, api_key)
             if scalyr_server is not None:
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -124,6 +124,10 @@ class Configuration(object):
                 self.__add_elements_from_array('monitors', content, self.__config)
                 self.__merge_server_attributes(fp, content, self.__config)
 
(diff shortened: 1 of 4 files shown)


scalyr_agent/configuration.py
```diff
@@ -124,6 +124,10 @@ def parse(self):
                 self.__add_elements_from_array('monitors', content, self.__config)
                 self.__merge_server_attributes(fp, content, self.__config)
 
+            # Check for api_key in the environment variable `scalyr_api_key`. Sometimes, injecting a secret key like
+            # this is preferable via environment variables
+            api_key = os.environ.get('scalyr_api_key') if os.environ.get('scalyr_api_key') else api_key
```
czerwingithub: Hmm... I'd rather only take the key from the environment if none is specified in the configuration file.  So, I think here you have to check to see if the api_key is none.

code-sauce: ok, makes sense. This avoids the case of a hierarchy, and there can only be one place where the key can be configured. Would you like to notify the user if both the config file and the environment variable have it set?

czerwingithub: Yes, that is a good idea.  If both are set, we use the configuration file (because that's preserves the current behavior), but emit a log warning to the agent.log that the environment key is being ignored.

code-sauce:
@czerwingithub updated with changes

scalyr_agent/configuration.py
```diff
@@ -29,6 +29,14 @@
 from scalyr_agent.json_lib import JsonConversionException, JsonMissingFieldException
 from scalyr_agent.util import JsonReadFileException
 
+import scalyr_agent.scalyr_logging as scalyr_logging
+
+# Set up the main logger.  We set it up initially to log to standard out,
+# but once we run fork off the daemon, we will use a rotating log file.
+log = scalyr_logging.getLogger('scalyr_agent')
+
+scalyr_logging.set_log_destination(use_stdout=True)
```
czerwingithub: You know, there's some dependency issues with using the scalyr logger from within this module.  Let's not import it here. Instead, have a reference to a logger passed into the method you need it below.. I think it is `parse`.  Default the argument to None (and don't log to it if is None).  However, then change the calls to `parse` in the agent main and other places to pass in the logger reference.

code-sauce:
@czerwingithub passed in the logger in the function and removed the logging dependency

PR #61, opened 2017-09-19
Run platform tests on differnt Python/Linux distributions

code-sauce:
Running the Scalyr Agent tests on different platforms. This structure can be expanded to cover all major Python/OS distributions. 

The base Docker images for Python are listed at https://store.docker.com/images/python

This should cover a lot of ground.

TODO:
- [x] in the `tearDown` method, remove the created image to reduce the memory footprint.

### DIFF
--- a/scalyr_agent/platform_tests/alpine/Dockerfile
+++ b/scalyr_agent/platform_tests/alpine/Dockerfile
@@ -1,0 +1,17 @@
+FROM python:2-alpine
+
+MAINTAINER Scalyr Inc <support@scalyr.com>
+
+RUN apk update
+RUN apk add "git"
+
+
+WORKDIR /tmp
+
+RUN git init && \
+  git config --local user.name "Scalyr" && git config --local user.email support@scalyr.com && \
+  git clone -b release git://github.com/scalyr/scalyr-agent-2.git
+
+RUN pip install mock==2.0.0
+RUN chmod -R +x scalyr-agent-2
+WORKDIR scalyr-agent-2
--- a/scalyr_agent/platform_tests/run_platform_test.py
+++ b/scalyr_agent/platform_tests/run_platform_test.py
@@ -1,0 +1,49 @@
+# Copyright 2015 Scalyr Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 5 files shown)


scalyr_agent/platform_tests/run_platform_test.py
```diff
@@ -0,0 +1,75 @@
...
+# author: Saurabh Jain <saurabh@scalyr.com>
+
+__author__ = 'saurabh@scalyr.com'
+
+import unittest
+from subprocess import call
+import os
+
+
+class cd:
```
czerwingithub: `cd` is a confusing name to use because it is a common Unix command.  I would suggest you name it `WorkingDirectory`.

scalyr_agent/platform_tests/run_platform_test.py
```diff
@@ -0,0 +1,75 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------
+#
+# author: Saurabh Jain <saurabh@scalyr.com>
+
+__author__ = 'saurabh@scalyr.com'
```
czerwingithub: Please add some documentation as to what this script does.

scalyr_agent/platform_tests/alpine/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:2-alpine
```
czerwingithub: Can you create a `README.md` file that describes the purpose of this directory (does platform level tests using docker to verify .. blah blah).

code-sauce:
@czerwingithub done. added more documentation.

code-sauce:
Added some instructions on how to set-up vagrant and Windows to run agent tests on Windows. Making it automated is hard because there is no CLI installer via ssh, so this could be a little manual for now, but easy enough to follow in the normal workflow.

czerwingithub:
In the `testAgentInstall.py` script, I actually do install and run the Scalyr Agent on Windows using ssh.  Check that out to see if that helps you with the automation.  It is not quite fullproof because the installation using the `msi` from the commandline does not block the commandline until it is finished (the install happens asychronously).  However, I found it works well enough.

code-sauce:
@czerwingithub thanks, that is helpful. I'll incorporate the Windows ssh changes from `testAgentInstall` in a separate PR.

PR #64, opened 2017-09-29
Process metrics

code-sauce:
This PR attempts to add a functionality for allowing:
1) optionally allow multiple processes to be aggregated (with or without) inclusion of child processes
2) optionally allow a single process to include child processes.

There are 2 new configuration options that are added for the [Linux Process Metrics](https://www.scalyr.com/help/monitors/linux-process-metrics):

1) `aggregate_multiple_processes` (optional/default False) that allows multiple processes (if they match) to be aggregated.
2) `include_child_processes` (optional/default False) that includes the child processes for both single/multi process situations.

### DIFF
--- a/scalyr_agent/builtin_monitors/linux_process_metrics.py
+++ b/scalyr_agent/builtin_monitors/linux_process_metrics.py
@@ -122,7 +122,7 @@ class Metric(namedtuple('Metric', ['name', 'type'])):
 
         A non-cumulative metric is one where the absolute observed value is reported every time.
 
-        :return:
+        @return: True if the metric is cumulative, False otherwise
         """
         return self.name in ('app.cpu', 'app.uptime', 'app.disk.bytes', 'app.disk.requests')
 
@@ -183,14 +183,14 @@ class BaseReader:
         Runs a single cycle of the sample collection.
 
         It should read the monitored file and extract all metrics.
-        :param collector: Optional - a dictionary to collect the metric values.
+        @param collector: Optional - a dictionary to collect the metric values.
                           The collector has key as Metric, and value as the actual observed scalar.
                           eg: {
                                 ("app.net.bytes", "in"): 12222,
                                 ("app.net.bytes", "out"): 20,
                               }
                           Note: a new collector will be instantiated if None is passed in.
-        :return: None or the optional collector with collected metric values
+        @return: None or the optional collector with collected metric values
         """
 
         self._timestamp = int(time.time())
@@ -608,7 +608,7 @@ class SockStatReader(BaseReader):
 # Reads stats from /proc/$pid/fd.
 class FileDescriptorReader:
     """Reads and records statistics from the /proc/$pid/fd directory.  Essentially it just counts the number of entries.
-
+sub_proc = Popen(cmd, shell=False, stdout=PIPE)
     The recorded metrics are listed below.  They all also have an app=[id] field as well.
       app.io.fds type=open:         the number of open file descriptors
     """
@@ -723,6 +723,79 @@ class ProcessTracker(object):
         return collector
 
 
+class ProcessList(object):
+    """
+    Class that encapsulates the listing of process(es) based on IDs, regular expressions etc.
+    """
+
+    def __init__(self):
+        # list of {'pid': processid, 'ppid': parentprocessid,  'cmd': command }
+        self.processes = []
+        # key -> parent process id, value -> [child process ids...]
+        self.parent_to_children_map = defaultdict(list)
+
+        cmd = ['ps', 'axo', 'pid,ppid,command']
+        sub_proc = Popen(cmd, shell=False, stdout=PIPE)
+        regex = r"\s+(\d+)\s+(\d+)\s+(.*)"
+        lines = sub_proc.stdout.readlines()
+        for line in lines:
+            match = re.search(regex, line)
+            if match:
+                _pid, _ppid, _cmd = match.groups()
+                self.processes.append({'pid': int(_pid), 'ppid': int(_ppid), 'cmd': _cmd})
+
+        self._init_parent_to_child_map()
+
+    def _init_parent_to_child_map(self):
+        for _process in self.processes:
+            self.parent_to_children_map[_process['ppid']].append(_process['pid'])
+
+    def get_matches_commandline(self, match_pattern):
+        """
+        Given a string, match the processes on the name
+        @param match_pattern: process command match pattern
+        @return: List of Process IDs
+        """
+
+        matches = []
+        for _process in self.processes:
+            if re.search(match_pattern, _process['cmd']):
+                matches.append(_process['pid'])
+        return matches
+
+    def get_child_processes(self, ppid):
+        """
+        Given a process pid, get the children pids
+        @param ppid: parent process id
+        @return: list of children process IDs
+        """
+
+        all_children = []
+        children_to_explore = []
+        for _pid in self.parent_to_children_map[ppid]:
+            all_children.append(_pid)
+            children_to_explore.append(_pid)
+
+        # get the children 'recursively'
+        while children_to_explore:  # the invariant
+            child_to_explore = children_to_explore.pop()
+            all_children.extend(self.parent_to_children_map[child_to_explore])
+        return list(set(all_children))
+
+    def get_matches_commandline_with_children(self, match_pattern):
+        """
+        Like get_matches_commandline method, given a string, match the processes on the name
+        but also returns the matched processes' children
+        @param match_pattern: process command match pattern
+        @return: List of Process IDs
+        """
+
+        matched_pids = self.get_matches_commandline(match_pattern)
+        for matched_pid in matched_pids:
+            matched_pids.extend(self.get_child_processes(matched_pid))
+        return list(set(matched_pids))
+
+
 class ProcessMonitor(ScalyrMonitor):
     """A Scalyr agent monitor that records metrics about a running process.
 
@@ -748,11 +821,6 @@ class ProcessMonitor(ScalyrMonitor):
     You can run multiple instances of this monitor per agent to monitor different processes.
     """
 
-    # poll interval (in seconds) for matching processes when we are monitoring multiple processes
-    # Needs to be initialized as a class variable, not an instance variable so that it is
-    # alive for all epochs of the monitor.
-    LAST_POLLED = {}
-
     def _initialize(self):
         """Performs monitor-specific initialization."""
         # The ids of the processes being monitored, if one has been matched.
@@ -766,9 +834,14 @@ class ProcessMonitor(ScalyrMonitor):
         self.__include_child_processes = self._config.get(
             'include_child_processes', default=False, convert_to=bool
         )
-
-        self.__poll_interval = self._config.get('poll_interval', default=120, convert_to=int)
+        if not getattr(self, 'last_discovered', False):
+            # poll interval (in seconds) for matching processes when we are monitoring multiple processes
+            # alive for all epochs of the monitor.
+            self.__last_discovered = {}
+        self.__process_discovery_interval = self._config.get('process_discovery_interval', default=120, convert_to=int)
         self.__datetime_format = "%Y-%m-%d %H:%M:%S"
+
+        # this is the key to the dictionary self.__last_discovered which denotes the key to the monitor process
         self.__last_polled_env_key = '%s_linux_process_monitor_last_polled' % self.__id
 
         __target_pids = self._config.get('pid', default=None, convert_to=str)
@@ -783,21 +856,21 @@ class ProcessMonitor(ScalyrMonitor):
                     else:
                         self.__target_pids.append(int(os.getpid()))
 
-        # history of all metrics which has form:
+        # Last 2 values of all metrics which has form:
         # {
         #   '<process id>: {
-        #                 <metric name>: [<metric at time 0>, <metric at time 1>.... ],
-        #                 <metric name>: [<metric at time 0>, <metric at time 1>.... ],
+        #                 <metric name>: [<metric at time 0>, <metric at time 1>],
+        #                 <metric name>: [<metric at time 0>, <metric at time 1>],
         #                 }
 
         #   '<process id>: {
-        #                 <metric name>: [<metric at time 0>, <metric at time 1>.... ],
-        #                 <metric name>: [<metric at time 0>, <metric at time 1>.... ],
+        #                 <metric name>: [<metric at time 0>, <metric at time 1>],
+        #                 <metric name>: [<metric at time 0>, <metric at time 1>],
         #                 }
         # }
-
         self.__metrics_history = defaultdict(dict)
-        # running total of metric values
+
+        # running total of metric values, both for cumulative and non-cumulative metrics
         self.__aggregated_metrics = {}
 
         if not (self.__commandline_matcher or self.__target_pids):
@@ -821,7 +894,7 @@ class ProcessMonitor(ScalyrMonitor):
         @param metrics: Collected metrics of the process
         @type pid: int
         @type metrics: dict
-        :return: None
+        @return: None
         """
 
         for _metric, _metric_value in metrics.items():
@@ -868,8 +941,6 @@ class ProcessMonitor(ScalyrMonitor):
                         if len(_metric_values) > 1:
                             # only report the cumulative metrics for more than one sample
                             self.__aggregated_metrics[_metric] += (_metric_values[-1] - _metric_values[-2])
-                        else:
-                            self.__aggregated_metrics[_metric] = 0
                 else:
                     if pid in running_pids_set:
                         # absolute metric - accumulate the last reported value
@@ -890,102 +961,28 @@ class ProcessMonitor(ScalyrMonitor):
         if not running_pids:
             self.__aggregated_metrics = {}
 
-    def get_pids_from_ps(self, subprocesses=False, ppid=None):
-        """
-        Get the list of pids running, or get the list of child processes given a parent process id
-        :param subprocesses: Get subprocesses?
-        :param ppid: parent process id, only if subprocesses is True
-        :type subprocesses: bool
-        :type ppid: int
-        :return: list of process ids
-        """
-
-        cmd = ['ps', 'ax', '-o', 'pid,command']
-        if subprocesses:
-            cmd = ["ps", "--ppid", str(ppid), "-o", "pid,command"]
-        sub_proc = Popen(cmd, shell=False, stdout=PIPE)
-        lines = sub_proc.stdout.readlines()
-        all_pids = []
-        try:
-            for line in lines:
-                line = line.strip()
-                if line.find(' ') > 0:
-                    pid = line[:line.find(' ')]
-                    if pid.lower() == 'pid':
-                        continue
-                    pid = int(pid)
-                    line = line[(line.find(' ') + 1):]
-                    if not subprocesses:
-                        if re.search(self.__commandline_matcher, line) is not None:
-                            all_pids.append(pid)
-                    else:
-                        all_pids.append(pid)
-        finally:
-            if sub_proc is not None:
-                sub_proc.wait()
-        return all_pids
-
-    def get_active_matched_pids(self, check_poll=True):
+    def get_active_matched_pids(self):
         """
         Get the active pids that match the commandline match expression.
         If there is no need to poll, then return the currently monitored pids
-        @:param check_poll: Should check for new matched processes?
-        @:param include_child_processes: Should include child processes of matching processes?
-        @:type check_poll: bool
-        @:type include_child_processes: bool
-        :return: List of running process IDs
+        @return: List of running process IDs
         """
 
-        if check_poll:
-            if not self.should_poll_for_processes():
-                return self.__pids
-
-        matching_pids = self.get_pids_from_ps()
-        if self.__include_child_processes:
-            for matching_pid in matching_pids:
-                matching_pids.extend(
-                    self.get_pids_from_ps(subprocesses=True, ppid=matching_pid)
-                )
-        matching_pids = list(set(matching_pids))
-        return matching_pids
-
-    def should_poll_for_processes(self):
-        """
-        We should poll for new processes in the following case:
-
-        * We do not currently have a process that we matched.
-        * This is a single process case, and the old process died.
-        * This is a multi-process case, and at least one process is alive and a certain time has expired.
-        :return: True if there is a need to poll, False otherwise
-        """
-
-        if not self.__pids:
-            # no processes are currently being monitored, poll again
-            return True
-
-        if not self.__aggregate_multiple_processes:
-            # single process case
-            active_matched_pids = self.get_active_matched_pids(check_poll=False)
-            if not active_matched_pids:
-                # no matched processes, try again after sometime
-                return False
-            if active_matched_pids[0] not in self.__pids:
-                # matched process, but not running, old process maybe died
-                return True
-        else:
+        if self.__aggregate_multiple_processes and self.__last_discovered.get(self.__last_polled_env_key):
             # multi-process case, poll every `poll_interval` seconds to
             # see if there are more processes to match
-            if (
-                    not self.LAST_POLLED.get(self.__last_polled_env_key) or
-                    (
-                        datetime.strptime(self.LAST_POLLED.get(self.__last_polled_env_key), self.__datetime_format) +
-                        timedelta(seconds=self.__poll_interval)
-                    ) < datetime.now()
-            ):
-                self.LAST_POLLED[self.__last_polled_env_key] = datetime.now().strftime(self.__datetime_format)
-                return True
+            _last_time_checked =  datetime.strptime(
+                self.__last_discovered.get(self.__last_polled_env_key), self.__datetime_format
+            )
+            if _last_time_checked + timedelta(seconds=self.__process_discovery_interval) > datetime.now():
+                return self.__pids
 
-        return False
+        # get the processes that matched
+        ps = ProcessList()
+        self.__last_discovered[self.__last_polled_env_key] = datetime.now().strftime(self.__datetime_format)
+        if self.__include_child_processes:
+            return ps.get_matches_commandline_with_children(self.__commandline_matcher)
+        return ps.get_matches_commandline(self.__commandline_matcher)
 
     def gather_sample(self):
         """Collect the per-process tracker for the monitored process(es).
@@ -1065,11 +1062,6 @@ class ProcessMonitor(ScalyrMonitor):
                 selected_pids = [os.getpid()]
             else:
                 selected_pids = self.__target_pids
-
-        # include children processes, if applicable
-        if self.__include_child_processes:
-            for pid in selected_pids:
-                selected_pids.extend(self.get_pids_from_ps(subprocesses=True, ppid=pid))
         return set(selected_pids)
 
 
--- a/scalyr_agent/builtin_monitors/linux_process_metrics.py
+++ b/scalyr_agent/builtin_monitors/linux_process_metrics.py
@@ -856,6 +856,10 @@ class ProcessMonitor(ScalyrMonitor):
             # for all the absolute metrics, decrease the count that the dead processes accounted for
             del self.metrics_history[_pid_to_remove]
 
+        # if no processes are running, there is no reason to report the running metric data
+        if not running_pids:
+            self.running_total_metrics = {}
+
     def gather_sample(self):
         """Collect the per-process tracker for the monitored process(es).
 
--- a/scalyr_agent/builtin_monitors/linux_process_metrics.py
+++ b/scalyr_agent/builtin_monitors/linux_process_metrics.py
@@ -200,14 +200,13 @@ class BaseReader:
                 # don't have permissions for it, then just don't collect this
                 # stat from now on.  If the user changes the configuration file
                 # we will try again to read the file then.
+                self._failed = True
                 if e.errno == 13:
                     self._logger.error("The agent does not have permission to read %s.  "
                                        "Maybe you should run it as root.", filename)
-                    self._failed = True
                 elif e.errno == 2:
                     self._logger.error("The agent cannot read %s.  Your system may not support that proc file type",
                                        filename)
-                    self._failed = True
                 else:
                     raise e
 
@@ -218,7 +217,7 @@ class BaseReader:
             except IOError, e:
                 self._logger.error( "Error gathering sample for file: '%s'\n\t%s" % (filename, str( e ) ) );
 
-                #close the file.  This will cause the file to be reopened next call to run_single_cycle
+                # close the file. This will cause the file to be reopened next call to run_single_cycle
                 self.close()
 
     def gather_sample(self, my_file):
@@ -268,6 +267,7 @@ class StatReader(BaseReader):
       app.threads:           the number of threads being used by the process
       app.nice:              the nice value for the process
     """
+
     def __init__(self, pid, monitor_id, logger):
         """Initializes the reader.
 
@@ -279,6 +279,7 @@ class StatReader(BaseReader):
         @type monitor_id: str
         @type logger: scalyr_agent.AgentLogger
         """
+
         BaseReader.__init__(self, pid, monitor_id, logger, "/proc/%ld/stat")
         # Need the number of jiffies_per_sec for this server to calculate some times.
         self._jiffies_per_sec = os.sysconf(os.sysconf_names['SC_CLK_TCK'])
@@ -295,6 +296,7 @@ class StatReader(BaseReader):
         @return: The number of centiseconds for the specified number of jiffies.
         @rtype: int
         """
+
         return int((jiffies * 100.0) / self._jiffies_per_sec)
 
     def calculate_time_ms(self, jiffies):
@@ -307,6 +309,7 @@ class StatReader(BaseReader):
         @return: The number of milliseconds for the specified number of jiffies.
         @rtype: int
         """
+
         return int((jiffies * 1000.0) / self._jiffies_per_sec)
 
     def __get_uptime_ms(self):
@@ -315,6 +318,7 @@ class StatReader(BaseReader):
         @return: The number of milliseconds the system has been up.
         @rtype: int
         """
+
         if self._boot_time_ms is None:
             # We read /proc/uptime once to get the current boot time.
             uptime_file = None
@@ -346,7 +350,7 @@ class StatReader(BaseReader):
         line = line[(line.find(") ")+2):]
         fields = line.split()
         # Then the fields we want are just at fixed field positions in the
-        # string.  Just grap them.
+        # string.  Just grab them.
         self.print_sample("app.cpu", self.__calculate_time_cs(int(fields[11])), "user")
         self.print_sample("app.cpu",  self.__calculate_time_cs(int(fields[12])), "system")
         # The uptime is calculated by reading the 'start_time from stat which is expressed as the
@@ -355,6 +359,7 @@ class StatReader(BaseReader):
         self.print_sample("app.uptime", process_uptime)
         self.print_sample("app.nice", float(fields[16]))
         self.print_sample("app.threads", int(fields[17]))
+
 
 class StatusReader(BaseReader):
     """Reads and records statistics from the /proc/$pid/status file.
@@ -365,6 +370,7 @@ class StatusReader(BaseReader):
       app.mem.bytes type=peak_vmsize:   the maximum number of bytes used for virtual memory for process
       app.mem.bytes type=peak_resident: the maximum number of bytes of resident memory ever used by process
     """
+
     def __init__(self, pid, monitor_id, logger):
         """Initializes the reader.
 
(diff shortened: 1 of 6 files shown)


code-sauce:
@czerwingithub please take a look at this. Thanks for the help. One doubt remaining is that for dead processes, since we do not subtract their contribution, should we still keep reporting the metrics say if all processes we are monitoring die?

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +732,214 @@ class ProcessMonitor(ScalyrMonitor):
...
+        for _metric, _metric_value in metrics.items():
+            if not self.metrics_history[pid].get(_metric):
+                self.metrics_history[pid][_metric] = []
+            self.metrics_history[pid][_metric].append(_metric_value)
+            # only keep the last 2 running history for any metric
+            self.metrics_history[pid][_metric] = self.metrics_history[pid][_metric][-2:]
+
+    def _reset_absolute_metrics(self):
+        """
+        At the beginning of each process metric calculation, the absolute metrics
```
czerwingithub: Let's make it clear here what you mean by absolute... So, say `.. each process metric calculation, the absolute (non-cumulative) metrics ...`

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +732,214 @@ class ProcessMonitor(ScalyrMonitor):
...
+        running_pids_set = set(running_pids)
+
+        for pid, process_metrics in self.metrics_history.items():
+            for _metric, _metric_values in process_metrics.items():
+                if not self.running_total_metrics.get(_metric):
+                    self.running_total_metrics[_metric] = 0
+                if _metric.is_cumulative:
+                    if pid in running_pids_set:
+                        if len(_metric_values) <= 1:
+                            self.running_total_metrics[_metric] += (_metric_values[-1] if _metric_values else 0)
```
czerwingithub: We have only one metric value.. then this is for a new process.. we should not add anything into `running_total_metrics` for this process.  We should only add in a value (the delta) when we have two samples.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +732,214 @@ class ProcessMonitor(ScalyrMonitor):
...
-        if pid is not None:
-            self.__gathers.append(StatReader(self.__pid, self.__id, self._logger))
-            self.__gathers.append(StatusReader(self.__pid, self.__id, self._logger))
-            self.__gathers.append(IoReader(self.__pid, self.__id, self._logger))
-            self.__gathers.append(FileDescriptorReader(self.__pid, self.__id, self._logger))
-        # TODO: Re-enable these if we can find a way to get them to truly report
-        # per-app statistics.
-        #        self.gathers.append(NetStatReader(self.pid, self.id, self._logger))
-        #        self.gathers.append(SockStatReader(self.pid, self.id, self._logger))
+    def _calculate_running_total(self, running_pids):
```
czerwingithub: I don't think we should call this `running_total`.  In my mind, that's something different.  I would suggest we call it `calculate_aggregated_metrics`... where it is the metrics for all for all processes.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +732,214 @@ class ProcessMonitor(ScalyrMonitor):
...
+                    self.running_total_metrics[_metric] = 0
+                if _metric.is_cumulative:
+                    if pid in running_pids_set:
+                        if len(_metric_values) <= 1:
+                            self.running_total_metrics[_metric] += (_metric_values[-1] if _metric_values else 0)
+                        else:
+                            self.running_total_metrics[_metric] += (_metric_values[-1] - _metric_values[-2])
+                else:
+                    # absolute metric - accumulate the last reported value
+                    self.running_total_metrics[_metric] += _metric_values[-1]
```
czerwingithub: I think there's a bug here if process is dead.  We only check if it is not dead for the cumulative case.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +732,214 @@ class ProcessMonitor(ScalyrMonitor):
...
+        trackers = []
+        for _pid in list(self.__select_processes()):
+            trackers.append(ProcessTracker(_pid, self._logger, self.__id))
+
+        self._reset_absolute_metrics()
 
-        for gather in self.__gathers:
-            gather.run_single_cycle()
+        for _tracker in trackers:
+            _tracker.set_gathers()
```
czerwingithub: Can we just have the gathers set at `ProcessTracker` initialization?  It seems a little odd to have to call `set_gathers` to use it.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +732,214 @@ class ProcessMonitor(ScalyrMonitor):
...
+            _metrics = _tracker.collect()
+            self.record_metrics(_tracker.pid, _metrics)
 
-    def __select_process(self):
-        """Returns the proces id of a running process that fulfills the match criteria.
+        running_pids = [x.pid for x in trackers]
```
czerwingithub: I'm not sure if this syntax is allowed in Python 2.4.  I guess you can double check with your new unit test.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +732,214 @@ class ProcessMonitor(ScalyrMonitor):
...
+        and we should keep polling for it.
         """
-        if self.__pid is not None and not self.__is_running():
-            self.__set_process(None)
 
-        if self.__pid is None:
-            self.__set_process(self.__select_process())
+        trackers = []
+        for _pid in list(self.__select_processes()):
+            trackers.append(ProcessTracker(_pid, self._logger, self.__id))
```
czerwingithub: We shouldn't create a new ProcessTracker every time through the gather sample if we already had one for the process last time through it.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +732,214 @@ class ProcessMonitor(ScalyrMonitor):
...
-        @return: The process id of the matching process, or None
-        @rtype: int or None
+        @return: The set of process id(s) of the matching process, or set()
+        @rtype: set()
         """
+
         sub_proc = None
-        if self.__commandline_matcher is not None:
+
+        if self.__commandline_matcher:
```
czerwingithub: We need to add in a new configuration option for the monitor that is something like `aggregate_multiple_processes`, which should default to `False`.  We do not want existing customers to all of the sudden start having all processes that match their expression aggregated together.  We always have to keep in mind to preserve old behavior.

Also, add back in the check we have (for when we are only matching one process) to emit a warning message when we match multiple processes.  You should also mention that they can turn on the `aggregate_multiple_processes` option if the want in that log message.

czerwingithub: Also, we should only actually only look for new processes under certain conditions:
* We do not currently have a process we matched
* This is a single process instance and the old process died
* We are in the multiple process case and we are already watching at least one process... and a certain time period has expired (configured via another option).  Essentially, maybe only once every 2 minutes will we look to see if there are more processes we should be monitoring.

czerwingithub: Also, we need to add in some option to monitor child processes as well (match all child processes of any process currently matched).

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -192,44 +184,53 @@ def run_single_cycle(self):
 
         filename = self._file_pattern % self._pid
 
+        if not collector:
+            collector = {}
         if self._file is None:
             try:
                 self._file = open(filename, "r")
             except IOError, e:
+                print e
```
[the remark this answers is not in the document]
code-sauce: remnants of debugging. removing.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +732,214 @@ class ProcessMonitor(ScalyrMonitor):
...
         self.__id = self._config.get('id', required_field=True, convert_to=str)
         self.__commandline_matcher = self._config.get('commandline', default=None, convert_to=str)
-        self.__target_pid = self._config.get('pid', default=None, convert_to=str)
+        self.__target_pids = self._config.get('pid', default=None, convert_to=str)
 
-        if self.__commandline_matcher is None and self.__target_pid is None:
-            raise BadMonitorConfiguration('At least one of the following fields must be provide: commandline or pid',
-                                          'commandline')
+        # convert target pid into a list. Target pids can be a single pid or a CSV of pids
+        self.__target_pids = [int(x.strip()) for x in self.__target_pids.split(',')] if self.__target_pids else []
```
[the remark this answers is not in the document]
code-sauce: good point. Ternary operators were introduced in Python 2.5

scalyr_agent/builtin_monitors/linux_process_metrics.py
code-sauce: i think again it was 2.5. i . will change it.

code-sauce:
@czerwingithub - I _think_ I addressed your PR comments, and added support for including the optional children processes. Let me know what you think.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -105,9 +105,32 @@
...
+
+    @property
+    def is_cumulative(self):
+        """
+        A cumulative metric is one that needs a prior value to calculate the next value.
+        i.e. only the deltas for the current observed values are reported.
+
+        A non-cumulative metric is one where the absolute observed value is reported every time.
+
+        :return:
```
czerwingithub: To be consistent, use `@return` style of documentation return values.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -177,11 +178,21 @@ def __init__(self, pid, monitor_id, logger, file_pattern):
...
 
         It should read the monitored file and extract all metrics.
+        :param collector: Optional - a dictionary to collect the metric values.
+                          The collector has key as Metric, and value as the actual observed scalar.
+                          eg: {
+                                ("app.net.bytes", "in"): 12222,
+                                ("app.net.bytes", "out"): 20,
+                              }
+                          Note: a new collector will be instantiated if None is passed in.
+        :return: None or the optional collector with collected metric values
```
czerwingithub: Use `@return` here and elsewhere.

czerwingithub: I don't think it ever return `None`...  I think it either returns the collector that was passed in or the one it created.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
     In additional to the fields listed above, each metric will also have a field 'app' set to the monitor id to specify
     which process the metric belongs to.
 
     You can run multiple instances of this monitor per agent to monitor different processes.
     """
+
+    # poll interval (in seconds) for matching processes when we are monitoring multiple processes
+    # Needs to be initialized as a class variable, not an instance variable so that it is
+    # alive for all epochs of the monitor.
+    LAST_POLLED = {}
```
czerwingithub: I think it should just be an instance variable.  I'm not quite sure why it would need to be a class var.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
-            raise BadMonitorConfiguration('At least one of the following fields must be provide: commandline or pid',
-                                          'commandline')
+        self.__aggregate_multiple_processes = self._config.get(
+            'aggregate_multiple_processes', default=False, convert_to=bool
+        )
+        self.__include_child_processes = self._config.get(
+            'include_child_processes', default=False, convert_to=bool
+        )
+
+        self.__poll_interval = self._config.get('poll_interval', default=120, convert_to=int)
```
czerwingithub: We need a better name for this config option (and also variable name).  How about `process_discovery_interval`.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
+
+        #   '<process id>: {
+        #                 <metric name>: [<metric at time 0>, <metric at time 1>.... ],
+        #                 <metric name>: [<metric at time 0>, <metric at time 1>.... ],
+        #                 }
+        # }
+
+        self.__metrics_history = defaultdict(dict)
+        # running total of metric values
+        self.__aggregated_metrics = {}
```
czerwingithub: Update comment to say if this is for all metrics or just the cumulative ones.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
+                if _t:
+                    if _t != '$$':
+                        self.__target_pids.append(int(_t))
+                    else:
+                        self.__target_pids.append(int(os.getpid()))
+
+        # history of all metrics which has form:
+        # {
+        #   '<process id>: {
+        #                 <metric name>: [<metric at time 0>, <metric at time 1>.... ],
```
czerwingithub: This makes it sound like we are keeping more than two samples... I would describe it just as the last two values for all the metrics.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
+        self.__aggregate_multiple_processes = self._config.get(
+            'aggregate_multiple_processes', default=False, convert_to=bool
+        )
+        self.__include_child_processes = self._config.get(
+            'include_child_processes', default=False, convert_to=bool
+        )
+
+        self.__poll_interval = self._config.get('poll_interval', default=120, convert_to=int)
+        self.__datetime_format = "%Y-%m-%d %H:%M:%S"
+        self.__last_polled_env_key = '%s_linux_process_monitor_last_polled' % self.__id
```
czerwingithub: Document what this is.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
+            for _metric, _metric_values in process_metrics.items():
+                if not self.__aggregated_metrics.get(_metric):
+                    self.__aggregated_metrics[_metric] = 0
+                if _metric.is_cumulative:
+                    if pid in running_pids_set:
+                        if len(_metric_values) > 1:
+                            # only report the cumulative metrics for more than one sample
+                            self.__aggregated_metrics[_metric] += (_metric_values[-1] - _metric_values[-2])
+                        else:
+                            self.__aggregated_metrics[_metric] = 0
```
czerwingithub: I don't think you should be setting this to zero... that means anytime a new process comes along, we would be resetting everything to zero.  Instead, just don't do anything if `len(_metric_values) <= 1`

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
+            del self.__metrics_history[_pid_to_remove]
+
+        # if no processes are running, there is no reason to report the running metric data
+        if not running_pids:
+            self.__aggregated_metrics = {}
+
+    def get_pids_from_ps(self, subprocesses=False, ppid=None):
+        """
+        Get the list of pids running, or get the list of child processes given a parent process id
+        :param subprocesses: Get subprocesses?
```
czerwingithub: Use `@param` notation here and below.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
+        all_pids = self.__metrics_history.keys()
+        for _pid_to_remove in list(set(all_pids) - set(running_pids)):
+            # for all the absolute metrics, decrease the count that the dead processes accounted for
+            del self.__metrics_history[_pid_to_remove]
+
+        # if no processes are running, there is no reason to report the running metric data
+        if not running_pids:
+            self.__aggregated_metrics = {}
+
+    def get_pids_from_ps(self, subprocesses=False, ppid=None):
```
czerwingithub: I think this is getting too complicated and would be better if broke it out into a smaller abstraction.

Call it `ProcessList`.  When it is instantiated, it invokes `ps -axo pid,ppid,command` to get the complete list in one go.  It saves that information in its state variables.

Then `ProcessList` should have the following method:
  * `get_matches_commandline(input_string)`
  * `get_child_processes(pid)`

You could even then define a final method that uses the first two:
  * get_matches_commandline_and_children(input_string)

You can test this abstraction separately.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
-            gather.run_single_cycle()
+        cmd = ['ps', 'ax', '-o', 'pid,command']
+        if subprocesses:
+            cmd = ["ps", "--ppid", str(ppid), "-o", "pid,command"]
+        sub_proc = Popen(cmd, shell=False, stdout=PIPE)
+        lines = sub_proc.stdout.readlines()
+        all_pids = []
+        try:
+            for line in lines:
+                line = line.strip()
```
czerwingithub: I know you didn't write this code, but when you update it, I believe you can greatly simplify it with a regular expression and matching groups.  Match just on two integers followed by a string that is the commandline.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
+            if not self.should_poll_for_processes():
+                return self.__pids
 
-        This will either use the commandline matcher or the target pid to find the process.
-        If no process is matched, None is returned.
+        matching_pids = self.get_pids_from_ps()
+        if self.__include_child_processes:
+            for matching_pid in matching_pids:
+                matching_pids.extend(
+                    self.get_pids_from_ps(subprocesses=True, ppid=matching_pid)
```
czerwingithub: The drawback here is you keep having to invoke `ps aux` for every child process.

czerwingithub: You should also include the children of the children, recursively.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
-                if self.__pid in matching_pids:
-                    self._logger.warning( "Multiple processes match the command '%s'.  Returning existing pid." % self.__commandline_matcher, limit_once_per_x_secs=300, limit_key='linux-process-monitor-existing-pid' )
-                    return self.__pid
+        if not self.__pids:
+            # no processes are currently being monitored, poll again
+            return True
+
+        if not self.__aggregate_multiple_processes:
+            # single process case
+            active_matched_pids = self.get_active_matched_pids(check_poll=False)
```
czerwingithub: You have `get_active_matched_pids` invoking this method, and this method invoking `get_active_matched_pids`.  That seems like a recipe for infinite recursion.

You should just simplify this and use `self.__pids` directly.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +741,336 @@ class ProcessMonitor(ScalyrMonitor):
...
+                # no matched processes, try again after sometime
+                return False
+            if active_matched_pids[0] not in self.__pids:
+                # matched process, but not running, old process maybe died
+                return True
+        else:
+            # multi-process case, poll every `poll_interval` seconds to
+            # see if there are more processes to match
+            if (
+                    not self.LAST_POLLED.get(self.__last_polled_env_key) or
```
czerwingithub: This is a bit complex.  You should simplify this by just checking if `poll_interval` plus  some `__last_time_check` defined for this instance is less than the current time.  If it is, then it is time to poll.

scalyr_agent/builtin_monitors/linux_process_metrics.py
code-sauce: great idea

scalyr_agent/builtin_monitors/linux_process_metrics.py
code-sauce: maintaining a state of the older ProcessTracker and calling collect on them has some unwanted side effects, also maintaining the state consumes some extra memory. I didn't think there was a lot of overhead.

code-sauce:
@czerwingithub please take a look. I split out the abstraction to create a `ProcessList`

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -588,19 +630,172 @@ def __init__(self, pid, monitor_id, logger):
...
+
+    def __init__(self):
+        # list of {'pid': processid, 'ppid': parentprocessid,  'cmd': command }
+        self.processes = []
+        # key -> parent process id, value -> [child process ids...]
+        self.parent_to_children_map = defaultdict(list)
+
+        cmd = ['ps', 'axo', 'pid,ppid,command']
+        sub_proc = Popen(cmd, shell=False, stdout=PIPE)
+        regex = r"\s+(\d+)\s+(\d+)\s+(.*)"
```
czerwingithub: Document what this format is with an example.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -588,19 +630,172 @@ def __init__(self, pid, monitor_id, logger):
...
+        lines = sub_proc.stdout.readlines()
+        for line in lines:
+            match = re.search(regex, line)
+            if match:
+                _pid, _ppid, _cmd = match.groups()
+                self.processes.append({'pid': int(_pid), 'ppid': int(_ppid), 'cmd': _cmd})
+
+        self._init_parent_to_child_map()
+
+    def _init_parent_to_child_map(self):
```
czerwingithub: I think this method is only used once.  Just inline it directly above.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -588,19 +630,172 @@ def __init__(self, pid, monitor_id, logger):
...
+        for _process in self.processes:
+            self.parent_to_children_map[_process['ppid']].append(_process['pid'])
```
czerwingithub: Should not add ppid to pid if they are equal.  I think sometimes process id = 1 is sometimes reported as a parent of itself.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -588,19 +630,172 @@ def __init__(self, pid, monitor_id, logger):
...
+        all_children = []
+        children_to_explore = []
+        for _pid in self.parent_to_children_map[ppid]:
+            all_children.append(_pid)
+            children_to_explore.append(_pid)
+
+        # get the children 'recursively'
+        while children_to_explore:  # the invariant
+            child_to_explore = children_to_explore.pop()
+            all_children.extend(self.parent_to_children_map[child_to_explore])
```
czerwingithub: To do true recursive, I think we must invoke `get_child_process` here.  After all, don't we want the children of the children as well?

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +814,255 @@ class ProcessMonitor(ScalyrMonitor):
...
-            raise BadMonitorConfiguration('At least one of the following fields must be provide: commandline or pid',
-                                          'commandline')
+        self.__aggregate_multiple_processes = self._config.get(
+            'aggregate_multiple_processes', default=False, convert_to=bool
+        )
+        self.__include_child_processes = self._config.get(
+            'include_child_processes', default=False, convert_to=bool
+        )
+        if not getattr(self, 'last_discovered', False):
+            # poll interval (in seconds) for matching processes when we are monitoring multiple processes
```
czerwingithub: You say this is a poll interval, but you initalize `__last_discovered` to a dict.  I'm not sure why you are treating `last_discovered` as special.  You should just have an instance variable named `__last_discovered` set to be None at first, but then after that, the time (in milliseconds since epoch) of when you last tried to discover new matching processes.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +814,255 @@ class ProcessMonitor(ScalyrMonitor):
...
+            # alive for all epochs of the monitor.
+            self.__last_discovered = {}
+        self.__process_discovery_interval = self._config.get('process_discovery_interval', default=120, convert_to=int)
+        self.__datetime_format = "%Y-%m-%d %H:%M:%S"
+
+        # this is the key to the dictionary self.__last_discovered which denotes the key to the monitor process
+        self.__last_polled_env_key = '%s_linux_process_monitor_last_polled' % self.__id
```
czerwingithub: Remove this.  You just need to once every 120 seconds try to see what processes should be monitored by this tracker.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +814,255 @@ class ProcessMonitor(ScalyrMonitor):
...
+        )
+        self.__include_child_processes = self._config.get(
+            'include_child_processes', default=False, convert_to=bool
+        )
+        if not getattr(self, 'last_discovered', False):
+            # poll interval (in seconds) for matching processes when we are monitoring multiple processes
+            # alive for all epochs of the monitor.
+            self.__last_discovered = {}
+        self.__process_discovery_interval = self._config.get('process_discovery_interval', default=120, convert_to=int)
+        self.__datetime_format = "%Y-%m-%d %H:%M:%S"
```
czerwingithub: Remove this.  Just keep track of time in milliseconds since epoch.  That's how we do it throughout the rest of the code base for this type of problem.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +814,255 @@ class ProcessMonitor(ScalyrMonitor):
...
+
+        # if no processes are running, there is no reason to report the running metric data
+        if not running_pids:
+            self.__aggregated_metrics = {}
+
+    def get_active_matched_pids(self):
+        """
+        Get the active pids that match the commandline match expression.
+        If there is no need to poll, then return the currently monitored pids
+        @return: List of running process IDs
```
czerwingithub: Is it the list of running process ids or the list to monitor?

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +814,255 @@ class ProcessMonitor(ScalyrMonitor):
...
+        """
+
+        if self.__aggregate_multiple_processes and self.__last_discovered.get(self.__last_polled_env_key):
```
czerwingithub: I'm not sure why having the `__last_polled_env_key` is useful.  I really think we just need once `I tried to discover new processes X seconds ago` variable per monitor instance.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +814,255 @@ class ProcessMonitor(ScalyrMonitor):
...
+            # multi-process case, poll every `poll_interval` seconds to
+            # see if there are more processes to match
+            _last_time_checked =  datetime.strptime(
```
czerwingithub: Just keep the time as the number of millisecons past epoch.  Much easier than having to do time formatting stuff.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -619,163 +814,255 @@ class ProcessMonitor(ScalyrMonitor):
...
-
-                self._logger.warning( "Multiple processes match the command '%s'.  Returning the pid of the first matching process" % self.__commandline_matcher, limit_once_per_x_secs=300, limit_key='linux-process-monitor-first-pid' )
-
-                return matching_pids[0]
 
-            finally:
-                # Be sure to wait on the spawn process.
-                if sub_proc is not None:
-                    sub_proc.wait()
+        selected_pids = []
```
czerwingithub: Ok, I think we need to simplify this.  There's too much overlap between `__select_process` and `get_active_matched_pids` and I believe we are missing some functionality here that we require.

Here's how I see the desired logic:
* First, iterate over the current selected pid list.
* For each pid, see if it is still alive.  If it is, stop iterating.  This determine if we have at least one pid still alive that we were last monitoring.
* If we have one pid alive and not `__aggregate_multiple_processes` return the current list of selected pids (which will be just one pid)
* If we have one pid alive and `_aggregate_multiple_processes` then check to `_last_checked_time`.  If it has expired get the set of all matching processes and swap it out for the current set.
* If we do not have one pid alive, then get the set of all matching processes and swap it out for the current set.  (If we are not `__aggregate_multiple_processes`, then emit the log warning if we matched more than one pid and take just the first one.)

Does that make sense?  We can discuss on Monday as needed.

scalyr_agent/builtin_monitors/linux_process_metrics.py
czerwingithub: What unwanted side effects?  It is not the overhead I am worried about in terms of reusing the ProcessTracker.  Rather, it's about the naturalness of the abstraction and doing things in a common way.  If something is called ProcessTracker, than I expect it to track that process over a period of time.  Additionally, if we know we want to track a process, it just feels natural to me that we should create something to track it over its lifetime.. and that object only is let go once the process dies.

Also, there is an efficiency thing as well.  The stat readers will only open up their corresponding /proc/pid/XXX file once when the reader is created.  After that, everytime it wants to read the file, it seeks to zero and reads the contents.  The Linux operating system is suppose to like this better.

scalyr_agent/builtin_monitors/linux_process_metrics.py
code-sauce: I was avoiding the recursive solution, but you're right - there is a bug. I forgot to iterate over the children and add them to the list. Will make the change and add the tests.

```
def get_child_processes(ppid, node):
    all_children = []
    children_to_explore = set()
    for _pid in parent_to_children_map[ppid]:
        all_children.append(_pid)
        children_to_explore.add(_pid)

    # get the children 'recursively'
    while children_to_explore:  # the invariant
        child_to_explore = children_to_explore.pop()
        if not parent_to_children_map.get(child_to_explore):
            continue
        unvisited = parent_to_children_map[child_to_explore]
        for node in unvisited:
            children_to_explore.add(node)
            all_children.append(node)
    return list(set(all_children))

if __name__ == '__main__':

    # p1 -> p2 -> p4 -> p5
    #        |
    #        p3 

    # p6 -> p7

    parent_to_children_map = {'p1': ['p2'], 'p2': ['p3', 'p4'], 'p4': ['p5'], 'p6': ['p7']}
    print get_child_processes('p1', parent_to_children_map)
```

this returns the recursed list of children:
```
['p2', 'p3', 'p4', 'p5']`
```

scalyr_agent/builtin_monitors/linux_process_metrics.py
code-sauce: makes sense!

scalyr_agent/builtin_monitors/linux_process_metrics.py
code-sauce: yup, epoch time makes sense

scalyr_agent/builtin_monitors/linux_process_metrics.py
code-sauce: @czerwingithub I think I have a basic question about keeping state and monitoring when was the last time we checked for something. I put some print statements on when the `__initialize__` function is called and it seems like it is called on every epoch, so I am not sure how can we maintain state across epochs in memory by repurposing an instance variable.

```

 __initialize called.. self:  139941875320912 sjaintest

 record metrics called:  agent 139941874867024

 record metrics called:  sjaintest 139941874681360


 __initialize called.. self:  139941874896080 agent

 record metrics called:  sjaintest 139941874681360

 record metrics called:  sjaintest 139941874681360

 record metrics called:  sjaintest 139941874681360

 record metrics called:  sjaintest 139941874681360

 record metrics called:  sjaintest 139941874681360

 record metrics called:  sjaintest 139941874681360

 record metrics called:  agent 139941874867024


 __initialize called.. self:  139941874895632 sjaintest```

File: scalyr_agent/builtin_monitors/linux_process_metrics.py (hunk 32 above)
Comment (code-sauce) [reply scalyr_agent/builtin_monitors/linux_process_metrics.py]: monitored. will clarify

**Comment by code-sauce** [comment]
@czerwingithub see if the last 3 commits make sense. Been looking at the code for long, so I'll take a look later again to see if I missed anything.

File: scalyr_agent/builtin_monitors/linux_process_metrics.py [hunk 36]
```diff
@@ -765,23 +748,38 @@ def get_matches_commandline(self, match_pattern):
 
     def get_child_processes(self, ppid):
         """
-        Given a process pid, get the children pids
-        @param ppid: parent process id
-        @return: list of children process IDs
+        Given a process id, return all children processes (recursively)
+        :param ppid: parent process id
```
czerwingithub: Change to using `@` instead of `:` to be consistent.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -765,23 +748,38 @@ def get_matches_commandline(self, match_pattern):
...
         # get the children 'recursively'
         while children_to_explore:  # the invariant
             child_to_explore = children_to_explore.pop()
-            all_children.extend(self.parent_to_children_map[child_to_explore])
+            if not self.parent_to_children_map.get(child_to_explore):
+                continue
+            unvisited = self.parent_to_children_map[child_to_explore]
+            for node in unvisited:
+                children_to_explore.add(node)
+                all_children.append(node)
```
czerwingithub: Just for safety, check first to see if `node` is in `all_children` and only add it to `children_to_explore` if it is not.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -834,19 +832,19 @@ def _initialize(self):
         self.__include_child_processes = self._config.get(
             'include_child_processes', default=False, convert_to=bool
         )
-        if not getattr(self, 'last_discovered', False):
+
+        self.__trackers = {}  # key -> process id, value -> ProcessTracker object
+        if not getattr(self, 'last_discovered', None):
```
czerwingithub: Just write this as `self.__last_discovered = None`.  No need to be checking this attribute now.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -834,19 +832,19 @@ def _initialize(self):
...
         self.__process_discovery_interval = self._config.get('process_discovery_interval', default=120, convert_to=int)
-        self.__datetime_format = "%Y-%m-%d %H:%M:%S"
-
-        # this is the key to the dictionary self.__last_discovered which denotes the key to the monitor process
-        self.__last_polled_env_key = '%s_linux_process_monitor_last_polled' % self.__id
 
         __target_pids = self._config.get('pid', default=None, convert_to=str)
         self.__target_pids = []
 
+        self.__running_pids = []  # list of all process ids that are currently running
```
czerwingithub: How is this different than `self.__pids`?  I think we can remove `self.__running_pids`.

scalyr_agent/builtin_monitors/linux_process_metrics.py
code-sauce: I can do that, but for the same reason of duplicity,  we're removing duplicates 2 lines later `list(set(all_children))` but it doesn't hurt checking.

scalyr_agent/builtin_monitors/linux_process_metrics.py
code-sauce: `__running_pids` is a list of *all* the running processes at a particular time. `__pids` is the list of running processes we are interested in.

code-sauce: but you're right. it can be refactored and we don't really need `__running_pids`

code-sauce:
@czerwingithub please take a look. I took your suggestion and removed the `__running_pids` in lieu of `__pids`.

PR #68, opened 2017-11-05
Split status update message into three parts

Stephan202:
This way the agent won't claim to have verified the connection when in fact it was started with `--no-check-remote-server`.

(When I run the tests using the command in `.circleci/config.yml`, 20 errors and 47 failures are reported, both before and after this change. If you'd like me to do any of this differently, don't hesitate to ask!)

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -368,6 +368,8 @@ class ScalyrAgent(object):
             finally:
                 client.close()
 
+            print "Server connection verified."
+
     def __start(self, quiet, no_fork, no_check_remote):
         """Executes the start command.
 
@@ -413,7 +415,7 @@ class ScalyrAgent(object):
             self.__fail_if_already_running()
 
             if not quiet:
-                print "Configuration and server connection verified, starting agent in background."
+                print "Starting agent in background."
             self.__controller.start_agent_service(self.__run, quiet, fork=True)
         else:
             self.__controller.start_agent_service(self.__run, quiet, fork=False)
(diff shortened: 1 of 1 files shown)


scalyr_agent/agent_main.py
```diff
@@ -413,7 +415,7 @@ def __start(self, quiet, no_fork, no_check_remote):
             self.__fail_if_already_running()
 
             if not quiet:
-                print "Configuration and server connection verified, starting agent in background."
+                print "Starting agent in background."
```
czerwingithub: We would like to emit as few lines as possible -- users tend to not read the output so it's better in our opinion to put as much information as we can in each line.

How about if you just rewrite this as:

    if not quiet:
        if no_check_remote:
            print "Configuration verified, starting agent in background."
        else
            print "Configuration and server connection verified, starting agent in background."

scalyr_agent/agent_main.py
```diff
@@ -368,6 +368,8 @@ def __perform_config_checks(self, no_check_remote):
             finally:
                 client.close()
 
+            print "Server connection verified."
```
czerwingithub: This will emit the output to stdout even if the user has selected the `quiet` option.  I have a suggest below where we can do what you are trying to do, but in a little more compact of a form.  If you follow that suggestion, you can remove this line.

scalyr_agent/agent_main.py
```diff
@@ -849,6 +851,8 @@ def __verify_can_write_to_logs_and_data(self, config):
         if not os.access(config.agent_data_path, os.W_OK):
             raise Exception('User cannot write to agent data directory \'%s\'.' % config.agent_data_path)
 
+        print "Configuration verified."
```
czerwingithub: We can remove this line in favor of the suggested fix above.

Stephan202:
Thanks for the review! My thinking was that splitting out the print statements ensures that logging happens on each code path, but I understand you want to avoid that. As for the quiet mode; I simply overlooked this. I have pushed a commit with the suggested changes.

The Travis CI failure points to a private repo; I assume the failure is unrelated to my changes. If not, let me know :)

Stephan202:
You're welcome :+1: . Thanks for making the software that runs client-side open-source.

PR #70, opened 2017-12-04
Imron/k8s

imron:
Here is the documentation for running under Kubernetes.  It's based off my previous imron/raw-docker branch and should merge cleanly into the current master.

The documentation makes the assumption that  k8s/scalyr-agent-2.yaml exists on the master branch of the main scalyr/scalyr-agent-2 repository.

It also assumes that Dockerhub contains an image scalyr/scalyr-k8s-agent which is built from the  docker/Dockerfile.k8s_config

Please adjust commands as appropriate to use local versions while testing.


docker/Dockerfile.custom_k8s_config
```diff
@@ -0,0 +1,18 @@
+# Dockerfile for building an image of scalyr/scalyr-docker-agent with custom configuration
```
czerwingithub: I thought we were going to have a different image name for the k8 build of the agent, such as `scalyr/scalyr-k8-agent`

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -103,6 +103,12 @@
                      'legacy behavior.\n',
                      convert_to=str, default="docker_api")
 
+define_config_option( __monitor__, 'docker_raw_logs',
+                     'Optional (defaults to False). If True, the docker monitor will use the raw log files on disk to read logs.'
```
czerwingithub: I would also document here how we determine the path to the log file.  Reading this, I start looking for the option on how I specify that.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -456,6 +483,9 @@ def __update_checkpoints( self ):
         to file.
         """
 
+        if self._use_raw_logs:
```
czerwingithub: I'm always wary where we add in return points in the middle of a method. I would suggest we instead move the code below into a method like `__watch_container_logs` and then only invoke that here if `!self._use_raw_logs`.

PR #73, opened 2018-01-14
Reduce client errors when servers overloaded

sgrimm-sg:
When Scalyr's servers are overloaded, they return a combination of "too
busy" HTTP responses and internal errors, both of which result in
ERROR-level log messages in the agent log.

This commit adds explicit checking for the specific error message that
accompanies internal errors when the servers are too busy, and
downgrades the log messages to warnings since the agent retries
automatically and no human intervention is required.


imron:
Hi sgrimm-sg, thanks for submitting this pull request, and sorry it slipped through the cracks and didn't get a response for so long.  I'll be taking a look at this some time over the next few days and will update you then.

PR #75, opened 2018-02-23
k8s monitor

imron:
The kubernetes monitor is currently based entirely off the
docker monitor, with support added for querying the k8s API
to retrieve information about the pods currently running.

Eventually common functionality between the docker and k8s
monitors needs to be factored out.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -19,6 +19,7 @@ __author__ = 'imron@imralsoftware.com'
 import datetime
 import docker
 import fnmatch
+import hashlib
 import traceback
 import logging
 import os
@@ -40,6 +41,7 @@ from scalyr_agent.log_watcher import LogWatcher
 from scalyr_agent.monitor_utils.server_processors import LineRequestParser
 from scalyr_agent.monitor_utils.server_processors import RequestSizeExceeded
 from scalyr_agent.monitor_utils.server_processors import RequestStream
+from scalyr_agent.monitor_utils.k8s import Kubernetes
 
 from scalyr_agent.util import StoppableThread
 
@@ -102,13 +104,6 @@ define_config_option( __monitor__, 'log_mode',
                      'preferred method due to bugs/issues found with the docker API.  It is not the default to protect '
                      'legacy behavior.\n',
                      convert_to=str, default="docker_api")
-
-define_config_option( __monitor__, 'docker_raw_logs',
-                     'Optional (defaults to False). If True, the docker monitor will use the raw log files on disk to read logs.'
-                     'The location of the raw log file is obtained by querying the path from the Docker API. '
-                     'If false, the logs will be streamed over the Docker API.'
-                     convert_to=bool,
-                     default=False)
 
 define_config_option( __monitor__, 'metrics_only',
                      'Optional (defaults to False). If true, the docker monitor will only log docker metrics and not any other information '
@@ -310,6 +305,37 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
 
     return result
 
+class PodInfo( object ):
+    """
+        A collection class that stores label and other information about a kubernetes pod
+    """
+    def __init__( self, name='', namespace='', uid='', nodeName='', labels={}, containers=[] ):
+
+        self.name = name
+        self.namespace = namespace
+        self.uid = uid
+        self.nodeName = nodeName
+        self.labels = labels
+        self.containers = containers
+
+        # generate a hash we can use to compare whether or not
+        # any of the pod info has changed
+        md5 = hashlib.md5()
+        md5.update( name )
+        md5.update( namespace )
+        md5.update( uid )
+        md5.update( nodeName )
+
+        # flatten the labels dict in to a single string
+        flattened = []
+        for k,v in labels.iteritems():
+            flattened.append( k )
+            flattened.append( v )
+        md5.update( ''.join( flattened ) )
+
+        self.digest = md5.digest()
+
+
 class ContainerChecker( StoppableThread ):
     """
         Monitors containers to check when they start and stop running.
@@ -319,8 +345,6 @@ class ContainerChecker( StoppableThread ):
 
         self._config = config
         self._logger = logger
-
-        self._use_raw_logs = config.get('docker_raw_logs')
 
         self.__delay = self._config.get( 'container_check_interval' )
         self.__log_prefix = self._config.get( 'docker_log_prefix' )
@@ -342,7 +366,8 @@ class ContainerChecker( StoppableThread ):
         self.__glob_list = config.get( 'container_globs' )
 
         self.containers = {}
-        self.__checkpoints = {}
+
+        self.__k8s = Kubernetes()
 
         self.__log_watcher = None
         self.__module = None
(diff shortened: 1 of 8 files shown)


scalyr_agent/monitor_utils/k8s.py
```diff
@@ -0,0 +1,77 @@
...
+            self._session.headers.update( self._standard_headers )
+
+    def query_api( self, path, pretty=0 ):
+        """ Queries the k8s API at 'path', and converts OK responses to JSON objects
+        """
+        self._ensure_session()
+        url = self._http_host + path + '?pretty=%d' % (pretty)
+        response = self._session.get( url, verify=self._verify_connection(), timeout=self._timeout )
+        if response.status_code != 200:
+            raise Exception( "Invalid response from Kubernetes API when querying '%s': %s" %( path, str( response ) ) )
```
czerwingithub: Let's define an exception type like K8ApiException.. that way we can more narrowly catch it.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+define_config_option( __monitor__, 'report_container_metrics',
+                      'Optional (defaults to True). If true, metrics will be collected from the container and reported  '
+                      'to Scalyr.', convert_to=bool, default=True)
+
+# for now, always log timestamps to help prevent a race condition
+#define_config_option( __monitor__, 'log_timestamps',
+#                     'Optional (defaults to False). If true, stdout/stderr logs will contain docker timestamps at the beginning of the line\n',
+#                     convert_to=bool, default=False)
+
+define_metric( __monitor__, "docker.net.rx_bytes", "Total received bytes on the network interface", cumulative=True, unit="bytes", category="Network" )
```
czerwingithub: Is it still right to refer to this as docker metrics?  Do people running k8s think of it as also running docker?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+                     limit_key='docker-api-running-containers' )
+        result = None
+
+    return result
+
+class PodInfo( object ):
+    """
+        A collection class that stores label and other information about a kubernetes pod
+    """
+    def __init__( self, name='', namespace='', uid='', nodeName='', labels={}, containers=[] ):
```
czerwingithub: Can we change to `node_name`?  We should be consistent about variable naming convention.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+
+        self.__host_hostname = host_hostname
+
+        self.__readback_buffer_size = self._config.get( 'readback_buffer_size' )
+
+        self.__glob_list = config.get( 'container_globs' )
+
+        self.containers = {}
+
+        self.__k8s = Kubernetes()
```
czerwingithub: `Kubernetes` seems too generic of a name.  Maybe something that indicates it for the api?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+            path = logger['log_config']['path']
+            if self.__log_watcher:
+                self.__log_watcher.remove_log_path( self.__module, path )
+            self._logger.log(scalyr_logging.DEBUG_LEVEL_1, "Stopping %s" % (path) )
+
+        self.raw_logs = []
+
+    def _process_namespaces( self, namespaces ):
+        """ Iterates over all namespaces on this node and then queries
+        every pod in every namespace.
```
czerwingithub: Say something about the fact it's going to process all of the pods and what processing means.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -0,0 +1,77 @@
...
+            self._session = requests.Session()
+            self._session.headers.update( self._standard_headers )
+
+    def query_api( self, path, pretty=0 ):
+        """ Queries the k8s API at 'path', and converts OK responses to JSON objects
+        """
+        self._ensure_session()
+        url = self._http_host + path + '?pretty=%d' % (pretty)
+        response = self._session.get( url, verify=self._verify_connection(), timeout=self._timeout )
+        if response.status_code != 200:
```
czerwingithub: Let's also add some scalyr debugging down here in case we need it... maybe level 3 or something.  Log what the URL was and what the status code was with byte count of response.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+        """
+        # get the list of namespaces
+        items = namespaces.get( 'items', [] )
+        containers = {}
+
+        # iterate over all namespaces, querying and processing all pods
+        # in the namespace
+        for namespace in items:
+            metadata = namespace.get( 'metadata', {} )
+            if 'name' in metadata:
```
czerwingithub: Would it be an error if `name` is not in `metadata`?  If so, let's emit an error message if we do not find it with a limit of once per 5 mins.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+        @return A dict of PodInfo objects, keyed on docker container ID
+        """
+        # get the list of namespaces
+        items = namespaces.get( 'items', [] )
+        containers = {}
+
+        # iterate over all namespaces, querying and processing all pods
+        # in the namespace
+        for namespace in items:
+            metadata = namespace.get( 'metadata', {} )
```
czerwingithub: For all of these methods.. let's add in a scalyr debugging at the most fine level to emit a log line for each item we are iterating over.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+        containerStatuses = status.get( 'containerStatuses', [] )
+
+        containers = []
+
+        # build list of container ids used by this pod, making sure to
+        # strip docker:// from the front of each one
+        prefix = 'docker://'
+        for container in containerStatuses:
+            if 'containerID' not in container:
+                continue
```
czerwingithub: Is this an error?  If so, then let's emit a log line with a limit of one per 5 min

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+
+        container_attributes = base_attributes.copy()
+        container_attributes['containerName'] = info['name']
+        container_attributes['containerId'] = cid
+
+        parser = 'docker'
+        if cid in k8s_data:
+            pod = k8s_data[cid]
+
+            container_attributes['podName'] = pod.name
```
czerwingithub: Change `podName` to `pod_name` unless if there is a really strong convenient of using camel case for k8s.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+            container_attributes['namespace'] = pod.namespace
+            container_attributes['podUid'] = pod.uid
```
czerwingithub: `pod_uid`

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+            container_attributes['nodeName'] = pod.nodeName
```
czerwingithub: `node_name`

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -0,0 +1,1226 @@
...
+
+        self.__glob_list = self._config.get( 'container_globs' )
+
+        self.__report_container_metrics = self._config.get('report_container_metrics')
+
+        self.__container_checker = None
+        if self._config.get('log_mode') != 'syslog':
+            self.__container_checker = ContainerChecker( self._config, self._logger, self.__socket_file, self.__docker_api_version, host_hostname, data_path, log_path )
+
+        self.__network_metrics = self.__build_metric_dict( 'docker.net.', [
```
czerwingithub: Same question here.. is it still appropriate to call them `docker` metrics?  Now... for our customers, maybe this is easier because they could have dashboards built using these metric names.  What do you think?  Do people generally think of this as not running docker?

imron: As mentioned via email, I think it's still worth calling them docker metrics, because they are metrics for each docker container, and there are likely separate k8s specific metrics we could be collecting too, so calling the docker metrics 'docker_metrics' helps differentiate.

czerwingithub: That sounds fine to me.  You are more in touch with what the community will expect.

imron:
These are all now fixed.

PR #81, opened 2018-04-11
K8s preview

imron:
This updates the preview branch of k8s to have all the debug logging and the container-mapping used by the recent debug build.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -256,7 +256,7 @@ def _split_datetime_from_line( line ):
     return (dt, log_line)
 
 def _get_containers(client, ignore_container=None, restrict_to_container=None, logger=None,
-                    only_running_containers=True, glob_list=None, include_log_path=False):
+                    only_running_containers=True, glob_list=None, include_log_path=False, include_k8s_info=False):
     """Gets a dict of running containers that maps container id to container name
     """
     if logger is None:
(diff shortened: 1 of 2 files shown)


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -558,8 +583,15 @@ def check_containers( self, run_state ):
...
                     pod = None
-                    if cid in k8s_data:
-                        pod = k8s_data[cid]
+                    if 'k8s_info' in info:
+                        pod_name = info['k8s_info'].get( 'pod_name', 'invalid_pod' )
+                        pod_namespace = info['k8s_info'].get( 'pod_namespace', 'invalid_namespace' )
+                        pod = k8s_data.get( pod_namespace, {} ).get( pod_name, None )
+
+                        if not pod:
+                            self._logger.warning( "No pod info for container %s.  pod: '%s/%s'" % (cid[:8], pod_namespace, pod_name),
```
czerwingithub: You refer to cid[:8] in a bunch of places.. maybe just once in this method but several times in others.  I would create a helper function that is named whatever this is extracting out, then just invoke it on `cid` anytime you need cid[:8]... in methods where you use it a bunch of times, use a variable to hold it.  Also, add error checking in the method to handle the case where `len(cid) < 8` -- I know it probably can't, but just to be safe.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -80,12 +119,24 @@ def query_api( self, path, pretty=0 ):
 
     def query_pod( self, namespace, pod ):
         """Wrapper to query a pod in a namespace"""
-        return self.query_api( '/api/v1/namespaces/%s/pods/%s' % (namespace, pod) )
+        if not pod or not namespace:
```
czerwingithub: Is this an error?  If they pass in an empty `pod` or `namespace`, should we just thrown an exception?

imron:
Ok, will do.  For reference, this is just getting a shortened container id to use for logging.  Using the full cid just adds noise to the logging or splits lines so I'm taking the first 8 chars of the cid which will be unique enough for identification without pushing the useful information from the log message off the screen or on to the following line.

Regarding the second item about checking for len < 8, this is not needed.

It is not an error for slices to specify an end value beyond the length.  See slicing documentation here:

https://docs.python.org/2/reference/expressions.html#slicings

Relevant quote:

"The lower and upper bound expressions, if present, must evaluate to plain integers; defaults are zero and the sys.maxint, respectively. If either bound is negative, the sequence’s length is added to it. The slicing now selects all items with index k such that i <= k < j where i and j are the specified lower and upper bounds. This may be an empty sequence. It is not an error if i or j lie outside the range of valid indexes (such items don’t exist so they aren’t selected)."

PR #86, opened 2018-05-09
Checking for short lived containers

imron:


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -275,9 +280,23 @@ def _get_short_cid( container_id ):
     # https://docs.python.org/2/reference/expressions.html#slicings
     return container_id[:8]
 
+def _ignore_old_dead_container( container, threshold ):
```
czerwingithub: Add some quick documentation for the method, including what threshold is.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -275,9 +280,23 @@ def _get_short_cid( container_id ):
...
+    result = False
+    # check for recently finished containers
+    if threshold is not None:
+        state = container.get( 'State', {} )
+
+        #ignore any that are finished and that are also too old
+        if state != 'running':
+            created = container.get( 'Created', 0 ) # default to a long time ago
+            if created < threshold:
+                result = True
```
czerwingithub: how about just return True here?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -275,9 +280,23 @@ def _get_short_cid( container_id ):
...
+
+    return result
 
 def _get_containers(client, ignore_container=None, restrict_to_container=None, logger=None,
-                    only_running_containers=True, glob_list=None, include_log_path=False, include_k8s_info=False):
+                    only_running_containers=True, running_or_finished_since=None, glob_list=None, include_log_path=False, include_k8s_info=False):
```
czerwingithub: Add some argument documentation here as well.

scalyr_agent/util.py
```diff
@@ -220,6 +220,9 @@ def rfc3339_to_datetime( string ):
     # split the string in to main time and fractional component
     parts = string.split(".")
 
+    if parts[0].endswith( 'Z' ):
```
czerwingithub: Add a comment to describe what this is doing.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -582,7 +608,12 @@ def check_containers( self, run_state ):
                 k8s_data = self.get_k8s_data()
 
                 self._logger.log(scalyr_logging.DEBUG_LEVEL_2, 'Attempting to retrieve list of containers:' )
-                running_containers = _get_containers(self.__client, ignore_container=self.container_id, glob_list=self.__glob_list, include_log_path=True, include_k8s_info=True)
+
+                running_or_finished_since = None
+                if self.__short_lived_seconds is not None:
```
czerwingithub: At one point, we had discussed checking for all containers that were created after the last time we polled the container list (and are already dead).  I forget, why can't we do that?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -85,6 +85,11 @@
                      'Optional (defaults to docker). Prefix added to the start of all docker logs. ',
                      convert_to=str, default='docker')
 
+define_config_option( __monitor__, 'docker_short_lived_interval',
```
czerwingithub: I think we should default for this to be on.  I'm not as concerned about existing customers seeing more containers show up then they used to having... it's probably not common folks have short lived containers.  I'm more concerned about people not getting all of this containers.

imron: No problem.

scalyr_agent/util.py
imron: Will do.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Done.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Done.  Also changed this to running_or_created_after to reflect that it's actually using the create time not the finish time to filter containers.

imron:
Made the suggested changes and rebased of current master for a clean merge.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
czerwingithub: Hmm.. in my mind, this points at a larger problem.  What happens if a container we are watching exits?  Do we make sure we finish copying its logs, or do we stop where ever we were when the container exits?  It would be better for us to put such a container in a state of "keep copying until you reach the end of the log file and then no longer watch the log file."

imron:
Yep agree.

I've modified the the code to fix that, but there's still one small bug I can't track down yet, where it's still skipping the log.

I should have a fix for it tomorrow (my time) we can discuss details in our meeting.

Regards,
Imron

imron:
Ok. This should be good now.

Short lived files should all be being picked up, and there is no longer a configuration option needed.

PR #87, opened 2018-05-10
allow global config items to be specified in agent.d config fragments

imron:


czerwingithub:
Also, can you check out the test cases in configuration_test.py?   I could have sworn we had test cases to check duplicate keys.  At least add two cases -- one to verify you can set a given config option in a file other than agent.json and another to verify we throw an error if you specify the same key across multiple files.

imron:
Comments and tests updated.

For reference, there didn't appear to be config tests for duplicate keys (but there is now).

PR #88, opened 2018-05-11
fixed linux process metrics exceptions

imron:
There were 2 issues causing this.

The underlying issue was the call to os.listdir with a path containing the process id, which would not exist if the process had exited.

This was throwing an exception and marking the metric gathering as failed for that metric.

Then, the BaseReader was returning None during `run_single_cycle` if metric gathering had failed for a given metric reader, but the result of `run_single_cycle` was always being passed unchecked in to a call to dict.update (which expects an iterator).

Because None is not iterable, then whenever the reader failed there would then be multiple errors about not being able to iterate NoneType.

This is now fixed so that:

1. If 'file not found' occurs during os.listdir for a path with a pid, we catch the exception, log a warning and return 0.  

2. BaseReader now returns an empty dict on failure rather than None because of recent changes to have the readers return dicts.

3. For future proofing, in case a new Reader is created that accidentally returns None rather than a dict, the return value from `run_single_cycle` is checked and only passed to dict.update if it is non None.

For good measure, I also replaced a number of hard-coded error numbers with the appropriate python errno.CONSTANT


scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -638,7 +639,18 @@ def run_single_cycle(self, collector=None):
...
 
-        num_fds = len(os.listdir(self.__path))
+        num_fds = 0
+        try:
+            num_fds = len(os.listdir(self.__path))
+        except OSError, e:
+            if e.errno == errno.ENOENT:
+                # ignore file not found, it just means the process
+                # is dead so log a warning, but still continue
+                self._logger.warn( "Path '%s' not found.  This likely means that process '%ld' is no longer running" % (self.__path, self.__pid ) )
```
czerwingithub: Let's not even bother to log the warning here.. this is kind of expect behavior.. a process may exit that is being monitored.  

I think it's ok to consider it having zero fds if the process has exited.

imron:
Warning removed.  I've also rebased off master for a clean merge.

PR #92, opened 2018-05-29
added option to disable ssl verification for internal k8s api queries

imron:
Provide the option to disable ssl certificate verification when performing queries on the internal k8s api.

This is added because some older versions of k8s appear to fail verification, so this gives people using legacy versions of k8s a workaround for that.

@czerwingithub, @johnpadilla can provide more details.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -121,7 +121,8 @@ define_config_option( __monitor__, 'report_container_metrics',
 
 define_config_option( __monitor__, 'verify_k8s_api_queries',
                       'Optional (defaults to True). If true, then the ssl connection for all queries to the k8s API will be verified using '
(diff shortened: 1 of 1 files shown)


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -119,6 +119,11 @@
                       'Optional (defaults to True). If true, metrics will be collected from the container and reported  '
                       'to Scalyr.', convert_to=bool, default=True)
 
+define_config_option( __monitor__, 'verify_k8s_api_queries',
+                      'Optional (defaults to True). If true, then the ssl connection for all queries to the k8s API will be verified using '
+                      'the ca.crt certificate found in the service account directory. If false, no verification will be performed ',
```
czerwingithub: Add in a comment saying this can be used to get around issues with older k8 releases.

imron:
I've updated the configuration option description to explain why it is useful.

PR #93, opened 2018-05-30
Imron/journald

imron:
Initial journald plugin.


scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,258 @@
...
+
+global_log = scalyr_logging.getLogger(__name__)
+__monitor__ = __name__
+
+define_config_option( __monitor__, 'journal_path',
+                     'Optional (defaults to /var/log/journal). Location on disk of the journald logs.',
+                     convert_to=str, default='/var/log/journal')
+
+define_config_option( __monitor__, 'journal_poll_timeout_milliseconds',
+                     'Optional (defaults to 0). The number of milliseconds to wait for data while polling the journal file. '
```
czerwingithub: I think you should have a little explanation in the main documentation for the monitor on how the polling from journald works and reference it here.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,258 @@
...
+                   )
+
+global_log = scalyr_logging.getLogger(__name__)
+__monitor__ = __name__
+
+define_config_option( __monitor__, 'journal_path',
+                     'Optional (defaults to /var/log/journal). Location on disk of the journald logs.',
+                     convert_to=str, default='/var/log/journal')
+
+define_config_option( __monitor__, 'journal_poll_timeout_milliseconds',
```
czerwingithub: Let's put this in units of seconds and allow a float.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,258 @@
...
+
+define_config_option( __monitor__, 'journal_fields',
+                     'Optional dict containing a list of journal fields to include with each message, as well as a field name to map them to.\n'
+                     'Note: Not all fields need to exist in every message and only fields that exist will be included.\n'
+                     'Defaults to {:\n'
+                     '  "_SYSTEMD_UNIT": "unit"\n'
+                     '  "_PID": "pid"\n'
+                     '  "_MACHINE_ID": "machine_id"\n'
+                     '  "_BOOT_ID": "boot_id"\n'
+                     '  "_HOSTNAME": "hostname"\n'
```
czerwingithub: Will it be normal for journald to have logs from other hosts?

This hostname will be in addition to the `serverHost` field already uploaded by the agent.  The `serverHost` field is what determines what server the log came from in the UI.  I want to make sure we aren't confusing the user in the common case.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,258 @@
...
+                     '  "_SOURCE_REALTIME_TIMESTAMP": timestamp"\n'
+                     '}\n',
+                     default=None
+                     )
+
+define_config_option( __monitor__, 'journal_matches',
+                      'Optional list containing match strings for filtering entries.  A match string consists of "FIELD=value" where FIELD '
```
czerwingithub: I believe you said we are relying on a standard journald filtering mechanism here.  Our documentation here should point to their documentation.  I'm not quite understanding how this works from the description here.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,258 @@
...
+                      'is a field of the journal entry e.g. _SYSTEMD_UNIT, _HOSTNAME, _GID and "value" is the value to filter on. '
+                      'All matches of different field are combined with logical AND, and matches of the same field are automatically combined with logical OR. '
+                      'If empty or None then no filtering occurs.',
+                      default=None )
+
+define_config_option( __monitor__, 'checkpoint_name',
```
czerwingithub: I think we should do away with this and just rely on the existing monitor instance name mechanism to allow them to specify unique names for the monitor using the `id` configuration option.  If they don't specify an `id` then that's on them.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,258 @@
...
+
+        Note: This function should only be called directly after a call to _has_pending_entries()
+        returns True.
+        """
+        # read all the entries
+        for entry in self._journal:
+            try:
+                msg = entry.get('MESSAGE', '')
+                extra = self._get_extra_fields( entry )
+                self._logger.emit_value('message', msg, extra_fields=extra)
```
czerwingithub: `message` is a reserved field in scalyr meant to hold the entire log line .  So, use something else here.. maybe `content` or `payload`?

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,258 @@
...
+        try:
+            process = self._journal.process()
+        except Exception, e:
+            # early return if there was an error
+            global_log.warn( "Error processing journal entries: %s" % str(e),
+                             limit_once_per_x_secs=60, limit_key='journald-process-error' )
+
+            return False
+
+        if process == journal.NOP:
```
czerwingithub: Add comment here saying what a NOP from journald means.

scalyr_agent/monitor_utils/journald_checkpoints.py
```diff
@@ -0,0 +1,74 @@
...
+# author:  Imron Alston <imron@scalyr.com>
+
+__author__ = 'imron@scalyr.com'
+
+import random
+import threading
+import scalyr_agent.util as scalyr_util
+import scalyr_agent.scalyr_logging as scalyr_logging
+
+global_log = scalyr_logging.getLogger(__name__)
```
czerwingithub: Why have this as a separate file?

scalyr_agent/monitor_utils/journald_checkpoints.py
```diff
@@ -0,0 +1,74 @@
...
+
+_global_lock = threading.Lock()
```
czerwingithub: document that it must be held to access the next field.

scalyr_agent/monitor_utils/journald_checkpoints.py
```diff
@@ -0,0 +1,74 @@
...
+        if filename not in _global_checkpoints:
+            checkpoints = {}
+            try:
+                checkpoints = scalyr_util.read_file_as_json( filename )
+            except:
+                global_log.log( scalyr_logging.DEBUG_LEVEL_1, "No checkpoint file '%s' exists.\n\tAll journald logs for '%s' will be read starting from their current end.", filename )
+                checkpoints = {}
+
+            result = Checkpoint( filename, checkpoints )
+            _global_checkpoints[filename] = result
```
czerwingithub: I'm not a big fan of holding a lock while we are doing I/O operations.  How about we hold the lock to check to see if `filename` is in `_global_checkpoints` and then again when we go to try to insert in what we just read into `_global_checkpoints`.  If someone else has added it to `_global_checkpoints` while we did the read operation, then we just drop our `Checkpoint` object on the ground and return the one in the the `_global_checkpoints` dict.

scalyr_agent/monitor_utils/journald_checkpoints.py
```diff
@@ -0,0 +1,74 @@
...
+    finally:
+        _global_lock.release()
+
+    return result
+
+class Checkpoint( object ):
```
czerwingithub: Add documentation for what this class does.

scalyr_agent/monitor_utils/journald_checkpoints.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def __init__( self, filename, checkpoints ):
+        self._lock = threading.Lock()
+        self._filename = filename
+        self._checkpoints = checkpoints
+
+    def get_checkpoint( self, name ):
```
czerwingithub: Document this method and other methods, including arguments.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,258 @@
+# Copyright 2014 Scalyr Inc.
```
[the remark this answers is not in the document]
imron: Good catch.  That's what I get for copying/pasting the copyright from the syslog monitor (which probably shouldn't be 2014 either).

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,258 @@
...
+from scalyr_agent import ScalyrMonitor, define_config_option
+from scalyr_agent.scalyr_monitor import BadMonitorConfiguration
+from scalyr_agent.monitor_utils import journald_checkpoints
+from scalyr_agent.monitor_utils.journald_checkpoints import Checkpoint
+import scalyr_agent.scalyr_logging as scalyr_logging
+
+try:
+    from systemd import journal
+except ImportError:
+    raise Exception('Python systemd library not installed.\n\nYou must install the systemd python library in order '
```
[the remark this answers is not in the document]
imron: No other places where it's raised (but several where it's caught).  That being said, the monitor modules are loaded dynamically, so it's already the case that the exception will only be raised if the user has configured the agent to use the journald monitor.  I just double checked this by setting it to always raise that exception and the error message only manifested when the journald_monitor was configured, otherwise the agent started normally.  I can make it throw in the initialize method if necessary, but the behaviour is already the same as you described so I'm not sure if it's necessary.

scalyr_agent/builtin_monitors/journald_monitor.py
imron: I'm not sure,  however journald supports it and Docker (the company) make extensive use of journald so maybe it's something they'll encounter?  I can change it to originating_host or similar or just leave it out?  If they need it, I guess they can always specify it manually in their agent config.

scalyr_agent/builtin_monitors/journald_monitor.py
imron: There is no existing mechanism per se, instead it appears the `id` field is just a common convention used by some monitors (and each one implements as necessary).  I've updated the agent to also use id instead of checkpoint_name, or use the module_name if no id is specified.

scalyr_agent/builtin_monitors/journald_monitor.py
imron: `details` appears to be used by other plugins, and I saw that during development and thought `message` is more suitable.  Now I now why that's the case :-)  Maybe switch it to `details`?

scalyr_agent/monitor_utils/journald_checkpoints.py
imron: Not for anything relevant to this plugin, but as I was writing it, I thought that this is not the only plugin that requires saving checkpoints (docker plugin does it also) and future plugins will probably need this functionality at some point too, so my thinking was it might make sense to have a generic way of handling monitor checkpoints and ended up putting this in its own file in case it gets expanded for other plugins later.

scalyr_agent/monitor_utils/journald_checkpoints.py
imron: I've fixed this up.  I agree, I'm not a big fan of locking while doing I/O either.  What about the case further down where we write the file atomically?  The atomic_write function writes to a temp file first, so I guess we'd then need to be not locking and creating unique temporary files per thread, and relying on the fact that the OS will take care of renames atomically for us without needing the lock?

I had a look at doing that initially, but the standard python tempfile mechanism didn't really have something suitable for how we use it here, and so I thought holding the lock for that I/O is probably going to be ok - at least for the journald plugin which has relatively small checkpoint files and relatively little that can go wrong during the I/O phase.  Thoughts?

scalyr_agent/builtin_monitors/journald_monitor.py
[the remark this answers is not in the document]
czerwingithub: Ok, we can leave it where it is as long as the error message is being shown at the right place.  Just add in a comment saying this should only be thrown during monitor configuration.

scalyr_agent/builtin_monitors/journald_monitor.py
czerwingithub: Yes, let's leave it out for now with a comment saying that that field is available for use by customers.  I suspect (but could very well be wrong) that this will be duplicate with the serverHost so I don't want to confuse folks by having duplicate information.

scalyr_agent/builtin_monitors/journald_monitor.py
czerwingithub: Sure `details` sounds good to me.

scalyr_agent/monitor_utils/journald_checkpoints.py
czerwingithub: I hear you about wanting to reuse it on other plugins.  However, until we actually have multiple plugins using it, I suggest we don't put it in its own file.  Essentially, don't try to generalize until we actually have multiple uses.

scalyr_agent/monitor_utils/journald_checkpoints.py
czerwingithub: Yup, I noticed the lock being held for the writes as well... and yes, there's not as easy of a way to avoid holding it there IMO.  I prefer to not hold locks during I/O but sometimes it's just too hard to avoid it.. and yes, at least this is just on the write path so should not get much contention.

imron:
Ok, I've incorporated all those changes, done basic smoke-testing and also added documentation.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -0,0 +1,379 @@
...
+            try:
+                msg = entry.get('MESSAGE', '')
+                extra = self._get_extra_fields( entry )
+                self._logger.emit_value('details', msg, extra_fields=extra)
+                self._last_cursor = entry.get('__CURSOR', None)
+            except Exception, e:
+                global_log.warn( "Error getting journal entries: %s" % str(e),
+                                 limit_once_per_x_secs=60, limit_key='journald-entry-error' )
+
+    def gather_sample(self):
```
czerwingithub: Ok, after reading this in the documentation, it just seems too weird to overload `sample_interval` for polling journald entries.  How about you just create a `journal_poll_interval` config option.  Then, just invoke the monitor's `set_sample_interval` during the configuration?  Then document here that we just rely on `gather_sample` being invoked once ever `journal_poll_interval`.

Sorry, you'll have to update the documentation, but I think this makes more sense for now.

imron:
Ok, I've updated the documentation and monitor, and have rebased against the current master.

PR #100, opened 2018-06-20
Imron/garbage monitor

imron:
Calling `gc.collect()` in the main agent loop based on a new config option: `garbage_collection_interval`.

This defaults to 300 seconds (5 minutes).  Set it to <= `config_change_check_interval` to collect garbage every iteration of the main loop.  Set it to <= 0 to disable collection.

The collecter will run exactly every `garbage_collect_interval` seconds.  Rather it runs on the next iteration of the main loop after the `garbage_collect_interval` has expired.

For example, if the `config_change_check_interval` is set to 10 and the `garbage_collect_interval` is set to 15, then it will take at least 20 seconds before garbage collection is performed - the first iteration occurs at t+10 and `garbage_collect_interval` hasn't expired, then the garbage collect interval expires at t+15 but the main loop is still blocked.  When the main loop wakes at t+20 it notices that the garbage collect interval has expired and so performs the collection.  The next collection will occur at t+30.

Also, the `garbage_monitor` has a new flag `disable_garbage_collection_before_dump` that disables garbage collection before doing a dump of garbage objects.

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -34,6 +34,7 @@
 __author__ = 'czerwin@scalyr.com'
 
 import errno
+import gc
 import os
 import sys
 import time
@@ -733,6 +734,9 @@ class ScalyrAgent(object):
 
                 config_change_check_interval = self.__config.config_change_check_interval
 
+                gc_interval = self.__config.garbage_collect_interval
+                next_gc_time = current_time + gc_interval
+
                 while not self.__run_state.sleep_but_awaken_if_stopped( config_change_check_interval ):
 
                     current_time = time.time()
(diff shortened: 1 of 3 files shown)


imron:
Added the ability to monitor live objects.

Unfortunately there doesn't seem to be a way to hot reload monitors for an already running agent so you won't be able to monitor live objects without stopping and starting the agent.

scalyr_agent/agent_main.py
```diff
@@ -790,6 +794,12 @@ def __run(self, controller):
                             # Update the debug_level based on the new config.. we always update it.
                             self.__update_debug_log_level(new_config.debug_level)
 
+                        # see if we need to perform a garbage collection
+                        if gc_interval > 0 and current_time > next_gc_time:
```
czerwingithub: Why not just write the second clause as `current_time > last_gc_time + gc_interval`.  If you do that, then you do not need to do the little bit of odd dance you do below to make sure you properly update the `next_gc_time` based on the new `gc_interval`.

imron: Good idea!  Sometimes you miss the forest because of all the trees.  Update coming shortly.

imron:
Ok, that change above is now in.

PR #103, opened 2018-07-10
K8s monitor - Add option to filter containers by namespace name

Eyjafjallajokull:
Added new option `namespace_globs` to Kubernetes monitor. It will allow users to stream logs only from containers running in matching namespaces. In case where Kubernetes cluster runs multiple workloads and only subset of them should be integrated with Scalyr.

Let me know if there is something to improve in this patch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ xunit.xml
 # Sphinx documentation
 docs/_build/
 
+docker/root
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -111,7 +111,12 @@ define_config_option( __monitor__, 'metrics_only',
                      convert_to=bool, default=False)
 
 define_config_option( __monitor__, 'container_globs',
-                     'Optional (defaults to None). If true, a list of glob patterns for container names.  Only containers whose names '
+                     'Optional (defaults to None). If true, a list of glob patterns for container names. Only containers whose names '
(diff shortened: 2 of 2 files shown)


imron:
Thanks Eyjafjallajokull, I'll take a look at this shortly.

As an aside, we are currently also working on a feature whereby you'll be able to exclude pods from being monitored by the agent by specifying an annotation in the pod's yaml file. e.g. you'd specify something like:

annotations:
  config.agent.scalyr.com/exclude: true

And then that pod will be excluded from the agent.

I imagine that could also solve the problem you mentioned above, except you'd need to add that annotation to all the pods you wanted excluded, which might be more hassle than just being able to exclude/include entire namespaces.

Eyjafjallajokull:
Thanks for reply. In my case option to include selected containers (via `include` label) and exclude all other would be more suitable.

imron:
Ok, so just to make sure I understand your use case, if you have the ability to include specific containers in a pod's yaml file, and all other containers are ignored then you no longer need filtering by namespace?

Eyjafjallajokull:
Yes, that feature would be sufficient for me.

imron:
It will hopefully land in the main repo some time over the next few days, and then will take a week or more after that to make it in to an official release.

Eyjafjallajokull:
It's now possible to use the following pod lables:
```
log.config.scalyr.com/exclude: true
```
or
```
log.config.scalyr.com/include: false
```
And scalyr agent option:
```
k8s_include_all_containers: false
```
https://www.scalyr.com/help/install-agent-kubernetes#furtherReading

PR #104, opened 2018-07-31
added config item check_remote_if_no_tty

imron:
If this config option is False, and stdout is not a tty, then the agent will not ping the remote scalyr servers when it starts


imron:
Updated so that the command line setting now overrides the config file setting.

PR #105, opened 2018-09-04
Imron/k8s annotations

imron:
This pull request enables log file configuration using kubernetes annotations.

It processes all annotations, and for any annotations that start with the prefix `config.agent.scalyr.com/` it extracts the entries (minus the prefix) and maps them to a dict which is later applied to the log_config used by the log files processed by the k8s monitor.  See below for mapping behaviour.

Supported fields are:

* parser
* attributes
* sampling_rules
* rename_logfile
* redaction_rules

Which behave in the same way as specified in the main [Scalyr help docs](https://www.scalyr.com/help/scalyr-agent?teamToken=y_9E6DfWhCq%2FXYr7DN2zCw--#logUpload).  Items that do not work as specified in help are

* exclude (see below)
* lineGroupers (not supported at all)
* path (the path is always fixed for k8s container logs)

Excluding Logs
------

Containers and pods can be specifically included/excluded from logging.  Unlike the normal log_config `exclude` option which takes an array of log path exclusion globs, annotations simply support a Boolean true/false for a given container/pod.  Both `include` and `exclude` are supported, with `include` always overriding `exclude` if both are set. e.g.

    config.agent.scalyr.com/exclude: true

has the same effect as

    config.agent.scalyr.com/include: false

By default the agent monitors the logs of all pods/containers, and you have to manually exclude pods/containers you don't want.  You can also set a value in agent.json `k8s_include_all_containers: false`, in which case all containers are excluded by default and have to be manually included.

Mapping of Annotation Strings to Config Options
--------

Items separated by a period are mapped to dict keys e.g. if the annotation was specified as:

      config.agent.scalyr.com/attributes.parser: accessLog

It would be mapped to a dict

    result = {
        "attributes": {
            "parser": "accessLog"
        }
    }

Arrays can be specified by using one or more digits as the key, e.g. if the annotation was

      config.agent.scalyr.com/sampling_rules.0.match_expression: INFO
      config.agent.scalyr.com/sampling_rules.0.sampling_rate: 0.1
      config.agent.scalyr.com/sampling_rules.1.match_expression: FINE
      config.agent.scalyr.com/sampling_rules.1.sampling_rate: 0

This will be mapped to the following structure:

    result = {
        "sampling_rules": [
            {
                "match_expression": "INFO",
                "sampling_rate": 0.1
            },
            {
                "match_expression": "FINE",
                "sampling_rate": 0
            }
        ]
    }

Array keys are sorted by numeric order before processing and unique objects need to have different digits as the array key. If a sub-key has an identical array key as a previously seen sub-key, then the previous value of the sub-key is overwritten

There is no guarantee about the order of processing for items with the same numeric array key, so if the config was specified as:

      config.agent.scalyr.com/sampling_rules.0.match_expression: INFO
      config.agent.scalyr.com/sampling_rules.0.match_expression: FINE

It is not defined or guaranteed what the actual value will be (INFO or FINE).

Applying config options to specific containers in a pod
-------

If a pod has multiple containers and you only want to apply log configuration options to a specific container you can do so by prefixing the option with the container name, e.g. if you had a pod with two containers `nginx` and `helper1` and you wanted to exclude `helper1` logs you could specify the following annotation:

    config.agent.scalyr.com/helper1.exclude: true

Config items specified without a container name are applied to all containers in the pod, but container specific settings will override pod-level options, e.g. in this example:

    config.agent.scalyr.com/exclude: true
    config.agent.scalyr.com/nginx.include: true

All containers in the pod would be excluded *except* for the nginx container which is included.

This technique is applicable for all log_config options, not just include/exclude, so for example you could set line sampling rules for all containers in a pod, but use a different set of line sampling rules for one specific container in the pod if needed.

Dynamic Updates
------

Currently all annotation config options except `exclude: true`/`include: false` can be dynamically updated using the `kubectl annotate` command.

For `exclude: true`/`include: false` once a pod/container has started being logged, then while the container is still running, there is currently no way to dynamically start/stop logging of that container using annotations without updating the config yaml, and applying the updated config to the cluster.

Documentation is still to come, and will be included during the code review stage.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -1,4 +1,4 @@
-# Copyright 2014 Scalyr Inc.
+# Copyright 2018 Scalyr Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,14 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ------------------------------------------------------------------------
-# author:  Imron Alston <imron@imralsoftware.com>
+# author:  Imron Alston <imron@scalyr.com>
 
-__author__ = 'imron@imralsoftware.com'
+__author__ = 'imron@scalyr.com'
 
 import datetime
 import docker
 import fnmatch
-import hashlib
 import traceback
 import logging
 import os
@@ -41,7 +40,7 @@ from scalyr_agent.log_watcher import LogWatcher
 from scalyr_agent.monitor_utils.server_processors import LineRequestParser
 from scalyr_agent.monitor_utils.server_processors import RequestSizeExceeded
 from scalyr_agent.monitor_utils.server_processors import RequestStream
-from scalyr_agent.monitor_utils.k8s import KubernetesApi
+from scalyr_agent.monitor_utils.k8s import KubernetesApi, KubernetesCache, PodInfo
 
 from scalyr_agent.util import StoppableThread
 
@@ -118,6 +117,28 @@ define_config_option( __monitor__, 'container_globs',
 define_config_option( __monitor__, 'report_container_metrics',
                       'Optional (defaults to True). If true, metrics will be collected from the container and reported  '
                       'to Scalyr.', convert_to=bool, default=True)
+
+define_config_option( __monitor__, 'k8s_include_all_containers',
+                      'Optional (defaults to True). If True, all containers in all pods will be monitored by the kubernetes monitor '
+                      'unless they have an include: false or exclude: true annotation. '
+                      'If false, only pods/containers with an include:true or exclude:false annotation '
+                      'will be monitored. See documentation on annotations for further detail.', convert_to=bool, default=True)
+
+define_config_option( __monitor__, 'k8s_cache_expiry_secs',
+                     'Optional (defaults to 120). The amount of time to wait between fully updating the k8s cache from the k8s api. '
+                     'Increase this value if you want less network traffic from querying the k8s api.  Decrease this value if you '
+                     'want dynamic updates to annotation configuration values to be processed more quickly.',
+                     convert_to=int, default=120)
+
+define_config_option( __monitor__, 'k8s_max_cache_misses',
+                     'Optional (defaults to 20). The maximum amount of single query k8s cache misses that can occur in `k8s_cache_miss_interval` '
+                     'secs before performing a full update of the k8s cache',
+                     convert_to=int, default=20)
+
+define_config_option( __monitor__, 'k8s_cache_miss_interval',
+                     'Optional (defaults to 10). The number of seconds that `k8s_max_cache_misses` cache misses can occur in before '
+                     'performing a full update of the k8s cache',
+                     convert_to=int, default=10)
 
 define_config_option( __monitor__, 'verify_k8s_api_queries',
                       'Optional (defaults to True). If true, then the ssl connection for all queries to the k8s API will be verified using '
@@ -306,7 +327,8 @@ def _ignore_old_dead_container( container, created_before=None ):
     return False
 
 def _get_containers(client, ignore_container=None, restrict_to_container=None, logger=None,
-                    only_running_containers=True, running_or_created_after=None, glob_list=None, include_log_path=False, include_k8s_info=False):
+                    only_running_containers=True, running_or_created_after=None, glob_list=None, include_log_path=False, k8s_cache=None,
+                    k8s_include_by_default=True, current_time=None):
     """Queries the Docker API and returns a dict of running containers that maps container id to container name, and other info
         @param client: A docker.Client object
         @param ignore_container: String, a single container id to exclude from the results (useful for ignoring the scalyr_agent container)
@@ -318,7 +340,10 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
             dead containers that were created after the specified time.  Used to pick up short-lived containers.
         @param glob_list: String.  A glob string that limit results to containers whose container names match the glob
         @param include_log_path: Boolean.  If true include the path to the raw log file on disk as part of the extra info mapped to the container id.
-        @param include_k8s_info: Boolean.  If true include k8s information (if it exists) as part of the extra info mapped to the container id
+        @param k8s_cache: KubernetesCache.  If not None, k8s information (if it exists) for the container will be added as part of the extra info mapped to the container id
+        @param k8s_include_by_default: Boolean.  If True, then all k8s containers are included by default, unless an include/exclude annotation excludes them.
+            If False, then all k8s containers are excluded by default, unless an include/exclude annotation includes them.
+        @param current_time.  Timestamp since the epoch
     """
     if logger is None:
         logger = global_log
@@ -366,7 +391,7 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
                     log_path = None
                     k8s_info = None
                     status = None
-                    if include_log_path or include_k8s_info:
+                    if include_log_path or k8s_cache is not None:
                         try:
                             info = client.inspect_container( cid )
                             log_path = info['LogPath'] if include_log_path and 'LogPath' in info else None
@@ -374,7 +399,7 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
                             if not only_running_containers:
                                 status = info['State']['Status']
 
-                            if include_k8s_info:
+                            if k8s_cache is not None:
                                 config = info.get('Config', {} )
                                 labels = config.get( 'Labels', {} )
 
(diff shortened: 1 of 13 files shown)


imron:
Just a quick note to say I decided to change the annotation prefix from `config.agent.scalyr.com/` to `log.config.scalyr.com`.  This is so we can have separate annotations `agent.config.scalyr.com/` for agent specific config changes that shouldn't be handled as log configuration.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -17,6 +22,624 @@ class K8sApiException( Exception ):
...
+        md5.update( namespace )
+        md5.update( uid )
+        md5.update( node_name )
+
+        # flatten the labels dict in to a single string
+        flattened = []
+        for k,v in labels.iteritems():
+            flattened.append( k )
+            flattened.append( v )
+        md5.update( ''.join( flattened ) )
```
czerwingithub: Why not just move the `md5.update` up into the loop?  It seems a little unnecessary to create flattened.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -17,6 +22,624 @@ class K8sApiException( Exception ):
...
+        # any of the pod info has changed
+        md5 = hashlib.md5()
+        md5.update( name )
+        md5.update( namespace )
+        md5.update( uid )
+        md5.update( node_name )
+
+        # flatten the labels dict in to a single string
+        flattened = []
+        for k,v in labels.iteritems():
```
czerwingithub: Is `iteritems` guaranteed to return the items in a deterministic order?  that impacts the hash.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -17,6 +22,624 @@ class K8sApiException( Exception ):
...
+        # flatten the container names
+        md5.update( ''.join( container_names ) )
+
+        # flatten the annotations dict in to a single string
+        flattened = []
+        for k,v in annotations.iteritems():
+            flattened.append( k )
+            flattened.append( str(v) )
+
+        md5.update( ''.join( flattened ) )
```
czerwingithub: Same here about `flattened`.. seems like we can skip creating the string.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -17,6 +22,624 @@ class K8sApiException( Exception ):
...
+            flattened.append( k )
+            flattened.append( v )
+        md5.update( ''.join( flattened ) )
+
+        # flatten the container names
+        md5.update( ''.join( container_names ) )
+
+        # flatten the annotations dict in to a single string
+        flattened = []
+        for k,v in annotations.iteritems():
```
czerwingithub: Same here about the deterministic order of `iteritems`.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -127,9 +785,37 @@ def query_pod( self, namespace, pod ):
 
     def query_pods( self, namespace=None, filter=None ):
         """Wrapper to query all pods in a namespace, or across the entire cluster"""
-        query = '/api/v1/pods'
+        return self.query_objects( '/api/v1/pods', '/api/v1/namespaces/%s/pods', namespace, filter )
```
czerwingithub: If seems like namespace=None would be bad here.  Maybe we shouldn't have it default to None in the argument list?

scalyr_agent/copying_manager.py
```diff
@@ -208,6 +208,8 @@ def __init__(self, configuration, monitors):
         # log file has been processed yet
         self.__logs_pending_removal = {}
 
+        # a list of log_matches for logs with configs that need to be reloaded
```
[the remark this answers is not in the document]
imron: Done.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -119,6 +119,28 @@
...
+                      'Optional (defaults to True). If True, all containers in all pods will be monitored by the kubernetes monitor '
+                      'unless they have an include: false or exclude: true annotation. '
+                      'If false, only pods/containers with an include:true or exclude:false annotation '
+                      'will be monitored. See documentation on annotations for further detail.', convert_to=bool, default=True)
+
+define_config_option( __monitor__, 'k8s_cache_expiry_secs',
+                     'Optional (defaults to 120). The amount of time to wait between fully updating the k8s cache from the k8s api. '
+                     'Increase this value if you want less network traffic from querying the k8s api.  Decrease this value if you '
+                     'want dynamic updates to annotation configuration values to be processed more quickly.',
+                     convert_to=int, default=120)
```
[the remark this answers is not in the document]
imron: Done

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -852,22 +827,38 @@ def __get_base_attributes( self ):
...
         result = None
 
         container_attributes = base_attributes.copy()
         container_attributes['containerName'] = info['name']
         container_attributes['containerId'] = cid
         parser = 'docker'
+        common_annotations = {}
+        container_annotations = {}
+        pod_name = '--'
+        pod_namespace = '--'
```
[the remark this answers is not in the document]
imron: This is used as a marker of a clearly invalid name.  It is for the case where a log message might be generated that uses the pod name and namespace, but no pod name or namespace exists (because of an error querying the API).  I have added a comment to this effect in the source.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -877,25 +868,106 @@ def __get_log_config_for_container( self, cid, info, k8s_data, base_attributes )
...
+        invalid_keys = [ 'path', 'lineGroupers' ];
+
+        # set config items, ignoring invalid options and taking care to
+        # handle attributes
+        for key, value in annotations.iteritems():
+            if key in skip_keys:
+                continue
 
+            if key in invalid_keys:
+                self._logger.warning( "Invalid key '%s' found in annotation config for '%s/%s'. Configuration of '%s' is not currently supported via annotations" % (key, pod_namespace, pod_name, key ),
```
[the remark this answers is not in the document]
imron: Done

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1271,20 +1341,20 @@ def __log_metrics( self, container, metrics_to_emit, metrics, extra=None ):
                 # for each running container.
                 self._logger.emit_value( key, metrics[value], extra, monitor_id_override=container )
 
-    def __log_network_interface_metrics( self, container, metrics, interface=None ):
-        extra = {}
+    def __log_network_interface_metrics( self, container, metrics, interface=None, k8s_extra={} ):
```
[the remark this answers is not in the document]
imron: Done.  Same as for later methods.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1362,6 +1457,11 @@ def gather_sample( self ):
                             extra = { 'pod_uid': pod.uid,
                                       'pod_namespace': pod.namespace,
                                       'node_name': pod.node_name }
+
+                            deployment_info = self.__build_k8s_deployment_info( k8s_cache, pod )
```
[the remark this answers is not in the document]
imron: One is for querying the deployment information based on the container info, and one is for querying the deployment information based on the pod information (obtained after querying the container information).  The updated method comments now explain this briefly.

scalyr_agent/monitor_utils/k8s.py
imron: Update requires a string.  If you just call str( labels ) then if labels is a JsonObject you get the object id and type rather than a key/value textual representation of the object.

Flattening them out like this ensures the md5 is based on the contents of the key/value strings.

There is no guarantee of ordering in a dict, but for the same set of keys it seems to be consistent, at least within a single running instance.

scalyr_agent/monitor_utils/k8s.py
imron: See previous comment

scalyr_agent/monitor_utils/k8s.py
imron: See previous comment

scalyr_agent/monitor_utils/k8s.py
imron: See previous comment

scalyr_agent/monitor_utils/k8s.py
imron: No, None is ok here and if None, then the entire cluster is queried rather than just the namespace.  This function is just a convenience wrapper that passes in the appropriate urls.  The None checking is performed in query_objects, rather than duplicating the exact same logic in each of these query_*s methods.  I've updated the comments of the methods to mention this.

PR #106, opened 2018-09-06
Imron/performance update

imron:
This set of commits contains work from the performance improvement research done about a year ago, updated to merge cleanly with the current master, as well as new changes to disable json parsing of k8s logs on the client.

New configuration options
------

* k8s_parse_json - Defaults to True.  An option of the kubernetes_monitor.  If True log files will be parsed as json to extract the log message and timestamp.  If False, the raw json will be uploaded as the log message.  This option can be turned on to avoid json parsing on the client.

* minimum_scan_interval - Defaults to `None`.  This can be specified as either a global or a log_config option, and the log_config option will override the global one if specified.  This option is a value in seconds that controls how often to scan a log file for new bytes.   If not specified, logs are scanned every iteration of the main loop.  If you have a large number of low frequency logs and a handful of high frequency logs, you can set the minimum_scan_interval for the low frequency logs to a high value (e.g. 10-15 seconds or even longer).  That way the agent is mostly processing the high frequency logs, and only looks at the low frequency logs every `minimum_scan_interval` seconds.

* full_checkpoint_interval_in_seconds.  Defaults to 60 seconds, a global option specified in agent.json (or agent.d/*) that controls how often a full checkpoint is written to disk.  Previously, a full checkpoint was written at the end of each main loop.  Now each loop it will only write checkpoints for active files (i.e. those that have logged data), and then every `full_checkpoint_interval_in_seconds` it will write the full set of checkpoints.

Testing
------

To test performance improvements.  Firstly I would run the agent with this set of patches applied.  There should be some performance improvement gains in terms of reduced CPU usage.

After checking those results. I would then test with k8s_parse_json set as `False` to see if further gains can be had by not parsing json logs on the client.

Finally if the node has a large number of containers running on it, most of which log with a low frequency log, but one or two that log with a high frequency, I would try looking at configuring the minimum_scan_interval for low frequency logs.  This might be tricky to configure for individual logs under k8s due to needing to create `log` sections in the agent.d/* files.  Probably the best way to do this for testing purposes would be to set it at 5-10 seconds as a global option (which will apply to all log files) and then have a specific `log` rules to match the container id of high frequency logs, and for those logs setting the value to 0.  Setting this value will become much easier once #105 is merged, which allows setting log_config values in the k8s pod yaml files.  If a node mostly has containers that log with high frequency and only a handful that log with low frequency, it is probably not worth setting this option.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -119,6 +119,11 @@ define_config_option( __monitor__, 'report_container_metrics',
                       'Optional (defaults to True). If true, metrics will be collected from the container and reported  '
                       'to Scalyr.', convert_to=bool, default=True)
 
+define_config_option( __monitor__, 'k8s_parse_json',
+                      'Optional (defaults to True). If True, the log files will be parsed as json before uploading to the server '
+                      'to extract log and timestamp fields.  If False, the raw json will be uploaded to Scalyr.',
+                      convert_to=bool, default=True)
+
 define_config_option( __monitor__, 'verify_k8s_api_queries',
                       'Optional (defaults to True). If true, then the ssl connection for all queries to the k8s API will be verified using '
                       'the ca.crt certificate found in the service account directory. If false, no verification will be performed. '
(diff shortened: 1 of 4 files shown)


czerwingithub:
I rebased this and merged in through a separate PR (#109).

imron:
Thanks.  Hopefully it was a clean rebase and didn't involve much cleaning up.

czerwingithub:
Nope, just one line.

PR #107, opened 2018-09-14
k8s metrics

imron:
Added network metrics for k8s


echee2:
@imron, I'm just going through older PRs for cleanup -- is there any code in this PR that needs to be folded in? Otherwise we can close this?

imron:
I don't think so and it can probably be closed.  This is probably a hangover from when I did development on my local fork, and Steve merged in from his local fork.

PR #110, opened 2018-10-10
Allow unspecified Journal Paths

lowjoel:
According to the documentation for the journal.Reader class
https://www.freedesktop.org/software/systemd/python-systemd/journal.html#systemd.journal.Reader
the path flag is optional. If the path is set to `None` it defaults to the system journal.

### DIFF
--- a/scalyr_agent/builtin_monitors/journald_monitor.py
+++ b/scalyr_agent/builtin_monitors/journald_monitor.py
@@ -198,6 +198,8 @@ class JournaldMonitor(ScalyrMonitor):
 
     def _initialize(self):
         self._journal_path = self._config.get( 'journal_path' )
+        if len(self._journal_path) == 0:
+            self._journal_path = None
         if self._journal_path and not os.path.exists( self._journal_path ):
             raise BadMonitorConfiguration( "journal_path '%s' does not exist or is not a directory" % self._journal_path, 'journal_path' )
 
--- a/scalyr_agent/builtin_monitors/journald_monitor.py
+++ b/scalyr_agent/builtin_monitors/journald_monitor.py
@@ -198,7 +198,7 @@ class JournaldMonitor(ScalyrMonitor):
 
     def _initialize(self):
         self._journal_path = self._config.get( 'journal_path' )
-        if not os.path.exists( self._journal_path ):
+        if self._journal_path and not os.path.exists( self._journal_path ):
             raise BadMonitorConfiguration( "journal_path '%s' does not exist or is not a directory" % self._journal_path, 'journal_path' )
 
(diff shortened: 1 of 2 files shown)


lowjoel:
Currently the Scalyr agent is not booting for us out of the box:
```
Oct 10 03:18:41 systemd[1]: Starting LSB: Manages the Scalyr Agent 2, which provides log copying...
Oct 10 03:18:41 scalyr-agent-2[18232]: Error reading configuration file: journal_path '/var/log/journal' does not exist or is not a directory
Oct 10 03:18:41 scalyr-agent-2[18232]: Terminating agent, please fix the configuration file and restart agent.
Oct 10 03:18:41 systemd[1]: scalyr-agent-2.service: Control process exited, code=exited status=1
Oct 10 03:18:41 systemd[1]: Failed to start LSB: Manages the Scalyr Agent 2, which provides log copying.
Oct 10 03:18:41 systemd[1]: scalyr-agent-2.service: Unit entered failed state.
Oct 10 03:18:41 systemd[1]: scalyr-agent-2.service: Failed with result 'exit-code'.
```

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -198,7 +198,7 @@ class JournaldMonitor(ScalyrMonitor):
 
     def _initialize(self):
         self._journal_path = self._config.get( 'journal_path' )
-        if not os.path.exists( self._journal_path ):
+        if self._journal_path and not os.path.exists( self._journal_path ):
```
czerwingithub: Hmm.. there's currently no way that I can see of specifying `None` for the journal path via the configuration file (since it defaults to `/var/log/journald`).  It can be empty string (which would at least let your condition pass) but I think we need to pass in None to the library to be safe.  How about we add a line above this that says:

    if len(self._journal_path) == 0:
        self._journal_path = None

Then just change the documentation for the `journal_path` configuration option to indicate empty string will use the system library.

The other choice, of course, is to change the default of `journal_path` to be None.  However, I'm not a big fan of changing default values in case customers depend on the current defaults.

lowjoel: Your suggestion sounds good. Let me change the code.

lowjoel:
@czerwingithub any chance of merging this and doing a release?

czerwingithub:
@lowjoel My apologies, I thought we would be releasing the agent earlier this week.  However, we hit a few snags with some other changes which delayed the release.  We should be on schedule to push the new release which will include your change on Thursday Oct 18.

BTW, thanks very much for submitting this fix.  We definitely appreciate your contribution here.

PR #131, opened 2019-01-17
Ticket #AGENT-32

imron:
Forces https for the scalyr server, unless the `--force-http option` is
passed on the command line.

The code for this has been placed in the Configuration object, to ensure
that anywhere that queries the scalyr_server from the config will be
given the https version of the url.

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -213,7 +213,7 @@ class ScalyrAgent(object):
         self.__config_file_path = config_file_path
 
         try:
-            self.__config = self.__read_and_verify_config(config_file_path, force_http=command_options.force_http )
+            self.__config = self.__read_and_verify_config(config_file_path)
 
             # check if not a tty and override the no check remote variable
             if not sys.stdout.isatty():
@@ -275,24 +275,21 @@ class ScalyrAgent(object):
                 raise Exception('Caught exception when attempt to execute command %s.  Exception was %s' %
                                 (command, str(e)))
 
-    def __read_and_verify_config(self, config_file_path, force_http ):
+    def __read_and_verify_config(self, config_file_path):
         """Reads the configuration and verifies it can be successfully parsed including the monitors existing and
         having valid configurations.
 
         @param config_file_path: The path to read the configuration from.
         @type config_file_path: str
 
-        @param force_http: Whether or not to force the scalyr server to use 'http'
-        @type force_http: boolean
-
         @return: The configuration object.
         @rtype: scalyr_agent.Configuration
         """
-        config = self.__read_config(config_file_path, force_http )
+        config = self.__read_config(config_file_path)
         self.__verify_config(config)
         return config
 
-    def __read_config(self, config_file_path, force_http ):
+    def __read_config(self, config_file_path):
         """Reads the configuration from the file path but does not do a full verification.
 
         This will only throw an error if the files themselves cannot be read.  This includes the configuration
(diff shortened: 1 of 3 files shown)


scalyr_agent/agent_main.py
```diff
@@ -213,7 +213,7 @@ def main(self, config_file_path, command, command_options):
         self.__config_file_path = config_file_path
 
         try:
-            self.__config = self.__read_and_verify_config(config_file_path)
+            self.__config = self.__read_and_verify_config(config_file_path, force_http=command_options.force_http )
```
czerwingithub: Ok, sorry to ask for a variable rename, but let's call this `allow_http` instead of `force_http`.  It should be a configuration option that we can have in agent.json.

scalyr_agent/agent_main.py
```diff
@@ -275,21 +275,24 @@ def main(self, config_file_path, command, command_options):
                 raise Exception('Caught exception when attempt to execute command %s.  Exception was %s' %
                                 (command, str(e)))
 
-    def __read_and_verify_config(self, config_file_path):
+    def __read_and_verify_config(self, config_file_path, force_http ):
```
czerwingithub: I don't think we need to support this from the commandline, so you can probably just get rid of this here. (by it I mean passing in the `force_http` variable)

scalyr_agent/agent_main.py
```diff
@@ -1214,6 +1220,9 @@ def stop(self):
                       help="For status command, prints detailed information about running agent.")
     parser.add_option("", "--no-fork", action="store_true", dest="no_fork", default=False,
                       help="For the run command, does not fork the program to the background.")
+    parser.add_option("", "--force-http", action="store_true", dest="force_http", default=False,
```
czerwingithub: No need to for commandline option.  This should just be an option in the configuration.

scalyr_agent/configuration.py
```diff
@@ -153,6 +156,21 @@ def parse(self, logger=None):
...
+                parts = urlparse.urlparse( server )
+                if not parts.scheme:
+                    https_server = 'https://' + server
+                elif parts.scheme == 'http':
+                    https_server = re.sub( "^http://", "https://", server )
+
+                if https_server != server:
+                    self.__config['scalyr_server'] = https_server
+                    if logger:
+                        logger.info( "Forcing https protocol for server url: %s -> %s.  You can override this with the --force-http flag, but be mindful that there are security implications with doing this, including tramsitting your Scalyr api key over an insecure connection." % (server, self.__config['scalyr_server'] ) )
```
czerwingithub: If we put this logging statement here, I think it will be emitted everytime we check to see if the configuration file has changed.  Correct?  Maybe instead we just have the Configuration object remember the `raw_server_host` and have a check in the ScalyrClient that sees if the raw host is diferent than the one we are using.. if we so, we emit the warning when we init the ScalyrClient.  Or, we just have a boolean that remembers if we forced it to secure or not.  You get the idea.

scalyr_agent/agent_main.py
imron: No problem.

scalyr_agent/agent_main.py
imron: Done.

scalyr_agent/configuration.py
imron: Ok I've changed this.  The check now happens in the main agent loop.  It needs to happen there because the `ScalyrClient` doesn't have a reference to the config object.

The log message is only output if the server and the raw server are different, *and* the server is different from value the previous time through the loop.

This ensures that the message will be printed once, unless scalyr_server is changed in the config file, and then it will print it again (once) if the value is different from the raw value.

scalyr_agent/configuration.py
```diff
@@ -153,6 +154,21 @@ def parse(self, logger=None):
             self.__verify_or_set_optional_string(self.__config, 'scalyr_server', 'https://agent.scalyr.com',
                                                  'configuration file %s' % self.__file_path)
 
+            self.__config['raw_scalyr_server'] = self.__config['scalyr_server']
```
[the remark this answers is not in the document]
imron: Unit tests added.

PR #135, opened 2019-02-11
allow event collection from remote machines

imron:
Please make sure to use the updated win32evtlog.pyd binary on the build machine (sent via email), otherwise the agent will crash when an attempt is made to read events from a remote machine.

### DIFF
--- a/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
+++ b/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
@@ -33,6 +33,7 @@ except ImportError:
     windll = None
 
 from scalyr_agent import ScalyrMonitor, define_config_option
+import scalyr_agent.scalyr_logging as scalyr_logging
 
 __author__ = 'imron@imralsoftware.com'
 
@@ -76,6 +77,18 @@ define_config_option(__monitor__, 'server_name',
                      'Optional (defaults to ``localhost``). The remote server where the event log is to be opened\n',
                      default='localhost', convert_to=str)
 
+define_config_option(__monitor__, 'remote_user',
+                     'Optional (defaults to ``None``). The username to use for authentication on the remote server.  This option is only valid on Windows Vista and above\n',
+                     default=None, convert_to=str)
+
+define_config_option(__monitor__, 'remote_password',
+                     'Optional (defaults to ``None``). The password to use for authentication on the remote server.  This option is only valid on Windows Vista and above\n',
+                     default=None, convert_to=str)
+
+define_config_option(__monitor__, 'remote_domain',
+                     'Optional (defaults to ``None``). The domain to which the remote user account belongs.  This option is only valid on Windows Vista and above\n',
+                     default=None, convert_to=str)
+
 class Api( object ):
     def __init__( self, config, logger ):
         self._checkpoints = {}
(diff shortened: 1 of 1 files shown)


imron:
Pinging @czerwingithub.  This is the commit needed to fix querying events on remote machines.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -232,7 +246,29 @@ def __init__( self, config, logger, channels ):
...
+        # see if we need to create a remote connection
+        if self._server is not None and self._server != 'localhost':
+            username = config.get( 'remote_user' )
+            password = config.get( 'remote_password' )
+            domain = config.get( 'remote_domain' )
+            flags = win32evtlog.EvtRpcLoginAuthDefault
+
+            # login object is a tuple
+            login = (self._server, username, domain, password, flags )
+            self._logger.log( scalyr_logging.DEBUG_LEVEL_0, "Performing remote login: server - %s, user - %s, domain - %s" % (self._server, username, domain) )
```
czerwingithub: DEBUG_LEVEL_0 is the same level as INFO... let's do DEBUG_LEVEL_1 here.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -232,7 +246,29 @@ def __init__( self, config, logger, channels ):
 
         self.__bookmarks = {}
 
-    def load_checkpoints( self, checkpoints ):
+    def load_checkpoints( self, checkpoints, config ):
+
+        # see if we need to create a remote connection
+        if self._server is not None and self._server != 'localhost':
```
czerwingithub: It seems to me connecting to a remote host when we are loading the checkpoint file is odd.  Why is this necessary?

At the very least, let's make another method called `open_remote_session_if_necessary` that sets `self._session` with the opened session.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
imron: Done.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
imron: For the old event log API, we needed to poll events.  `load_checkpoints` loads a set of checkpoints and then that is used later on when calling `read_event_log`.

For the newer event log API, there is no mechanism to poll events.  Instead, you subscribe to an event channel that calls a user supplied callback function for each new event.

For the code that handles the newer api, the subscription happens in `load_checkpoints` because we subscribe from a specific checkpoint (or `bookmark`).

If you want to query events from a remote machine, then you need to pass in a remote session handle when setting up the event subscriptions, and because this is only needed in the `load_checkpoints` code, that's where it was put.

I've updated the code to shift things in to separate functions, but the calls still happen inside the `load_checkpoints`, mainly because `session` needs to be in a member variable so that it is persistent for the lifetime of the object, but we don't populate it in the `__init__` method for this class, because the `__init__` method gets called in the monitor's `_initialize` method, whereas `load_checkpoints` is only called once when the monitor starts running.

imron:
Ready for code review.

PR #137, opened 2019-02-13
Ticket: AGENT-44

imron:
Don't add logs dynamically if a log path or glob already exist that match the
path.

The purpose of this is so that log entries specified in `agent.json` or
`agent.d/*.json` can override the parser and other attributes of dynamically
added logs.

This is a temporary solution.  The real solution is to allow people to override
log config options in docker labels, similar to how it works with k8s
annotations.

### DIFF
--- a/scalyr_agent/copying_manager.py
+++ b/scalyr_agent/copying_manager.py
@@ -281,14 +281,13 @@ class CopyingManager(StoppableThread, LogWatcher):
         """
         return self.__log_matchers
 
-    def __path_in_all_paths( self, path, all_paths ):
-        """ Returns the key in `all_paths` if `path` is in `all_paths` or if `path` matches any globs in `all_paths` """
-        result = None
-        for p in all_paths.keys():
+    def __path_in_globs( self, path, path_globs ):
+        """ Returns the first glob pattern in 'path_globs' that matches 'path' if one exists."""
+        for p in path_globs:
             if fnmatch.fnmatch( path, p ):
-                result = p
-                break
-        return result
+                return p
+
+        return None
 
     def add_log_config( self, monitor, log_config ):
         """Add the log_config item to the list of paths being watched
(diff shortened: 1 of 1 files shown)


scalyr_agent/copying_manager.py
```diff
@@ -280,6 +281,15 @@ def log_matchers(self):
         """
         return self.__log_matchers
 
+    def __path_in_all_paths( self, path, all_paths ):
+        """ Returns the key in `all_paths` if `path` is in `all_paths` or if `path` matches any globs in `all_paths` """
+        result = None
+        for p in all_paths.keys():
+            if fnmatch.fnmatch( path, p ):
+                result = p
```
czerwingithub: Just return p here?

scalyr_agent/copying_manager.py
```diff
@@ -289,12 +299,16 @@ def add_log_config( self, monitor, log_config ):
...
+            path = self.__path_in_all_paths( log_config['path'], self.__all_paths )
+            if path is None:
                 log.log(scalyr_logging.DEBUG_LEVEL_0, 'Adding new log file \'%s\' for monitor \'%s\'' % (log_config['path'], monitor.module_name ) )
                 self.__pending_log_matchers.append( LogMatcher( self.__config, log_config ) )
                 self.__all_paths[log_config['path']] = 1
+                path = log_config['path']
             else:
-                self.__all_paths[log_config['path']] += 1
+                log.log(scalyr_logging.DEBUG_LEVEL_0, 'New log file \'%s\' already matches \'%s\' for monitor \'%s\'' % (log_config['path'], path, monitor.module_name ) )
+                if path == log_config['path']:
```
czerwingithub: Add comment as to why you check whether or not `path` is equal to `log_config['path']`

scalyr_agent/copying_manager.py
```diff
@@ -280,6 +281,15 @@ def log_matchers(self):
         """
         return self.__log_matchers
 
+    def __path_in_all_paths( self, path, all_paths ):
+        """ Returns the key in `all_paths` if `path` is in `all_paths` or if `path` matches any globs in `all_paths` """
```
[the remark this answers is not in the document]
imron: Done.  I've also renamed the function to __path_in_globs

scalyr_agent/copying_manager.py
imron: Done.

scalyr_agent/copying_manager.py
imron: Done.

imron:
Code review changes have been pushed.

PR #138, opened 2019-03-05
Imron/events

imron:
Here is the initial version of the event leader election, ready for code review.

I've fixed the problem where it wasn't switching to a new leader, and the leader can now change dynamically.

There are still debugging statements left in at the info level - I will switch them to debug level later - maybe flag them in review to make sure I don't forget.

This currently uses the existing k8s cache.  I will merge the new cache code with this later today.  It will be only a few lines change from this code and won't effect the leader election logic at all.

You may need to update the code that adds deployment info - I'm not sure how that changed on your end.

You'll also need to update the service-account - there was an extra endpoint I needed to add permission for.


scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+
+            Check to see if the leader is specified as a node label.
+            If not, then if there is not a current leader node, or if the current leader node no longer exists
+            then query all nodes for the leader
+        """
+
+        try:
+            leader = None
+            # first check to see if the leader node is specified as a label
+            # we do this first in case the label has been moved to a different node but the old node is still
```
czerwingithub: In this model, you are allowing the label to be on multiple nodes and the leader election picks the oldest of all of those nodes, right?

The drawback is I believe you have to always query on that label to determine if the set of nodes with the label has changed.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+            If there is currently a leader node, then only that node is polled to see if it still
+            exists.
+
+            Check to see if the leader is specified as a node label.
+            If not, then if there is not a current leader node, or if the current leader node no longer exists
+            then query all nodes for the leader
+        """
+
+        try:
+            leader = None
```
czerwingithub: Let's change this to `new_leader`

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+            # first check to see if the leader node is specified as a label
+            # we do this first in case the label has been moved to a different node but the old node is still
+            # alive
+            if self._check_labels:
+                query = "?labelSelector=%s" % (urllib.quote( 'agent.config.scalyr.com/events_leader=true' ) )
+                leader = self._check_nodes_for_leader( k8s, query )
+
+            # if no labelled leader was found, then check to see if the previous leader is still alive
+            if leader is None:
+                if self._check_if_alive( k8s, self._current_leader ):
```
czerwingithub: Do a check here to see if `_current_leader` is none.  I understand that the method might check that itself, but it still feels better to do the check before the method is invoked.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+            leader = None
+            # first check to see if the leader node is specified as a label
+            # we do this first in case the label has been moved to a different node but the old node is still
+            # alive
+            if self._check_labels:
+                query = "?labelSelector=%s" % (urllib.quote( 'agent.config.scalyr.com/events_leader=true' ) )
+                leader = self._check_nodes_for_leader( k8s, query )
+
+            # if no labelled leader was found, then check to see if the previous leader is still alive
+            if leader is None:
```
czerwingithub: Make this condition `if leader is not None` so that you can move the `leader set from label` up here.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+                            json = json_lib.parse(line)
+                        except Exception, e:
+                            global_log.warning( "Error parsing event json: %s, %s, %s" % (line, str(e), traceback.format_exc() ) )
+                            continue
+
+                        obj = json.get( "object", JsonObject() )
+
+                        # check to see if the resource version we are using has expired
+                        if json.get( 'type', '' ) == 'ERROR':
+                            if obj.get( 'kind', '' ) == 'Status' and obj.get( 'reason', '' ) == 'Expired':
```
czerwingithub: Right a method called `is_resource_expired` (or something like that) and use it here.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+                        resource_version = metadata.get_int( "resourceVersion", None, none_if_missing=True )
+                        if resource_version and resource_version <= last_resource:
+                            global_log.log( scalyr_logging.DEBUG_LEVEL_2, "Skipping older resource events" )
+                            continue
+
+                        last_resource = resource_version
+                        last_event = resource_version
+
+                        # see if this event is about an object we are interested in
+                        involved = obj.get( "involvedObject", JsonObject())
```
czerwingithub: How about write a method called `parse_involved` (or something like that) that returns `kind, name, namespace` and use it here.  This method has a bit too many lines and I think that gets in the place of readability.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+                        kind = involved.get( 'kind', None, none_if_missing=True )
+                        if kind and kind not in self.__event_object_filter:
+                            global_log.log( scalyr_logging.DEBUG_LEVEL_1, "Ignoring event due to unknown kind %s - %s" % (kind, str(metadata) ) )
+                            continue
+
+                        # get cluster and deployment information
+                        k8s_fields = "cluster-name=%s" % cluster_name
+                        if kind:
```
czerwingithub: If kind is None, should we log the event at all?

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+                            name = involved.get( 'name', None, none_if_missing=True )
+                            namespace = involved.get( 'namespace', None, none_if_missing=True )
```
czerwingithub: Shouldn't we also be checking to see if this namespace is not part of the ignored list?

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Yes, that's correct, but before doing that, I'm also sorting nodes by node name first in case there are two equally old nodes, in which case it will always return a consistent result.

For the second point, I'm not sure that's a drawback.  

Label filtering/selecting is an integral part of k8s.  According to [the docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) label selectors are "the core grouping primitive in Kubernetes" so I would imagine this kind of query is going to be acceptably performant - even for large clusters.

I imagine the nodes that have this label set will be limited to a small handful.  The smallest case 1, but probably 2-3 just for fallback purposes.

When this function is called the `leader_check_interval` has expired.  This means we need to be hitting the api anyway firstly to determine if the `current_leader` is still alive, and if not, hitting it again to query *all the nodes*.

So instead of hitting the api to check if a single node is alive, we instead first query to get a short list of alive nodes that have this label attached.

In a best case scenario we are doing the same amount of calls to the API (because if we get nodes back, we don't have to check the alive status of the previously current node, or query the entire list of nodes in the cluster).

In the worse case scenario (no labels set and the current node is dead) we are doing one more API call (query on labels - which fails, query on current node - which fails, then query on all nodes), however a worst case scenario is predicated on configuring the monitor to use labels (which is off by default) and then not setting any labels.

The last possibility is the case where the previous leader node goes down, then assuming labels have been set on a currently running node, querying by this label is likely going to be more efficient, because it avoids having to then query and process all the nodes.

czerwingithub: Ok, we should make sure we call this out in the documentation that the number of nodes with the `could be leader` label should be small.  Also, let's rename that label.  `events_leader=true` makes it look like it is the event leader, where it could just be a candidate.  How about `events_leader_candidate=true`.  Kind of long, but not sure which if we really want to remove `events` here.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+                     'Increase this value if you want less network traffic from querying the k8s api.  Decrease this value if you '
+                     'want dynamic updates to annotation configuration values to be processed more quickly.',
+                     convert_to=int, default=30)
+
+define_config_option( __monitor__, 'k8s_cache_purge_secs',
+                     'Optional (defaults to 300). The number of seconds to wait before purging unused items from the k8s cache',
+                     convert_to=int, default=300)
+
+define_config_option( __monitor__, 'k8s_ignore_namespaces',
+                      'Optional (defaults to "kube-system"). A comma-delimited list of the namespaces whose pods\'s '
```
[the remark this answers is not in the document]
imron: It's all going global once the cache update is merged so it will apply to both.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+        """
+            Checks to see if the node specified by `node` is alive
+        """
+        if node is None:
+            return False
+
+        result = {}
+        try:
+            # this call will throw an exception on failure
+            result = k8s.query_api( '/api/v1/nodes/%s' % node )
```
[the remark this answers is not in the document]
imron: It doesn't guarantee that pods can be scheduled on that node.  The customer might have configured a node with various taints to prevent pods running on it.  I think this will come down to the problem we discussed earlier about whether or not the agent is running on a particular node, and the solution being to manually specify the event leader node in that case.

The node could also potentially be shutting down (so no more scheduling allowed) in which case it will be picked up the next iteration through the loop so it's probably ok.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+            if create_time is not None and create_time < oldest_time:
+                oldest_time = create_time
+                oldest = name
+
+        return oldest
+
+    def _check_nodes_for_leader( self, k8s, query_fields='' ):
+        """
+            Checks all nodes on the cluster to see which one has the oldest creationTime
+            If `ignore_master` is True then any node with the label `node-role.kubernetes.io/master`
```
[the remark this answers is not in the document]
imron: Done.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -0,0 +1,478 @@
...
+        nodes = response.get( 'items', [] )
+
+        # sort the list by node name to ensure consistent processing order
+        # if two nodes have exactly the same creation time
+        nodes.sort( key=lambda n: n.get( 'metadata', {} ).get( 'name', '' ) )
+
+        # the leader is the longest running node
+        return self._get_oldest_node( nodes, self._ignore_master )
+
+    def _get_current_leader( self, k8s ):
```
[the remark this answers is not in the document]
imron: Done.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Done

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Done.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Done.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Done

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Done.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Probably not.  Kind shold always be specified, so will early `continue` if None.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: good catch

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Done.

PR #141, opened 2019-03-11
Ticket #AGENT-52

imron:
Added EventID to log messages.

Also fixed duplicate logs if config changes

### DIFF
--- a/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
+++ b/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
@@ -302,12 +302,20 @@ class NewApi( Api ):
 
         result = {}
 
-        # build the EventId
+        # In the new event log api, EventIds were replaced by an InstanceId.
+        # The InstanceID is made by combining the old EventId with any
+        # SystemQualifiers associated with the event, to create a new 32bit value
+        # with the EventId in the lower 16bits and the SystemQualifiers
+        # in the high 16bits.
         event_id_val = vals[win32evtlog.EvtSystemEventID]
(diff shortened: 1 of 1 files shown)


scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -294,8 +302,14 @@ def GetFormattedEventAsDict( self, render_context, event ):
...
 
-        event_id = vals[win32evtlog.EvtSystemEventID]
-        qualifiers = vals[win32evtlog.EvtSystemQualifiers]
+        # build the EventId
+        event_id_val = vals[win32evtlog.EvtSystemEventID]
+        if event_id_val[1] != win32evtlog.EvtVarTypeNull:
+            event_id = event_id_val[0]
+            qualifiers_val = vals[win32evtlog.EvtSystemQualifiers]
+            if qualifiers_val[1] != win32evtlog.EvtVarTypeNull:
+                event_id = win32api.MAKELONG( event_id, qualifiers_val[0] )
```
czerwingithub: Can you add some comments explaining what is going on here?

imron: Done.  If you'd prefer, we could also split this in to 3 separate fields:

1. EventID
2. SystemQualifiers
3. InstanceID

(see comments in the new commit to see how these are related).

czerwingithub: If we strongly thought folks would want to search on those independently, then it would make sense. I don't know if we have strong data one way or the other, so let's just keep it simple for now.

PR #143, opened 2019-03-11
Ticket #AGENT-14

imron:
The postgres monitor was keeping a connection and cursor open per instance of
the monitor.  By default statements executed by a connection are executed
inside of a transaction.

To ensure consistency, all database statistics remain constant inside a single
transaction and so when we were querying the statistics because it was still
in the same transaction, the same values were being returned.

I've fixed this by opening and closing the db connection each gather_sample.

### DIFF
--- a/scalyr_agent/builtin_monitors/postgres_monitor.py
+++ b/scalyr_agent/builtin_monitors/postgres_monitor.py
@@ -413,4 +413,10 @@ instance."""
                             self._logger.emit_value(self._db._database_stats[table][key][0], dbstats[key], extra)
                         else:
                             self._logger.emit_value(self._db._database_stats[table][key][0], timestamp_ms(dbstats[key]))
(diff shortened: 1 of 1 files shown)


scalyr_agent/builtin_monitors/postgres_monitor.py
```diff
@@ -401,3 +413,4 @@ def timestamp_ms(dt):
                             self._logger.emit_value(self._db._database_stats[table][key][0], dbstats[key], extra)
                         else:
                             self._logger.emit_value(self._db._database_stats[table][key][0], timestamp_ms(dbstats[key]))
+        self._db.close()
```
czerwingithub: Add in a comment explaining why you close this.  We don't want the bug to sneak back in .

imron: Done

PR #145, opened 2019-03-18
Ticket #AGENT-6, #AGENT-47, #AGENT-51

imron:
This commit adds support for docker labels to the docker monitor and the
syslog_monitor.

The following config options have been added:

* `labels_as_attributes` - defaults to false.  When true, labels will be
  uploaded as log attributes if the pass the `label_include_globs` and
  `label_exclude_globs` filters.

* `label_prefix` - defaults to ''.  A prefix to add to the key of any labels
  uploaded as log attributes.

* `label_include_globs` - a list of globs that determines which labels to
  include as attributes.  If a label matches any of the globs on the list then
  it will be marked for inclusion.  Note: `label_exclude_logs` is processed
  *after* this option, so the label might still be excluded.  Default [ '*' ]

* `label_exclude_globs` - a list of globs that is used to exclude specific
  labels from inclusion as log attributes.  If a label matches any of the
  globs on this list then it is excluded.  This option is applied *after*
  `label_include_globs`.  Default [ 'com.scalyr.config.*' ]

* `use_labels_for_log_config` - Defaults to True.  When True any label
  starting with com.scalyr.config.log is used to configure the log config for
  the container.  This follows the same rules and the annotation configuration
  for k8s, except that docker keys cannot have underscores in them, and so any
  config keys that contain hyphens are converted to use underscores instead,
  e.g. `com.scalyr.config.log.rename-logfile` will be converted to a log config
  key `rename_logfile`.

These configuration options are placed in the config for the docker monitor.
If the syslog monitor is being used in docker mode, then it will read these
configurations from the docker monitor, or use the default values if the
docker monitor isn't running.


scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -124,6 +128,35 @@
                       'Optional (defaults to True). If true, metrics will be collected from the container and reported  '
                       'to Scalyr.', convert_to=bool, default=True)
 
+define_config_option( __monitor__, 'label_include_globs',
+                     'Optional (defaults to [\'*\']). If `labels_as_attributes` is True then this option is a list of glob strings used to '
+                     'include labels that should be uploadeded as log attributes.  The docker monitor first gets all container labels that '
+                     'match a glob in this list and then filters out any labels that match a glob in `label_exclude_globs`, and the final list is then '
```
czerwingithub: I would replace with `match any glob in this list`

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -124,6 +128,35 @@
...
+                     'uploaded as log attributes. ',
+                      default=['*'])
+
+define_config_option( __monitor__, 'label_exclude_globs',
+                     'Optional (defaults to [\'com.scalyr.config.*\']). If `labels_as_attributes` is True, then this is a list of glob strings used to '
+                     'exclude labels from being uploaded as log attributes.  Any label whose key matches a glob on this list will not be added as a '
```
czerwingithub: I would again say `matches any glob`

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -310,6 +349,50 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
 
     return result
 
+def get_attributes_and_config_from_labels( labels, docker_options ):
+    """
+        Takes a dict of labels and splits it in to two separate attributes and config dicts.
+
+        @param: labels - a dict containing all the labels
```
czerwingithub: I would say `a dict containing all the Docker labels for a container`

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -596,14 +692,23 @@ def __get_last_request_for_log( self, path ):
...
+        if 'parser' not in result and 'parser' not in attributes:
+            result['parser'] = parser
+
+        result['path'] = path
+        result['parse_lines_as_json'] = parse_as_json
+
+        if 'attributes' in result:
+            result['attributes'].update( attributes )
+        else:
+            result['attributes'] = JsonObject( attributes )
```
czerwingithub: Why does this have to be a `JsonObject`?  Can you add a comment.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -596,14 +692,23 @@ def __get_last_request_for_log( self, path ):
 
         return scalyr_util.seconds_since_epoch( result )
 
-    def __create_log_config( self, parser, path, attributes, parse_as_json=False ):
+    def __create_log_config( self, parser, path, attributes, base_config={}, parse_as_json=False ):
         """Convenience function to create a log_config dict from the parameters"""
```
czerwingithub: Can you add in some documentation for the arguments.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -1074,6 +1279,19 @@ def _initialize( self ):
 
         metrics_only = self._config.get('metrics_only')
 
+        self.label_exclude_globs = self._config.get( 'label_exclude_globs' )
+        self.label_include_globs = self._config.get( 'label_include_globs' )
+
+        if not scalyr_util.is_list_of_strings( self.label_include_globs):
```
czerwingithub: Why not move this into some sort of validate method on `DockerOptions`?

scalyr_agent/monitor_utils/annotation_config.py
```diff
@@ -174,6 +173,13 @@ def sort_numeric( pair ):
         if _is_int( key ):
             return int( key )
         return key
+
+    def key_name( key, convert_hyphens ):
```
czerwingithub: How about `docker_key_name` and document what this method does and why.

scalyr_agent/scalyr_monitor.py
```diff
@@ -200,6 +200,17 @@ def increment_counter(self, reported_lines=0, errors=0):
         self.__errors += errors
         self.__lock.release()
 
+    def monitor_manager( self, manager ):
```
czerwingithub: Let's call this `config_from_monitors`.  At least it will better describe what this is used for.

scalyr_agent/builtin_monitors/docker_monitor.py
imron: The only reason is that validating a list of strings is something that I'm sure I've used elsewhere in the code (but I couldn't find which monitor), and is likely to be used in other config options in other monitors in the future.

Happy to move it in to DockerOptions for now, and move it back out if it turns out to be needed elsewhere later.

scalyr_agent/builtin_monitors/docker_monitor.py
imron: Done.

scalyr_agent/builtin_monitors/docker_monitor.py
imron: Done

scalyr_agent/builtin_monitors/docker_monitor.py
imron: Done

scalyr_agent/builtin_monitors/docker_monitor.py
imron: Done.

scalyr_agent/builtin_monitors/docker_monitor.py
imron: Done

scalyr_agent/builtin_monitors/docker_monitor.py
imron: Actually, upon further inspection, the call to this is happening in the `_initialize` method before any DockerOptions are created (they are created in the ContainerChecker at a different point in the code, so I'm not sure if DockerOptions is the best place to put it - otherwise we'll have to start passing another docker_options option around.

Currently `_initialize` ensures that the config items are valid, and then later the ContainerChecker is passed the config object and it reads the previously validated values and creates the DockerOptions.

I think leaving it as is for the moment is probably a better option - though happy to discuss (or have you override).

scalyr_agent/monitor_utils/annotation_config.py
imron: `docker_key_name` is probably not appropriate as this method is also used by the k8s annotations.  Maybe `normalize_key_name` as it is used to ensure keys use the correct form - on k8s they will already contain underscores and on docker it will convert hyphens to underscores.  The function doesn't know beforehand whether it is called by docker or k8s, and relies on the `convert_hyphens` param to decide whether to convert hyphens or not.

scalyr_agent/scalyr_monitor.py
imron: Done

imron:
Code review changes pushed.

imron:
Now with documentation for docker labels

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -693,7 +693,12 @@ def __get_last_request_for_log( self, path ):
         return scalyr_util.seconds_since_epoch( result )
 
     def __create_log_config( self, parser, path, attributes, base_config={}, parse_as_json=False ):
-        """Convenience function to create a log_config dict from the parameters"""
+        """Convenience function to create a log_config dict
+        @param: parser - string - a parser to use if no parser is found in the attributes or base_config
```
czerwingithub: You should use the `@type` jsdoc tag to label the parameter types.  Here and the next few as well.

czerwingithub: Call this `default_parser` as well since it is only used if there is none specified in `attributes` or `base_config`

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -1165,9 +1176,9 @@ def __init__(self, labels_as_attributes=False, label_prefix='', label_include_gl
         """
         @param: labels_as_attributes - boolean - if True any labels that are not excluded will be added to the attributes result dict
```
czerwingithub: Sorry, i didn't notice the types being slipped in here the first time.  Just move the parameter types to `@type` tags.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -1234,6 +1245,120 @@ class DockerMonitor( ScalyrMonitor ):
     It can also collect all log messages written to stdout and stderr by those containers, in conjunction with syslog.
     See the online documentation for more details.
 
+    ## Docker Labels
+
+    You can get the Scalyr agent to upload custom attributes for your containers by setting labels on the container, and then configuring the agent to upload labels as attributes.  This is controlled by a number of config options:
```
czerwingithub: `You can configure the Scalyr Agent to upload customer attributes for your containers based on the labels you set on the container itself.  This can be used to easily set the parser that should be used to parse the container's log, as well as adding in arbitrary labels on the container's log.

To use this functionality, you must properly configure the agent by setting the `labels_as_attributes` configuration option to `true`.  All of the docker monitor configuration options related to this feature are as follows:`

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -1234,6 +1245,120 @@ class DockerMonitor( ScalyrMonitor ):
...
+
+    Note: Log attributes contribute towards your total log volume, so it is wise to limit the labels to a small set of approved values, to avoid paying for attributes that you don't need.
+
+    ## Using Docker Labels for Configuration
+
+    You can also use docker labels to configure the log settings for a specific container, such as setting the parser or setting redaction rules.
+
+    The agent takes any label on a container that begins with `com.scalyr.config.log.` and maps it to the corresponding option in the `log_config` stanza for that container's logs (minus the prefix). 
+
+    For example, if you add the following label to your container:
```
czerwingithub: Do they have to turn anything on in the agent to use the configuration-options-from-labels feature?

scalyr_agent/builtin_monitors/docker_monitor.py
czerwingithub: Ok, that sounds fine to me.

scalyr_agent/builtin_monitors/docker_monitor.py
imron: Done.

imron: Done.

scalyr_agent/builtin_monitors/docker_monitor.py
imron: Done.  Will pay more attention to this going forward.

imron:
Ready for code review

PR #147, opened 2019-03-20
k8s events documentation

imron:
Added documentation about node leaders for the kubernetes events monitor

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
@@ -130,10 +130,103 @@ class EventLogFormatter(BaseFormatter):
 
 class KubernetesEventsMonitor( ScalyrMonitor ):
     """
-# Kubernetes Events Monitor
+    # Kubernetes Events Monitor
 
-The Kuberntes Events monitor streams Kubernetes events from the Kubernetes API
+    The Kuberntes Events monitor streams Kubernetes events from the Kubernetes API
 
+    This monitor fetches Kubernetes Events from the Kubernetes API server and uploads them to Scalyr. [Kubernetes Events](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#event-v1-core) record actions taken by the Kubernetes master, such as starting or killing pods. This information can be extremely useful in debugging issues with your Pods and investigating the general health of your K8s cluster. By default, this monitor collects all events related to Pods (including objects that control Pods such as DaemonSets and Deployments) as well as Nodes.
+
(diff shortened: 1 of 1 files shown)


scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -130,10 +130,103 @@ def __init__(self):
...
+
+    The second approach is to manually assign a leader node in the agent config.  This is done via the `leader_node` option of the Kubernetes Events monitor:
+
+    ```
+    ...
+    monitors: [
+      {
+        module: "scalyr_agent.builtin_monitors.kubernetes_events_monitor",
+        ...
+        leader_node: <name of leader node>
```
czerwingithub: Add quotes around the <name of leader node>

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -130,10 +130,103 @@ def __init__(self):
...
+      }
+    ]
+    ...
+    ```
+
+    When the leader node is explicitly set like this, no API queries are made to the K8s API server and only the node with that name will be used to query K8s events.  The downside to this approach is that it is less flexible, especially in the event of the node shutting down unexpectedly.
+
+
+    Since the leader election relies on some assumptions and could impact load on the K8s API server, The leader election algorithm does rely on a few assumptions to work correctly and could, for large clusters, impact performance.
```
czerwingithub: This paragraph needs to be fixed.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Quote around the.......?

I guess <name of the leader node> but in case you also meant the keys, I added them around the keys also.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
imron: Fixed in what way?  Do you mean the extra blank line or is there more explanation needed?

In any case, I've removed the blank line.

imron:
Pushed the above changes.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
czerwingithub: That paragraph starting with `Since the leader election relies on...` looks like it has two similar sentences that were accidentally joined together.  At the very least, the `T` in `The` should be lower-cased.  Also, there's a lot of repetitiveness to the sentence (repeating of `assumptions`, `impact`, `relies`, etc).

imron: Ah sorry, didn't read the full paragraph.  That was just a straight copy/paste from [the existing documentation](https://www.scalyr.com/help/monitors/kubernetes-events).  I've modified it to run nicely in to the next section on disabling the events.

PR #148, opened 2019-03-21
k8s global cache

imron:
Global cache for k8s

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -270,7 +270,7 @@ class Configuration(object):
 
     @property
     def k8s_verify_api_queries(self):
-        return self.__get_config().get_bool('verify_k8s_api_queries')
+        return self.__get_config().get_bool('k8s_verify_api_queries')
 
     @property
     def k8s_cache_expiry_secs(self):
@@ -999,7 +999,7 @@ class Configuration(object):
 
         self.__verify_or_set_optional_string(config, 'k8s_ignore_namespaces', 'kube-system', description, apply_defaults )
         self.__verify_or_set_optional_string(config, 'k8s_api_url', 'https://kubernetes.default', description, apply_defaults )
-        self.__verify_or_set_optional_bool(config, 'verify_k8s_api_queries', True, description, apply_defaults )
+        self.__verify_or_set_optional_bool(config, 'k8s_verify_api_queries', True, description, apply_defaults )
         self.__verify_or_set_optional_int(config, 'k8s_cache_expiry_secs', 30, description, apply_defaults )
         self.__verify_or_set_optional_int(config, 'k8s_cache_purge_secs', 300, description, apply_defaults )
 
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -259,6 +259,27 @@ class Configuration(object):
 
         return monitor_config
 
+    # k8s cache options
+    @property
+    def k8s_ignore_namespaces(self):
+        return self.__get_config().get_string('k8s_ignore_namespaces')
+
+    @property
+    def k8s_api_url(self):
+        return self.__get_config().get_string('k8s_api_url')
+
+    @property
+    def k8s_verify_api_queries(self):
+        return self.__get_config().get_bool('verify_k8s_api_queries')
+
+    @property
+    def k8s_cache_expiry_secs(self):
+        return self.__get_config().get_int('k8s_cache_expiry_secs')
+
+    @property
+    def k8s_cache_purge_secs(self):
+        return self.__get_config().get_int('k8s_cache_purge_secs')
+
     # Debug leak flags
     @property
     def disable_send_requests(self):
@@ -975,6 +996,12 @@ class Configuration(object):
         self.__verify_or_set_optional_bool(config, 'verify_server_certificate', True, description, apply_defaults)
         self.__verify_or_set_optional_string(config, 'http_proxy', None, description, apply_defaults)
         self.__verify_or_set_optional_string(config, 'https_proxy', None, description, apply_defaults)
+
+        self.__verify_or_set_optional_string(config, 'k8s_ignore_namespaces', 'kube-system', description, apply_defaults )
+        self.__verify_or_set_optional_string(config, 'k8s_api_url', 'https://kubernetes.default', description, apply_defaults )
+        self.__verify_or_set_optional_bool(config, 'verify_k8s_api_queries', True, description, apply_defaults )
+        self.__verify_or_set_optional_int(config, 'k8s_cache_expiry_secs', 30, description, apply_defaults )
+        self.__verify_or_set_optional_int(config, 'k8s_cache_purge_secs', 300, description, apply_defaults )
 
         #Debug leak flags
         self.__verify_or_set_optional_bool(config, 'disable_leak_send_requests', False, description, apply_defaults)
--- a/scalyr_agent/monitor_utils/k8s.py
+++ b/scalyr_agent/monitor_utils/k8s.py
@@ -604,6 +604,22 @@ class _CacheConfig( object ):
         self.cache_purge_secs = cache_purge_secs
         self.namespaces_to_ignore = namespaces_to_ignore
 
+    def __eq__( self, other ):
+        """Equivalence method for _CacheConfig objects so == testing works """
+        return (self.api_url == other.api_url and
+                self.verify_api_queries == other.verify_api_queries and
+                self.cache_expiry_secs == other.cache_expiry_secs and
+                self.cache_purge_secs == other.cache_purge_secs and
+                self.namespaces_to_ignore == other.namespaces_to_ignore)
+
+    def __ne__( self, other ):
+        """Non-Equivalence method for _CacheConfig objects because Python 2 doesn't
+        automatically generate != if == is defined
+        """
+        # return result based on negation of `==` rather than negation of `__eq__`
+        return not (self == other)
+
+
     def __str__( self ):
         return "\n\tapi_url: %s\n\tverify_api_queries: %s\n\tcache_expiry_secs: %d\n\tcache_purge_secs: %d\n\tnamespaces_to_ignore: %s\n" % ( str(self.api_url),
             str( self.verify_api_queries), self.cache_expiry_secs, self.cache_purge_secs, str( self.namespaces_to_ignore ) )
@@ -634,14 +650,11 @@ class KubernetesCache( object ):
     def __init__( self, api_url="https://kubernetes.default", verify_api_queries=True, cache_expiry_secs=30, cache_purge_secs=300, namespaces_to_ignore=None, start_caching=True ):
 
         self._lock = threading.Lock()
-        self._namespace_filter = None
         self._node_filter = None
         self._k8s = None
-        self._cache_expiry_secs=cache_expiry_secs
-        self._cache_purge_secs=cache_purge_secs
-        self._config_update_count = 0
+        self._pending_config = None
 
-        empty_config = _CacheConfig( api_url='', namespaces_to_ignore=[] )
+        self._config = _CacheConfig( api_url='', namespaces_to_ignore=[] )
 
 
         new_config = _CacheConfig( api_url=api_url,
@@ -651,7 +664,7 @@ class KubernetesCache( object ):
                                    namespaces_to_ignore=namespaces_to_ignore )
 
         # perform initial configuration of the cache
-        self._configure( old_config=empty_config, new_config=new_config )
+        self._configure( new_config )
 
         # create the controller cache
         self._controller_processor = ControllerProcessor()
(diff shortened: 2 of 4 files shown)


scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
+
+        self._lock = threading.Lock()
+        self._namespace_filter = None
+        self._node_filter = None
+        self._k8s = None
+        self._cache_expiry_secs=cache_expiry_secs
+        self._cache_purge_secs=cache_purge_secs
+        self._config_update_count = 0
+
+        empty_config = _CacheConfig( api_url='', namespaces_to_ignore=[] )
```
czerwingithub: You should just set `self._config` to this value.  See discussion below, but I don't think you need to be passing in `old_config` to the `_configure` method.  You should just rely on `self._config`.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
 class KubernetesCache( object ):
 
-    def __init__( self, k8s, logger, cache_expiry_secs=30, cache_purge_secs=300, namespaces_to_ignore=None ):
+    def __init__( self, api_url="https://kubernetes.default", verify_api_queries=True, cache_expiry_secs=30, cache_purge_secs=300, namespaces_to_ignore=None, start_caching=True ):
+
+        self._lock = threading.Lock()
+        self._namespace_filter = None
+        self._node_filter = None
+        self._k8s = None
+        self._cache_expiry_secs=cache_expiry_secs
```
czerwingithub: You should just read these values (`cache_expiry_secs` and `cache_purge_secs`) from `self._config`.  Having them in two places can lead to bugs.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
-        self._lock = threading.Lock()
         self._initialized = False
 
-        self._thread = StoppableThread( target=self.update_cache, name="K8S Cache" )
+        self._thread = None
+
+        if start_caching:
+            self.start()
+
+    def _configure( self, old_config, new_config ):
```
czerwingithub: I don't think you need to be passing in `old_config` here.  Just use what is in `self._config` as the current config.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
-    def __init__( self, k8s, logger, cache_expiry_secs=30, cache_purge_secs=300, namespaces_to_ignore=None ):
+    def __init__( self, api_url="https://kubernetes.default", verify_api_queries=True, cache_expiry_secs=30, cache_purge_secs=300, namespaces_to_ignore=None, start_caching=True ):
+
+        self._lock = threading.Lock()
+        self._namespace_filter = None
+        self._node_filter = None
+        self._k8s = None
+        self._cache_expiry_secs=cache_expiry_secs
+        self._cache_purge_secs=cache_purge_secs
+        self._config_update_count = 0
```
czerwingithub: See discussion below but you should not need this variable.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
+        old_k8s = None
+        old_namespace_filter = None
+        old_node_filter = None
+        old_config_update_count = 0
+        self._lock.acquire()
+        try:
+            old_k8s = self._k8s
+            old_namespace_filter = self._namespace_filter
+            old_node_filter = self._node_filter
+            old_config_update_count = self._config_update_count
```
czerwingithub: I'm pretty sure there is a flaw in using the `config_update_count` approach to detecting if some other thread has updated the cache.  Let's say this happens, two threads are trying to update the cache config at the same time.  Which thread should win?  I would argue that it's the thread that invoked this method last (last caller wins).  Your approach doesn't guarantee that.

It's a little hard to show, but let me try.  Suppose Thread A invokes `_configure` first.  It will get to this critical block first and record `old_config_update_count` as 5.  It then releases the lock to do its work to build up the filters.  Thread B now invokes `_configure`.  It has invoked it after A.  It also gets to the first critical block and records `old_config_update_count` as 5.  It then releases the lock to build the filters.... Well, suppose A then actually updates the cache and increments `_config_update_count` to 6.  B's update to the cache config will fail.. breaking the invariant that I want the latest call to win.

czerwingithub: One way to do this, (and I think it is less complex) is to have another instance variable called `self._pending_config`.  This will end up replacing `self._config_update_count`.  It represents the new configuration that will next be applied to the cache.  This variable is set in this first block when you first have the lock -- thread A will set it to the configuration object it has .. and then thread B will set it to the configuration object it has.  In the second lock block below, you check to see if `self._pending_config` still points to the same configuration object you gave it.  If so, then you are allowed to update the cache.. otherwise, you cannot.

Does that make sense?  Sorry, it's a little hard to explain like this.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
+        finally:
```
czerwingithub: So, here you would set `self._pending_config = new_config`.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
+        changed = False
+
+        # get old state values
+        old_k8s = None
+        old_namespace_filter = None
+        old_node_filter = None
+        old_config_update_count = 0
+        self._lock.acquire()
+        try:
+            old_k8s = self._k8s
```
czerwingithub: I believe there are only two ways we will not return True in this method.  Either, the two configurations are the same so no change is needed... or this thread does not end up having its configuration recognized because some other thread updated the configuration over it.

I think you can simplify your code by checking right here (while you hold the lock) if the new configuration is equal to `self._config`.  If it is, then you can just return False here.  Then.. the only way you could return False is if your thread doesn't win (which we will handle down below).

So, to ease this, you should have an equals method on the CacheConfig object.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
+        """
+        Configures the cache based on any changes in the configuration
+        @param old_config: the old configuration
+        @param new_config: the new configuration
+        @type old_config: _CacheConfig
+        @type new_config: _CacheConfig
+        @return: True if the config changed, False otherwise
+        """
+
+        changed = False
```
czerwingithub: As explained below, I don't think you need this variable.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
+        if old_config.cache_expiry_secs != new_config.cache_expiry_secs or old_config.cache_purge_secs != new_config.cache_purge_secs:
+            changed = True
+
+        # update with new values
+        self._lock.acquire()
+        try:
+            # if config update count has changed, it means a newer config change
+            # came through on another thread before we finished this call, so only make our changes
+            # if the update counts are the same
+            if self._config_update_count != old_config_update_count:
```
czerwingithub: Replace this with `if self._config is not new_config:  return False`.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
+
+                self._cache_expiry_secs = new_config.cache_expiry_secs
+                self._cache_purge_secs = new_config.cache_expiry_secs
+
+                self._config = new_config
+
+                if changed:
+                    global_log.log( scalyr_logging.DEBUG_LEVEL_1, "Got new config %s", str( self._config ) )
+
+                self._config_update_count += 1
```
czerwingithub: Just `return True` here.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
+        finally:
+            self._lock.release()
+
+        return changed
```
czerwingithub: You shouldn't need this final return statement here.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -566,42 +580,209 @@ def process_object( self, obj ):
...
+        self._lock.acquire()
+        try:
+            old_config = self._config
+        finally:
+            self._lock.release()
+
+        changed = self._configure( old_config=old_config, new_config=new_config )
+
+        if changed:
+            global_log.info( "K8s Cache configuration has been updated" )
```
czerwingithub: Maybe this should be a debug log?  I don't know if we really need to log this in normal cases.

scalyr_agent/configuration.py
```diff
@@ -259,6 +259,27 @@ def parse_monitor_config(self, monitor_config, context_description='uncategorize
...
+    def k8s_ignore_namespaces(self):
+        return self.__get_config().get_string('k8s_ignore_namespaces')
+
+    @property
+    def k8s_api_url(self):
+        return self.__get_config().get_string('k8s_api_url')
+
+    @property
+    def k8s_verify_api_queries(self):
+        return self.__get_config().get_bool('verify_k8s_api_queries')
```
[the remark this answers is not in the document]
imron: Done.  Only reason I didn't do this earlier was because it was an existing parameter and I wasn't sure if changing it would break anyone's current setup.

scalyr_agent/monitor_utils/k8s.py
imron: I was in two minds about this.  The reason I did it like this was to differentiate "current state" from "configuration options".

The idea being that threads can happily change the configuration options without affecting state and vice versa, and then having a fixed point at which configurations options are turned in to state, allowing a fixed state throughout the caching operations.

Happy to update this, but perhaps moving all cache state (k8s, filters and expiry) to a separate helper class would make the separation I was aiming for clearer?

scalyr_agent/monitor_utils/k8s.py
imron: Yep makes sense.  That's a better way to do things.

scalyr_agent/monitor_utils/k8s.py
imron: Fair point.  It was drilled in to us in university to try to have single points of return per function and old habits die hard.  This is something I've noticed cropping up in a few different code reviews, so I'll pay more attention to it, but just wanted to provide you with an explanation of why you see me doing this.

scalyr_agent/monitor_utils/k8s.py
imron: Done.

scalyr_agent/monitor_utils/k8s.py
imron: Done.

scalyr_agent/monitor_utils/k8s.py
imron: Done.

scalyr_agent/monitor_utils/k8s.py
imron: That being the case, the only reason we were returning True or False in the _configure method was to output an update.  I've therefore gone through and updated the previous changes to no longer return True/False and just keep the debug output in that function instead.

scalyr_agent/monitor_utils/k8s.py
imron: This came in part from an earlier version of the code and also from wanting to read the values/call functions outside of the lock (e.g. the calls to need_new_k8s_object and need_new_filters).

I've removed this parameter and also moved testing for those values inside the lock, but then acting on them outside of the lock.

scalyr_agent/monitor_utils/k8s.py
imron: See previous comments.

scalyr_agent/monitor_utils/k8s.py
imron: Done.

scalyr_agent/monitor_utils/k8s.py
imron: Done.

scalyr_agent/monitor_utils/k8s.py
imron: Yep.

scalyr_agent/monitor_utils/k8s.py
imron: Done.  Added both __eq__ and __ne__ (because __ne__ is not defined by default based on __eq__).

imron:
These changes are now commited.

scalyr_agent/monitor_utils/k8s.py
czerwingithub: Yeah, I was debating this approach as well.. having a separate class that gets all of the state.  If you think it's a better approach, yes, then please go ahead and do that.

scalyr_agent/monitor_utils/k8s.py
czerwingithub: Thanks.  I can very much understand the notion that you should only return in one place.  And, I'm a big fan of minimizing the number of exit points.  However, I do think it's a fine pattern to do some checks early on in a method and return as necessary.  So, yes, go by the spirit of having a few number of deliberate places where you can exit.

scalyr_agent/monitor_utils/k8s.py
czerwingithub: That's fine... yes, you can remove the return value here and just keep the debug version of this in the function itself.

imron:
Ok, I think I might make that abstraction then, just to make things clearer.

Imron

imron:
I've added the cache configuration state class.

This now handles all the updates based on new configurations, and also allows callers to make a local copy of relevant state, which can be used in a separate thread without worrying about it being changed underneath the caller.

PR #150, opened 2019-03-24
AGENT-40 k8s user-agent info

echee2:
Fixes k8s aspect of AGENT-40, wherein additional `user-agent` fragments are periodically polled from monitors and added to global connection settings.

# Background

Marketing team wants additional insight as to whether events are coming from k8s, docker. Currently, the `user-agent` only contains local host platform info. This PR seeks to refactor the code to enable extraction of said info from monitors. Additionally, a fix is implemented for k8s.

# Changes

1. Refactor the monitor manager/monitors such that `MonitorsManager` periodically polls all agents for additional user-agent "fragments". Fragments are immediately applied to global connection settings. Thereafter, all events from this agent will reflect those fragments.

2. Unless I'm missing something, `MonitorsManager` was not a thread (it merely started up monitors, each running in their own threads). This PR gives `MonitorsManager`  it's own thread. I have tried to think of potential pitfalls such as:
    - what happens if `SystemExit` signal is encountered. 
    - is coordination/barriers needed to control startup flow between this thread and others (I don't think so).
    - pitfalls of catching broad `Exception` (ie if system errors occur and are swallowed? But it seems to me we have to catch broadly when dealing with monitors that may be written by customers).
   Kindly review this section carefully!

3. Added callback mechanism so `MonitorsManager` can modify user agent without needing to know anything about `ScalyrClientSession` (these classes should not know about each other).

4. Helper inner-function to startup `WorkerThread`. Why?
     - to setup aforementioned callback
     - eliminate code duplication and avoid future mistakes where we forget to update startup logic in 2 different places.

5. As discussed, `MonitorsManager` should poll more frequently than on every config change. I thought about each monitor having it's own cadence (for modifying user-agent -- and in future, modifying other kinds of state). However, I decided to keep things simple and minimize knobs. So I fixed polling frequency to 60s. We should document this and maybe advise monitor implementers to performing caching if necessary if this period is too frequent (just as k8s monitor does).

6. Modified `k8s_monitor` to access `/version` endpoint ([as described in AGENT-40](https://scalyr.myjetbrains.com/youtrack/issue/AGENT-40)). 
   - Tried to follow existing `KubernetesCache` code (hope I'm doing it right)
   - user-agent fragment is then extracted from the simple dict that `/version` returns.

# How was this tested?

- Created an EKS cluster in AWS dev account.
- built the k8s docker image with dev code, pointing to `qatesting`
- deployed DaemonSet
- observed user agent in qatesting search UI
image

# TODO

- get initial feedback that above approach is acceptable
- unit tests for `MonitorsManager` thread (in terms of agent lifecycle)
- other unit tests

### DIFF
--- a/scalyr_agent/monitors_manager.py
+++ b/scalyr_agent/monitors_manager.py
@@ -18,11 +18,11 @@
 __author__ = 'czerwin@scalyr.com'
 
 import os
-import threading
 
 from scalyr_agent.agent_status import MonitorManagerStatus
 from scalyr_agent.agent_status import MonitorStatus
 from scalyr_agent.scalyr_monitor import load_monitor_class, ScalyrMonitor
+from scalyr_agent.util import StoppableThread
 
 from __scalyr__ import get_package_root
 
@@ -31,8 +31,12 @@ import scalyr_agent.scalyr_logging as scalyr_logging
 log = scalyr_logging.getLogger(__name__)
 
 
-class MonitorsManager(object):
-    """Maintains the list of currently running ScalyrMonitor instances and allows control to start and stop them."""
+class MonitorsManager(StoppableThread):
+    """Maintains the list of currently running ScalyrMonitor instances and allows control to start and stop them.
+
+    Also periodically polls all monitors for user-agent fragments that are then immediately reflected in all requests
+    sent to Scalyr.
+    """
     def __init__(self, configuration, platform_controller):
         """Initializes the manager.
         @param configuration: The agent configuration that controls what monitors should be run.
@@ -41,6 +45,7 @@ class MonitorsManager(object):
         @type configuration: scalyr_agent.Configuration
         @type platform_controller: scalyr_agent.platform_controller.PlatformController
         """
+        StoppableThread.__init__(self, name='monitor manager thread')
         if configuration.disable_monitors_creation:
             log.log( scalyr_logging.DEBUG_LEVEL_0, "Creation of Scalyr Monitors disabled.  No monitors created." )
             self.__monitors = []
@@ -49,24 +54,18 @@ class MonitorsManager(object):
 
         self.__disable_monitor_threads = configuration.disable_monitor_threads
 
-        # This lock protects __running_monitors.
-        self.__lock = threading.Lock()
-        # The list of monitors.  Each element is a ScalyrMonitor instance.  __lock must be held to modify this var.
         self.__running_monitors = []
+        self.__user_agent_callback = None
+        self._user_agent_refresh_interval = configuration.user_agent_refresh_interval
 
     def find_monitor( self, module_name ):
         """Finds a monitor with a specific name
            @param: module_name - the module name of the monitor to find
            @return: a monitor object if a monitor matches `module_name`, or None
         """
-
-        self.__lock.acquire()
-        try:
-            for monitor in self.__monitors:
-                if monitor.module_name == module_name:
-                    return monitor
-        finally:
-            self.__lock.release()
+        for monitor in self.__monitors:
+            if monitor.module_name == module_name:
+                return monitor
 
         return None
 
@@ -76,25 +75,21 @@ class MonitorsManager(object):
         @return:  The status for all the monitors.
         @rtype: MonitorManagerStatus
         """
-        try:
-            self.__lock.acquire()
-            result = MonitorManagerStatus()
+        result = MonitorManagerStatus()
 
-            for monitor in self.__monitors:
-                if monitor.isAlive():
-                    result.total_alive_monitors += 1
+        for monitor in self.__monitors:
+            if monitor.isAlive():
+                result.total_alive_monitors += 1
 
-                status = MonitorStatus()
-                status.monitor_name = monitor.monitor_name
-                status.reported_lines = monitor.reported_lines()
-                status.errors = monitor.errors()
-                status.is_alive = monitor.isAlive()
+            status = MonitorStatus()
+            status.monitor_name = monitor.monitor_name
+            status.reported_lines = monitor.reported_lines()
+            status.errors = monitor.errors()
+            status.is_alive = monitor.isAlive()
 
-                result.monitors_status.append(status)
+            result.monitors_status.append(status)
 
-            return result
-        finally:
-            self.__lock.release()
+        return result
 
     def start_manager(self):
         """Starts all of the required monitors running.
@@ -111,22 +106,30 @@ class MonitorsManager(object):
                 if self.__disable_monitor_threads:
                     log.log( scalyr_logging.DEBUG_LEVEL_0, "Scalyr Monitors disabled.  Skipping %s" % monitor.monitor_name )
                     continue
-                # Check to see if we can open the metric log.  Maybe we should not silently fail here but instead
-                # fail.
+                # Check to see if we can open the metric log.  Maybe we should not silently fail here but instead fail.
                 if monitor.open_metric_log():
                     monitor.config_from_monitors( self )
                     log.info('Starting monitor %s', monitor.monitor_name)
                     monitor.start()
                     self.__running_monitors.append(monitor)
                 else:
-                    log.warn('Could not start monitor %s because its log cold not be opened', monitor.monitor_name)
+                    log.warn('Could not start monitor %s because its log could not be opened', monitor.monitor_name)
+            # Start the monitor manager thread. Do not wait for all monitor threads to start as some may misbehave
+            if not self.__disable_monitor_threads:
+                log.log(scalyr_logging.DEBUG_LEVEL_0, "Starting Scalyr Monitors manager thread")
+                self.start()
+            else:
+                log.log(scalyr_logging.DEBUG_LEVEL_0, "Scalyr Monitors disabled.  Skipping monitor manager thread")
         except:
             log.exception('Failed to start the monitors due to an exception')
 
-    def stop_manager(self):
+    def stop_manager(self, wait_on_join=True, join_timeout=5):
         """Stops all of the monitors.
 
         This will only return after all the threads for the monitors have been stopped and joined on.
+
+        @param wait_on_join: If True, will block on a join of of the thread running the manager.
+        @param join_timeout: The maximum number of seconds to block for the join.
         """
         # TODO:  Move these try statements out of here.  Let higher layers catch it.
         for monitor in self.__running_monitors:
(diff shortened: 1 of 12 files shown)


echee2:
@czerwingithub, posting this for early feedback in case I'm way off.

echee2:
Need to check regarding situations where multiple instances of a monitor type are created. Specifically the semantics of `get_user_agent_fragment` are to return an additional user-agent fragment per monitor instance (not monitor class).

IIUC customers can run multiple instances of a monitor. E.g. if monitoring multiple mysql databases, we would need one MysqlMonitor for each database. These types of monitors don't need to augment user-agent though. User-agent augmentation is mainly for platform scenarios (docker, k8s).

For docker/k8s, each monitor instance _should_ return a different fragment (assuming it's even possible to have multiple instances of these monitor types). Consequently, current semantics of `get_user_agent_fragment` -- i.e. the fragment is different for each monitor instance -- are appropriate.

echee2:
@czerwingithub, should be ready for review. Please let me know if you think the unit tests provide enough coverage. Thx.

scalyr_agent/tests/monitors_manager_test.py
```diff
@@ -74,6 +83,133 @@ def test_module_with_id(self):
...
+            self.assertEquals(fragments, [test_frag])
+        test_manager.set_user_agent_augment_callback(augment_user_agent)
+
+        # Fake the clock to fast-forward MonitorsManager sleep loop
+        fake_clock = scalyr_util.FakeClock()
+        for obj in [test_manager, patched_monitor_0, unpatched_monitor]:
+            obj._run_state = scalyr_util.RunState(fake_clock=fake_clock)
+
+        # Override Agent Logger to increment counter
+        # We will use this as a simple synchronization barrier to know when the monitors have finally started emitting
```
czerwingithub: I'll suggest an alternative approach here for this multi-thread testing.  I would argue that we don't really care about when the monitors start logging.  What we really care about is when we start getting `fragment_polls`.  So, what we should do it just block until we start getting those.

Basically, create a small thread-safe object that stores how many polls have happened.  Give that object a way of saying `block_until_n_polls` where you specify the number of polls and then a timeout (real-time) of how long you are willing to wait for that happen.  You want the timeout in case the test is broken..  you don't want to be waiting forever.

So, you start the test and block until you get the first poll.  Then, you can also do that loop you have done below where you keep incrementing the fake time.. you just block for n+1 pulls.  If you always specify a real-time timeout, you also don't have to have your check for the real clock.  If the test fails if the poll doesn't happen in 5 seconds of real time.. then it will definitely fail if it is not using a fake clock (because it takes 60 seconds between polls.)

scalyr_agent/tests/monitors_manager_test.py
```diff
@@ -74,6 +83,133 @@ def test_module_with_id(self):
...
+            return test_frag
+        for mon in [patched_monitor_0, patched_monitor_1]:
+            mon.get_user_agent_fragment = mock_get_user_agent_fragment
+            self.assertEquals(mon.get_user_agent_fragment(), test_frag)  # monkey patched
+        self.assertEquals(unpatched_monitor.get_user_agent_fragment(), None)  # not patched
+
+        # Mock the callback (that would normally be invoked on ScalyrClientSession)
+        # Check that the exact fragment returned by the 2 patched monitors are deduplicated.
+        def augment_user_agent(fragments):
+            counter['callback_invocations'] += 1
```
czerwingithub: Don't you have to hold the `counter_lock` here?

scalyr_agent/tests/monitors_manager_test.py
```diff
@@ -74,6 +83,133 @@ def test_module_with_id(self):
...
+
+            def emit_value(self, key, value, extra_fields):
+                counter['patched_monitor_started'] = 1
+
+        for monitor in [patched_monitor_0, patched_monitor_1, unpatched_monitor]:
+            monitor._logger = FakeAgentLogger('fake_agent_logger')
+
+        # Start the manager and wait for monitors to start emitting. The wait should be really quick.
+        test_manager.start_manager()
+        while not counter['patched_monitor_started']:
```
czerwingithub: Do you have to hold the `counter` lock here?

scalyr_agent/monitors_manager.py
```diff
@@ -151,6 +169,46 @@ def stop_manager(self):
...
+
+    def run(self):
+        """Poll monitors for specific state changes (e.g. User-Agent fragment)"""
+
+        prev_user_agent_frags = None
+
+        # noinspection PyBroadException
+        try:
+            while self._run_state.is_running():
+                self.__lock.acquire()
```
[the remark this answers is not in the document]
echee2: Makes sense. Unrelated to this PR but should we also avoid holding locks when querying monitors in `general_status`?

scalyr_agent/monitors_manager.py
```diff
@@ -151,6 +169,46 @@ def stop_manager(self):
...
+                        if frag:
+                            user_agent_frags.add(frag)
+                    if user_agent_frags != prev_user_agent_frags:
+                        self.__user_agent_callback(sorted(user_agent_frags))
+                    prev_user_agent_frags = user_agent_frags
+                except Exception:
+                    log.exception('Monitor manager failed to query monitor %s' % monitor.monitor_name)
+                finally:
+                    self.__lock.release()
+                self._run_state.sleep_but_awaken_if_stopped(MonitorsManager.POLL_MONITORS_SECS)
```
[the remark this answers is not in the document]
echee2: BTW, should we make this a generic interval where monitors are polled for all sorts of information (right now it's only user-agent, but in the future it could be be more). Or do we want the different settings for each type of information pollable from monitors? (If so, we would have to keep track of elapsed time in the manager poll loop and then selectively poll different types of info as appropriate to the individual settings.

scalyr_agent/tests/monitors_manager_test.py
echee2: I think we're ok here as only 1 thread accesses this state (`MonitorsManager`)

echee2: I've moved the state that's accessed by multiple threads into the new `FragmentPolls` class. Hopefully this is easier to read/maintain.

imron:
I wonder if rather than creating separate thread in the MonitorManager, we should just be getting the fragments in the CopyingManager, which already has its own thread and has copies of both the monitors and the client.

Each iteration of the main loop it could check if appropriate time has passed and then query all monitors and update the client as necessary.

Thoughts?

scalyr_agent/monitors_manager.py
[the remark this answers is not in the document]
czerwingithub: I think we spoke about this in person, but yes, we should avoid holding the lock for that as well.  Thanks for catching that.

scalyr_agent/monitors_manager.py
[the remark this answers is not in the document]
czerwingithub: My suggestion is that we don't over-generalize too early (kind of like not optimizing too early).  Let's just code up this specific feature in the way that makes sense for it to be coded up.  So, let's use `user_agent_refresh_interval` since it communicates to the user that it does, etc.  After all, this is the first time in 4 years that we are adding something like this, so it's not clear to me if we will need a generic mechanism for these type of things.

czerwingithub:
We've found the overhead for creating a thread is not that high.   I'd rather keep it in the monitor class since that, from a software engineering point-of-view, is the best thing (the monitor manager class should be the thing looping over the monitor instances.)

scalyr_agent/tests/monitors_manager_test.py
```diff
@@ -74,7 +83,146 @@ def test_module_with_id(self):
...
+            def poll_until_target(self, target_polls, maxwait):
+                """Blocks until n fragment polls are simulated by the MonitorsManager, or an absolute cutoff time.
+                whichever comes first.
+
+                @param target_polls: Number of polls to reach
+                @param maxwait: Time (seconds) to wait for target to be reached
+                @return: True if number of polls reaches target_polls, else False
+                """
+                deadline = time.time() + maxwait
+                while self.count() < target_polls and time.time() < deadline:
```
czerwingithub: Sorry, I should have been a little more clear in my last review.  I think you should be using a `Condition` here to wait until the count has reached `target_polls`.  Take a look at the documentation: 
 https://docs.python.org/2/library/threading.html#condition-objects

Essentially, you can use the `wait` call on the condition to wait until the count has reached the right level.  Essentially, every call that increments the count should notify on the on the condition object.  Then, here, first see if the level has been reached, and if it has not, then invoke `wait` with the `maxwait` time, looping until `target_polls` has been reached or the timeout has occurred.

IMO, you should not need to invoke `advance_time` here.  You really should just have to advance the time to poll interval outside of the loop.. and then wait for the count to catch up.  If you find that doesn't work, let's discuss.

echee2: Hmm, I believe the approach you're mentioning here is exactly what  I tried to implement (after the initial review). I used a condition (instead of lock), then in this loop I wait for the count, up until a maximum time.

The problem I encountered is this: Which thread calls `fake_clock.advance_time`?

Here are the threads I'm aware of:
1. MonitorsManager thread
2. Individual Monitor threads
3. Main (test) thread

The MonitorsManager and individual monitor threads are configured to use the fake clock. They each go to sleep at the end of their main loops and rely on an external thread to advance the fake clock (in order to be awakened).

That leaves the main thread to advance fake clock. So if we use a condition in this here and go to sleep (by calling wait()), nobody is advancing the fake clock, hence MonitorsManager never wakes up to poll and the test always fails.

I then created yet another thread to drive the fake clock, but then realized this loop itself could drive the fake clock. I then realized the condition was overkill and so replaced it with a simple lock. 

Finally, the time.sleep(0.01) isn't absolutely necessary but I put it there just so the testrunner doesn't completely hog 100% CPU in the event of some weird situation where it fails but the maxwait was accidentally set too far out.

I hope this all makes sense. I'm having trouble seeing how I can switch this back to a condition without creating yet new thread which I feel makes this more complex.

czerwingithub: Sorry it took a bit for me to get back to this.  Yup, I understand what you are talking about in terms of the difficulty of advancing the clock just once.  Yes, to fix this the "right" way would involve too much effort.

What's making me cringe a bit about the current approach is that you are doing a busy wait loop.. and just relying on that sleep 0.01 call giving the monitor thread time to wake up and do the poll.

You could change this loop to rely on the condition variable and get rid of the busy wait loop.  The trick is that you have to make sure the monitor thread is asleep before you advance the clock.  It would look something like this:

```
   while self.count() < target_polls and time.time() < deadline
       fake_clock.block_until_n_waiting_threads(1)    # I think you just need the 1 for the monitor thread
       fake_clock.advance_time(increment_by=poll_interval)
       wait_for_poll_increment(timeout=deadline - time.time()
```

The `wait_for_poll_increment` would be implemented using the condition variable approach with the cv on the poll count.

Does that make sense?  I think that would work.  I would advocate taking the time to implement it that way because it will help set precedents in the code base to use things like `block_until_n_waiting_threads` instead of relying on busy wait loops.

echee2: Makes sense. Agree with setting precedents. I'll give it a shot!

echee2: @czerwingithub : I've incorporated your suggestions. Hopefully this LGTY.

echee2:
Thanks for the detailed review @czerwingithub . Since we don't have a framework in place for testing k8s api server endpoints, the existing unit tests missed a bug caused by concurrent code change.  But I caught it when I manually tested the k8s docker image. Here's a screen capture of qatesting showing the user-fragment with k8s version info:
image

PR #151, opened 2019-03-24
allow setting the scalyr server from an env variable

imron:


scalyr_agent/configuration.py
```diff
@@ -149,8 +149,23 @@ def parse(self, logger=None):
                 api_key = env_api_key
 
             self.__set_api_key(self.__config, api_key)
+
+            # if we don't have a scalyr_server from the config, check if `scalyr_server` is defined as an
+            # environment variable
+            if scalyr_server is None:
```
czerwingithub: Why not just write this as `os.environ.get('scalyr_server') or os.environ.get('SCALYR_SERVER')`

imron: Done.

PR #156, opened 2019-03-27
Fix unit tests by including ujson in development

echee2:
IIUC the json lib includes have to account for various situations where json lib may or may not exist. If Json lib doesn't exist, use the fallback `json_lib`.

`scalyr_client_test.py` are currently breaking because of extra space. I think this occurs in the following scenario in development/CI:

- `ujson` library is missing
- `json` lib version adds spaces by default when encoding (to eliminate spaces, you need to add 
```
json.dumps(separators=(',', ':'))
```

The solution, I _think_, is to include ujson in `dev-requirements.txt`. Doing so now results in all tests passing.

Question: if a customer setup mimics the current breaking tests (i.e. customer has `json` installed, there will be extra spaces in the JSON. Does this break anything when it's sent over? (I would not think so?)

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -12,7 +12,6 @@ pysnmp
 requests==2.20.0
 redis==2.10.6
 Twisted==17.9.0
-ujson==1.35
 unittest2==1.1.0
 virtualenv==15.1.0
 webapp2==2.5.2
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -12,6 +12,7 @@ pysnmp
 requests==2.20.0
 redis==2.10.6
 Twisted==17.9.0
+ujson==1.35
 unittest2==1.1.0
 virtualenv==15.1.0
 webapp2==2.5.2
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -75,7 +75,14 @@ except ImportError:
     try:
         import json
         _json_lib = 'json'
-        _json_encode = json.dumps
+
+        def dumps_no_space(*args, **kwargs):
+            """Eliminate spaces by default. Python 2.4 does not support partials."""
+            if 'separators' not in kwargs:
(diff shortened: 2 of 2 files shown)


czerwingithub:
@imron What do you think in terms of the fix?  We could require `ujson` but that would mean there would be slight differences if `ujson` is not available.  The alternative is to use the `json.dumps(separators=...` that Ed points out.  I lean towards that approach myself.

@echee2 No, the extra spaces won't break anything when it's sent over the wire to Scalyr.

imron:
Do the tests all pass when using the (separators=) approach?

That would be my preference also.

echee2:
I'll give the separators approach a go.

echee2:
All tests pass with `separators=` approach. Since `functools.partial` was new in 2.5, I'm rolling my own simple partial. WDYT?

imron:
I think that approach seems fine.

You mention all tests are passing, was this *all* tests or just tests relevant to json?

I only ask because on my machine I just checked out this branch and ran the tests (python2.7 only) and there are 21 failures and 5 errors (vs 22 failures and 5 errors from before your changes).  

And I just wanted to check there wasn't something else about my setting causing all those failures (this is the first time I've run all the tests in a few months 😬)

czerwingithub:
I am seeing more failures beyond the JSON related ones.  In general, we
should work to clean this up.  We need to have faith in our tests and also
it just looks sloppy that we have these failures.  And I say that with the
knowledge that I probably broke a bunch of these myself without realizing.

echee2:
Ah, the cause of discrepancy might be that I was running the `nosetests` command from `.circleci/config.yml' which had `--ignore-files=.*log_processing_test.*`

What command are you guys using to run all tests?

czerwingithub:
In the top-level directory, I run `python run_tests.py`.  There's obviously
been some bit rot.

-Steven

PR #159, opened 2019-04-03
AGENT-40: Docker user-agent fragment

echee2:
# Background

This is a follow up to recently merged PR https://github.com/scalyr/scalyr-agent-2/pull/150.

In this PR, I add the following:

1. user-agent support for docker
     -  returns fragment as `docker=true` if docker_monitor is running but version cannot be determined.
     - no specific support for ECS yet

2. modify user-agent support for k8s by returning `k8s=true` if k8s_monitor is running but version cannot be determined.

3. Unit tests to lock-down expected behavior for what k8s & docker monitors are expected to return.

# Details

In this PR, I add user-agent fragment support for Docker.  As discussed with @imron, there is potential for race condition between manager and monitor.  To handle this as well as potential failure during version API call, I follow a lazy-instantiation "self-healing" approach that's aligned with the rest of the monitor behavior.

The basic idea is once the version is successfully obtained (non-null), it will never be checked again for the lifetime of the agent.  The rationale being that the docker version cannot change without an agent reboot.

The version, once set, is then decorated with extra info (syslog vs docker_api mode, raw vs api) etc. 

Finally, if the version is not set, we still return `docker=true` since it tells us that _some_ docker version is running. Note:  I also piggy back a similar modification to the k8s_monitor in this PR and will add a unit test for that.

# How was this tested?

1. Created docker image using `python build_package.py docker_syslog_builder`
2. Launched docker image with API key pointing to `qatesting`
3. Verified `docker=xxxx` appeared in user agent as follows:

image

Note: I did not manual-test the different config permutations (syslog vs dockerapi-raw vs dockerapi-api).   However, my unit test cover the permutations.

### DIFF
--- a/scalyr_agent/builtin_monitors/docker_monitor.py
+++ b/scalyr_agent/builtin_monitors/docker_monitor.py
@@ -26,26 +26,16 @@ import re
 import random
 import socket
 import stat
-import struct
-import sys
 import time
 import threading
 from scalyr_agent import ScalyrMonitor, define_config_option, define_metric
 import scalyr_agent.util as scalyr_util
-import scalyr_agent.json_lib as json_lib
 import scalyr_agent.scalyr_logging as scalyr_logging
 from scalyr_agent.json_lib import JsonObject, ArrayOfStrings
-from scalyr_agent.json_lib import JsonConversionException, JsonMissingFieldException
-from scalyr_agent.log_watcher import LogWatcher
-from scalyr_agent.monitor_utils.server_processors import LineRequestParser
-from scalyr_agent.monitor_utils.server_processors import RequestSizeExceeded
-from scalyr_agent.monitor_utils.server_processors import RequestStream
 import scalyr_agent.monitor_utils.annotation_config as annotation_config
 from scalyr_agent.scalyr_monitor import BadMonitorConfiguration
 
 from scalyr_agent.util import StoppableThread
-
-from scalyr_agent.util import RunState
 
 from requests.packages.urllib3.exceptions import ProtocolError
 
@@ -102,9 +92,9 @@ define_config_option( __monitor__, 'log_mode',
                      'Optional (defaults to "docker_api"). Determine which method is used to gather logs from the '
                      'local containers. If "docker_api", then this agent will use the docker API to contact the local '
                      'containers and pull logs from them.  If "syslog", then this agent expects the other containers '
-                     'to push logs to this one using the syslog Docker log plugin.  Currently, "syslog" is the '
-                     'preferred method due to bugs/issues found with the docker API.  It is not the default to protect '
-                     'legacy behavior.\n',
+                     'to push logs to this one using the Docker syslog logging driver.  Currently, "syslog" is the '
+                     'preferred method due to bugs/issues found with the docker API (To protect legacy behavior, '
+                     'the default method is "docker_api").',
                      convert_to=str, default="docker_api")
 
 define_config_option( __monitor__, 'docker_raw_logs',
@@ -1534,6 +1524,8 @@ class DockerMonitor( ScalyrMonitor ):
             "throttled_time"
         ])
 
+        self.__version = None
+        self.__version_lock = threading.RLock()
 
     def set_log_watcher( self, log_watcher ):
         """Provides a log_watcher object that monitors can use to add/remove log files
@@ -1624,7 +1616,56 @@ class DockerMonitor( ScalyrMonitor ):
         for cid, info in containers.iteritems():
             self.__gather_metrics_from_api_for_container( info['name'] )
 
+    def get_user_agent_fragment(self):
+        """This method is periodically invoked by a separate (MonitorsManager) thread and must be thread safe."""
+
+        def _version_to_fragment(ver):
+            # Helper function to transform docker version to user-agent fragment
+            if not ver:
+                # version not yet set: return 'docker=yes' to signify docker
+                return 'docker=true'
+            log_mode = self._config.get('log_mode')
+            if log_mode == 'syslog':
+                extra = ''
+            else:
+                extra = '|raw' if self._config.get('docker_raw_logs') else '|api'
+            return 'docker=%s|%s%s' % (ver, log_mode, extra)
+
+        self.__version_lock.acquire()
+        try:
+            return _version_to_fragment(self.__version)
+        finally:
+            self.__version_lock.release()
+
+    def _fetch_and_set_version(self):
+        """Fetch and record the docker system version via the docker API"""
+        ver = None
+        try:
+            ver = self.__client.version().get('Version')
+        except Exception:
+            self._logger.exception('Could not determine Docker system version')
+
+        if not ver:
+            return
+
+        self.__version_lock.acquire()
+        try:
+            self.__version = ver
+        finally:
+            self.__version_lock.release()
+
     def gather_sample( self ):
+        """Besides gathering data, this main-loop method also queries the docker API for version number.
+
+        Due to potential race condition between the MonitorsManager and this manager at start up, we set the version
+        in lazy fashion as follows:  On each gather_sample() the DockerMonitor thread queries the docker API for
+        the version until the first success (In most cases, this means a single query shortly after start up). Once
+        set, it will never be queried again as the docker system version cannot change without necessitating an agent
+        restart.
+        """
+        if not self.__version:
+            self._fetch_and_set_version()
+
         # gather metrics
         if self.__report_container_metrics:
             containers = _get_containers(self.__client, ignore_container=None, glob_list=self.__glob_list )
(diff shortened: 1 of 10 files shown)


echee2:
Question: do you know if the `docker-py` client api is thread-safe?  In other words, do I need to acquire lock before calling `self.__client.<api_call>`??

imron:
I don't know off the top of my head, but I believe it is.  That said, even if it's not, you probably shouldn't acquire a lock before calling out to the network because any timeout or delay in the network could potentially cause everything to stop.

I think a better way to do this would be to call out to the API shortly after the docker.Client is created.  Note, this should be in `ContainerChecker` `start` and not `__init__` because the `__init__` gets called by the monitor's `_initialize` method which runs everytime the agent checks for new configurations (whether there is a new configuratin or not), whereas `start` will only be called once per successful configuration change.

From there set a member variable on the ContainerChecker object that can be read at any time from any thread because it will never be modified.  It safe to have this constant because in order for the docker version to change, the docker daemon needs to restart, and if the daemon restarts then the agent container will also restart.

The only question then becomes is there a data race between when the MonitorManager starts polling monitors for user agent fragments and when the Docker monitor starts and queries the Docker version.  I looked at this briefly and I think there might be? but perhaps needs some further investigation.

imron:
Also, regarding fragments, I guess it would be useful to know which input path the customer has configured - syslog, api or raw_logs.

echee2:
Great points @imron.  Will take all these into account.

echee2:
Some follow up thoughts:

1. IIUC doesn't `ContainerChecker` only apply to non-syslog use case? (user-agent is higher level 
 and should belong to DockerMonitor class I would think)

2. I think race condition between `monitor_manager` and monitor thread will happen if we perform the version check in `DockerMonitor.run()` (I don't think we use any barriers to gate monitor threads).  We may need a one-time lazy-instantiation strategy to set the version.

3. If we do decide on one-time, I then wonder if we should make the k8s monitor do the same.  The one scenario where I could see k8s monitor needing to poll is if the k8s api server can be upgraded independently of agent pods (ie live k8s server change). I don't know enough about k8s to know if this is possible now or in future.

echee2:
@imron, I've moved the version check into `gather_sample` like we discussed. 

@czerwingithub, for this PR, I wrote a very similar unit test to the previous user-agent PR.  In the process, I fixed a non-deterministic bug (uncaught in the first PR) that was causing the tests to hang sporadically.  I haven't fully analyzed but I'm pretty sure `block_until_n_waiting_threads` must wait for all monitor + manager threads to sleep at the end of their main loops  before advancing the fake clock.  I think failure to "barrier" all threads leads to race situation where the condition variable never notifies.  This may be caused by the `increment()` calls happening too early/quickly.  Anyway, I've moved the reusable code into `FakeClockCounter` for reuse.  But the only real fix was to simply make `block_until_n_waiting_threads` wait for the correct number of threads.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -102,9 +92,9 @@
                      'Optional (defaults to "docker_api"). Determine which method is used to gather logs from the '
                      'local containers. If "docker_api", then this agent will use the docker API to contact the local '
                      'containers and pull logs from them.  If "syslog", then this agent expects the other containers '
-                     'to push logs to this one using the syslog Docker log plugin.  Currently, "syslog" is the '
-                     'preferred method due to bugs/issues found with the docker API.  It is not the default to protect '
-                     'legacy behavior.\n',
+                     'to push logs to this one using the Docker syslog logging driver.  Currently, "syslog" is the '
+                     'preferred method due to bugs/issues found with the docker API (To protect legacy behavior, '
+                     'the default method is "docker_api").',
```
echee2: I reworded this a bit to:
a) Use the official docker name for their syslog plugin
b) Slightly sharpen the English in the last sentence.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1691,9 +1691,7 @@ def get_user_agent_fragment(self):
         ver = None
         if k8s_cache:
             ver = k8s_cache.get_api_server_version()
-            if ver:
-                ver = 'k8s=%s' % ver
-        return ver
+        return 'k8s=%s' % (ver if ver else 'true')
```
echee2: I owe you guys a unit test for this one (or at least manual testing before Friday release)

scalyr_agent/test_util.py
```diff
@@ -0,0 +1,135 @@
...
+
+from scalyr_agent import json_lib
+from scalyr_agent.configuration import Configuration
+from scalyr_agent.platform_controller import DefaultPaths
+from scalyr_agent.json_lib import JsonArray, JsonObject
+from scalyr_agent.monitors_manager import MonitorsManager
+from scalyr_agent.scalyr_logging import AgentLogger
+
+
+class ScalyrTestUtils(object):
```
echee2: I refactored some reusable utility functions (for tests) into this class.

scalyr_agent/tests/monitors_manager_test.py
```diff
@@ -101,80 +91,37 @@ def test_user_agent_polling(self):
...
-                    fake_clock.block_until_n_waiting_threads(1)  # wait for monitor thread loop to complete
-                    old_count = self.count()
-                    fake_clock.advance_time(increment_by=poll_interval)
-                    self.wait_for_increment(old_count, timeout=deadline-time.time())
-
-                return self.count() == target_polls
-
-        fragment_polls = FragmentPolls()
+        # The MonitorsManager + 3 monitor threads will wait on the fake clock
+        fragment_polls = FakeClockCounter(fake_clock, num_waiters=4)
```
echee2: The related Fake Clock and Counter have been encapsulated into `FakeClockCounter` for reuse across multiple tests.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -1624,7 +1616,56 @@ def __gather_metrics_from_api( self, containers ):
...
+            # Helper function to transform docker version to user-agent fragment
+            if not ver:
+                # version not yet set: return 'docker=yes' to signify docker
+                return 'docker=true'
+            log_mode = self._config.get('log_mode')
+            if log_mode == 'syslog':
+                extra = ''
+            else:
+                extra = '|raw' if self._config.get('docker_raw_logs') else '|api'
+            return 'docker=%s|%s%s' % (ver, log_mode, extra)
```
echee2: The rationale for pipe `|` as a separator is that this character is unlikely to be part of any text returned by docker version, k8s version and any other system user-agent fragments.  This would facilitate common text processing across user-agent strings (first split by semicolon, then split by key=value, then split value on pipe).

echee2:
@imron, @czerwingithub this PR should be ready for final review.  

There's a fair amount of code but mainly refactoring of common test code.

The meat of the testing is in `DockerMonitorTest` and `KubernetesMonitorTest`

scalyr_agent/util.py
```diff
@@ -651,6 +651,69 @@ def _increment_waiting_count(self, increment):
...
+        @param target: Target count to reach
+        @param fake_increment_sec: Seconds to increment the fake clock on each poll
+        @param maxwait: Time (seconds) to wait for target to be reached
+        @return: True if number of polls reaches target_polls, else False
+        """
+        deadline = time.time() + maxwait
+        while self.count() < target and time.time() < deadline:
+            # Important: wait for monitor thread & all monitor loops to complete
+            # Failing to do so leads to non-deterministic test hangs
+            self.__fake_clock.block_until_n_waiting_threads(self.__num_waiters)
```
echee2: This fixes AGENT-74 test bug (which causes unit tests to hang non-deterministically)

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -26,27 +26,17 @@
...
 from scalyr_agent import ScalyrMonitor, define_config_option, define_metric
 import scalyr_agent.util as scalyr_util
-import scalyr_agent.json_lib as json_lib
 import scalyr_agent.scalyr_logging as scalyr_logging
 from scalyr_agent.json_lib import JsonObject
-from scalyr_agent.json_lib import JsonConversionException, JsonMissingFieldException
-from scalyr_agent.log_watcher import LogWatcher
-from scalyr_agent.monitor_utils.server_processors import LineRequestParser
-from scalyr_agent.monitor_utils.server_processors import RequestSizeExceeded
-from scalyr_agent.monitor_utils.server_processors import RequestStream
```
echee2: These imports are no longer needed.

echee2:
@imron I've resolved conflicts and rebased.

imron:
I still get a message that branch has conflicts in `docker_monitor.py`

echee2:
That was weird.  I didn't see this as a conflict when I rebased.  Will do it again.

echee2:
OK @imron, conflicts should be resolved ( I think the previous one came about when I merged in one of the smaller PRs you had recently approved )

echee2:
Thanks for reviewing this @imron

PR #160, opened 2019-04-04
AGENT-65 and AGENT-66 - Broken unit test (json encoding)

echee2:
1. Json encoder/decoder issue breaking `TestLogFileProcessor.test_log_attributes`
2. _fallback_json_encode missing `sort_keys` 

Error for (1) is 
```
======================================================================
ERROR: test_log_attributes (scalyr_agent.tests.log_processing_test.TestLogFileProcessor)
  File "/Users/echee/Documents/code/scalyr/scalyr-agent-2/scalyr_agent/tests/log_processing_test.py", line 1448, in test_log_attributes
    log_attributes=vals['attributes'])
  File "/Users/echee/Documents/code/scalyr/scalyr-agent-2/scalyr_agent/log_processing.py", line 1330, in __init__
    self.__base_event = Event(thread_id=self.__thread_id, attrs=log_attributes)
  File "/Users/echee/Documents/code/scalyr/scalyr-agent-2/scalyr_agent/scalyr_client.py", line 1217, in __init__
    self.__set_attributes( thread_id, attrs )
  File "/Users/echee/Documents/code/scalyr/scalyr-agent-2/scalyr_agent/scalyr_client.py", line 1250, in __set_attributes
    tmp_buffer.write( scalyr_util.json_encode(attributes) )
  File "/Users/echee/Documents/code/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 88, in json_encode
    return _json_encode( obj, sort_keys=True )
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 250, in dumps
    sort_keys=sort_keys, **kw).encode(obj)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py", line 209, in encode
    chunks = list(chunks)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py", line 442, in _iterencode
    o = _default(o)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: {'host': 'scalyr-1'} is not JSON serializable
```

# ANALYSIS

`scalyr_agent.util.json_encode` does not know how to encode our custom `JsonArray` and `JsonObject` by default, if the underlying lib is `json` or `ujson`.  (If the underlying lib is the fallback `json_lib`, we're OK)

Furthermore,  if underlying is `json`, we get the loud `TypeError` as above.  But if underlying is `ujson`, `json_encode()` insiduously does not raise exception **but encodes incorrectly**

# STATUS

Per discussion with Steven, I have:

1. Made ujson complain loudly if you try to make it encode a JsonArray/Object
2. Modified the failing test to correctly pass native dict instead of JsonObject
3. Added a check to `LogFileProcessor` to raise exception if you incorrectly pass in a JsonArray/Object

### DIFF
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -82,7 +82,10 @@ def get_json_implementation(lib_name):
 
         return lib_name, ujson_dumps_custom, ujson.loads
 
-    if lib_name == 'json':
+    else:
+        if lib_name != 'json':
+            raise ValueError('Unsupported json library %s' % lib_name)
+
         import json
 
         def json_dumps_custom(*args, **kwargs):
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -72,26 +72,23 @@ def get_json_implementation(lib_name):
 
     elif lib_name == 'ujson':
         import ujson
-        return lib_name, ujson.dumps, ujson.loads
+
+        def ujson_dumps_custom(*args, **kwargs):
+            # ujson does not raise exception if you pass it a JsonArray/JsonObject while producing wrong encoding.
+            # Detect and complain loudly.
+            if isinstance(args[0], (json_lib.JsonObject, json_lib.JsonArray)):
+                raise TypeError('ujson does not correctly encode objects of type: %s' % type(args[0]))
+            return ujson.dumps(*args, **kwargs)
+
+        return lib_name, ujson_dumps_custom, ujson.loads
 
     if lib_name == 'json':
         import json
-
-        class JsonObjArrayEncode(json.JSONEncoder):
-            def default(self, obj):
-                if isinstance(obj, json_lib.JsonObject):
-                    return obj.asDict()
-                if isinstance(obj, json_lib.JsonArray):
-                    return obj.asList()
-                # Let the base class default method raise the TypeError
-                return json.JSONEncoder.default(self, obj)
 
         def json_dumps_custom(*args, **kwargs):
             """Eliminate spaces by default. Python 2.4 does not support partials."""
             if 'separators' not in kwargs:
                 kwargs['separators'] = (',', ':')
-            if 'cls' not in kwargs:
-                kwargs['cls'] = JsonObjArrayEncode
             return json.dumps(*args, **kwargs)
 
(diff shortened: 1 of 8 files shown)


echee2:
Need you guys' input on this @czerwingithub @imron .  Particularly because the failing test pertains to `LogFileProcessor` which seems risky?

echee2:
I talked with Steven.  Here's the summary:

1. `JsonObject` and `JsonArray` really aren't necessary and possibly be eliminated entirely
2. Our fallback lib is only a requirement for parsing our JSON (more permissive, comments etc)
      - but once parsed, it does not need to be returning `JsonObject` and `JsonArray`.  Investigate returning regular python objects.
3. This failing unit test can just be changed to supply dict instead of `JsonObject`.   Furthermore the code path to sending over logs should be ensuring that `JsonObject` / `JsonArray` are NOT passed.

scalyr_agent/util.py
```diff
@@ -55,47 +55,82 @@
...
+        def json_dumps_custom(*args, **kwargs):
             """Eliminate spaces by default. Python 2.4 does not support partials."""
             if 'separators' not in kwargs:
                 kwargs['separators'] = (',', ':')
             return json.dumps(*args, **kwargs)
 
-        _json_encode = dumps_no_space
-        _json_decode = json.loads
+        return lib_name, json_dumps_custom, json.loads
+
```
imron: If someone passes in a `lib_name` value that is not `json` `ujson` or `json_lib` (for whatever reason, maybe a spelling mistake, or maybe a new json library is added), then the function will reach the end and return `None`, instead of an expected 3-value tuple.

It's not a big issue, because it'll probably never happen and if it does we'll get a compile error, but maybe if we get an unrecognized name then we raise an exception with a message that the json `lib_name` is unsupported.

echee2: Thanks, good suggestion.  Change has been made & unit test installed to capture.

PR #161, opened 2019-04-04
Ticket: #AGENT-20

imron:
Removes python 2.7 specific formatting code from the log processor and related
tests.

Also updates the `md5_digest` utility function to return a hex string by default
rather than returning the raw md5 bytes, which could contain null-bytes and
other invalid ASCII.

### DIFF
--- a/scalyr_agent/log_processing.py
+++ b/scalyr_agent/log_processing.py
@@ -2051,10 +2051,7 @@ class LogLineRedacter(object):
                 for _group_index, _group in enumerate(_groups):
                     # for each group in a match, replace the `replacement_ex` with either its `group` content, or
                     # the hash of the `group` depending on the hash indicator \\1 vs \\H1 etc.
-                    group_hash_indicator = "\{}{}".format(
-                        LogLineRedacter.HASH_GROUP_INDICATOR,
-                        _group_index + 1
-                    )
+                    group_hash_indicator = "\\%s%s" % (LogLineRedacter.HASH_GROUP_INDICATOR, _group_index + 1)
                     replacement_matches += 1
                     if group_hash_indicator in replacement_ex:
                         # the group needs to be hashed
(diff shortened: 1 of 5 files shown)


imron:
@czerwingithub , I don't know if you remember the details of the commit that added the hashed replacement values, but there was a md5_digest function added to utils that returned the raw md5 bytes, rather than a hex-string representation. 

It only appears to be used by the replacement rule code hashing, but I changed it up to return a hex string by default as it was showing invalid chars in the uploaded logs - which is probably not ideal.  I'm not sure however if that was originally the intended behaviour?

I made it so that the return type (raw bytes or hex string) is determined by a parameter, but in the Python md5 api, it distinguishes these by 2 separate calls `md5.digest` and `md5.hexdigest`.  We could probably just call our utility function `md5_hexdigest` because it's not called anywhere else and I suspect that what we should really be doing anyway.

With these changes, the only places in the code that still use `.format` are win32 code and third-party libraries that we include, which I think is acceptable.

scalyr_agent/log_processing.py
```diff
@@ -2051,10 +2051,7 @@ def __replace_groups_with_hashed_content(re_ex, replacement_ex, line):
                 for _group_index, _group in enumerate(_groups):
                     # for each group in a match, replace the `replacement_ex` with either its `group` content, or
                     # the hash of the `group` depending on the hash indicator \\1 vs \\H1 etc.
-                    group_hash_indicator = "\{}{}".format(
-                        LogLineRedacter.HASH_GROUP_INDICATOR,
-                        _group_index + 1
-                    )
+                    group_hash_indicator = "\\%s%s" % (LogLineRedacter.HASH_GROUP_INDICATOR, _group_index + 1)
```
czerwingithub: Why not use %d for the second param since it is a number?

scalyr_agent/log_processing.py
```diff
@@ -2066,7 +2063,7 @@ def __replace_groups_with_hashed_content(re_ex, replacement_ex, line):
                         )
                     else:
                         # the group does not need to be hashed
-                        replaced_group = replaced_group.replace("\{}".format(_group_index + 1), _group, 1)
+                        replaced_group = replaced_group.replace("\\%s" % (_group_index + 1), _group, 1)
```
czerwingithub: Use %d?

scalyr_agent/util.py
```diff
@@ -196,11 +196,14 @@ def create_unique_id():
     return result
 
 
-def md5_digest(data):
+def md5_digest(data, hexdigest=True):
```
czerwingithub: I'm sure it was just an oversight that we weren't using the hexdigest.  Yes, I think you can just make this always return the hex digest.

imron:
All done.  For consistency with the md5 class I've renamed the `md5_digest` method to `md5_hexdigest`.

PR #166, opened 2019-04-09
Agent 38 - extend environment variable configuration to top-level & k8s params

echee2:
# Background

This PR extends environment variable support for config params to all top-level params + select module-level params (k8s & k8s_events).

The mental model is env variables can be mapped to specific config params.  To ease code management, mapping is done in centralized locations via `ENVIRONMENT_AWARE_CONFIG_PARAMS` maps on the `Configuration` class and individual monitors (optional).

## Precedence

Environment variables do not change. This is why config file value must take precedence over environment value.

## Naming 

If no custom name is provided,  config params are mapped to env variables `scalyr_<param_name>'.   If custom name is provided, that specific name is used.  Both upper case and lower case env vars are checked.  

(Note: the custom mapping is needed for `SCALYR_SERVER` (makes no sense to prepend SCALYR_).  Also, the k8s_events params are very generic names and definitely need special casing.

# Strategy

## A.  Code management

All config params that must be overridable via env variable must now be defined in class-level `ENVIRONMENT_AWARE_CONFIG_PARAMS` maps.  These maps are present on:
- `Configuration`
- `KubernetesMonitor`
- 'KubernetesEventMonitor`
In future, other monitors that wish to have overridable environment variables must define this map.

Critically, these static maps are used by unit test to ensure correct overriding. (Unit tests must fail if a param is registered in the map but not "exercised" by the unit test.)

## B.  Logging conflicts between env & config file

If values are defined in both environment & JSON config files, I log a warning if values differ.  Only log during configuration refresh (otherwise too noisy)

## C. Parsing environment variables

Since env vars are strings, customer is responsible for ensuring correctness (must be parseable into bool, int, float etc). 

To maintain consistency with JSON config files, I use same parser code for env vars.  This offers the ability to support JSON snippets in env variables -- although we currently do not use it. (Note: [ElasticSearch supports this](https://www.elastic.co/guide/en/beats/metricbeat/current/using-environ-vars.html))

`event_object_filter` is a list of strings.  We could theoretically require customer to enter full string representation of a JSON array.  But this is unwieldy and I feel we should support simple comma-separated lists.  As such, I introduce `ArrayOfStrings` type.  (A new type was beneficial for stricter type-checking and testing.  Note we also previously introduced the notion of this type in `__verify_or_set_optional_array_of_strings`.)

# How tested

1. See "Central to long-term code-maintenance"

2. Exhaustive tests to exercise all centrally defined environment-aware variables.  General strategy: set the env variable to something different from config file or default value.  Ensure config file wins if defined.  Otherwise, ensure env val is taken.

3. Expand testing of `MonitorConfig.get` to reflect augmented type-conversion support.
- e.g. see `test_str_conversion` for `str` -> `ArrayOfStrings` conversion.

# Past notes

1. After lengthy discussion, we decided that yet another route of configuration for k8s would unacceptably complicate the mental model for customers.  Since environment variable configuration is a widely accepted mechanism and since we already have such support for a few variables,  we decided to expand this instead of introducing a (third) methodology.

2. To conform with unofficial standards, environment variables should be of the form `SCALYR_<PARAM_NAME>`  (all upper case).  However, some customers may have problems with upper case so we will also support falling back to lower case.

3. Unfortunately, existing env var support already exists for `SCALYR_SERVER`, which does not match the pattern `SCALYR_<PARAM_NAME>` (The environment variable name that correctly matches should be `SCALYR_SCALYR_SERVER`).  We will special-case this variable.


echee2:
@imron @czerwingithub this PR is ready for final review pending the final issue of precedence:

Environment variables cannot be changed easily.  Hence, customer cannot dynamically edit config files to alter environment-aware config params.  An agent restart is necessary. (Customer must obviously edit environment variables before restart).

I can see how this mental model might still work. i.e. "If you're going to override params via environment variables, managing your environment variables _should_ be the main point of edit (for these params)"

The alternative is to have config files take priority:-
A potential downside there for _environment-aware_ params is that an outdated config file value would always override customer-specified environment var, forcing customer to edit the config file. (If you think about it, if they are defining environment variables for these params in the first place, wouldn't that suggests it is more convenient for them to do so via environment and that editing config files is the less convenient option?)

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1251,6 +1237,27 @@ class KubernetesMonitor( ScalyrMonitor ):
...
+    ENVIRONMENT_AWARE_CONFIG_PARAMS = {
+        'container_check_interval': None,
+        'docker_max_parallel_stats': None,
+        'container_globs': None,
+        'report_container_metrics': None,
+        'report_k8s_metrics': None,
+        'k8s_ignore_pod_sandboxes': None,
+        'k8s_include_all_containers': None,
+        'k8s_use_v2_attributes': None,
+        'k8s_use_v1_and_v2_attributes': None,   # echee TODO (double check w steven)
```
echee2: Need input from @czerwingithub whether this param needs to be environment-aware

Background notes from Imron: 
v1 did not group by deployments. In order to prevent breakages, this flag was added so that customers who had dashboards that relied on the old attribute names would still work.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -213,6 +213,18 @@ class KubernetesEventsMonitor( ScalyrMonitor ):
...
+        "max_log_rotations": "SCALYR_K8S_MAX_LOG_ROTATIONS",
+        "log_flush_delay": "SCALYR_K8S_LOG_FLUSH_DELAY",
+        "message_log": "SCALYR_K8S_MESSAGE_LOG",
+        "event_object_filter": "SCALYR_K8S_EVENT_OBJECT_FILTER",
+        "leader_check_interval": "SCALYR_K8S_LEADER_CHECK_INTERVAL",
+        "leader_node": "SCALYR_K8S_LEADER_NODE",
+        "check_labels": "SCALYR_K8S_CHECK_LABELS",
+        "ignore_master": "SCALYR_K8S_IGNORE_MASTER",
+    }
+
```
echee2: In contrast to Global and `kubernetes_monitor` params, these `kubernetes_events_monitor` param names are too generic.  So instead of relying on default prepending of `SCALYR_`,  I specify custom names.

scalyr_agent/agent_main.py
```diff
@@ -285,17 +285,14 @@ def __read_and_verify_config(self, config_file_path):
...
         """
-        config = self.__read_config(config_file_path)
+        config = self.__make_config(config_file_path)
         self.__verify_config(config)
         return config
 
-    def __read_config(self, config_file_path):
-        """Reads the configuration from the file path but does not do a full verification.
+    def __make_config(self, config_file_path):
+        """Make Configuration object. Does not read nor verify.
```
echee2: Fixed outdated comment.

scalyr_agent/configuration.py
```diff
@@ -50,7 +51,75 @@ class Configuration(object):
     This also handles reporting status information about the configuration state, including what time it was
     read and what error (if any) was raised.
     """
-    def __init__(self, file_path, default_paths):
+
+    # Authoritative list of params that can be overridden by environment variables.
+    ENVIRONMENT_AWARE_CONFIG_PARAMS = {
+        # Special-case global params
+        'api_key': None,
+        'scalyr_server': 'SCALYR_SERVER',
```
echee2: Except for `scalyr_server`, all other params don't need custom names ( can default to prepending `SCALYR_` )

czerwingithub:
<snip>
> The alternative is to have config files take priority:-
> A potential downside there for _environment-aware_ params is that an outdated config file value would always override customer-specified environment var, forcing customer to edit the config file. (If you think about it, if they are defining environment variables for these params in the first place, wouldn't that suggests it is more convenient for them to do so via environment and that editing config files is the less convenient option?)

Yes, I think this is good point and pushes us in the direction the configs should override the environment variables.  I was thinking of a similar argument yesterday.

In the K8s scenario, what are we trying to avoid by having the env variables take precedence?  We are trying to avoid a situation where they have created config edits that they now want to override with environment variables.  For this case, I think we don't have to examine the problem of them trying to override the default config files because we can make an effort to never set things in the config that need to be overridden.  So, if the created config edits in the first place.. they are knowledgeable enough to undo those config edits... or just keep using config edits instead of env variables.

So... yes, I think all around, it just makes more sense to have config files override the env variables.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
echee2: The conclusion during today's standup:

k8s_use_v2_attributes & k8s_use_v1_and_v2_attributes don't need to be environment-aware.  Reasoning: users of k8s config maps are likely on Starbuck/v2 anyways.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -32,11 +31,8 @@
 import logging
 import logging.handlers
 
-from scalyr_agent import ScalyrMonitor, define_config_option, AutoFlushingRotatingFileHandler
-from scalyr_agent.log_watcher import LogWatcher
+from scalyr_agent import types, ScalyrMonitor, define_config_option, AutoFlushingRotatingFileHandler
 from scalyr_agent.scalyr_logging import BaseFormatter
-from scalyr_agent.scalyr_logging import DEBUG_LEVEL_0, DEBUG_LEVEL_1, DEBUG_LEVEL_2
-from scalyr_agent.monitor_utils.auto_flushing_rotating_file import AutoFlushingRotatingFile
```
echee2: Removed stale imports

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -78,11 +74,14 @@
                      'absolute path',
                      convert_to=str, default='kubernetes_events.log')
 
-define_config_option( __monitor__, 'event_object_filter',
-                     'Optional (defaults to [ \'CronJob\', \'DaemonSet\', \'Deployment\', \'Job\', \'Node\', \'Pod\', \'ReplicaSet\', \'ReplicationController\', \'StatefulSet\' ] ). A list of event object types to filter on. '
-                     'If set, only events whose `involvedObject` `kind` is on this list will be included.',
-                     default=[ 'CronJob', 'DaemonSet', 'Deployment', 'Job', 'Node', 'Pod', 'ReplicaSet', 'ReplicationController', 'StatefulSet' ]
-                     )
+EVENT_OBJECT_FILTER_DEFAULTS = [
+    'CronJob', 'DaemonSet', 'Deployment', 'Job', 'Node', 'Pod', 'ReplicaSet', 'ReplicationController', 'StatefulSet'
```
echee2: Refactor default `event_object_filter` list to prevent divergence from description

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1251,6 +1237,27 @@ class KubernetesMonitor( ScalyrMonitor ):
...
+        # 'k8s_use_v2_attributes': None,
+        # 'k8s_use_v1_and_v2_attributes': None,
+        # The following params have deprecated (moved to global config)
+        # 'k8s_ignore_namespaces': None,
+        # 'k8s_api_url': None,
+        # 'k8s_verify_api_queries': None,
+        # 'k8s_cache_expiry_secs': None,
+        # 'k8s_cache_purge_secs': None,
+    }
+
```
echee2: Explicitly comment-out params that should not be environment-aware (to indicate a conscious decision)

scalyr_agent/configuration.py
```diff
@@ -134,20 +206,6 @@ def parse(self, logger=None):
...
-            if api_key and env_api_key and api_key != env_api_key:
-                # ignore and key in the env variable and warn the user
-                if logger:
-                    logger.warn("You have different api keys in the config file and env variable `scalyr_api_key`."
-                             " Ignoring the env variable.")
-            elif not api_key and env_api_key:
-                if logger:
-                    logger.debug("Using the api key from the env variable `scalyr_api_key`")
-                api_key = env_api_key
-
```
echee2: Relocated this specialized handling into general case.

scalyr_agent/configuration.py
```diff
@@ -1035,6 +1093,80 @@ def __verify_main_config_and_apply_defaults(self, config, file_path, apply_defau
         self.__verify_or_set_optional_int(config, 'disable_leak_verify_config_create_copying_manager', None, description, apply_defaults)
         self.__verify_or_set_optional_bool(config, 'disable_leak_verify_config_cache_config', False, description, apply_defaults)
 
+    def __get_config_or_environment_val(self, config_object, param_name, param_type):
```
echee2: Encapsulate getting of config file value with getting of environment value as well.
Also add support for newly defined `ArrayOfStrings` type (for type safety)

scalyr_agent/tests/scalyr_monitor_test.py
```diff
@@ -18,24 +18,31 @@
...
+from scalyr_agent.json_lib.objects import JsonArray, JsonObject, ArrayOfStrings
 from scalyr_agent.scalyr_monitor import MonitorConfig, BadMonitorConfiguration, define_config_option
 from scalyr_agent.test_base import ScalyrTestCase
 
 
 class MonitorConfigTest(ScalyrTestCase):
 
     def test_base(self):
+        test_array = ["a", 1, False]
+        test_obj = {"a": 100, "b": 200}
```
echee2: I have augmented this test class to reflect additional support for converting between types.
The main type we need to support is `ArrayOfStrings`.  This gives us capability to support comma-separated list of strings in environment variable such as for `event_object_filter`.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -213,6 +214,18 @@ class KubernetesEventsMonitor( ScalyrMonitor ):
        ```
     """
 
+    ENVIRONMENT_AWARE_CONFIG_PARAMS = {
```
czerwingithub: Why not modify the `define_config_option` method to accept an optional parameter specifying what environment variable it defaults to?  That way all information about the configuration parameter is located together.

scalyr_agent/configuration.py
```diff
@@ -50,7 +52,75 @@ class Configuration(object):
     This also handles reporting status information about the configuration state, including what time it was
     read and what error (if any) was raised.
     """
-    def __init__(self, file_path, default_paths):
+
+    # Authoritative list of params that can be overridden by environment variables.
+    ENVIRONMENT_AWARE_CONFIG_PARAMS = {
```
czerwingithub: Do we need this dict?  I think we are going to allow all global parameters to be configured via an environment variable.. so the only useful information here is None vs non-None.  And only one param has a non-None value.   We could modify the `_verify_or_set....` config methods to allow you to override the env var name used for the variable.

scalyr_agent/configuration.py
```diff
@@ -1035,6 +1093,80 @@ def __verify_main_config_and_apply_defaults(self, config, file_path, apply_defau
...
+        )
+
+        # Not set in environment
+        if env_val is None:
+            return config_val
+
+        # Report conflicting values
+        if self.__logger:
+            if config_val is not None and config_val != env_val:
+                self.__logger.warn(
```
czerwingithub: Because we check config once every 30 seconds, we would be logging this every 30 seconds.  Use the `limit_once_per_x_secs`  logging features to emit this once every 5 minutes.

scalyr_agent/scalyr_monitor.py
```diff
@@ -740,28 +766,56 @@ def get(self, field, required_field=False, max_value=None, min_value=None,
...
+                env_aware_param_map = self.__monitor_class.ENVIRONMENT_AWARE_CONFIG_PARAMS
+                envar_name, envar_strval, envar_val = get_config_from_env(
+                    field, env_aware_param_map.get(field), convert_to=convert_to
+                )
+                if result is None:
+                    result = envar_val
+                elif report_conflicting_environment_value and envar_val:
+                    # Both config param and environment variables defined.  Warn if conflict.
+                    if result != envar_val and log:
+                        log.warn(
```
czerwingithub: Same here about the using the `limit_once_per_x_secs` feature to limit to once per 5 minutes.

scalyr_agent/util.py
```diff
@@ -1269,3 +1272,87 @@ def close(self):
...
+    @param strlist: list to be converted
+    @return: None if strlist is empty, else return a JsonArray of strings
+    @raise TypeError if element_type is specified and conversion of any element fails
+    """
+    if not strlist:
+        return None
+    strlist = strlist.strip()
+
+    # Remove surrounding square brackets
+    if strlist[0] == '[' and strlist[-1] == ']':
```
czerwingithub: Add a check for strlist being an empty string after the strip operation?

scalyr_agent/util.py
```diff
@@ -1269,3 +1272,87 @@ def close(self):
...
+
+def get_config_from_env(param_name, custom_env_name=None, convert_to=None):
+    """Returns the environment variable value for a config param
+
+    If a custom environment variable name is defined, use it instead of prepending 'SCALYR_'.
+
+    @param param_name: Config param name (may be global or module-level)
+    @custom_env_name: Custom environment variable name
+    @expected_type: if not None, will convert and validate to this type.  Otherwise, will leave as string.
+    @return: 3-tuple containing (param_name, env string val, converted value)
```
czerwingithub: Just personal preference.. returning a tuple like this is a little bit of a pain.  Do we return all of this just to support the error logging if there is a conflict in a config provided value?  If so, maybe we should move that logging here?  After all, we do have two instances of that logging.. it might be better to consolidate it.

scalyr_agent/util.py
```diff
@@ -1269,3 +1272,87 @@ def close(self):
...
+        strval = os.getenv(key)
+
+    if not strval or convert_to is None:
+        return key, strval, None
+
+    try:
+        if convert_to in {unicode, str}:
+            return key, strval, convert_to(strval)
+
+        # Special-case handling for ArrayOfStrings
```
czerwingithub: It seems there's a overlap between this code and the conversion code you modified in the `configuration.py`.  Is there a way to move this method in there so that it can rely on the `_perform_conversion` method?

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
echee2: My initial design was actually an optional param in `define_config_option` and there are certainly benefits of colocating it with the rest of the param definition.  However, I later realized 2 things:

1) we should adopt a uniform style for both global and monitor-level (ie both centralized map or else both during `define_config_option` or `__verify_or_set_optional_xxx`.

2) As the number of total config options balloons, there's benefit in being able to quickly identify the complete list of environment aware params.  I felt it was hard to track them when spread out.  

2) For testing.  Without this, I think the individual calls to `__verify_or_set_optional_xxx` would have to build this map at runtime anyway.

WDYT? If you still feel strongly about this,  I can look into changing it.

scalyr_agent/configuration.py
echee2: (related to the above)

scalyr_agent/configuration.py
echee2: will fix

scalyr_agent/util.py
echee2: Consolidating logging sgtm. Will try.

scalyr_agent/util.py
echee2: I will investigate.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
czerwingithub: I do agree with your point number `1`... we should move to a `define_config_option` for the global parameters.  However, I'm not sure if you want to do all of that work in this PR.  Adding the environment aware arguments to `__verify_or_set..` is at least a step towards that.

As for point `2`, I'm not sure if I agree.  I can't really see me saying "what are the environment aware params?".  It's much more likely that I will ask myself "for this particular param, what is it's default and can I set it via an environment variable?".   If you are worried that we will need to know the complete list to avoid collision.. I would suggest we should solve that through code.  If we attempt to define two options with the same environment variable, throw an assertion.

I do agree that the `__verify_or_set_optional_xxx` will have to build the map at runtime, but that's fine.

I do think we should make this change now.  I think we want to be going in the right eventual direction and this is a step away from that.  WDYT?

echee2: All fair points.  I will change.

echee2:
@czerwingithub I've included all of your suggestions. 

 I also fixed a pre-2.7 compatibility issue (cannot initialize set with curly braces). It was causing  `python setup.py` to fail.

Note that the current circleCI tests fail because of https://scalyr.myjetbrains.com/youtrack/issue/AGENT-71   (short circuit bug).   

But this PR doesn't cause any additional failures.

echee2:
Thanks for the solid guidance!  Will collapse commits and merge.

PR #167, opened 2019-04-17
AGENT-84: Display environment variables in agent status-v

echee2:
# BACKGROUND

With the expansion of environment variable support ([AGENT-38)](https://scalyr.myjetbrains.com/youtrack/issue/AGENT-38), we should list environment vars to facilitate debugging and customer support.

- Important variables should be listed up front
- Missing important variables should be listed as missing
- SCALYR_API_KEY and other sensitive values should be redacted
- The rest should be displayed in sorted order (ignoring case)
- When displayed, do not alter case (to reflect actual environment)

# ASSUMPTION

As long as there is a formula for all scalyr-related env variables, there is no need to programmatically derive a comprehensive list of keys.   Except for `K8S_EVENT_DISABLE`,  as of now,  all env variables start with `SCALYR_`.    **We must preserve this invariant** as we define new custom mappings.

# IMPLEMENTATION

With the above assumption, we can easily filter for env vars starting with `SCALYR_` (case insensitive), then special casing the one or two vars like `K8S_EVENT_DISABLE`.

# TESTING

See the modified unit tests in this PR

Also tested on k8s (minikube).  Logged into pod and did `agent status -v`

image

### DIFF
--- a/scalyr_agent/agent_status.py
+++ b/scalyr_agent/agent_status.py
@@ -27,6 +27,8 @@
 
 __author__ = 'czerwin@scalyr.com'
 
+import os
+
 import scalyr_agent.util as scalyr_util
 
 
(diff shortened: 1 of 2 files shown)


scalyr_agent/agent_status.py
```diff
@@ -283,6 +285,43 @@ def report_status(output, status, current_time):
...
+        upper2actualkey = {}
+        for k in os.environ.keys():
+            kup = k.upper()
+            if kup.startswith('SCALYR') or kup in special_case_keys:
+                upper2actualkey[kup] = k
+
+        # Sorted list of all scalyr keys, including main_keys which may not be present
+        # Sort order does not consider letter case.
+        sorted_upperkeys = main_keys + sorted(set(upper2actualkey.keys()) - set(main_keys))
+        # sorted_actualkeys = [upper2actualkey.get(k) or k for k in sorted_upperkeys]
```
imron: Looks like some leftover code was left in a comment.

Perhaps delete rather than leaving it in?

echee2: Thanks for catching it.  Updated.

PR #168, opened 2019-04-17
Ticket AGENT-42

imron:
This commit adds support for container runtimes other than docker.

By default, the monitor will detect whether docker is being used or whether
another CRI compatible runtime (e.g. containerd) is running.  You can force the
use of the CRI codepath with the option `k8s_always_use_cri`.

When using the CRI codepath, you need to map in both `/var/log/containers` and
`/var/log/pods` so that the agent can read the log files from disk.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -169,7 +169,9 @@ define_config_option( __monitor__, 'k8s_parse_format',
                       'Optional (defaults to `auto`). Valid values are: `auto`, `json`, `cri` and `raw`. '
                       'If `auto`, the monitor will try to detect the format of the raw log files, '
                       'e.g. `json` or `cri`.  Log files will be parsed in this format before uploading to the server '
-                      'to extract log and timestamp fields.  If `raw`, the raw contents of the log will be uploaded to Scalyr.',
+                      'to extract log and timestamp fields.  If `raw`, the raw contents of the log will be uploaded to Scalyr. '
+                      '(Note: An incorrect setting can cause parsing to fail which will result in raw logs being uploaded to Scalyr, so please leave '
+                      'this as `auto` if in doubt.)',
                       convert_to=str, default='auto', env_aware=True)
 
 define_config_option( __monitor__, 'k8s_always_use_cri',
@@ -845,6 +847,14 @@ class ContainerChecker( StoppableThread ):
         self._container_enumerator = None
         self._container_runtime = 'unknown'
 
+    def _is_running_in_docker( self ):
+        """
+        Checks to see if the agent is running inside a docker container
+        """
+        # whether or not we are running inside docker can be determined
+        # by the existence of the file /.dockerenv
+        return os.path.exists( "/.dockerenv" )
+
     def start( self ):
 
         try:
@@ -899,7 +909,9 @@ class ContainerChecker( StoppableThread ):
                 self._container_enumerator = CRIEnumerator( self.container_id, k8s_api_url, query_filesystem=self.__cri_query_filesystem )
 
             if self.__parse_format == 'auto':
-                if self._container_runtime == 'docker':
+                # parse in json if we detect the container runtime to be 'docker' or if we detect
+                # that we are running in docker
+                if self._container_runtime == 'docker' or self._is_running_in_docker():
                     self.__parse_format = 'json'
                 else:
                     self.__parse_format = 'cri'
@@ -1921,10 +1933,14 @@ class KubernetesMonitor( ScalyrMonitor ):
                     controller_info.update( cluster_info )
                 self.__gather_k8s_metrics_for_pod( pod, info, controller_info )
 
+        except ConnectionError, e:
+            self._logger.warning( "Error connecting to kubelet API: %s.  No Kubernetes stats will be available" % str( e ),
+                                  limit_once_per_x_secs=300,
+                                  limit_key='kubelet-api-connection-stats' )
         except KubeletApiException, e:
             self._logger.warning( "Error querying kubelet API: %s" % str( e ),
                                   limit_once_per_x_secs=300,
-                                  limit_key='kubelet-api-query' )
+                                  limit_key='kubelet-api-query-stats' )
 
     def __get_k8s_cache(self):
         k8s_cache = None
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -180,7 +180,7 @@ define_config_option( __monitor__, 'k8s_always_use_cri',
                      convert_to=bool, default=False, env_aware=True)
 
 define_config_option( __monitor__, 'k8s_cri_query_filesystem',
-                     'Optional (defaults to False). If True, then when in CRI mode, the monitor will query the filesystem for the list of active containers, rather than the Kubelet API.  This is useful in environments where the Kubelet API has been disabled.',
+                     'Optional (defaults to False). If True, then when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than first querying the Kubelet API. This is a useful optimization when the Kubelet API is known to be disabled.',
                      convert_to=bool, default=False, env_aware=True)
 
 define_config_option( __monitor__, 'k8s_verify_api_queries',
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -159,9 +159,21 @@ define_config_option( __monitor__, 'k8s_cache_init_abort_delay',
                      convert_to=int, default=20)
 
 define_config_option( __monitor__, 'k8s_parse_json',
-                      'Optional (defaults to True). If True, the log files will be parsed as json before uploading to the server '
-                      'to extract log and timestamp fields.  If False, the raw json will be uploaded to Scalyr.',
-                      convert_to=bool, default=True, env_aware=True)
+                      'Deprecated, please use `k8s_parse_format`. If set, and True, then this flag will have the equivalent of setting `k8s_parse_format` to `auto`. '
+                      'If set and False, then this flag will be equivalent to setting `k8s_parse_format` to `raw`',
+                      convert_to=bool, default=None)
+
+define_config_option( __monitor__, 'k8s_parse_format',
+                      'Optional (defaults to `auto`). Valid values are: `auto`, `json`, `cri` and `raw`. '
+                      'If `auto`, the monitor will try to detect the format of the raw log files, '
+                      'e.g. `json` or `cri`.  Log files will be parsed in this format before uploading to the server '
+                      'to extract log and timestamp fields.  If `raw`, the raw contents of the log will be uploaded to Scalyr.',
+                      convert_to=str, default='auto')
+
+define_config_option( __monitor__, 'k8s_always_use_cri',
+                     'Optional (defaults to False). If True, the kubernetes monitor will always try to read logs using the container runtime interface '
+                     'even when the runtime is detected as docker',
+                     convert_to=bool, default=False)
 
 define_config_option( __monitor__, 'k8s_verify_api_queries',
                       'Optional (defaults to True). If true, then the ssl connection for all queries to the k8s API will be verified using '
@@ -262,6 +274,9 @@ _CONTROLLER_KEYS = {
     'ReplicationController': 'k8s-replication-controller',
     'StatefulSet': 'k8s-stateful-set'
 }
+
+# A regex for splitting a container id and runtime
+_CID_RE = re.compile( '^(.+)://(.+)$' )
 
 class K8sInitException( Exception ):
     """ Wrapper exception to indicate when the monitor failed to start due to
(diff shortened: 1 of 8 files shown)


echee2:
Looking.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -872,6 +883,24 @@ def _update_api_server_version(self, k8s):
         finally:
             self._lock.release()
 
+    def _get_scalyr_container_id_and_runtime( self, k8s ):
+        pod_name = k8s.get_pod_name()
+        pod = k8s.query_pod( k8s.namespace, pod_name )
+        if pod is None:
+            return '<none>', '<none>'
```
echee2: See FOCUS 3

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -872,6 +883,24 @@ def _update_api_server_version(self, k8s):
...
+        containers = status.get( 'containerStatuses', [] )
+        for container in containers:
+            name = container.get( 'name' )
+            if name and name == 'scalyr-agent':
+                containerId = container.get( 'containerID', '' )
+                m = _CID_RE.match( containerId )
+                if m:
+                    return m.group(2), m.group(1)
+
+        return '<none>', '<none>'
```
echee2: See FOCUS 3

scalyr_agent/log_processing.py
```diff
@@ -183,7 +221,14 @@ def __init__(self, path, config, log_config, file_system=None, checkpoint=None):
         self.__log_deletion_delay = config.log_deletion_delay  # Defaults to 10 * 60
         self.__page_size = config.read_page_size  # Defaults to 64 * 1024
 
-        self.__parse_as_json = log_config.get('parse_lines_as_json', False)
+        self.__parse_format = log_config.get( 'parse_format' )
+        parse_as_json = log_config.get('parse_lines_as_json', None, none_if_missing=True)
```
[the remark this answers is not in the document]
imron: > 2. But in k8s and docker monitors, __create_log_config seems to be creating dicts.

Which are then converted to JsonObjects later on when `parse_log_config` is called (see e.g. in `copying_manager.py`, `def add_log_config`

So it's definitely expecting a JsonObject, it's just until the `parse_lines_as_json` option was deprecated, we never needed to use JsonObject specific code (the `none_if_missing` param), and so all the tests that were incorrectly passing in `dicts` didn't have a problem.

I've made it so the code will handle either.  Probably it would have been better to make the tests pass in JsonObjects but there were a lot of cases doing this, and it's not clear if there is anywhere else doing this either.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -597,14 +781,12 @@ def __init__( self, config, global_config, logger, socket_file, docker_api_versi
         self.__module = None
         self.__start_time = time.time()
         self.__thread = StoppableThread( target=self.check_containers, name="Container Checker" )
+        self._container_enumerator = None
+        self._container_runtime = 'none'
```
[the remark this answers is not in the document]
imron: That's a fair point.  I've made these return `None`.  The only thing though is that for `self._container_runtime = 'none'` this still needs to be set to a non-None value as it is used in various string substitutions and error messages, and I think it's better to have an invalid value rather than checking for `None` before each use case.  To avoid confusion with `None` I've renamed the default value to `unknown`.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -541,6 +556,163 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
...
+class ContainerEnumerator( object):
+    """
+    Base class that defines an api for enumerating all containers running on a node
+    """
+
+    def __init__( self, container_id ):
+        self._container_id = container_id
+
+    def get_containers( self, running_or_created_after=None, glob_list=None, k8s_cache=None, k8s_include_by_default=True, k8s_namespaces_to_exclude=[], current_time=None ):
+        return {}
```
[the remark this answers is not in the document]
imron: Good point.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -541,6 +556,163 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
...
+                    continue
+
+                # get the list of containers for this pod
+                status = pod.get( 'status', {} )
+                containers = status.get( 'containerStatuses', [] )
+                for container in containers:
+
+                    # get the id of the container
+                    cid = container.get( 'containerID' )
+                    if cid is None:
```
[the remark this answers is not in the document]
imron: It will never be an empty string.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -541,6 +556,163 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
...
+                    short_cid = _get_short_cid( cid )
+
+                    # ignore the scalyr-agent container
+                    if cid == self._container_id:
+                        global_log.log( scalyr_logging.DEBUG_LEVEL_2, "Excluding container because cid '%s' is the scalyr_agent container" % short_cid )
+                        continue
+
+                    # make sure the container has a name
+                    cname = container.get( 'name' )
+                    if cname is None:
```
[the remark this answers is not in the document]
imron: It will never be an empty string.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -541,6 +556,163 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
...
+                        pod = k8s_cache.pod( pod_namespace, pod_name, current_time )
+                        if pod:
+                            # check to see if we should exclude this container
+                            default_exclude = not k8s_include_by_default
+                            exclude = pod.exclude_pod( container_name=cname, default=default_exclude)
+
+                            # exclude if necessary
+                            if exclude:
+                                if pod.annotations:
+                                    global_log.log( scalyr_logging.DEBUG_LEVEL_2, "Excluding container '%s' based on pod annotations, %s" % (short_cid, str(pod.annotations)) )
```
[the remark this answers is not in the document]
imron: > doesn't the %s implicitly run a str() on whatever you pass?

Hmm, I'm sure i've encountered parts of the code where it was necessary to put a str around some objects - specifically when printing exception objects, and you'll see that's where the majority of this usage happens in the code base - though maybe it's just something I picked up from the existing codebase, because you're right, according to the docs and it appears python does call `str` when `%s` is used, and I did a quick test and didn't notice any problems.  I wonder if it's something that was relevant to older versions?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -173,9 +173,21 @@
                      convert_to=int, default=20)
 
 define_config_option( __monitor__, 'k8s_parse_json',
-                      'Optional (defaults to True). If True, the log files will be parsed as json before uploading to the server '
-                      'to extract log and timestamp fields.  If False, the raw json will be uploaded to Scalyr.',
-                      convert_to=bool, default=True)
+                      'Deprecated, please use `k8s_parse_format`. If set, and True, then this flag will have the equivalent of setting `k8s_parse_format` to `auto`. '
```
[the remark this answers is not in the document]
imron: Done.

scalyr_agent/log_processing.py
```diff
@@ -79,6 +79,44 @@
...
+    if index < 0:
+        return None, None, None, None
+
+    tags = line[:index]
+
+    # skip over the ' ' delimiter and get the log line
+    line = line[index+1:]
+
+    return timestamp, stream, tags, line
+
```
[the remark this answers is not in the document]
imron: For context, the current code is almost a line for line conversion of the [Go code used by k8s](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/logs/logs.go#L125-L169) to parse CRI logs.

There is no issue with multiple spaces, because the format currently uses single space as the delimiter.

Given that there isn't really a clear specification for this log format outside of the code, I think it's clearer to have variables explicitly named so people reading the code can easily read what each slice (and return value) represents.

`return s[0:locs[1]], s[locs[1]+1:locs[2]], s[locs[2]+1:locs[3]], s[locs[3]+1:]` is significantly less readable than `return timestamp, stream, tags, line`

Having them all bunched up like that also makes it easier for errors to slip in, for example running the suggested replacement there appears to be an out of bounds error because after getting rid of the initial -1, the array only contains 3 elements, zero indexed, and they are being accessed as if they are 1-indexed.  An easy issue to fix, but it does illustrate the point about how complexity makes it easy for bugs to creep in.

Finally, I'm not sure the concern about performance is warranted.  This code *is* on a hot path, but that path has been profiled heavily and running fine previously doing full json-parsing on log lines (which by default are json objects).  Previous profiling revealed that turning off json parsing and just uploading the raw line didn't have much impact on overall performance (for reference, the biggest hit to performance on hot paths is debug statements that are never printed).  

I'd also caution against any performance issues without any profiling to compare versions because it can reveal interesting things.   In the most recent round of performance improvments, the performance hit was on something completely unexepected (the json response from the docker/k8s apis didn't specify an encoding so python was doing encoding detection on the returned string (some 2MB+) and that was causing long delays).

Out of curiousity, I did some quick profiling of both the current version of this function and the suggested changes (with the OOB error fixed) and it revealed:

* string slicing has negliable impact on performance here - the biggest hit to performance in that function is the conversion of the timestamp to nanoseconds and the amount of slicing going on is insignificant compared to that.

* when removing the conversion to nanoseconds and just returning the raw string for that field, the original version of the code runs faster than the suggested changes (twice as fast on my machine).

I'll send you the test harness I used separately so you can check the results yourself, but assuming I haven't made an error in how I was timing it (just using timeit), then given that the original code is easier to read and also appears to be faster, I think it's probably ok to leave it as is.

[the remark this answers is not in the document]
echee2: Good points.   The locs-array was just an example you can of course assign explicit variables to the location indexes, as well as the slices, and return a 4-tuple of nice variable names.  

Readability matters aside, I fully agree about avoiding premature performance optimization.   However, my concern is not performance (speed), rather memory leaks.  IIUC we have had past problems where the cause is still unknown and the fix was to explicitly force garbage collection. 

With this in mind, I feel further discussion on this point is warranted:

- Go creates slice-by-reference.  It [does not copy data](https://blog.golang.org/go-slices-usage-and-internals).
- In contrast, python makes [a completely new copy](https://stackoverflow.com/questions/5722006/does-python-do-slice-by-reference-on-strings).

Imagine a situation where these logs are heavy and each line is long -- say 1000 chars. The go example will chop off a few chars for the metadata columns.  And each time it does, it creates entirely new strings, say, 990, 980, 970 chars long.  Multiply that by a large number of lines and you get the picture.  

In contrast, the approach I propose is simply slice exactly what you need without creating these intermediate long strings:-
3 short slices for the metadata
A single 970 char slice for the log line.

In summary,  I feel there is really no downside and this isn't an optimization per-se, rather avoiding creating unnecessary objects given how Python string slices work.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1678,6 +1918,9 @@ def get_user_agent_fragment(self):
             ver = k8s_cache.get_api_server_version()
             if ver:
                 ver = 'k8s=%s' % ver
+            runtime = k8s_cache.get_container_runtime()
```
echee2: I packaged the most recent code into a k8s image and deployed to minikube.
I see this error on startup and the the k8s & k8se log files remain 0 length.

image

I think at least one thing to fix is to set `self._container_runtime = None` in `KubernetesCache`

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -159,9 +161,25 @@
...
+                      'e.g. `json` or `cri`.  Log files will be parsed in this format before uploading to the server '
+                      'to extract log and timestamp fields.  If `raw`, the raw contents of the log will be uploaded to Scalyr.',
+                      convert_to=str, default='auto', env_aware=True)
+
+define_config_option( __monitor__, 'k8s_always_use_cri',
+                     'Optional (defaults to False). If True, the kubernetes monitor will always try to read logs using the container runtime interface '
+                     'even when the runtime is detected as docker',
+                     convert_to=bool, default=False, env_aware=True)
+
+define_config_option( __monitor__, 'k8s_cri_query_filesystem',
```
echee2: There's still one failing unit test (that wasn't pre-existing).  This one's an easy fix and related to my recent env variable test.  Basically just need to add the newly env-aware variables into `test_environment_aware_module_params`.


image

scalyr_agent/log_processing.py
```diff
@@ -493,8 +547,22 @@ def readline(self, current_time=None):
...
         #                                              expected_size, actual_size)
 
-        # check to see if we need to parse the line as json
-        if self.__parse_as_json:
+        # check to see if we need to parse the line as cri.
+        # Note: we don't handle multi-line lines.
+        if self.__parse_format == 'cri':
+            timestamp, stream, tags, message = _parse_cri_log( result.line )
+            if message is None:
+                log.warn("Didn't find a valid log line in CRI format for log %s.  Logging full line." %
```
echee2: I created a test cluster in EKS and deployed your branch.   This is docker (not containerD).  The only environment variables I've set are May have found a bug? =>  `kubernetes_monitor.log` is empty.  `agent.log` suggests it has gone into CRI mode (but this is docker):

```
2019-04-19 01:11:28.672Z ERROR [monitor:kubernetes_monitor] [kubernetes_monitor.py:768] Error connecting to kubelet API endpoint - HTTPConnectionPool(host='192.168.45.216', port=10255): Max retries exceeded with url: /pods (Caused by NewConnectionError('<scalyr_agent.third_party.requests.packages.urllib3.connection.HTTPConnection object at 0x7fca8dc52fd0>: Failed to establish a new connection: [Errno 111] Connection refused',)). The Scalyr Agent will now monitor containers from the filesystem
...
2019-04-19 01:11:38.182Z WARNING [core] [log_processing.py:556] Didn't find a valid log line in CRI format for log /var/log/containers/currencyservice-668f49f985-kwlnm_default_server-a50fbbd13977584b58efa8f96a6be4c48784edb1029c41394036446aa1fd8d08.log.  Logging full line.
...
```

image

I also pulled in my status-v PR which verifies only the following env vars set:

image

imron: Can you invite me to look at the logs on your account?

echee2: I've emailed you a .tgz of configs, logs, statusv output

echee2: Update:  I deployed the exact image on a different test k8s docker cluster (that @czerwingithub normally uses). This time it works fine.  My scripts are pretty much exactly the same (for both clusters).  The only difference might possibly be the RBAC service account settings which I have not changed for a long time in the failing cluster? (I will inspect them for differences).

Regardless, it sounds to me what might be happening is agent fails to ping kublet API, then causing it  to enter CRI parsing mode.  The files in `/var/log/containers` are there but are in JSON format hence the errors.

imron: > it sounds to me what might be happening is agent fails to ping kublet API, then causing it to enter CRI parsing mode

I'm not sure that's it, because even if it enters CRI mode, docker provides CRI support and the monitor should detect that it is docker under CRI and still parse as json.  The `parse_format` is distinct from whether or not we are checking CRI locations for container and log information.

The only reason it should be trying to parse the logs in CRI is if it didn't detect that the containers were running under docker, so I'd like to try and figure out what it thought the container runtime was.

imron: Aaaand the container runtime is of type `unknown` :-/

Which is not `docker` and so it defaults to parsing in CRI format.

Let's move this to a discussion more amenable to back-and-forth conversation.  If you're still up, I'll ping you in Slack.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -159,9 +161,25 @@
...
-                      convert_to=bool, default=True, env_aware=True)
+                      'Deprecated, please use `k8s_parse_format`. If set, and True, then this flag will override the `k8s_parse_format` to `auto`. '
+                      'If set and False, then this flag will override the `k8s_parse_format` to `raw`.',
+                      convert_to=bool, default=None)
+
+define_config_option( __monitor__, 'k8s_parse_format',
+                      'Optional (defaults to `auto`). Valid values are: `auto`, `json`, `cri` and `raw`. '
+                      'If `auto`, the monitor will try to detect the format of the raw log files, '
+                      'e.g. `json` or `cri`.  Log files will be parsed in this format before uploading to the server '
+                      'to extract log and timestamp fields.  If `raw`, the raw contents of the log will be uploaded to Scalyr.',
```
echee2: As discussed, I propose we add a disclaimer warning users not to set this unless they know what they are doing.  For example:
"(Note: An incorrect setting will cause parsing to fail so please leave this as `auto` if in doubt.)"

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -580,17 +586,125 @@ class CRIEnumerator( ContainerEnumerator ):
...
+    def _get_containers_from_filesystem( self, k8s_namespaces_to_exclude ):
+
+        result = []
+
+        # iterate over all files that match our container glob
+        for filename in glob.iglob( self._container_glob ):
+
+            # ignore any files that don't match the format we are expecting
+            m = self._info_re.match( filename )
+            if m is None:
```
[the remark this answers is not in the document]
imron: Not here because it's DEBUG_LEVEL_2, and anything DEBUG_LEVEL_1 or above goes in a separate log file that does not get uploaded to Scalyr servers, and in the event we need to turn on debugging at this level we'll probably want to see all debug messages without limits.

For reference, anything with `error`, `warn`, `info` or `DEBUG_LEVEL_0` will probably need a limit because those things all go to the main agent.log.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Yes.  That was an oversight.  I've also added `agent_container_id = None` too, both of which should be set to `None` in `__init__`.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Fixed.

scalyr_agent/log_processing.py
imron: The agent now checks to see if it can detect docker by other means, and also sets the parse format to json if that's the case.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Done.  But modified slightly to add what will happen if parsing fails:

> Note: An incorrect setting can cause parsing to fail which will result in raw logs being uploaded to Scalyr, so please leave this as `auto` if in doubt.

imron:
Ready for review again.  Note: rebased from current master so any old working copies/branches need to be reset.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
echee2: 👍

scalyr_agent/log_processing.py
echee2: I deployed a fresh image with your latest branch to my test EKS cluster.  Verified JSON docker json logs are being parsed as json.

image

Note to self:  When we have automated end-to-end testing for k8s, the test should verify JSON is parsed correctly.

scalyr_agent/log_processing.py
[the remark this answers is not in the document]
echee2: Are we still going to reduce the number of slices as discussed, or were you planning on addressing this in a future PR?

echee2:
@imron, I rebuilt the k8s docker image from your latest code and deployed. 
- I saved the previous cluster's log files. Lemme know if you need them.
- New pod is `scalyr-agent-2-fjhv5`
- The kublet API is still disabled as expected.
- But I verified correct `json` mode detection (screenshot provided in separate comment)

In terms of code review, thanks for addressing prior comments. The only outstanding is the string slice.

In terms of readiness for production, the goals would be validate this on IBM cloud / CRI.

I know we're still waiting for IBM support to help grant access to Kublet API.  In the meantime, DYT it's worth getting Joel to test the directory-polling fallback?  IIUC your current code should work out of the box, just no metrics right?

(Also, we should discuss whether https://scalyr.myjetbrains.com/youtrack/issue/AGENT-90 should be part of the next release now that we realize Kublet API could be disabled more frequently than we realize.

scalyr_agent/log_processing.py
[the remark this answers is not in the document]
imron: I did some memory testing with the `guppy` lib and heap usage appears to be identical for both versions.  I'll send you the test harness so you can confirm, but if it looks ok then considering the original is slightly faster, I think we can leave as is for now.

imron:
Ok this is ready for review again and also addresses #AGENT-90

scalyr_agent/log_processing.py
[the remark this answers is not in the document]
echee2: sgtm.

echee2:
@imron, the latest code change LGTM. 

I think to avoid divergence from master we can probably merge even before C.S. finish their testing on IBM as long as we do the following: 
1) Resolving conflicts .  
2) Ensure new code doesn't break existing (non-containerD) agent.

Unit tests don't capture end-to-end sanity check for (2) unfortunately so if you haven't already done so with the latest code changes, you might want to sanity-check on k8s and/or docker json for example. 

Also Bill was wondering on Monday whether there are other containerD platforms besides IBM that we can test on (e.g. ECS/EKS).

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -159,9 +161,27 @@
...
+                      'this as `auto` if in doubt.)',
+                      convert_to=str, default='auto', env_aware=True)
+
+define_config_option( __monitor__, 'k8s_always_use_cri',
+                     'Optional (defaults to False). If True, the kubernetes monitor will always try to read logs using the container runtime interface '
+                     'even when the runtime is detected as docker',
+                     convert_to=bool, default=False, env_aware=True)
+
+define_config_option( __monitor__, 'k8s_cri_query_filesystem',
+                     'Optional (defaults to False). If True, then when in CRI mode, the monitor will query the filesystem for the list of active containers, rather than the Kubelet API.  This is useful in environments where the Kubelet API has been disabled.',
```
echee2: Recommend changing the English a bit for correctness (the existing text is from a previous iteration that did not auto-switch to filesystem I think):

"If True, then when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than **first querying** the Kubelet API.  This is a useful optimization when the Kubelet API is known to be disabled."

PR #169, opened 2019-04-18
AGENT-85: Follow-on minor fixes to AGENT-38.

echee2:
Follow on fixes to AGENT-38:

1. container_globs needs to become an ArrayOfStrings + unit tests
2. Add limit_key to the 2 log lines that use limit_once_per_second but fail to add the limit_key


echee2:
Interesting observation that reinforces the need for automated end-to-end testing ASAP:

unit tests don't detect the following error but when I built an RPM from current master and tried to run I get this error which is almost certainly related to this PR (ie this PR should fix the error):

image

(I was running in a freshly built docker image)

Also if I'm not mistaken, the previous release (and possibly earlier releases) fail with Python 2.4 and Python 2.5 because of the missing `sort_keys` arg in one of the json_lib functions.  Again, because we don't have automated testing across platforms that was not caught.

echee2:
Thanks for the review @imron

PR #171, opened 2019-04-23
Ticket #AGENT-88

imron:
Fixes a problem where we were checking the existence of a key in a dict, but the
variable could sometimes be None.

I'm still not sure of the root cause of this, but I suspect it could be caused
if there is a container that uses a non-json logging driver, meaning it won't
have a log path, and the lack of the log path was causing the monitor not to
create the abovementioned dict.

I've fixed the stack trace and also added extra debug information to help
identify the root cause of the issue.


imron:
Hi Ed,  I know you mentioned you'd kick off a build early Tuesday.  I was wondering if there was still time to get this PR included?

echee2:
Spot fix LGTM.  I'll file a follow up ticket where we should identify the cause.

PR #173, opened 2019-04-24
AGENT-93: circle python versions

echee2:
# WHAT

This PR establishes the framework for multi-python-version unit tests.

It's not meant to fix all tests to be python 24,25,26 compatible. But I do fix one problem (using unittest2 instead of unittest) as an example.  One outstanding incompatibility for python26 has been noted with a comment (pertains to `str.decode`) and I've requested input from @imron 

# BACKGROUND

We need testing across versions.  After attempting various permutations including statically vs dynamically compiled versions of older python, I believe this is best tradeoff that I can get working that will move us forward:

1. create a custom docker image with pre-compiled python 2.x  (for now stored in edwardchee public docker repo but easily moved to scalyr repo later)
2. parallelize CircleCI to run different version tests
3. Include all pip packages IN the build to save startup time (we can modify this later to run dynamic pip installs of missing packages.  This reduces startup time from 2-3 mins -> 1min. But for now, at least this gets us going.)

This PR only has 27 and 26, but establishes the framework for when I can build the 25 and 24 images.

Note, I also studied caching of docker images (without paying for CircleCI's Docker layer caching).  It's certainly doable but current startup without caching is only 1 min. With parallel jobs, it's not urgent.

# HOW TESTED

Python27 - no additional unit test failures
Python26 - A persisting failure is the aforementioned `str.decode` incompatibility.  There are some others like dictionary comprehensions and usage of multiple `with mock as x`

# Notes

Pointers on implementing caching:
https://blog.jondh.me.uk/2018/04/strategies-for-docker-layer-caching-in-circleci/

CircleCI docker layer caching
https://circleci.com/docs/2.0/docker-layer-caching/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@
 #
 version: 2
 jobs:
-  test:
+  test27:
     working_directory: ~/scalyr-agent-2
     docker:
       - image: circleci/python:2.7-jessie-browsers
@@ -35,8 +35,19 @@ jobs:
       - store_artifacts:
           path: cover/
 
+  test26:
+    docker:
+      - image: circleci/python:2.7-jessie-browsers
+    steps:
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          command: docker run -it -e TEST_BRANCH=${CIRCLE_BRANCH} edwardchee/scalyr-agent-ci:python26 /tmp/run_unittests.sh
+
+
 workflows:
   version: 2
   unittest:
     jobs:
-      - test
+      - test27
+      - test26
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,18 +3,18 @@
 # running test suites, and mocking Python Objects for the Scalyr Agent.
 asyncio==3.4.3
 coverage==4.4.1
-mock==2.0.0
+mock==0.8.0
 nose==1.3.7
 nose-exclude
 nose_xunitmp==0.4.0
 parameterized==0.7.0
 pg8000==1.11.0
 pysnmp
-requests==2.20.0
 redis==2.10.6
-Twisted==17.9.0
 ujson==1.35
 unittest2==1.1.0
 virtualenv==15.1.0
 webapp2==2.5.2
 WebTest==2.0.28
+unittest2==1.1.0
+future==0.17.1
--- a/scalyr_agent/all_tests.py
+++ b/scalyr_agent/all_tests.py
@@ -19,7 +19,7 @@
 
 __author__ = 'czerwin@scalyr.com'
 
-import unittest
+import unittest2 as unittest
 import os
 import sys
 import traceback
--- a/scalyr_agent/builtin_monitors/redis_monitor.py
+++ b/scalyr_agent/builtin_monitors/redis_monitor.py
@@ -20,6 +20,7 @@
 __author__ = 'imron@scalyr.com'
 
 import binascii
+import codecs
 import re
 import time
 
@@ -206,7 +207,7 @@ class RedisHost( object ):
         except UnicodeDecodeError, e:
             if self.utf8_warning_interval:
                 logger.warn( "Redis command contains invalid utf8: %s" % binascii.hexlify( entry['command'] ), limit_once_per_x_secs=self.utf8_warning_interval, limit_key="redis-utf8" )
-            command = entry['command'].decode( 'utf8', errors="replace" )
+            command = entry['command'].decode('utf8', errors="replace")
 
         time_format = "%Y-%m-%d %H:%M:%SZ"
         logger.emit_value( 'redis', 'slowlog', extra_fields={
--- a/scalyr_agent/scalyr_logging.py
+++ b/scalyr_agent/scalyr_logging.py
@@ -175,7 +175,7 @@ if hasattr(sys, '_getframe'):
 
 
 # noinspection PyPep8Naming
-class AgentLogger(logging.Logger):
+class AgentLogger(logging.Logger, object):
     """Custom logger to use for logging information, errors, and metrics from the Scalyr agent.
 
     These instances are returned by scalyr_agent.getLogger.  You should use that method in place of logging.getLogger.
--- a/scalyr_agent/scalyr_logging.py
+++ b/scalyr_agent/scalyr_logging.py
@@ -175,7 +175,7 @@ if hasattr(sys, '_getframe'):
 
 
 # noinspection PyPep8Naming
-class AgentLogger(logging.Logger, object):
+class AgentLogger(logging.Logger):
     """Custom logger to use for logging information, errors, and metrics from the Scalyr agent.
 
     These instances are returned by scalyr_agent.getLogger.  You should use that method in place of logging.getLogger.
--- a/scalyr_agent/tests/configuration_test.py
+++ b/scalyr_agent/tests/configuration_test.py
@@ -1428,7 +1428,7 @@ class NullHandler(logging.Handler):
         pass
 
 
-class FakeAgentLogger(AgentLogger):
+class FakeAgentLogger(AgentLogger, object):
     def __init__(self, name):
         super(FakeAgentLogger, self).__init__(name)
         if not len(self.handlers):
--- a/scalyr_agent/tests/configuration_test.py
+++ b/scalyr_agent/tests/configuration_test.py
@@ -1423,11 +1423,16 @@ class TestConfiguration(ScalyrTestCase):
         return self.make_path(None, path)
 
 
+class NullHandler(logging.Handler):
+    def emit(self, record):
+        pass
+
+
 class FakeAgentLogger(AgentLogger):
     def __init__(self, name):
         super(FakeAgentLogger, self).__init__(name)
         if not len(self.handlers):
-            self.addHandler(logging.NullHandler())
+            self.addHandler(NullHandler())
 
 
 class FakePlatform(object):
--- a/scalyr_agent/json_lib/tests/encode_decode_test.py
+++ b/scalyr_agent/json_lib/tests/encode_decode_test.py
@@ -18,7 +18,7 @@
 __author__ = 'echee@scalyr.com'
 
 
-import unittest
+import unittest2 as unittest
 
 from scalyr_agent import util
 from scalyr_agent.json_lib import JsonObject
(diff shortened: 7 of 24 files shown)


.circleci/config.yml
```diff
@@ -35,8 +35,19 @@ jobs:
...
           path: cover/
 
+  test26:
+    docker:
+      - image: circleci/python:2.7-jessie-browsers
+    steps:
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          command: docker run -it -e TEST_BRANCH=AGENT-93-circle-python-versions edwardchee/scalyr-agent-ci:python26 /tmp/run_unittests.sh
```
echee2: I'm currently using my own public repo to host this custom docker image. We can later move it under the public Scalyr account but I don't yet have authorization.  No security risk for now (this image has nothing sensitive, just 26-compiled python & pip)

dev-requirements.txt
```diff
@@ -3,18 +3,18 @@
 # running test suites, and mocking Python Objects for the Scalyr Agent.
 asyncio==3.4.3
 coverage==4.4.1
-mock==2.0.0
+mock==0.8.0
```
echee2: Downgraded mock library to work with lower versions of python. Verified it still works with 2.7

scalyr_agent/builtin_monitors/redis_monitor.py
```diff
@@ -206,7 +207,11 @@ def log_entry( self, logger, entry ):
         except UnicodeDecodeError, e:
             if self.utf8_warning_interval:
                 logger.warn( "Redis command contains invalid utf8: %s" % binascii.hexlify( entry['command'] ), limit_once_per_x_secs=self.utf8_warning_interval, limit_key="redis-utf8" )
-            command = entry['command'].decode( 'utf8', errors="replace" )
+            command = entry['command'].decode('utf8', errors="replace")
+            # echee: The following replacement should work as a pre 2.7
+            # replacement for the above line but I'm not comfortable activating
+            # it until we have a unit test given that it could mess up REDIS
+            # command = codecs.decode(entry['command'], 'utf8', 'replace')
```
echee2: I _think_ this alternative line is equivalent to the existing (line 210) line which is not python 2.6 compatible because of the kwarg.  However I'm nervous about including this fix until confident that we have test coverage.  WDYT?

scalyr_agent/all_tests.py
```diff
@@ -19,7 +19,7 @@
 
 __author__ = 'czerwin@scalyr.com'
 
-import unittest
+import unittest2 as unittest
```
echee2: Unittest2 needed for backward compatibility

scalyr_agent/builtin_monitors/redis_monitor.py
imron: I'm not sure the alternative line works.  According to the [2.6 docs](https://docs.python.org/2.6/library/codecs.html#codecs.Codec.decode), the encoding is not specified as a parameter, rather it appears you need to manually get the decoder for a specific encoding from a factory function [`get_decoder`](https://docs.python.org/2/library/codecs.html#codecs.getdecoder).

I'll do some manual testing shortly, but we may need to look at writing our own wrapper shims that get loaded depending on the version.

Also, is this the only place we use `decode`?  I think it appears in other places in the code also.

echee2: I think you're right that it doesn't work.  I must have been mistakenly looking at 2.7 `Codec` documentation.  And yes we'll have to fix elsewhere.  (I think the unit test suite just exercises this one and the others might not be covered.)

imron:
I noticed that a bunch of code in the `third_party` repos has also been modified to work with unittest2.

Are you planning to run the tests on the third_party repos as part of our tests?  We have not traditionally done this, and `run_tests.py` only runs our own tests (well, technically, any file that ends in `_test.py` which only our tests do).

I'm not sure how much effort we should expend on making third_party tests work on all previous versions.  Customers that use a monitor that relies on a third party lib can be assumed to be running a version of python that supports it.  Customers that would like to use a monitor that uses a third party lib that doesn't run on their older version of python should be encouraged to update to a newer version of python, or we can look at compatibility options when it becomes an issue for them (which in the 4 years I've been working on the agent has been never).

If we aren't going to run third_party tests, then I think anything in `third_party` should be left as is.

I think there's room for further discussion in which approach we wish to take.

echee2:
Ah good catch.  Didn't mean to modify third_party.  (Pycharm did a replace and since third_party is a subdir they got replaced). I'll fix.

As an aside, we should start thinking about Bill's suggestion to remove unmodified 3rd party code from our repo.  (We can still include it in in the published packages).

scalyr_agent/scalyr_logging.py
```diff
@@ -175,7 +175,7 @@ def alternateCurrentFrame():
 
 
 # noinspection PyPep8Naming
-class AgentLogger(logging.Logger):
+class AgentLogger(logging.Logger, object):
```
imron: Does logger not already inherit from object?

Looking through the 2.7 source, it would appear to go `object -> logging.Filterer -> logging.Logger -> AgentLogger`

scalyr_agent/tests/configuration_test.py
```diff
@@ -1423,11 +1423,16 @@ def convert_path(self, path):
         return self.make_path(None, path)
 
 
+class NullHandler(logging.Handler):
```
imron: It looks like `FakeAgentLogger` is identical to the one in `scalyr_agent/test_util.py`.  Can we just import that one rather than defining it again (same then will `NullHandler` and `FakePlatform`)

echee2: Yup, I'm aware of the code duplication . (This is from an earlier PR. There was some reason why I chose to leave it in there for the time being but with intention to deduplicate.  I'll deduplicate in a different PR . (There's actually more stuff that needs fixing/cleaning up especially pertaining to python backwards compatibility.  For example with context manager usage.  I should try to keep all that separate from this PR and have this PR be mainly about the CircleCI config.)

scalyr_agent/scalyr_logging.py
echee2: I did see that and was initially confused too.  But when run on 2.6 it fails (not 2.7) . So it has to be the case that one of the parent classes in 2.6 fails to inherit object.

echee2:
@imron Thanks again for catching the 3rd party issue.  I believe I've fixed all of them (unittest imports).  

I'll refactor/dedup the test utility classes in another soon-to-come PR.

echee2:
Also removed the inadequate suggestion for `str.decode` backward compatibility.

scalyr_agent/scalyr_logging.py
imron: You're right. I just looked through the 2.6 source, and yes, it turns out `Filterer` doesn't inherit from `object`.

What problem were you encountering without inheriting from object and what happens in python if the same parent class exists at multiple points in the inheritence hierarchy?

I probably wouldn't be asking this if it was happening on a class in the test library, but AgentLogger is used everywhere, so just want to make sure there aren't going to be any side-effects in 2.7, which presumably is the common use case for all our clients.

Also, if it's just a testing thing, is it possible to fix the problem in 2.6 by having the `FakeAgentLogger` inherit from object instead, therefore not affecting `AgentLogger` at all?

scalyr_agent/tests/configuration_test.py
imron: Sounds good.

.circleci/config.yml
imron: Do you not have access to dockerhub from doing the build process?  If you can publish the official scalyr-agent builds, you should be able to add other public repos too.

Also, with regards to this, I think we should change the name of the image slightly.  Common practice generally seems to be to include the tech stack in the name, and use the tag for the version only e.g. assuming the image gets moved to the public scalyr repo, something like `scalyr/scalyr-agent-ci-python:2.6`

imron: P.S commands for doing this would be:

`docker login`
(username `scalyr`, password `<you have this already somewhere>`

`docker push scalyr/scalyr-agent-ci-python:2.6`

imron: And other thought, do you have the Dockerfile you used to generate this image?  That should probably be added to the main repo too, maybe under the `docker` directory.

echee2: Let's hold off docker image naming/Dockerfile as those are in flux and might change in future so we should avoid unnecessary code review.  Also not just a single dockerfile but current implementation (which in future might change) requires a .tgz with compiled python binaries -- since this repo is public and really should be for agent-specific code, others cloning this repo shouldn't have to download largish unrelated files.

I'm also starting a separate "ingest devops" private scalyr repo for CD/release automation related files and the above might be better suited there. 

Can discuss more in next standup.

scalyr_agent/scalyr_logging.py
echee2: This is a good thought. 

I wouldn't _think_ there'd be problems with repeat inheritance but definitely needs to be better understood. 

Lemme fire up 2.6 and look more closely again at whether it's possible to isolate this to only test code.

imron: > I wouldn't think there'd be problems with repeat inheritance but definitely needs to be better understood.

That would be my assumption too, but I don't know enough about the details to be sure.  If we can isolate it to test only code that'd be the easiest way to deal with this for now.

.circleci/config.yml
echee2: BTW goal of this PR is so we can start seeing 26 agent code errors (and then 25, 24) as soon as possible.  For example, once this PR is merged, anyone can start making code fix PRs and immediately how it affects 26. 

So, note that I'm making a conscious decision to defer any devops / docker image issues and optimizations for the future. After 25,24, the top priority is install release automation.

Once we have all automation functionally working, we can chip away at cleanup or things like docker image caching.  For example, we might spend time to find a CircleCI python26 image and replace what I've installed with that.

imron: Yep.  Agree this doesn't need to be addressed yet, just wanted to be sure we are thinking about it.

scalyr_agent/scalyr_logging.py
echee2: Thanks again for this observation/catch.   Was able to move to `FakeAgentLogger` . btw -- the error happens when calling superclass __init__

image

imron: That makes sense.  `super` only works with classes derived from `object`.

All looks good.  I'll approve this in a minute.

echee2:
Thanks for the detailed review @imron.  With this merge we can start chipping away at 2.6 incompatibilities, if any.   Also I just noticed a test failure I didn't see earlier, which we should also address.

image

PR #175, opened 2019-04-24
Ticket: #AGENT-83

imron:
Fixed problem with hashed redaction rules, where lines that didn't match any
rule were discarded, and also problems where lines were truncated after the
final matched group of a line.

I'm not sure how many customers use this feature, but from what I can tell, it's
been broken like this since the very beginning.

### DIFF
--- a/scalyr_agent/tests/log_processing_test.py
+++ b/scalyr_agent/tests/log_processing_test.py
@@ -1357,6 +1357,51 @@ class TestLogFileProcessor(ScalyrTestCase):
         self.assertFalse(completion_callback(LogFileProcessor.SUCCESS))
         self.assertEquals(events.get_message(0), 'GET /foo&password=foo&start=true\n')
 
+    def test_hashed_redaction_in_middle_of_line(self):
+        # Testing for #AGENT-83 indicated that redaction rules with hashing was truncating
+        # lines after the end of the final hash.  This is an explicit test to catch that
+        # problem
+        log_processor = self.log_processor
+        log_processor.add_redacter( "\"([A-F0-9]{32})(\\.app1.)\"", "\"*****\\H2\"" )
+
+        line_base = '127.0.0.1 - - [15/Apr/2019:16:30:42 -0700] "GET /10/ HTTP/1.1" 200 113020 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/73.0.3683.86 Safari/537.36" D=63388 R=app1b/- '
+
+        line_unredacted = line_base + '"0123456789ABCDEF0123456789ABCDEF.app1b" something extra\n'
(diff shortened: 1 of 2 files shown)


scalyr_agent/tests/log_processing_test.py
```diff
@@ -1357,6 +1357,51 @@ def test_redaction_rule(self):
...
+        log_processor.add_redacter( "\"([A-F0-9]{32})(\\.app1.)\"", "\"*****\\H2\"" )
+
+        line_base = '127.0.0.1 - - [15/Apr/2019:16:30:42 -0700] "GET /10/ HTTP/1.1" 200 113020 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/73.0.3683.86 Safari/537.36" D=63388 R=app1b/- '
+
+        line_unredacted = line_base + '"0123456789ABCDEF0123456789ABCDEF.app1b"\n'
+        line_expected = line_base + '"*****9e174c63530b55ef27fcbfdbdde9c403"\n'
+        self.append_file( self.__path,
+            'ello\n',
+            line_unredacted
+            )
```
echee2: Could we tighten the formatting a bit here?  In addition to the nonstandard indent of the last paren, each arg is short enough that you might consider fitting in a single line.

On this note, another tool I would very much like us to install a code formatter (Black). This will eliminate any time wasted on subjective preferences and standardize our formatting for easier reading.

Some background on why code formatting:
https://medium.com/@ryconoclast/why-you-should-use-a-code-formatter-4f02dd40db14

In my previous company, I found flake8 to be extremely useful.  Even though we didn't use Black for the python code base, a similar code formatter  called `Prettier` was extremely useful for a large React project I worked on.  The uniformity eases code maintenance, readability, grepping for string patterns when debugging. It's especially useful in major refactors such as python 2->3.  

Black goes one step further than flake8.  Instead of just giving errors on violations, Black can actually reformat code.  Black can run within popular editors including Pycharm/ Vim to help formatting as you code. 

I feel appearances can make a difference for the agent, which is often the initial customer touch point.  Since the agent is open-source, it's to our benefit to make our code "look" as professional as possible -- even if the difference is just formatting. (These days, this can make a difference when a decision maker is a coder.)

https://www.mattlayman.com/blog/2018/python-code-black/

imron: I've fixed up the formatting.  That was a hangover from when I first wrote the test and *did* have long lines for each line I was appending to the file.

I'd be ok with a code formatter, the only potential issue I'd see is that running it initially would change a large number of lines, and that obscure the code history a little bit e.g. after running the formatter and committing, then doing `git blame`, you'll get the commit of the formatting change rather than the commit of the original code.  It's not a big issue, just something to be aware of.

Maybe it's something we can introduce as part of the 2->3 conversion, which will be touching most lines of code also.

PR #178, opened 2019-04-25
AGENT-96: Deduplicate some test classes

echee2:
Follow up to this discussion: https://github.com/scalyr/scalyr-agent-2/pull/173#discussion_r278401848

Consolidating some test classes that were duplicated because of prior divergence between code branches (sitting in review)


echee2:
Friendly ping @imron . (should be an easy review)

PR #179, opened 2019-04-26
Agent smoke tests

echee2:
IMO we really need these smoke tests.

# What

Launches a docker image wherein agent is installed and started up in a way that mimics a customer installing it.

Writes some test JSON into a test file. Agent is configured to look for this file.

Query Scalyr backend up to a certain time limit searching for specific log query results that would have been only uploaded from this host.

# What does this buy us?

Devs can tell if their code change has caused catastrophic failure of agent.

For example, 2.4 and 2.5 don't work currently as you can see from the test results.

# Implementation

This PR uses a self-contained docker image which contains smoke-test code as follows:
1. Build agent RPM from our `build_package.py` script.  
2. RPM install 
3. Activate python 2.x *without* any pip libraries installed (so that our included third_party lib code is used)
4. Populate a test file with 10K lines JSON -> qatesting.
5. Switch back to python2.7, using requests lib to query Scalyr backend.  Look for specific line that would have been uploaded.

# Orthogonal stuff

Kinda orthogonal to this PR but too lazy to start another one -- I had to further isolate environment variables from `configuration_test` as the *unit test* was picking up the CircleCI secret/environment variables used by the smoketest and failing.


scalyr_agent/tests/configuration_test.py
```diff
@@ -45,8 +45,9 @@ def setUp(self):
         self.__config_file = os.path.join(self.__config_dir, 'agent.json')
         self.__config_fragments_dir = os.path.join(self.__config_dir, 'agent.d')
         os.makedirs(self.__config_fragments_dir)
-        if os.environ.get('scalyr_api_key'):
-            del os.environ['scalyr_api_key']
+        for key in os.environ.keys():
+            if 'scalyr' in key.lower():
+                del os.environ[key]
```
echee2: Remove all env vars with `SCALYR` in them (not just scalyr_api_key)

echee2:
I will post some info on smoketest image on private agent-dev slack in case you're interested.

imron:
Yep, I was about to ask about that just now :-)

Specifically what smoketest.sh is doing.

Imron

PR #184, opened 2019-04-28
AGENT-103: Reduce docker images (alpine + python27 + ujson)

echee2:
@imron @czerwingithub I have incorporated @till's  and modified slightly to install ujson.

Latest more complete docker/README.md merged from master.

I ran the docker json/syslog smoketests locally and they passed.
Tested k8s manually and verified logs are uploaded to qatesting.

### DIFF
--- a/build_package.py
+++ b/build_package.py
@@ -88,7 +88,7 @@ def build_package(package_type, variant, no_versioned_file_name):
             # backward compatibility.)  We also publish this under scalyr/scalyr-docker-agent-syslog to help
             # with the eventual migration.
             artifact_file_name = build_container_builder(variant, version, no_versioned_file_name,
-                                                         'scalyr-docker-agent.tar.gz', 'docker/Dockerfile',
+                                                         'scalyr-docker-agent.tar.gz', 'docker/Dockerfile.syslog',
                                                          'docker/docker-syslog-config',
                                                          'scalyr-docker-agent-syslog',
                                                          ['scalyr/scalyr-agent-docker-syslog',
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,9 +1,5 @@
-FROM ubuntu:16.04
-ENV DEBIAN_FRONTEND noninteractive
+FROM python:2.7-alpine3.9 as scalyr
 MAINTAINER Scalyr Inc <support@scalyr.com>
-RUN apt-get update && \
-    apt-get install -y python && \
-    apt-get clean
 COPY scalyr-docker-agent.tar.gz /tmp/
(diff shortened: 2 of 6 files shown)


till:
Are these python versions?

echee2:
Hello @till .  Yes, these are python versions.  As you can see we have a few bugs that have crept in unnoticed so at least we know to fix them now.  Thank you for these suggestions by the way and your PR generally looks good (your PR doesn't seem to cause any additional problems beyond pre-existing).  

We might have something more to add to the readme.   I'll check with the rest of the team to see if they have any thoughts and get back to you.

till:
@echee2 Sounds good, thanks, let me know if you need anything else. I'm running with my fork for now, but would love to get rid off it. :)

echee2:
Thanks @till . We'll also be including `ujson` for more performant JSON parsing.  This should be incorporated by the next agent release sometime next week.

echee2:
@imron  does this LGTY?  If so I can merge this and start a bake-in test tomorrow morning.

imron:
Hi Ed,  I pushed a new commit that addresses a number of issues in the Dockerfiles.  Details are in the commit notes.  I'm not sure how this gets reviewed now?  I can't add you as a reviewer so maybe @czerwingithub needs to have a look at it?

The one concern I still have is exposing the syslog ports by default in the syslog image.  We weren't doing that previously, and I wonder how it will affect existing syslog installs?

till:
@imron `EXPOSE` serves as documentation, it doesn't do much. You still have to map the ports in compose or `docker run`.

imron:
You are right.  Not sure how I overlooked that.  I'll go back and edit the `README.md`.  The point still stands about not exposing them in the docker-json and k8s images though, because those images don't listen on those ports at all so documenting that they do would be misleading.

Also, I wonder if we should be exposing `514/udp` by default, because by default the syslog agent is only listening on `601/tcp`.

echee2:
@imron good catch on the misleading expose for non-syslog and thanks for restructuring. 

I have verified all images work.  

I have admin privileges and can merge without further review so we can start the bake-in test.

Thanks @till for starting this PR!

echee2:
For record, all unit and smoke tests pass
https://github.com/scalyr/scalyr-agent-2/pull/195

echee2:
Hi @till.  I think we're good.  I merged this branch (smoketest_pr_182) which was based off [your branch](https://github.com/scalyr/scalyr-agent-2/pull/182), which iiuc is I think why you got the impression that @imron's commits didn't make it in ? But his commits a9096d1 and 
4c49636d are in as you can see from the [master branch commit history](https://github.com/scalyr/scalyr-agent-2/commits/master)

PR #185, opened 2019-04-29
AGENT-101: Pass errors='replace' as an arg for pre-2.7 compatibility.

echee2:
Python 2.6 unittests are failing because of 2 issues.

This addresses one of them, namely `str.decode` errors="replace" should be passed as an arg, not a kwarg (only 2.7 onwards supports kwargs)


echee2:
Thanks for thinking about a test case.  Indeed, that test was the failure that alerted us to this.

PR #186, opened 2019-04-29
Ticket: #AGENT-100

imron:
Changed monitor.module_name to monitor_name - this was a side effect of some
refactoring (we used to pass in the monitor) that didn't get picked up.


imron:
I think what you mentioned the other day is right, if possible, we should be looking to add tests for every bug that is fixed.  This whole set of related features (dynamically adding/removing log_configs from the agent) has little/no testing.

Might be worth adding a low-priority ticket to address that, as this functionality is central to docker/k8s.

echee2:
👍for the new testing ticket

PR #189, opened 2019-05-01
AGENT-89: Pre-python 27 code compatibility

echee2:
This PR makes our code run on python 2.5 once again. 

0. **Whitelist certain test cases** that we know use 2.7 or else are incompatible with earlier versions.  (see `all_tests.py`).  Conservatively only skip a test if an exception is encountered when attempting to load the module. Also separated out k8s-related portion of `configuration_test` so it can be whitelist-excluded.

1. **Remove all third_party libraries we fork/clone** from `dev-requirements` to ensure third-party code is always used tests and development.  (Developers should also include third_party library in PYTHONPATH instead of installing potentially divergent newer versions.)

2. Revert unittest2 -> unittest because contrary to its claims, it does *not* work with python 2.5 or earlier (it is dependent on other libs that aren't compatible). 
   - Expand `ScalyrTestCase` to support `assertIs`
   - Create adapters for `unittest.skip`.  (If `unttest.skip` is not present, run custom code that simply prints out "Skipping test"

3. Replace `namedtuple` with a custom implementation in `scalyr_agent/builtin_monitors/linux_process_metrics.py`

4. Change syntax of exception names (from `except Exception as ex` -> from Exception, ex`)

5. Change `with open` -> try/finally

6. Change `with assert.Raises` context manager -> lambda

7. Change `with patch.object` -> function decorator.  Restructure test code as needed.

8. Implement `any` and `defaultdict` for 2.4

9. Remove ternary assignment statements

10. Reference return parts from `urlparse.urlparse` by index, not attribute name.
Python 2.4: https://docs.python.org/release/2.4.4/lib/module-urlparse.html
Python 2.7: https://docs.python.org/2/library/urlparse.html

11. Fixed `Logger.findCaller` to return a 2-tuple instead of 3-tuple for python 2.4

### DIFF
--- a/scalyr_agent/builtin_monitors/tests/docker_monitor_test.py
+++ b/scalyr_agent/builtin_monitors/tests/docker_monitor_test.py
@@ -48,10 +48,11 @@ class DockerMonitorTest(ScalyrTestCase):
             self._DockerMonitor__version_lock = threading.RLock()
             self._DockerMonitor__version = None
 
-        with mock.patch.object(DockerMonitor, '_initialize', fake_init):
+        manager_poll_interval = 30
+        fake_clock = FakeClock()
 
-            manager_poll_interval = 30
-            fake_clock = FakeClock()
+        @patch.object(DockerMonitor, '_initialize', new=fake_init)
+        def create_manager():
             manager = ScalyrTestUtils.create_test_monitors_manager(
                 config_monitors=[
                     {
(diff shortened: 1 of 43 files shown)


scalyr_agent/builtin_monitors/tests/docker_monitor_test.py
```diff
@@ -63,93 +64,95 @@ def fake_init(self):
...
-                    if fragment_polls.count() < 20:
-                        return 'syslog'
-                    else:
-                        return 'docker_api'
-                elif key == 'docker_raw_logs':
-                    if fragment_polls.count() < 30:
-                        return True
-                    else:
-                        return False
+            return manager
```
echee2: This large diff is merely unindenting since it's no longer with a context manager.

(Similarly for the other large diffs in unit tests)

echee2:
Hey @imron 

Thinking it would be good to get this PR in for next release too.   Even though I have not tracked down every failure, most of it is in this PR.  The outstanding are:
1. That `test_losing_read_access` failure in unittest-26 that I had asked for your input on
2. Some weird ifconfig issue with smoke-standalone-24
image

unittest-27 has been passing but still continues to occasionally hang, I suspect because of non-deterministic termination of threads in manager-related tests.  I am inclined to start a small PR to make all threads daemon (only in tests) to ensure termination.

echee2:
OK @imron , I think this PR should be ready for review.  All smoketests are fixed.  Unittest 26 is the sole deterministic failure as discussed (it also fails locally on python 24/25 but I have yet to create the docker images for CircleCI)

Also note unittest-27 still fails non-deterministically. I will fix that in separate PR with daemon threads.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1031,19 +1031,25 @@ def __init__( self, ca_file='/run/secrets/kubernetes.io/serviceaccount/ca.crt',
         try:
             # using with is ok here, because we need to be running
             # a recent version of python for various 3rd party libs
-            with open( token_file, 'r' ) as f:
+            f = open(token_file, 'r')
```
imron: k8s is guaranteed 2.7 or above, so `with` statements should be ok.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1031,19 +1031,25 @@ def __init__( self, ca_file='/run/secrets/kubernetes.io/serviceaccount/ca.crt',
...
+        except IOError:
             pass
 
         #get the namespace this pod is running on
         self.namespace = 'default'
         try:
             # using with is ok here, because we need to be running
             # a recent version of python for various 3rd party libs
-            with open( namespace_file, 'r' ) as f:
+            f = open( namespace_file, 'r' )
```
imron: Same comment regarding `with`

echee2:
Thanks for reviewing @imron . I'm merging this now so we start getting better regression signal.

Note, I think the AGENT-42 PR introduced a regression in unittest-27 (which I'll look at since I wrote the test).

As for context manager usage in `k8s.py`, iirc it may still break tests even though k8s tests are excluded (to to imports). I'll investigate later but low priority.

PR #190, opened 2019-05-02
Imron/agent 110 larger json buffer

imron:
Refactor of parsing as json so that json parsing is encapsulated in a `JsonLineMatcher`

This makes it easier to increase the buffer size for reading a full line of json independently from the maximum line size.

The json parser will now try to read at least `read_page_size` bytes from the file when parsing json lines, even if the `max_line_size` is less than this value (which it is by default).  Lines longer than `max_line_size` are still truncated to `max_line_size` bytes.

### DIFF
--- a/scalyr_agent/line_matcher.py
+++ b/scalyr_agent/line_matcher.py
@@ -21,6 +21,10 @@
 # author: Imron Alston <imron@scalyr.com>
 
 import re
+import scalyr_agent.util as scalyr_util
+import scalyr_agent.scalyr_logging as scalyr_logging
+
+log = scalyr_logging.getLogger(__name__)
 
 __author__ = 'imron@scalyr.com'
 
@@ -54,14 +58,15 @@ class LineMatcher(object):
     """
 
     @staticmethod
-    def create_line_matchers( log_config, max_line_length, line_completion_wait_time ):
+    def create_line_matchers( log_config, read_page_size, max_line_length, line_completion_wait_time ):
         """Creates line matchers based on the config passed in
         see: https://www.scalyr.com/help/parsing-logs#multiline for more info
 
         If no lineGroupers attribute is found, then it defaults to a single line matcher
 
         @param log_config: A JsonObject containing the log config
-        @param max_line_length: The maximum amount to read before returning a new line
+        @param read_page_size: The maximum page size to read - used by line matches that parse a specific line format
+        @param max_line_length: The maximum line length of a log message
         @param line_completion_wait_time: The maximum amount of time to wait
             if only a partial line is ready
         @return - a line matcher object based on the config
@@ -71,6 +76,17 @@ class LineMatcher(object):
 
         #return a single line matcher if line_groupers is empty or None
         if not line_groupers:
+            # return a json line parser if necessary
+            if log_config.get( 'parse_lines_as_json', False ):
+                return JsonLineMatcher(
+                    max_line_length=max_line_length,
+                    line_completion_wait_time=line_completion_wait_time,
+                    read_page_size=read_page_size,
+                    json_message_field=log_config.get( 'json_message_field', 'log' ),
+                    json_timestamp_field=log_config.get( 'json_timestamp_field', 'time' )
+                )
+
+            # otherwise return a normal parser
             return LineMatcher( max_line_length, line_completion_wait_time )
 
         #build a line matcher collection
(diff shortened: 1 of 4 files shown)


imron:
Requesting review from both @echee2  and @czerwingithub because the code touches a bunch of stuff in LogFileIterator that @echee2 is probably not very familiar with yet.

scalyr_agent/line_matcher.py
```diff
@@ -82,34 +116,44 @@ def __init__( self , max_line_length=5*1024, line_completion_wait_time=5*60 ):
         self.__line_completion_wait_time = line_completion_wait_time
         self.__partial_line_time = None
 
-    def readline( self, file_like, current_time ):
+    def readline( self, file_like, current_time, current_file=None ):
+        """
+        Reads a line from a `file_like` object
```
czerwingithub: Document `current_file`.  I know the other arguments aren't documented, but this one is not intuitive to me so worth documenting now IMO.

scalyr_agent/line_matcher.py
```diff
@@ -82,34 +116,44 @@ def __init__( self , max_line_length=5*1024, line_completion_wait_time=5*60 ):
...
                 file_like.seek( original_offset )
-                return ''
+                return LogLine('')
         else:
             self.__partial_line_time = None
 
         return line
 
-    def _readline( self, file_like, max_length=0 ):
+    def _readline( self, file_like, max_length=0, current_file=None ):
```
czerwingithub: `current_file` is not used in this function.  Is it needed?

scalyr_agent/line_matcher.py
```diff
@@ -125,6 +169,135 @@ def _readline( self, file_like, max_length=0 ):
         partial = len( line) > 0 and line[-1] != '\n' and line[-1] != '\r'
         return line, partial
 
+class JsonLineMatcher( LineMatcher ):
+    """
+    Parses a line as JSON
+
+    If the `parse_as_json` configuration item is True then
+    """
+    def __init__( self, max_line_length=5*1024, line_completion_wait_time=5*60, read_page_size=60*1024, json_message_field='log', json_timestamp_field="time" ):
```
czerwingithub: Document arguments.

scalyr_agent/line_matcher.py
```diff
@@ -125,6 +169,135 @@ def _readline( self, file_like, max_length=0 ):
...
+        LineMatcher.__init__( self, max_line_length, line_completion_wait_time )
+
+        self._json_message_field = json_message_field
+        self._json_timestamp_field= json_timestamp_field
+
+    def _truncated_log_line( self, line, max_length ):
+        """
+        Truncates a line if it is greater than `max_length`, and appends a message saying the number of bytes truncted
```
czerwingithub: I don't think we should be adding the truncate message into the actual log content here.  It's questionable, but it could seem like we are impacting the integrity of their data.  In the past, I think we just silently truncate.. or am I wrong about that?

scalyr_agent/line_matcher.py
```diff
@@ -125,6 +169,135 @@ def _readline( self, file_like, max_length=0 ):
...
+
+        @return: a tuple containing the content read from the file_like object and whether
+            this is a partial line or not.  `partial` will only be true if `read_page_size`
+            was exceeded and a full json object wasn't found.
+
+        @rtype: (LogLine, bool)
+        """
+
+        # set some sensible defaults
+        if max_length == 0:
```
czerwingithub: I think use `-1` or `None` to signal no `max_length` was given.  `0` is too close to something that could be passed in.

scalyr_agent/line_matcher.py
```diff
@@ -125,6 +169,135 @@ def _readline( self, file_like, max_length=0 ):
...
+
+        #return early if no line found
+        if len( line ) == 0:
+            return LogLine( '' ), False
+
+        # check for presence of newline chars to see if this is a partial line
+        partial = line[-1] != '\n' and line[-1] != '\r'
+
+        # early abort if this appears to be a partial line
+        if partial:
```
czerwingithub: What's going to happen after we truncate this line?  We still haven't seen the end of the JSON object, so I think we are going to keep reading the buffer and discarding it, in what, max line chunks?  Or is it in the page size chunk?

I do wonder if we should do something better here like keep reading the buffer (in future calls to this method) where we just keep looking for the end of a JSON object .. throwing away everything as we do.  Essentially, I'd like to emit just one bad line and then pick up at the next valid line.

scalyr_agent/line_matcher.py
```diff
@@ -125,6 +169,135 @@ def _readline( self, file_like, max_length=0 ):
...
+        # if we didn't find a valid message key/value pair
+        # throw a warning and treat it as a normal line
+        if message is None:
+            log.warn("Key '%s' doesn't exist in json object for log %s.  Logging full line. Please check the log's 'json_message_field' configuration" %
+                (self._json_message_field, current_file), limit_once_per_x_secs=300, limit_key=('json-message-field-missing-%s' % current_file))
+            return self._truncated_log_line( line, self.max_line_length ), False
+
+        # we found a key match for the message field, so use that for the log message
+        # and store any other fields in the attr dict
+        result = self._truncated_log_line( message, self.max_line_length )
```
czerwingithub: So, if the line is longer than max_line, I believe we will just write out the first part and throw everything else away.  A better approach (and what we normally do I believe) is just to emit the entire line in chunks of max_line size.  So, we'd just return the first portion here, and remember the remaining part of the line for future calls.  Is that doable, or are we not keeping enough state around?

PR #194, opened 2019-05-07
AGENT-102: Use unittest docker image running as user Scalyr

echee2:
I created a new `scalyr/scalyr-agent-ci-unittest` image which fixes the permissions problem.
This image installs python 2.4,5,6,7 using `pyenv`.
With this one image, we can unittest (and potentially smoketest) all versions.
(I will look to combining unittest/smoketest in one image shortly)

This now works for Python 26, 25 and 24.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,25 @@ jobs:
       - setup_remote_docker:
           docker_layer_caching: true
       - run:
-          command: docker run -it -e TEST_BRANCH=${CIRCLE_BRANCH} edwardchee/scalyr-agent-ci:python26 /tmp/run_unittests.sh
+          command: docker run -it -e TEST_BRANCH=${CIRCLE_BRANCH} -e PYTHON_VERSION=2.6 scalyr/scalyr-agent-ci-unittest:1 /tmp/unittest.sh
+
+  unittest-25:
+    docker:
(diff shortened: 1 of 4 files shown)


echee2:
@imron , all green!! 

Do note that I had to remove the `encoding` kwarg from `AutoFlushingRotatingFileHandler`.  AFAICT the `encoding` kwarg wasn't being used so it _seemed_ safe to remove?

Secondly, I'm following the policy of simply excluding tests that use language-incompatible third party libs.  In this case, I had to comment out `URL Monitor` for python 2.4.

Note, this PR is not critical for tomorrow's bake-in image, but it should hopefully be a quick review nonetheless.

echee2:
Thanks for the review!

imron:
> AFAICT the encoding kwarg wasn't being used so it seemed safe to remove?

Yes, it seems that this is not used at all.  I guess when it was written, it was written as a straight wrapper around `RotatingFileHandler`, with identical parameters in the init method.

Curiously there was also a previous fix for this function to get it working with earlier versions of python that left out the `delay` parameter, but I think it would start to get messy to have a similar set of conditionals inside each branch to take care of the `encoding` parameter also.

I think it's fine to leave it out, and if we ever need it (unlikely), we can cross the bridge of adding it then.

PR #197, opened 2019-05-08
Ticket #AGENT-115

imron:
Added the `container_globs_exclude` option to the docker monitor, which is a
blacklist of containers to exclude from monitoring, based on container name.

The blacklist is applied *after* processing `container_globs` to see which
containers should be included.

If `None` then no additional exclusions will be applied.

### DIFF
--- a/scalyr_agent/builtin_monitors/tests/docker_monitor_test.py
+++ b/scalyr_agent/builtin_monitors/tests/docker_monitor_test.py
@@ -19,16 +19,20 @@
 __author__ = 'echee@scalyr.com'
 
 
+import os
 import threading
 
 import mock
 from mock import patch
+from mock import Mock
+
+import scalyr_agent.util as scalyr_util
 
 from scalyr_agent.builtin_monitors.docker_monitor import DockerMonitor
+from scalyr_agent.builtin_monitors.docker_monitor import _get_containers
 from scalyr_agent.test_base import ScalyrTestCase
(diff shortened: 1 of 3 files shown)


imron:
Ready for review.

scalyr_agent/builtin_monitors/tests/docker_monitor_test.py
```diff
@@ -39,6 +43,111 @@ class DockerMonitorTest(ScalyrTestCase):
...
+        expected_ids = [
+            "b8757266fd6a9efcb40a2881215d8d3642e6f8a0cfdcbb941382d002f31dcca4",
+            "343c05b55c1b0cebca6df0309a4892974c5ed3013731433461ffe7442223659a",
+            "58ded582ebf14063bff3b15cd3962035be9904b73318796f1119d71472801d23",
+            "c4eacccbf687f408c708b12fa875ab55f4d6d0aad8e2eec489ba9712e041bd14",
+        ]
+
+        glob_list = {
+            'include': ['*include*'],
+            'exclude': ['*exclude*']
```
echee2: Small thing but I propose shortening the glob to `*exclude*` so that the suffix star actually does something (with current test data it doesn't exercise).

imron: I think we should add a bunch more unit tests for `_get_container` in general.  I'll create a new ticket, and handle the updating the exclude glob as part of that.

PR #198, opened 2019-05-08
AGENT-125: Daemonset and configMap for envFrom

echee2:
Current k8s Daemonset and instructions for defining configMap do not use envFrom to import all environment variables in a configMap. (Instead, individual variables from a configMap are explicitly mapped to an environment variable) using the env directive.

Now that we support an expanded list of environment variables, we should be directing customers toward the strategy of defining all desired environment variables in the (same) scalyr-config configMap, but instead of selectively mapping the keyvals -> environment variable, we should just stock import all keyvals.

The difference is that the configMap keys must now be the actual environment variable name, not an arbitrary string.

I hope to release this with 2.0.47.
The reason I would like this with 2.0.47 is in a couple of days we will hopefully update Agent docs and I can then refer customers to use these YAMLs.

# Testing

Since this PR simply adds 2 unused YAMLs, it should pose no risk.

Additionally, I manually used these 2 files to launch a test k8s cluster (successfully).  All it took was to 
1) replace <your_cluster_name> 
2) add a SCALYR_SERVER env variable.

### DIFF
--- a/k8s/scalyr-agent-2-configmap.yaml
+++ b/k8s/scalyr-agent-2-configmap.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+data:
+  SCALYR_K8S_CLUSTER_NAME: <your_cluster_name>
+  # Add any other Environment variables for environment-aware agent config variables
+  # Most environment variables start with SCALYR_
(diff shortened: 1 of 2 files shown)


echee2:
@imron hopefully this will be a quick one (if no concerns).  Ideally I'd like to include this commit in today's release.

PR #200, opened 2019-05-09
AGENT-128: Reference latest k8s image, not preview

echee2:
While writing docs, I ran the instructions to create a custom k8s image and observed it was referencing the preview image instead of latest.

This PR fixes that. 

Secondly, I realize the k8s.md file is a copy of our officials docs.  Eliminating to avoid getting out-of-sync.


docs/k8s.md
```diff
@@ -1,232 +0,0 @@
...
-
-## Further Reading
-
-### Creating custom Scalyr Agent configurations
-
-The default Kubernetes configuration file for the Scalyr Agent does the following:
-
-* Creates a Daemonset running the `scalyr/scalyr-k8s-agent:latest` container image (this image is available on Dockerhub).  This image differs from our standard docker container `scalyr-docker-agent` in that it has been configured to read the raw container logs from the Kubernetes node rather than relying on syslog or the Docker API.
-
-* Maps `/var/lib/docker/containers` of the node to `/var/lib/docker/containers` on the scalyr-k8s-preview-agent container.  This gives the container access to the raw logs from other pods and containers running on that node.
```
echee2: Note here that it also references preview

docker/Dockerfile.custom_k8s_config
```diff
@@ -4,13 +4,14 @@
 # This Dockerfile expects a gzipped tarball called `agent_config.tar.gz` to be in the local
 # directory.  This tarball should contain the desired configuration files (`agent.json` and all
 # `.json` files in `agent.d`) relative to the `/etc/scalyr-agent-2` configuration directory.
-# You can produce such a tarball using a running Scalyr Agent container by issuing the following command:
+# You can extract such a tarball from a running Scalyr Agent container by issuing the following command:
```
imron: 'extract a tarball' usually has a different meaning from what is meant here - the opposite in fact, because this command creates a new tarball from the config files, whereas extract normally means getting files out of the tarball, so I wonder if there's cause for confusion?

The config option is `--extract-config`, so it's extracting the *config*, and tarring it up, rather than extracting the tarball.

I wonder if 'produce' or 'create' might be the better wording here.  I'll leave it to you to decide.

echee2: Good point. Will revert.

PR #204, opened 2019-05-14
Add fuzz factors in K8s cache

czerwingithub:
To avoid query stampeding the API master, fuzz when
we issue queries to fill and revalidate the cache.

We have two separate controls, one for the start delay
and one for the revalidation.

These are default to off.  We will evaulate if we need
to set them based on K8s cluster size.

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -274,6 +274,14 @@ class Configuration(object):
         return self.__get_config().get_int('k8s_cache_expiry_secs')
 
     @property
+    def k8s_cache_expiry_fuzz_secs(self):
+        return self.__get_config().get_int('k8s_fuzz_cache_expiry_fuzz_secs')
+
+    @property
+    def k8s_cache_start_fuzz_secs(self):
+        return self.__get_config().get_int('k8s_cache_start_fuzz_secs')
+
+    @property
     def k8s_cache_purge_secs(self):
         return self.__get_config().get_int('k8s_cache_purge_secs')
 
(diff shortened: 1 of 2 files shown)


scalyr_agent/configuration.py
```diff
@@ -273,6 +273,14 @@ def k8s_verify_api_queries(self):
     def k8s_cache_expiry_secs(self):
         return self.__get_config().get_int('k8s_cache_expiry_secs')
 
+    @property
+    def k8s_cache_expiry_fuzz_secs(self):
+        return self.__get_config().get_int('k8s_fuzz_cache_expiry_fuzz_secs')
```
echee2: This looks to be a typo.  SHould be `get_int('k8s_cache_expiry_fuzz_secs')`

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -971,7 +994,13 @@ def update_cache( self, run_state ):
...
                 global_log.warn( "Exception occurred when updating k8s cache - %s\n%s" % (str( e ), traceback.format_exc()) )
 
-            run_state.sleep_but_awaken_if_stopped( local_state.cache_expiry_secs )
+            # Fuzz how much time we spend until the next cycle.  This should spread out when the agents query the
+            # API master over time in clusters with a larger number of agents.
+            if local_state.cache_expiry_fuzz_secs > 0:
+                fuzz_factor = random.uniform(0, local_state.cache_expiry_fuzz_secs)
+            else:
+                fuzz_factor = 0
+            run_state.sleep_but_awaken_if_stopped( local_state.cache_expiry_secs - fuzz_factor )
```
imron: What happens if `fuzz_factor` is greater than `cache_expiry_secs`?

imron:
> maybe doing a max(0, <potential_neg_number>)

*adding* rather than *subtracting* the fuzz factor would also work.  I think the cache refresh is not so sensitive that it couldn't be pushed out longer.

scalyr_agent/configuration.py
czerwingithub: Thanks, fixed.

scalyr_agent/monitor_utils/k8s.py
czerwingithub: Yup.  I changed this to do a max with zero so we never try to sleep a negative amount.  We might be sleeping zero.. but .. they asked for it.

PR #205, opened 2019-05-15
Imron/k8s profiling

imron:
This pull request adds the profiling code from our last round of profiling.  At the time it was decided not to merge it in to master, but perhaps that something to consider going forward?

Anyway, profiling can be turned on via the `enable_profiling` global configuration option.  There are also configuration options that can set the amount of time to profile for (`profile_duration_minutes` defaults to `2`) and the interval to wait between profiling sessions (`max_profile_interval_minutes` defaults to `60`)

There are a couple of other options, for setting the name of the log file containing profiling data (defaults to `agent.callgrind`) and also which clock to use for profiling (defaults to `random`, other options are `wall` and `cpu`.  Random will randomly pick `wall` or `cpu`).

### DIFF
--- a/docker/Dockerfile.profile
+++ b/docker/Dockerfile.profile
@@ -1,0 +1,18 @@
+# base image that creates all necessary dependencies for
+# the scalyr-agent
+# NOTE: multi-stage builds require Docker 17.05 or greater
+FROM python:2.7-alpine3.9 as scalyr-dependencies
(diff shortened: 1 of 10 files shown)


imron:
Note:  This was merged from my old forked repo, because that's where the profiling code was.  Going forward I'll be using this branch on the main repo.

PR #208, opened 2019-05-16
Imron/agent 141 removing non existent paths

imron:


echee2:
@imron can you maybe rebase this off master.  Right now it's hanging off a brach that I don't think we want to merge.

imron:
rebased on top of option-b/master

scalyr_agent/copying_manager.py
```diff
@@ -289,24 +289,24 @@ def __path_in_globs( self, path, path_globs ):
 
         return None
 
-    def add_log_config( self, monitor, log_config ):
+    def add_log_config( self, monitor_name, log_config ):
```
echee2: I see at least 3 calls to `add_log_config where the first param is a monitor, not a monitor_name.

image

imron: I see only 1 call in that list that was passing in the monitor rather than the monitor name (this is now fixed), unless I'm missing something, the other 2 are function definitions where self is always the first parameter.

echee2: You're right, LGTM. The tests are still failing because of the bug in option-b/master I presume?

imron:
I believe so.  The tests for this function work when run separately.  I'm fixing the other failing tests now.

imron:
I'm 99% sure.  I'll confirm they work first, and then merge.

imron:
Confirmed the tests all work.

PR #209, opened 2019-05-16
K8s profiling opt1

echee2:
Option 1


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1203,26 +1340,36 @@ def __get_log_config_for_container( self, cid, info, k8s_cache, base_attributes
...
-            pod = k8s_cache.pod( pod_namespace, pod_name )
+
+            rename_vars['pod_name'] = pod_name
+            rename_vars['namespace'] = pod_namespace
+            container_attributes['pod_name'] = pod_name
+            container_attributes['pod_namespace'] = pod_namespace
+
+            # get the cluster name
+            cluster_name = self.get_cluster_name( k8s_cache )
+            if self.__include_controller_info and cluster_name is not None:
```
echee2: Question : Do we have an opportunity to set `_k8s_cn` here? I'm currently running profiling.6 where the env variable is set, and no problems.  However, I _think_ `self.__include_controller_info` is False which is why `_k8s_cn` is not set.

Can we safely eliminate the `self.__include_controller_info` check here and therefore set `_k8s_cn`??

Will that give us UI searcheability by cluster?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -780,10 +829,38 @@ class ContainerChecker( StoppableThread ):
...
+        :type docker_api_version: str
+        :type agent_pod: QualifiedName
+        :type host_hostname: str
+        :type log_path: str
+        :type include_all: bool
+        :type include_controller_info: bool
+        :type namespaces_to_ignore: [str]
+        :type ignore_pod_sandboxes: bool
+        """
+        StoppableThread.__init__(self, name='container checker thread')
```
czerwingithub: Note, I added this init to the super class.  If someone thinks that's not the right thing to do, let me know.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -638,6 +674,21 @@ def get_containers( self, running_or_created_after=None, glob_list=None, k8s_cac
...
+                    # the log path should be a link to the /var/logs/pod/<pod_uid>/N.log
+                    # so read the link.  Note, use os.readlink rather than os.path.realpath
+                    # because we only want to traverse one level of links, rather than finding
+                    # the eventual path on disk (which may not be the /var/logs/pod/<pod_uid>/N.log)
+                    if os.path.islink( log_path ):
+                        link = os.readlink( log_path )
+                        m = self._pod_re.match( link )
+                        if m and cname == m.group(2):
+                            k8s_info['pod_uid'] = m.group(1)
+
```
czerwingithub: Does both the kublet api and filesystem approach not return the pod uid?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1677,6 +1844,13 @@ def __build_metric_dict( self, prefix, names ):
             result["%s%s"%(prefix, name)] = name
         return result
 
+    def __verify_required_env_var(self, env_var_name):
+        if len(os.environ.get(env_var_name, '')) == 0:
+            self._logger.error('ERROR: Missing required environment variable for kubenetes_monitor: %s' % env_var_name)
+            self._logger.error('Please restart with up-to-date Scalyr Agent Daemonset manifest (YAML) file.')
+            raise BadMonitorConfiguration('Required environment variables not set for kubernetes_monitor.',
```
echee2: Need a `%s` to marshall `env_var_name`

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2044,23 +2226,28 @@ def run( self ):
...
 
+            # disable controller info if k8s queries are disabled
+            if self.__k8s_disable_api_server:
+                self.__include_controller_info = False
+
             if self.__report_k8s_metrics:
 
                 runtime = None
                 if k8s_cache:
+                    # TODO(czerwin):  If the api server has been disabled, we need to decide what to do here.
```
echee2: Is this TODO still relevant?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2044,23 +2226,28 @@ def run( self ):
...
                     runtime = k8s_cache.get_container_runtime()
 
                 if runtime == 'docker':
                     self.__client = DockerClient( base_url=('unix:/%s'%self.__socket_file), version=self.__docker_api_version )
                     self.__metric_fetcher = DockerMetricFetcher(self.__client, self.__docker_max_parallel_stats )
 
                 k8s = KubernetesApi(k8s_api_url=self.__k8s_api_url)
-                self.__kubelet_api = KubeletApi( k8s )
+                self.__kubelet_api = KubeletApi( k8s, kubelet_api_host_ip=self._config.get('k8s_kubelet_host_ip') )
```
echee2: kwarg should be `host_ip`, not `kubelet_api_host_ip`

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1182,6 +1208,7 @@ def get_cluster_name( self ):
     def query_api( self, path, pretty=0 ):
         """ Queries the k8s API at 'path', and converts OK responses to JSON objects
         """
+        raise Exception( "K8s api queries disabled" ) # echee
```
echee2: Are you guys OK that we leave this in here to be 100% certain we're not making any API requests?

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1190,6 +1217,10 @@ def query_api( self, path, pretty=0 ):
             pretty = '?%s' % pretty
 
         url = self._http_host + path + pretty
+
+        import traceback
+        global_log.info('k8s api get %s, %s' % (url, traceback.format_exc()))
```
echee2: This code is currently never reached. Consider deleting.

czerwingithub: Ok, should this be removed since we throw the exception above?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Copying comments from slack for posterity:

If we pass `_k8s_cn` then we also need to pass `_k8s_dn` otherwise it gets automatically set to empty by the ingest pipeline.  This messes up the deployment names on the overview page and also leaves the search box without any autocomplete dropdowns.

My original code (version .5 I think) actually *did* include the `_k8s_cn`.  I took it out because of that problem.

We could try to discern the deployment name, and although we could make a decent guess, it adds too much uncertainty to the code - especially when we are hoping to pull in the actual deployment name from the API eventually.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -694,7 +743,7 @@ def _get_containers_from_filesystem( self, k8s_namespaces_to_exclude ):
                 continue
 
             # ignore the scalyr-agent container
-            if cid == self._container_id:
+            if self._ignore_pod is not None and pod_name == self._ignore_pod.name and pod_namespace == self._ignore_pod.namespace:
```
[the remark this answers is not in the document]
czerwingithub: Ah, yes, I guess you are right.  We can change that.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -746,7 +795,7 @@ def _get_containers_from_kubelet( self, k8s_namespaces_to_exclude ):
                     short_cid = _get_short_cid( cid )
 
                     # ignore the scalyr-agent container
-                    if cid == self._container_id:
+                    if self._ignore_pod is not None and pod_name == self._ignore_pod.name and pod_namespace == self._ignore_pod.namespace:
                         global_log.log( scalyr_logging.DEBUG_LEVEL_2, "Excluding container because cid '%s' is the scalyr_agent container" % short_cid )
```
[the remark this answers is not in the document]
czerwingithub: True, we can do that minor fix.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -855,14 +943,52 @@ def _is_running_in_docker( self ):
...
+        return (k8s_cache and k8s_cache.get_cluster_name()) or None
+
+    def _get_node_name( self ):
+        """ Gets the node name of the node running the agent """
+        node = os.environ.get( 'SCALYR_K8S_NODE_NAME' )
+        if node:
+            return node
+
+        result = None
+        if k8s_cache:
```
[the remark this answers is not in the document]
czerwingithub: Yeah, I saw this too ... we now fail starting up the monitor if the `SCALYR_K8S_NODE_NAME` environment variable is not set.. so this code path never gets run. We should still remove this dead code.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
czerwingithub: Ok, we should fix.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
czerwingithub: Yes.. if we turn on `report_k8s_metric`, then this part of the code will fail because `k8s_cache` is None if the api server is disabled.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
czerwingithub: Good catch.. ok, we should fix.

scalyr_agent/monitor_utils/k8s.py
imron: I say yes, for these reasons.

1. We'll get a searchable message in agent.log that we can track and fix
2. The big problem for Bukalapak is the CPU and network on master so I think we should try and avoid hitting the API at all for this.  If this causes us to miss out on a few logs, I feel that's a better option than having a high CPU and network load caused by unexpected queries.

Presumably we haven't seen this message at all when running the .6 on the test cluster?  If so I think we're ok.

czerwingithub: I think if we run this build on the load test cluser and this doesn't trip, I'm 100% certain.  I'm about 99% certain now.  But, maybe we should just replay this with a log message for now.  I'm open to that.  If there was some trouble we hit in tomorrow's run, we'd have to do a code change to remove this.

echee2: OK I will remove the exception raise and replace with a logging statement.   Since we have run this code on a 200 node vs 1 master cluster and it's clearly not pounding the master, it should be safe.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: The filesystem approach doesn't - this is the filesystem approach for getting the pod uid.

However you are right an it looks like the kubelet api is not currently returning the pod uid.  I can move this in to the `_get_containers_from_filesystem` and add pod uid information in to the call to `_get_containers_from_kubelet`.

docker/Dockerfile.profile
```diff
@@ -0,0 +1,18 @@
+# base image that creates all necessary dependencies for
```
imron: This file can be removed entirely.  Its changes have already been incorporated in to the main Dockerfiles.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -404,12 +412,13 @@ def _ignore_old_dead_container( container, created_before=None ):
...
     return False
 
-def _get_containers(client, ignore_container=None, restrict_to_container=None, logger=None,
+def _get_containers(client, ignore_container=None, ignore_pod=None, restrict_to_container=None, logger=None,
                     only_running_containers=True, running_or_created_after=None, glob_list=None, include_log_path=False, k8s_cache=None,
                     k8s_include_by_default=True, k8s_namespaces_to_exclude=None, ignore_pod_sandboxes=True, current_time=None):
     """Queries the Docker API and returns a dict of running containers that maps container id to container name, and other info
         @param client: A docker.Client object
         @param ignore_container: String, a single container id to exclude from the results (useful for ignoring the scalyr_agent container)
+        @param ignore_container: QualifiedPod, a pod name and namespace to exclude from results (useful for ignoring the scalyr_agent container)
```
imron: should be `@param ignore_pod`

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -566,14 +582,21 @@ class DockerEnumerator( ContainerEnumerator ):
     Container Enumerator that retrieves the list of containers by querying the docker remote API over the docker socket by usin
     a docker.Client
     """
-    def __init__( self, client, container_id ):
-        super( DockerEnumerator, self).__init__( container_id )
+    def __init__( self, client, ignore_pod ):
+        """
+        @param client The docker client to use for accessing docker
```
imron: Nitpick - missing : after `@param client`

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -566,14 +582,21 @@ class DockerEnumerator( ContainerEnumerator ):
...
+        @param ignore_pod A pod whose containers should not be included in the returned list.  Typically, this
```
imron: Another missing :

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -566,14 +582,21 @@ class DockerEnumerator( ContainerEnumerator ):
...
+            is the agent pod.
+        @type client DockerClient
```
imron: Missing :

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -566,14 +582,21 @@ class DockerEnumerator( ContainerEnumerator ):
...
+        @type ignore_pod QualifiedName
```
imron: Missing :

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -588,14 +611,27 @@ class CRIEnumerator( ContainerEnumerator ):
     Container Enumerator that retrieves the list of containers by querying the Kubelet API for a list of all pods on the node
     and then from the list of pods, retrieve all the relevant container information
     """
-    def __init__( self, container_id, k8s_api_url, query_filesystem ):
-        super( CRIEnumerator, self).__init__( container_id )
+    def __init__( self, ignore_pod, k8s_api_url, query_filesystem, kubelet_api_host_ip):
```
imron: Nitpick - All the doctypes are missing ':' after the name.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -780,10 +829,38 @@ class ContainerChecker( StoppableThread ):
         Monitors containers to check when they start and stop running.
     """
 
-    def __init__( self, config, global_config, logger, socket_file, docker_api_version, host_hostname, data_path, log_path,
+    def __init__( self, config, global_config, logger, socket_file, docker_api_version, agent_pod, host_hostname, log_path,
                   include_all, include_controller_info, namespaces_to_ignore,
                   ignore_pod_sandboxes ):
+        """
 
+        :param config: The configuration for the monitor which includes options for the container checker
```
imron: Nitpick should this be `@param` instead of `:param`.  Same for all following params.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: I think the correct thing is probably to have `ContainerChecker` inherit from `object`.  This is likely an artifact of a refactoring, but the main thread of ContainerChecker now has a `StoppableThread` as a member and it's that thread which does all the checking.  As far as I can tell, it never uses its inherited `StoppableThread` functionality.

Probably best to remove this (and inherit from object), but leaving it in doesn't look like it will cause harm.

scalyr_agent/monitor_utils/k8s.py
echee2: I'm keeping this since we are removing the exception instead

scalyr_agent/builtin_monitors/kubernetes_monitor.py
czerwingithub: @echee2 Ok, we probably should remove this then.  I'm sorry, that's my bad.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
czerwingithub: Ok, we can do that later.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
czerwingithub: We can do this later.

echee2: Filed bug https://scalyr.myjetbrains.com/youtrack/issue/AGENT-150

scalyr_agent/builtin_monitors/kubernetes_monitor.py
echee2: Filed https://scalyr.myjetbrains.com/youtrack/issue/AGENT-151

PR #212, opened 2019-05-21
Remove dependency on container id and warn on missing required env vars

czerwingithub:
Removed the dependency on the agent container id, which allowed
me to remove newly added filesystem code.  This hopefully
eliminates one more thing that can go wrong.

Also warn when critical required environment variables are missing.


czerwingithub:
I haven't finished testing this, but wanted to get the code in front of you guys.

PR #216, opened 2019-05-22
Fix Metric object memory leak

echee2:
# Background

I forgot to implement __hash__ and __eq__ for the `Metric` class when fixing our unit tests to work with earlier versions of Python. 

The `Metric` object was being used as keys in a dictionary.  Without those proper dunder methods the dictionary just keeps growing.

This manifests as multiple duplicate metrics (of the same key), leading to eventual omission by the Agent (which has protections against too much metrics data).  This symptom can be seen in the uploaded logs as follows:

**/var/log/scalyr-agent-2/linux_process_metrics.log .... Warning, skipped writing 810 log lines due to limit set by `monitor_log_write_rate` option...`**

# Fix

Implement __eq__ and __hash__

# Tests

A straightforward unit test that asserts a dictionary doesn't grow with repeat puts to the same key

### DIFF
--- a/scalyr_agent/builtin_monitors/linux_process_metrics.py
+++ b/scalyr_agent/builtin_monitors/linux_process_metrics.py
@@ -113,6 +113,12 @@ class Metric(object):
     """
     __slots__ = 'name', 'type', '_frozen'
 
+    def __hash__(self):
+        return hash(self.name) + 13 * hash(self.type)
+
+    def __eq__(self, other):
+        return other.name == self.name and other.type == self.type
+
     def __init__(self, name, _type):
         self.name = name
         self.type = _type
--- a/scalyr_agent/tests/linux_process_metrics_test.py
+++ b/scalyr_agent/tests/linux_process_metrics_test.py
@@ -24,6 +24,41 @@ import scalyr_agent.scalyr_logging as scalyr_logging
 
 
 class TestMetricClass(ScalyrTestCase):
+    def test_metric_as_object_key(self):
+        """
+        This protects against a bug introduced in v2.0.47 whereby the Metric object didn't implement __hash__ and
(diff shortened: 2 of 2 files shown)


scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -113,6 +113,12 @@ class Metric(object):
     """
     __slots__ = 'name', 'type', '_frozen'
 
+    def __hash__(self):
+        return hash(self.name) + hash(self.type)
```
czerwingithub: The purest in me says that you should multiple the second one by a prime number to get around periodicity in the sequence.  So,
`return hash(self.name) + 13 * hash(self.type)`

echee2:
@czerwingithub  ready for review again

scalyr_agent/tests/linux_process_metrics_test.py
```diff
@@ -6,7 +6,7 @@
 #
 #   http://www.apache.org/licenses/LICENSE-2.0
 #
-# Unless required by applicable law or agreed to in writing, software
+# Unless required by applicable law or agreed to     in writing, software
```
czerwingithub: Remove this white space addition.

scalyr_agent/tests/linux_process_metrics_test.py
```diff
@@ -24,6 +24,35 @@
 
 
 class TestMetricClass(ScalyrTestCase):
+    def test_metric_as_object_key(self):
+        dd = {}
```
czerwingithub: Add a comment here about what the bug was so that folks understand why you are adding a lot of instances of the object.

echee2: Comment added and whitespace error removed

PR #219, opened 2019-05-28
Make docker config params environment-aware

echee2:
# Background

We recently added expanded support to make global config params as well as k8s-related config params "environment-aware"

This needs to be expanded to docker config params as it is a major pain point for docker customers.

This PR makes docker config params environment-aware and also installs unit tests.

### DIFF
--- a/scalyr_agent/all_tests.py
+++ b/scalyr_agent/all_tests.py
@@ -57,6 +57,7 @@ PYTHON24_WHITELIST = [
 ]
 
 PRE_PYTHON27_WHITELIST = [
+    'scalyr_agent.tests.configuration_docker_test',
     'scalyr_agent.tests.configuration_k8s_test',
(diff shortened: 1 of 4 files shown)


echee2:
@imron I've fixed the broken unit test. Should be ready for review (in particular pls double check that the docker monitor config params that I'm making environment-aware are indeed suitable to become environment aware).

Also note I've added custom envar names if the config param itself is too "generic".  E.g. `readback_buffer_size` and `log_mode` are too generic so I've prepended SCALYR_DOCKER_ instead of just SCALYR_ for the env variable name. (For the others, I am assuming the semantics of the name itself identify them as belonging to docker (or k8s) monitor)

echee2:
OK. I removed the 2 params we discussed and also tested by launching docker json with following environment configs and then verified they propagated (with custom debug statements):

image

image

PR #220, opened 2019-05-30
AGENT-154: Rate Limiter

echee2:
Implements a rate limiter that blocks caller threads.

### DIFF
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -31,6 +31,7 @@ import os
 import random
 import threading
 import time
+from collections import deque
 
 import scalyr_agent.json_lib as json_lib
 from scalyr_agent.compat import custom_any as any
@@ -634,6 +635,7 @@ class FakeClock(object):
 
         @param set_to: The absolute time in seconds past epoch to set the time.
         @param increment_by: The number of seconds to advance the current time by.
+        @param notify_all: Whether to notifyAll() threads waiting on the time_condition
 
         @type set_to: float|None
         @type increment_by: float|None
@@ -647,12 +649,14 @@ class FakeClock(object):
         self._time_condition.release()
 
     def simulate_waiting(self):
-        """Will block the current thread until either the current time is changed or `wall_all_threads` is invoked.
+        """Will block the current thread until notified or until max_wait seconds has elapsed
+
+        Notification can occur if another thread invokes `advance_time` or `wake_all_threads`
 
         Since this can return even when the fake clock time has not changed, it is up to the calling thread to check
         to see if time has advanced far enough for any condition they wish.  However, it is typically expected that
         they will not only be waiting for a particular time but also on some other condition, such as whether or not
-        a condition is has been notified.
+        a condition has been notified.
         """
         self._time_condition.acquire()
         self._increment_waiting_count(1)
@@ -697,12 +701,12 @@ class FakeClock(object):
 class FakeClockCounter(object):
     """Helper class for multithreaded testing. Provides a method for a thread to block until a count has reached target
     value.  Moreover, for every successfully observed increment, it will advance the fake_clock and wait for all
-    other threads (driven the fake clock) to wait on the clock.  For example usage, see MonitorsManagerTest.
+    other threads (driven by the fake clock) to wait on the clock.  For example usage, see MonitorsManagerTest.
     """
 
     def __init__(self, fake_clock, num_waiters):
         """
-        @param fake_clock: FakeClock that will be advanced on every
+        @param fake_clock: FakeClock that will be advanced on every sleep_until_count_or_maxwait() call
         @param num_waiters: Number of threads that wait on fake clock
         """
         self.__fake_clock = fake_clock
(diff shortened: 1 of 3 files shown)


echee2:
@czerwingithub, this is not complete but in pretty good shape I hope. Still need to go through with fine tooth comb but unit tests suggest the overall functionality is working. 

Letting you know in case you need to look at the interface or comment on the overall design.  

I'm basically using a min heap of tokens where heap position is determined by "ripe" time.  Num tokens in heap corresponds to concurrency.  Every `acquire()` is blocked until the ripest token is available.  Every `release()` results in the token being pushed back onto the heap with the next ripe time.

echee2:
@czerwingithub I've changed this as discussed.

scalyr_agent/util.py
```diff
@@ -1337,3 +1341,265 @@ def close(self):
...
+        self._fake_clock = fake_clock
+        self._test_mode_lock = threading.Lock()
+
+        # Validate input
+        strategies = [self.INCREASE_STRATEGY_MULTIPLY, self.INCREASE_STRATEGY_RESET_THEN_MULTIPLY]
+        if increase_strategy not in strategies:
+            raise ValueError('Increase strategy must be one of %s' % strategies)
+
+        if self._max_concurrency < 1:
+            raise ValueError('max_concurrency must be greater than 1')
```
czerwingithub: Minor nit:  The error message doesn't match the code.. it must be 1 or greater.

scalyr_agent/util.py
```diff
@@ -1337,3 +1341,265 @@ def close(self):
...
+
+        if self._consecutive_success_threshold <= 1:
```
czerwingithub: Why not allow the `consecutive_success_threshold` to be 1?

scalyr_agent/util.py
```diff
@@ -1337,3 +1341,265 @@ def close(self):
...
+        try:
+            if success:
+                self._consecutive_successes += 1
+            else:
+                self._consecutive_successes = 0
+            if not success:
+                # Failure case: halve the rate
+                self._current_cluster_rate = max(
+                    self._lower_cluster_rate,
+                    self._current_cluster_rate / self._backoff_factor
```
czerwingithub: Most times I see a backoff factor in these type of algorithms, it is expressed as a multiplication instead of a division.  so, instead of having a backoff factor of 2, have it be 0.5.  Can you make that change?

scalyr_agent/util.py
```diff
@@ -1337,3 +1341,265 @@ def close(self):
...
+        @rtype: RateLimiterToken
+        """
+        if self._fake_clock:
+            return self._simulate_acquire_token()
+
+        # block until a token is available from the token heap
+        self._token_queue_cv.acquire()
+
+        try:
+            while len(self._token_queue) == 0 or self._time() < self._ripe_time:
```
czerwingithub: I think it would be better if you broke this up into two steps.. first, get a token from the token queue... you have to be holding the lock while you do that.  However, after you get the token (and record the ripe time when you get that token).. then you sleep until the ripe time has past.  You can do that without holding onto the token_queue_cv lock.

I think there are a couple of advantages with this approach.  First, in my mind simplifies the code here because you aren't trying to do two things in this loop.  Second, you don't have to call `_time()` while holding the lock... getting the system time can be weird now and then.. it is a system call and it does more than you'd think.  I try to avoid holding a lock if I have to do that.

I think the other change you'd have to make if you went this route is change `_get_next_ripe_time`.  Currently, you based the next ripe time on the current time, which works if you only invoke `_get_next_ripe_time` when the last ripe time has been reached.  With my approach, you'd be calling it when you acquire the token but before you actually sleep (to wait until the current time is ripe time).  But, I think that is easy to fix as well -- have `_get_next_ripe_time` be based on the max(current_time, last_ripe_time.)  you want to still use current time to cover the case when a thread calls `acquire_token` with the ripe time already past.

scalyr_agent/util.py
```diff
@@ -1337,3 +1341,265 @@ def close(self):
...
+                # no tokens available so sleep for a while
+                if len(self._token_queue) == 0:
+                    # queue contained no tokens so wait indefinitely
+                    self._token_queue_cv.wait()
+                else:
+                    # queue contained at least one token so sleep until the head token becomes ripe
+                    self._token_queue_cv.wait(max(0, self._ripe_time - self._time()))
```
czerwingithub: I'd be safer here and not even invoke the wait of the wait time is zero.

scalyr_agent/util.py
echee2: good catch.

scalyr_agent/util.py
echee2: @czerwingithub I'd like to understand if you see this change as a must-have.  I've spent too many hours trying to implement what I think satisfies your recommended change here: https://github.com/scalyr/scalyr-agent-2/pull/229

The problem I'm having is getting my tests to pass for multithreaded situation. I know I'm getting race conditions. What I don't know if it's just because of the way my test is written (to use the fake clock) of whether these will manifest in production code.  

Due to time pressure, unless you see this change as necessary, I would rather punt on this as there are other pressing things to work on. (iiuc your main concern is the extra cycles it takes to call system clock, but would that amount to *significant* problems in any scenario?)

echee2:
@czerwingithub , @imron  I'm going to merge this RateLimiter into master as it is pretty much standalone and doesn't affect existing code.  Any improvements to it can be made in separate PRs if needed.

PR #221, opened 2019-06-01
AGENT-139: Make k8s_events_disable an environment-aware global param

echee2:
Currently there is ad-hoc code that looks for `K8S_EVENTS_DISABLE` environment variable. 

There was a todo to make this a global param that's environment-aware. 

This PR adds that global param, taking in from `SCALYR_K8S_EVENTS_DISABLE`. 
It leaves `K8S_EVENTS_DISABLE` in place, thus allowing legacy customers to specify either.
It changes the k8s configmap to set `SCALYR_K8S_EVENTS_DISABLE`.

Additionally, I have updated AGENT-132 to signify the need to adjust documentation. 

# How tested

Unit tests. Also successful ran in minikube using the updated daemonset yaml that takes in from `SCALYR_K8S_EVENTS_DISABLE`

image

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
@@ -263,11 +263,16 @@ class KubernetesEventsMonitor( ScalyrMonitor ):
         if self.__max_log_rotations is None:
             self.__max_log_rotations = default_rotation_count
 
-        # TODO: This should be a config option but we will wait until we have general support for options set via
-        # K8s ConfigMap
-        disable = str(os.environ.get('K8S_EVENTS_DISABLE', "false")).lower()
+        # Support legacy disabling of k8s_events via the K8S_EVENTS_DISABLE environment variable
+        k8s_events_disable_envar = os.environ.get('K8S_EVENTS_DISABLE')
+        if k8s_events_disable_envar is not None:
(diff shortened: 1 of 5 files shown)


scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -263,11 +263,16 @@ def _initialize( self ):
...
+        k8s_events_disable_envar = os.environ.get('K8S_EVENTS_DISABLE')
+        if k8s_events_disable_envar is not None:
+            global_log.warn('The K8S_EVENTS_DISABLE environment variable is deprecated. Please use SCALYR_K8S_EVENTS_DISABLE instead.')
+            legacy_disable = str(k8s_events_disable_envar).lower()
+        else:
+            legacy_disable = 'false'
+
         # Note, accepting just a single `t` here due to K8s ConfigMap issues with having a value of `true`
-        self.__disable_monitor = disable == 'true' or disable == 't'
+        self.__disable_monitor = legacy_disable == 'true' or legacy_disable == 't' or self._global_config.k8s_events_disable
```
imron: Are we going to consider allowing `1` as true?

echee2: My vote is no.  The simpler we keep it the better.

PR #222, opened 2019-06-01
AGENT-124: Remove metrics_only config param from k8s monitor

echee2:
A trivial PR to remove what seems to be an unused config param in thee kubernetes monitor.

(I think this was copied over from docker monitor?)


imron:
At some point I should probably go through and remove the unused config options.  I think there are still another few out there.  I've added a ticket to [track this in youtrack](https://scalyr.myjetbrains.com/youtrack/issue/AGENT-156)

PR #224, opened 2019-06-01
K8s option b

echee2:
These commits include 
- profiler
- fuzz factors
- various other small improvements, almost exclusively k8s-related.

They will be needed by optionB feature branches but we should try to merge them in master ASAP to clearly distinguish from true optionB features.


scalyr_agent/monitor_utils/k8s.py
```diff
@@ -759,10 +804,14 @@ def configure( self, new_config ):
...
-            if new_config.verify_api_queries:
-                k8s = KubernetesApi(k8s_api_url=new_config.api_url)
-            else:
-                k8s = KubernetesApi( ca_file=None, k8s_api_url=new_config.k8s_api_url)
+            args = {
+                'k8s_api_url': new_config.api_url,
+                'query_timeout': new_config.query_timeout
+            }
+            if not new_config.verify_api_queries:
+                params['ca_file'] = None
```
echee2: This should be `args` not `params` I think

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1291,28 +1359,35 @@ class KubeletApi( object ):
...
-        host_ip = status.get( 'hostIP', None )
-
         if host_ip is None:
-            raise KubeletApiException( "Unable to get host IP for pod: %s/%s" % (k8s.namespace, pod_name) )
+            try:
+                pod_name = k8s.get_pod_name()
+                pod = k8s.query_pod( k8s.namespace, pod_name )
+                status = pod.get( 'status', {} )
+                host_ip = status.get( 'hostIP', None )
+                # Don't raise exception for now
```
echee2: What was the reason not to raise `KubletApiException` if host IP is null?  Does that reason still hold when merging this back into master?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2053,14 +2209,15 @@ def run( self ):
 
                 runtime = None
                 if k8s_cache:
+                    # TODO(czerwin):  If the api server has been disabled, we need to decide what to do here.
```
echee2: This TODO still needs to be handled.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1951,15 +2086,16 @@ def __get_k8s_cache(self):
...
-            runtime = k8s_cache.get_container_runtime()
+        #runtime = None
+        #if k8s_cache:
+        #    ver = k8s_cache.get_api_server_version()
+        #    runtime = k8s_cache.get_container_runtime()
         ver = 'k8s=%s' % (ver if ver else 'true')
-        if runtime:
-            ver += ';k8s-runtime=%s' % runtime
+        #if runtime:
+        #    ver += ';k8s-runtime=%s' % runtime
```
echee2: Recommend uncommenting these so unit tests pass.

I will remove version check in separate PR hanging off option B branch,

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2041,10 +2177,30 @@ def run( self ):
 
         k8s_cache = self.__get_k8s_cache()
 
+        envars_to_log = [
+            'SCALYR_K8S_NODE_NAME',
+            'SCALYR_K8S_POD_NAME',
+            'SCALYR_K8S_POD_NAMESPACE',
+            'SCALYR_K8S_POD_UID',
+            'SCALYR_K8S_KUBELET_HOST_IP',
```
echee2: Need to modify k8s YAML.

scalyr_agent/monitor_utils/k8s.py
imron: I think this might have been a hangover from a previous iteration of the code.  I did some quick testing, and didn't seem to see any issues with it.  I think we can leave it in.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: `k8s_cache` is always non-None now.  That being said getting the runtime can return `None` in which case metrics will be disabled.

It's probably ok to leave this as is with the merge to master (because it will still be calling out to the api).  It's only when the cache warmer stuff hits that it might be a problem and we can address it in a commit on that branch.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Yup.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2041,10 +2177,30 @@ def run( self ):
...
+            'SCALYR_K8S_CLUSTER_NAME',
+            'SCALYR_K8S_DISABLE_API_SERVER',
+            'SCALYR_K8S_CACHE_EXPIRY_SECS',
+            'SCALYR_K8S_CACHE_START_FUZZ_SECS',
+            'SCALYR_K8S_CACHE_EXPIRY_FUZZ_SECS',
+            'SCALYR_COMPRESSION_TYPE',
+            'SCALYR_ENABLE_PROFILING',
+            'SCALYR_PROFILE_DURATION_MINUTES',
+            'SCALYR_MAX_PROFILE_INTERVAL_MINUTES',
+            'SCALYR_K8S_ALWAYS_USE_DOCKER',
```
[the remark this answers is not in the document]
imron: Good point.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1291,28 +1359,35 @@ class KubeletApi( object ):
...
         }
         self._session.headers.update( headers )
 
-        self._http_host = "http://%s:%d" % ( host_ip, port )
-        self._timeout = 10.0
+        global_log.info('KubeletApi host ip = %s' % host_ip)
+        if host_ip:
+            self._http_host = "http://%s:%d" % ( host_ip, port )
+        else:
+            self._http_host = None
```
imron: Setting this to `None` has ramifications in later code e.g. [here](https://github.com/scalyr/scalyr-agent-2/blob/b6dbbd0ad76b2cd752cc8bea61ffebea146c95df/scalyr_agent/monitor_utils/k8s.py#L1395).  We need to make sure that if it's `None` then early abort future queries to the api.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2041,10 +2177,30 @@ def run( self ):
...
         k8s_cache = self.__get_k8s_cache()
 
+        envars_to_log = [
+            'SCALYR_K8S_NODE_NAME',
+            'SCALYR_K8S_POD_NAME',
+            'SCALYR_K8S_POD_NAMESPACE',
+            'SCALYR_K8S_POD_UID',
+            'SCALYR_K8S_KUBELET_HOST_IP',
+            'SCALYR_K8S_CLUSTER_NAME',
+            'SCALYR_K8S_DISABLE_API_SERVER',
```
echee2: Also this one should be removed since the `k8s_disable_api_server` feature isn't included in this branch.

echee2:
Looking at this PR, it seems most of the work is already in `option-b/master` .  However, some of the discussion comments may still need to be addressed.   Do not close this PR until addressed.

PR #225, opened 2019-06-03
Implement core of option_b approach

czerwingithub:
Implement the ControlledCacheWarmer, which prevents the
ContainerChecker from returning a pod unless it is warm (i.e.,
in the cache).  If it is not, the warmer runs a thread that
attempts to warm all active pods, one at a time.

TODO: Hook in Ed's token-based approach for accessing the API.
TODO: Reshape K8s temporary errors to all throw a K8sApiTemporaryError exception.
TODO: Implement retries on the path to fetch a pod and its controller info.


czerwingithub:
@echee2 @imron Suggested review path.  Review the ControlledCacheWarmer abstraction first defined in the kubernetes_monitor.py file and its corresponding tests.  The rest of this PR is essentially the small amount of work needed to hook that into the Docker-based ContainerChecker.

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -127,3 +129,389 @@ def start_test(m2, m1):
...
+        warmer = self.__warmer_test_instance
+        fake_cache = self.__fake_cache
+
+        with mock.patch.object(ControlledCacheWarmer, '_get_current_time', fake_get_current_time):
+
+            warmer.begin_marking()
+            warmer.mark_to_warm(self.CONTAINER_1, self.NAMESPACE_1, self.POD_1)
+            warmer.end_marking()
+
+            # Have to have 5 permanent errors before it is blacklisted.
```
echee2: 5 temporary errors

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -127,3 +129,389 @@ def start_test(m2, m1):
...
+            warmer.mark_to_warm(self.CONTAINER_1, self.NAMESPACE_1, self.POD_1)
+            warmer.end_marking()
+
+            self.assertEqual(warmer.active_containers(), [self.CONTAINER_1])
+            self.assertEqual(warmer.warming_containers(), [self.CONTAINER_1])
+            self.assertEqual(warmer.blacklisted_containers(), [])
+            self.assertFalse(warmer.is_warm(self.NAMESPACE_1, self.POD_1))
+
+            # To test that the entry was properly reset to error count = 0, see if it gets blacklisted after
+            # 5 new temporary errirs.
```
echee2: typo: errors

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -412,9 +430,428 @@ def _ignore_old_dead_container( container, created_before=None ):
...
+            if exception_to_report is not None and self.__logger is not None:
+                self.__logger.error('An error of type %s was seen when warming container %s.  Exception was %s: %s',
+                                    result_type, container_id, str(exception_to_report), str(traceback_report),
+                                    limit_once_per_x_secs=300, limit_key='warmer-record-result-%s' % result_type)
+
+    @staticmethod
+    def _get_current_time():
+        """Returns the current time.
+
+        This is used for testing.
```
echee2: nit: more precisely -- "This method exists so it can be overridden during testing."

czerwingithub:
Thanks for the review.  I've addressed all of your comments and it's ready to go for another look.  I did track down the cause of the hanging tests.  The only test now failing is the user agent test and I believe you took care of that in another commit.

As for your question about metrics, yes, you are right, if metrics are turned on we won't be rate limiting them.  However, we have metrics disabled for Buka so that should be ok for the option_b build.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1011,7 +1480,8 @@ def start( self ):
             if self.__always_use_docker or (self._container_runtime == 'docker' and not self.__always_use_cri):
                 global_log.info('kubernetes_monitor is using docker for listing containers')
                 self.__client = DockerClient( base_url=('unix:/%s'%self.__socket_file), version=self.__docker_api_version )
-                self._container_enumerator = DockerEnumerator( self.__client, self.__agent_pod )
+                self._container_enumerator = DockerEnumerator( self.__client, self.__agent_pod,
```
imron: Presumably this is only for Docker because that's all we're testing for at the moment and in the future this will also be added to other CRI runtimes?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -182,6 +185,21 @@
                      'containers using docker even when the runtime is detected to be something different.',
                      convert_to=bool, default=False, env_aware=True)
 
+define_config_option( __monitor__, 'k8s_use_controlled_warmer',
```
imron: Should these all be moved to the global config level, because presumably the controlled warmer is going to be used eventually for the kubernetes events monitor too?  Otherwise we'd need to duplicate the settings.

Another benefit of moving it to the global config is that it makes it easier to configure the cache to initialize straight away if it detects `k8s_use_controlled_warmer` is `True`

PR #229, opened 2019-06-04
AGENT-154: rate limiter version 2

echee2:
This PR examines an approach where the main loop in token acquire() does not check 2 things (queue size and time).

See https://github.com/scalyr/scalyr-agent-2/pull/220#discussion_r289867815


echee2:
Changed to WIP status to reflect that review is not needed.

PR #230, opened 2019-06-04
AGENT-147: Add extra_attributes to copying_manager

echee2:
We need to _not_ show all scalyr-agent-xxxx hosts in the main UI (since there could be a very large number of agents such as in k8s clusters).

This PR introduces the `extra_server_attributes` construct in the `CopyingManager`.
At init time, it will query the provided list of monitors for extra server attributes which it will 
add to its local copy of server attributes.  

The expanded set of server attributes is then supplied to the `_create_add_events_request` call (instead of the original config.server_attributes).

In this way, monitors can define extra server attributes that will be applied to all log lines uploaded by the agent. An example use case is the Kubernetes Monitor sets a server attribute which the backend can then use to modify the UI as needed.

# How tested?

Added a unit test but that does not check for addEvents()

So I added a temporary print statement just before [this line](https://github.com/scalyr/scalyr-agent-2/blob/master/scalyr_agent/copying_manager.py#L880):

It printed out the text "EXPANDED attribute = ...".

As you can see from the following screenshot it seems to work:

image

If you know of a better way to test this aspect please LMK.

### DIFF
--- a/scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
+++ b/scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
@@ -23,9 +23,11 @@ import mock
 from mock import patch
 
 from scalyr_agent.builtin_monitors.kubernetes_monitor import KubernetesMonitor
+from scalyr_agent.copying_manager import CopyingManager
 from scalyr_agent.util import FakeClock, FakeClockCounter
 from scalyr_agent.test_base import ScalyrTestCase
 from scalyr_agent.test_util import ScalyrTestUtils
+from scalyr_agent.tests.copying_manager_test import CopyingManagerInitializationTest
 
 
 class KubernetesMonitorTest(ScalyrTestCase):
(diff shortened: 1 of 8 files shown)


scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -126,4 +128,40 @@ def start_test(m2, m1):
 
                 manager.stop_manager(wait_on_join=False)
                 fake_clock.advance_time(increment_by=manager_poll_interval)
-            start_test()
\ No newline at end of file
+            start_test()
+
+
+class TestExtraServerAttributes(CopyingManagerInitializationTest):
+
```
imron: I think we should have a test for when two monitors define the same extra server attribute key with a different value.  In the current code, the 2nd monitor will overwrite the value of the 1st (see comments below).

echee2:
@imron this PR is also ready for re-review.

echee2:
@imron sorry, I just remembered you had moved this to a different branch based off option-b:
https://github.com/scalyr/scalyr-agent-2/pull/237
It is that branch that contains the review edits you asked for.

imron:
yep, it should be based off `option-b/master`

PR #233, opened 2019-06-06
Imron/profiler

imron:
This merges the fuzz factor and profiler code in to master.


imron:
Once this is merged in to master, I'll make all the new option-b branches from this.

PR #238, opened 2019-06-07
AGENT-163: Integrate RateLimiter -> CacheWarmer

echee2:
This is the main PR that integrates rate-limiter to cache-warmer.

## Status

Runs stably with rate limiting basically working.

Main bugs found:

## Bug 1:  controllers being re-queried unnecessarily (FIXED)
This is due to access_time not updated because `current_time` is not passed.
Eventually, these controllers are purged from the cache and subsequently requeried.

## Bug 2: Warmer main loop fails to catch exception (FIXED)
I observed an infrequent non-deterministic behavior where pod requeries fail to occur.  The best guess is the warmer main loop fails to catch exception and the thread dies.

## Bug 3: Warmer main loop races (Possibly fixed. Needs more testing)
`__pick_next_pod` sometimes unblocks when there are no pods to warm, resulting in race-looping.
It continues to happen in this PR. We should try to understand what causes it and whether the sleep call (to protect when it does happen) will interfere with rate limiting.

## Bug 4: log files repeatedly added/dropped (FIXED)
Happens when the cache's own thread expires a pod, then `_get_containers` excludes it.  Solution is to use a `soft purge` whenever warmer is present.  As far as possible, I try not to modify semantics of warmer.  All state that should belong in the cache is kept in the cache (e.g. marked as expired, access_time). 

Important: To avoid memory leak for unused pods, cache must still purge pods based on access time -- and access time must be pinged/kept alive (currently triggered by `warmer.mark_to_warm`.  The implication is `warmer.mark_to_warm` interval (currently 5s) must be < cache purge interval.

## Bug 5: rate limiter rate adjustment (FIXED)
Currently, a string of successes increases the target rate by the increase-factor. Thus, a long string of successes will quickly saturate target rate to max-rate, even if this target-rate is unrealistically high.  Eventually server failure will lead to backoff and it will take longer to backoff from a potentially unreasonably high max rate (despite logN).  I think a simple fix is to limit the rate increases to `min(actual-rate, target-rate) * increase-factor`.

## Bug 6: Other code paths need to be rate limited (FIXED)
a) If metrics is turned on, this code path will lead to a burst of API calls from each agent to the masters. (fixed)
b) kubernetes_events monitor

## Bug 7: Agent DaemonSet is being queried every cache interval when events/metrics are turned on (FIXED).
- This was caused by the metrics call to _get_containers *not* excluding the scalyr-agent pod, where as the cache-warmer call to _get_containers *does* exclude the scalyr agent pod daemonset.  Because the cache-warmer excludes the scalyr agent pod daemonset, it is never `mark_to_warmed`, and hence never kept alive.  So every cache interval loop, the daemonset is detected as expired and re-queried.

## Bug 8: (FIXED)
Exception was being swallowed - query_api exceptions were eventually swallowed. Serious bug, easy fix but a lot of work put into testing.

# Config map used

```
apiVersion: v1
kind: ConfigMap
data:
  SCALYR_K8S_CLUSTER_NAME: echee_ratelimit_clean
  SCALYR_SERVER: qatesting.scalyr.com
  SCALYR_K8S_LOG_API_RESPONSES_TO_DISK: "true"
  SCALYR_K8S_USE_CONTROLLED_WARMER: "true"
  SCALYR_K8S_EVENTS_DISABLE: "true"
  SCALYR_K8S_RATELIMIT_CLUSTER_NUM_AGENTS: "1"
  SCALYR_K8S_RATELIMIT_CLUSTER_RPS_INIT: "10"
  SCALYR_K8S_RATELIMIT_CLUSTER_RPS_MIN: "1.0"
  SCALYR_K8S_RATELIMIT_CLUSTER_RPS_MAX: "20000.0"
  SCALYR_K8S_RATELIMIT_CONSECUTIVE_INCREASE_THRESHOLD: "6"
  SCALYR_K8S_RATELIMIT_INCREASE_STRATEGY: "multiply"
  SCALYR_K8S_RATELIMIT_INCREASE_FACTOR: "2.0"
  SCALYR_K8S_RATELIMIT_BACKOFF_FACTOR: "0.5"
  SCALYR_K8S_RATELIMIT_MAX_CONCURRENCY: "1"
  SCALYR_DEBUG_INIT: "true"
  SCALYR_DEBUG_LEVEL: "5"
  SCALYR_K8S_QUERY_TIMEOUT_SECS: "30"
  SCALYR_K8S_CACHE_BATCH_POD_UPDATES: "true"
  SCALYR_K8S_CACHE_EXPIRY_SECS: "60"
  SCALYR_K8S_CACHE_PURGE_SECS: "10"
  SCALYR_K8S_CACHE_START_FUZZ_SECS: "0"
  SCALYR_K8S_CACHE_EXPIRY_FUZZ_SECS: "0"
  SCALYR_COMPRESSION_TYPE: bz2
  SCALYR_REPORT_CONTAINER_METRICS: "false"
  SCALYR_REPORT_K8S_METRICS: "false"
metadata:
  name: scalyr-config
  namespace: default
```

# k8s api queries made

Here are the contents of `/var/log/scalyr-agent-2/kapi/limited` after the first round of rate-limited queries are made:

image

Thereafter, every `SCALYR_K8S_CACHE_EXPIRY_SECS`, the pods are refetched:
image


scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -357,7 +357,7 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
                 result[cid] = {'name': cid, 'log_path': None, 'labels': None }
 
     except Exception, e:  # container querying failed
-        logger.error("Error querying running containers", limit_once_per_x_secs=300,
+        logger.exception("Error querying running containers", limit_once_per_x_secs=300,
```
echee2: We should use `.exception` as far as possible.  It was quite hard to debug a subtle bug from Steven's PR without this.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -848,6 +849,17 @@ def __init__(self, pod_namespace, pod_name):
             # The reason for the blacklisting.
             self.blacklist_reason = None
 
+        def __repr__(self):
+            return 'WarmingEntry:\n\tpod_namespace=%s\n\tpod_name=%s\n\tis_warm=%s\n\tis_recently_marked=%s\n\tfailure_count=%s\n\tblacklisted_until=%s\n\tblacklist_reason=%s\n' % (
```
echee2: TODO: I will change this __repr__ to be like CacheConfig's

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -86,7 +85,7 @@
     },
 }
 
-def cache( config ):
+def cache(global_config):
```
echee2: Especially for  newbies, it's hard to keep track of all the `config`s floating around -- global, monitor-specific or cache?!  Renamed this to `global_config` for clarity.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -609,7 +621,7 @@ def process_object( self, k8s, obj, query_options=None ):
             annotations = annotation_config.process_annotations( annotations )
         except BadAnnotationConfig, e:
             global_log.warning( "Bad Annotation config for %s/%s.  All annotations ignored. %s" % (namespace, pod_name, str( e )),
-                                  limit_once_per_x_secs=300, limit_key='bad-annotation-config-%s' % info.uid )
+                                  limit_once_per_x_secs=300, limit_key='bad-annotation-config-%s' % id(obj))
```
echee2: I think this is a bug (info does not exist).  Need to verify if id() will work.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -825,43 +850,44 @@ def _build_node_filter( self, k8s, namespaces_to_ignore ):
 
         return result
 
-    def configure( self, new_config ):
+    def configure(self, new_cache_config):
```
echee2: adding "cache" to the variable name avoids confusion with configuration/k8s_config objects

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1022,12 +1050,14 @@ def update_cache( self, run_state ):
             try:
                 # we only pre warm the pod cache and the cluster name
                 # controllers are cached on an as needed basis
-                if local_state.batch_pod_updates:
-                    self._pods.update( local_state.k8s, local_state.node_filter, 'Pod' )
+                if local_state.batch_pod_updates and not self._state.cache_config.use_controlled_warmer:
+                    self._pods_cache.update(local_state.k8s, local_state.node_filter, 'Pod')
```
echee2: Avoid calling `update` if using a warmer.  Correct?

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1022,12 +1050,14 @@ def update_cache( self, run_state ):
...
                 self._update_cluster_name( local_state.k8s )
-                self._update_api_server_version(local_state.k8s)
-
-                runtime = self._get_runtime( local_state.k8s )
+                if not self._state.cache_config.use_controlled_warmer:
+                    self._update_api_server_version(local_state.k8s)
+                    runtime = self._get_runtime( local_state.k8s )
```
echee2: Avoid checking runtime otherwise it is not rate limited.  Also for now, warming cache => docker.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1061,18 +1091,23 @@ def update_cache( self, run_state ):
 
             try:
                 current_time = time.time()
-                if local_state.batch_pod_updates:
-                    self._pods.update( local_state.k8s, local_state.node_filter, 'Pod' )
+                if not self._state.cache_config.use_controlled_warmer:
```
echee2: IIUC all of the following logic should be avoided when using warmer. 
Instead, the entire pod cache should be invalidated.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1101,8 +1136,8 @@ def pod( self, namespace, name, current_time=None, query_options=None ):
         if local_state.k8s is None:
             return
 
-        return self._pods.lookup( local_state.k8s, namespace, name, kind='Pod', current_time=current_time,
-                                  query_options=query_options )
+        return self._pods_cache.lookup(local_state.k8s, namespace, name, kind='Pod', current_time=current_time,
```
echee2: Rename `pods_cache` for clarity.  Another variable that is somewhat hard to keep straight (i.e a lot of `pods` floating around)

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -578,12 +569,20 @@ def __emit_report_if_necessary(self, current_time=None):
...
             warm_attempts_info = ''
             for category in ['total', 'success', 'temp_error', 'perm_error', 'unknown_error', 'unhandled_error']:
                 current_amount = self.__warming_attempts.get(category, 0)
+                global_log.info('echee: category={}, current_amount={}'.format(category, current_amount))
                 previous_amount = self.__last_reported_warming_attempts.get(category, 0)
+                global_log.info('echee: category={}, current_amount={}'.format(category, previous_amount))
                 warm_attempts_info += '%s=%d(delta=%d) ' % (category, current_amount, current_amount - previous_amount)
-            self.__logger.info('controlled_cache_warmer pending_warming=%d %blacklisted=%d %s',
+                global_log.info('echee: category={}, current_amount={}'.format(category, previous_amount))
+            self.__logger.info('controlled_cache_warmer pending_warming=%d blacklisted=%d %s',
```
echee2: Notice the very subtle bug in Steven's "blacklisted" . Changing exception handlers to log using `logger.exception` instead of `logger.error` were instrumental in finding this bug. 

We really should use logger.exception whenever possible and avoid logger.error unless we know for sure we want to suppress the stack. (alternatively include the stack via traceback)

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -528,6 +515,9 @@ def mark_to_warm(self, container_id, pod_namespace, pod_name):
         try:
             if not container_id in self.__active_pods:
                 self.__active_pods[container_id] = ControlledCacheWarmer.WarmingEntry(pod_namespace, pod_name)
+            else:
+                warming_entry = self.__active_pods[container_id]
+                warming_entry.is_warm = self.is_warm(pod_namespace, pod_name)
```
echee2: Had to add this to propagate cache invalidations into the warming entries.  Otherwise, warming entries seem to be always True and come cache interval time, the `k8s_cache.invalidate` does not result in any re-querying of API master.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1061,13 +1086,15 @@ def update_cache( self, run_state ):
...
             try:
                 current_time = time.time()
-                if local_state.batch_pod_updates:
-                    self._pods.update( local_state.k8s, local_state.node_filter, 'Pod' )
+                if not self._state.cache_config.use_controlled_warmer and local_state.batch_pod_updates:
+                    self._pods_cache.update(local_state.k8s, local_state.node_filter, 'Pod')
                 else:
                     global_log.log( scalyr_logging.DEBUG_LEVEL_1, "Purging unused pods" )
-                    self._pods.purge_expired( access_time=current_time )
+                    self._pods_cache.purge_expired(access_time=current_time)
```
echee2: controlled warmer code path should also use the same non-batched expiration logic.

echee2:
@imron I've changed the k8s cache invalidation bit so that the warmer code path follows the non-batch-update codepath.

I deleted/recreated 1 of 3 daemonsets and visually verified that API calls cease/resume as expected.

I think what remains is to check whether the API calls are complete (i.e. any missing API calls). I'll slack you a link of the kapi subdir so you can inspect.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -322,9 +337,11 @@ def shallow_copy(self):
 
         return result
 
-    def purge_expired( self, access_time ):
-        """
-            removes any items from the store who haven't been accessed since `access_time`
+    def purge_unused(self, access_time, soft_purge=False):
```
imron: I think this should probably be split in to two separate functions.

`purge` was used as name to explicitly refer to removing items from the cache that were no longer in use - as a form of garbage collection essentially.  This is no longer as clear because `mark expired` functionality has now been mixed in also.

With the way this is called by `update_cache` for soft purges, it's always going to mark ever item as stale (because it passes in the current time as the `access_time` so `obj.access_time < access_time` will be true for everything).

That being the case, it'd be clearer to have a `mark_as_expired` function that goes over every item in objects and marked it as expired.  That's what is happening here anyway for soft purges, it's just not clear. 

`update_cache` always sleeps for `cache_expire_secs` so that's an appropriate place for `mark_as_expired` to be called.

Then this function can always just remain a hard purge, for when `cache_purge_secs` expires, and `update_cache` is for what happens when `cache_expire_secs` expires.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -807,26 +801,35 @@ def run_and_propagate(self):
...
+                                                     traceback_report=traceback.format_exc())
+                    except Exception, e:
+                        self.__record_warming_result(container_id, unknown_error=e,
+                                                     traceback_report=traceback.format_exc())
+                else:
                     # Something else warmed the pod before we got a chance.  Just take the win.
-                    continue
+                    consecutive_warm_pods += 1
+                    if consecutive_warm_pods == 100:
+                        global_log.warn('echee: 100 consecutive warms in a row. Sleeping for 1 sec.')
```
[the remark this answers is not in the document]
echee2: Agree.  Meant as a todo really.  We should fully understand and eliminate the cause of this fast looping before merging.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
echee2: Agree we should discuss.  I think this  issue applies everywhere we log traceback.format_exc() or traceback.format_stack() as well so we need consistency

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1310,30 +1411,54 @@ def query_api( self, path, pretty=0, return_temp_errors=False ):
...
-                raise K8sApiAuthorizationException( path, status_code=response.status_code )
-            elif response.status_code == 404:
-                raise K8sApiNotFoundException( path, status_code=response.status_code )
-
-            global_log.log(scalyr_logging.DEBUG_LEVEL_3, "Invalid response from K8S API.\n\turl: %s\n\tstatus: %d\n\tresponse length: %d"
-                % ( url, response.status_code, len(response.text)), limit_once_per_x_secs=300, limit_key='k8s_api_query' )
-            if return_temp_errors:
-                raise K8sApiTemporaryError( "Invalid response from Kubernetes API when querying '%s': %s" %( path, str( response ) ), status_code=response.status_code )
+        # Optionally save api json to disk
+        if self.log_api_responses_to_disk:
```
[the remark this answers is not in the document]
echee2: I view this similarly to the debug logfile.  Do we truncate that?

scalyr_agent/monitor_utils/k8s.py
echee2: Good suggestion that helps long term maintenance. 👍

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1445,22 +1466,23 @@ def start( self ):
...
-                    message_time = current_time
-
-                # see if we need to abort the monitor because we've been waiting too long for init
-                elapsed = current_time - start_time
-                if elapsed > self.__k8s_cache_init_abort_delay:
-                    abort = True
-                    break
+            if not self._global_config.k8s_use_controlled_warmer:
+                # wait until the k8s_cache is initialized before aborting
+                # TODO: Ensure there is a way to disable blocking for the cache initialization.
```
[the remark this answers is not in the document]
echee2: Yeah I think no longer applicable, but I'll leave it in here just in case.

scalyr_agent/monitor_utils/k8s.py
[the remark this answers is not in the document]
imron: It has a RotatingFileHandler attached to it that automatically rotates when it gets to a certain size. See `scalyr_logging.py:1245`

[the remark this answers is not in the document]
echee2: We can implement deletion of earlier files but not in this PR due to time pressure. Perhaps we keep it as an undocumented param for the time being, default off an only turn it on if needed? If necessary can delete before merging into master?

echee2:
@imron ready for final review.

imron:
I've rebased the 404 branch off of this, put will come back and do the review in couple of hours 👍

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2576,12 +2675,14 @@ def get_user_agent_fragment(self):
         k8s_cache = self.__get_k8s_cache()
         ver = None
         runtime = None
-        if k8s_cache:
-           ver = k8s_cache.get_api_server_version()
-           runtime = k8s_cache.get_container_runtime()
+        if k8s_cache and not self._global_config.k8s_use_controlled_warmer:
+            # echee TODO: currently, if k8s_use_controlled_warmer is True, version checks are disabled.
```
echee2: There are some outstanding TODOs that must be resolved before merging back to master. This is one of them.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -807,26 +843,38 @@ def run_and_propagate(self):
...
+                else:
                     # Something else warmed the pod before we got a chance.  Just take the win.
-                    continue
+                    self.__record_warming_result(container_id, already_warm=True)
+                    consecutive_warm_pods += 1
+                    global_log.log(scalyr_logging.DEBUG_LEVEL_3,
+                                   'CacheWarmer.run_and_propagate: %s consecutive warms in a row: %s %s %s'
+                                   % (consecutive_warm_pods, container_id, pod_namespace, pod_name))
+                    if consecutive_warm_pods == 10:
+                        # TODO: Do not remove this circuit breaker until this issue is fully understood.
```
echee2: There are some outstanding TODOs that must be resolved before merging back to master. This is one of them.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1151,6 +1300,11 @@ def get_container_runtime( self ):
         result = None
         self._lock.acquire()
         try:
+            if self._state.cache_config.use_controlled_warmer:
+                # TODO: For now, controlled_warmer usage directly implies/requires 'docker' runtime
```
echee2: There are some outstanding TODOs that must be resolved before merging back to master. This is one of them.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1310,30 +1581,56 @@ def query_api( self, path, pretty=0, return_temp_errors=False ):
...
-            elif response.status_code == 404:
-                raise K8sApiNotFoundException( path, status_code=response.status_code )
-
-            global_log.log(scalyr_logging.DEBUG_LEVEL_3, "Invalid response from K8S API.\n\turl: %s\n\tstatus: %d\n\tresponse length: %d"
-                % ( url, response.status_code, len(response.text)), limit_once_per_x_secs=300, limit_key='k8s_api_query' )
-            if return_temp_errors:
-                raise K8sApiTemporaryError( "Invalid response from Kubernetes API when querying '%s': %s" %( path, str( response ) ), status_code=response.status_code )
-            else:
-                raise K8sApiException( "Invalid response from Kubernetes API when querying '%s': %s" %( path, str( response ) ), status_code=response.status_code )
+            # echee: TODO: remove this before merging
```
echee2: There are some outstanding TODOs that must be resolved before merging back to master. This is one of them.

scalyr_agent/configuration.py
```diff
@@ -1066,8 +1122,65 @@ def __verify_main_config_and_apply_defaults(self, config, file_path, apply_defau
...
+
+        # Whether to log api responses to disk (warning: this should be turned on only for debugging as it will
+        # fill up local disk over time
+        self.__verify_or_set_optional_bool(
+            config, 'k8s_log_api_responses_to_disk', False, description, apply_defaults, env_aware=True
+        )
+        # Optional (defaults to True). If true, will use the controlled warmer strategy to warm the pod
+        # cache.  This only applies when docker is being used as the container system
+        self.__verify_or_set_optional_bool(
+            config, 'k8s_use_controlled_warmer', True, description, apply_defaults, env_aware=True
```
echee2: Default values for rate limiter should be discussed and set for general clusters before merging to master.

scalyr_agent/monitor_utils/tests/k8s_test.py
```diff
@@ -123,6 +124,71 @@ def test_return_none_on_query_error_without_options( self ):
         obj = self.cache.lookup( self.k8s, self.clock.time(), self.NAMESPACE_1, self.POD_1 )
         self.assertIsNone( obj )
 
+class TestKubernetesApi( ScalyrTestCase ):
```
echee2: 💯

scalyr_agent/monitor_utils/k8s.py
echee2: Change to support other runtimes. 
Query path outside rate limiter. Need to rate limit.?

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1022,12 +1138,14 @@ def update_cache( self, run_state ):
...
-                    self._pods.update( local_state.k8s, local_state.node_filter, 'Pod' )
+                if local_state.batch_pod_updates and not local_state.use_controlled_warmer:
+                    self._pods_cache.update(local_state.k8s, local_state.node_filter, 'Pod')
                 self._update_cluster_name( local_state.k8s )
-                self._update_api_server_version(local_state.k8s)
-
-                runtime = self._get_runtime( local_state.k8s )
+                if not local_state.use_controlled_warmer:
+                    self._update_api_server_version(local_state.k8s)
+                    runtime = self._get_runtime( local_state.k8s )
```
echee2: Investigate how to get rate limiter

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1299,7 +1457,123 @@ def get_cluster_name( self ):
...
+                retries_left -= 1
+                if retry_error_context:
+                    global_log.warn('k8s API - retrying temporary error: %s' % retry_error_context,
+                                    limit_once_per_x_secs=300,
+                                    limit_key='k8s_api_retry-%s' % retry_error_limit_key)
+            finally:
+                # Any uncaught exceptions will result in an outcome of False
+                rate_limiter.release_token(token, rate_limit_outcome)
+
+    def __open_api_response_log(self, path, rate_limited):
```
echee2: Refactor logging of json responses

PR #242, opened 2019-06-13
404 is a temporary exception

imron:


imron:
With regards to 1. These changes refactor the tests so that we can test/catch that control issue we discovered when exceptions were being swallowed by the cache.  The `set_response` and `wait_for_response` now happen at the `KubernetesApi` level with a `FakeK8s` class.  `FakeCache` now has a `FakeK8s` and a `FakeProcessor` for dummying out the relevant internals.  I also added a number of tests for the `_K8sCache` itself.

With regards to 2.  `K8sApiNotFoundException` is caught before `K8sApiTemporaryError` and re-raises itself.  This allows it to bubble up and avoid the standard retry logic for temporary errors.

imron:
The other thing this PR does is add a timeout feature to all of the condition variable `wait` calls, so that in the event that the syncing logic gets out of whack, or code changes mess up the ordering (or have an exception that stops it going down the right code path), the test won't wait indefinitely, but will throw an exception that will fail the test.

It's quite easy to get tripped up by this and a failing test is much nicer than a hanging test.

scalyr_agent/monitor_utils/tests/k8s_test.py
```diff
@@ -232,3 +289,264 @@ def wait_for_requests(self, target_pending):
...
+
+    def wait_until_request_finished(self, namespace, name):
+        return self.k8s.wait_until_request_finished( namespace, name )
+
+    def simulate_add_pod_to_cache( self, pod_namespace, pod_name ):
+        """
+        Simulates adding a pod to the cache so that we can populate the
+        cache for testing purposes without going through the regular interface
+        """
+        self.k8s.set_response( pod_namespace, pod_name, success=True )
```
czerwingithub: If you are adding the pod to the cache (as you do below), why do you need to set an API response for it?  If it is in the cache, the api should never be queried for it, correct?

imron: This was accidentally left over from a previous revision of the code.  I've now removed it.

imron:
Merged and closed.

PR #243, opened 2019-06-14
Option b/already warm

imron:
Two commits.

The first one contains just the test that reproduces the problem (which will fail - or rather hang).
The second one contains the code that fixes the problem.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -790,16 +790,10 @@ class ControlledCacheWarmer(StoppableThread):
             if container_id in self.__active_pods:
                 entry = self.__active_pods[container_id]
                 if success:
-                    entry.is_warm = True
-                    entry.failure_count = 0
-                    entry.blacklisted_until = None
-                    entry.blacklist_reason = None
+                    entry.set_warm()
                     result_type = 'success'
                 elif already_warm:
-                    # make sure warm is set - normally redundant, but needed for tests
-                    # don't update failure counts or blacklists, as they would have been
-                    # set by a previous 'success'.
-                    entry.is_warm = True
+                    entry.set_warm()
                     result_type = 'already_warm'
                 else:
                     if permanent_error:
@@ -901,6 +895,15 @@ class ControlledCacheWarmer(StoppableThread):
             for key, val in self.__dict__.items():
                 s += '\n\t%s: %s' % (key, val)
             return s + '\n'
+
+        def set_warm( self ):
+            """
+            Sets the entry to warm, and resets failure count and black list status.
+            """
+            self.is_warm = True
+            self.failure_count = 0
+            self.blacklisted_until = None
+            self.blacklist_reason = None
 
 
 def _get_containers(client, ignore_container=None, ignored_pod=None, restrict_to_container=None, logger=None,
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -410,7 +410,6 @@ def _ignore_old_dead_container( container, created_before=None ):
 
     return False
 
-
 class ControlledCacheWarmer(StoppableThread):
     """A background thread that does a controlled warming of the pod cache.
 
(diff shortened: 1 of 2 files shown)


scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -299,10 +301,20 @@ def wait_until_request_finished(self, pod_namespace, pod_name):
             finally:
                 self.__lock.release()
 
+        def add_pod_to_cache( self, pod_namespace, pod_name ):
```
echee2: Recommend self-documenting rename of this function to capture semantics signifying we are simulating someone else populating the cache for us instead of the cachewarmer thread itself doing so (the idea being to speed up readability). 

Maybe something like `simulate_another_thread_adding_pod_to_cache` or something like that.

Alternatively just add docstring explaining the purpose of this function to support the test case.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -577,9 +607,8 @@ def __emit_report_if_necessary(self, current_time=None):
             self.__last_report_time = current_time
 
             warm_attempts_info = ''
-            for category in ['total', 'success', 'temp_error', 'perm_error', 'unknown_error', 'unhandled_error']:
-                current_amount = self.__warming_attempts.get(category, 0)
-                previous_amount = self.__last_reported_warming_attempts.get(category, 0)
+            stats = __gather_report_stats()
```
echee2: missing `self.`

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -763,6 +795,12 @@ def __record_warming_result(self, container_id, success=None, permanent_error=No
                     entry.blacklisted_until = None
                     entry.blacklist_reason = None
                     result_type = 'success'
+                elif already_warm:
+                    # make sure warm is set - normally redundant, but needed for tests
+                    # don't update failure counts or blacklists, as they would have been
+                    # set by a previous 'success'.
```
echee2: Question: Don't we need to set 
```
                    entry.failure_count = 0
                    entry.blacklisted_until = None
                    entry.blacklist_reason = None
```

iiuc these entries are propagate/maintain state so the previous iteration could be a failure right? 
So my thinking is whether this iteration is a `success` (if we warmed it ourselves) or a free `already_warm` if another thread did it for us, in both cases we should be resetting state to a success case?  Am I missing something?

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -359,6 +371,30 @@ def test_basic_case(self):
...
+        warmer.begin_marking()
+        warmer.mark_to_warm(self.CONTAINER_1, self.NAMESPACE_1, self.POD_1)
+        warmer.end_marking()
+        fake_cache.add_pod_to_cache( self.NAMESPACE_1, self.POD_1 )
+
+        warmer.block_until_idle()
+
+        stats = warmer.get_report_stats()
+
+        success_count = stats.get( 'success', (0, 0) )
```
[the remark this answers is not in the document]
imron: Done

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
imron: Renamed it to 'simulate_add_pod_to_cache' and added a docstring also.

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -359,6 +371,30 @@ def test_basic_case(self):
 
         self.assertTrue(warmer.is_warm(self.NAMESPACE_1, self.POD_1))
 
+    def test_already_warm(self):
+        warmer = self.__warmer_test_instance
+        fake_cache = self.__fake_cache
+
+        warmer.begin_marking()
+        warmer.mark_to_warm(self.CONTAINER_1, self.NAMESPACE_1, self.POD_1)
+        warmer.end_marking()
```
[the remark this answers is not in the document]
imron: Done.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Gah.  I had it in my working tree, just not committed 🤦‍♂️

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: I don't think we need to set them.

We *do* need to set `is_warm` so that when `update_containers_to_warm` is called (from within `record_warming_result`) the container is removed from the list of containers.

For the other values, I think it's better not to set them based on the assumption that `already_warm` is not a `success`, it's just 'something was already there' and so we leave whatever was already there - maybe it was the result of a success, maybe there was a success followed by a failure, and our `already_warm` shouldn't disturb that because it wasn't a success in its own right.

PR #246, opened 2019-06-18
Create separate ControlledCacheWarmer instance for metrics

czerwingithub:
Due to how the marking process works, we cannot use the same
ControlledCacheWarmer instances between both the metric
fetching and the log container listing.  To fix this, I
created a separate instance for the metrics.

The longer term fix is to move to a single ContainerIterator
instance used by both metrics and logs.  However, that is more
work because we have to make that abstraction thread safe.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -411,6 +411,7 @@ def _ignore_old_dead_container( container, created_before=None ):
 
     return False
 
+
 class ControlledCacheWarmer(StoppableThread):
     """A background thread that does a controlled warming of the pod cache.
 
@@ -420,10 +421,11 @@ class ControlledCacheWarmer(StoppableThread):
     errors are seen.  While on the blacklist, a pod's information will not be fetched in order to
     populate the cache.
     """
-    def __init__(self, max_failure_count=5, blacklist_time_secs=300, max_query_retries=3, rate_limiter=None,
-                 logger=None):
+    def __init__(self, name='warmer', max_failure_count=5, blacklist_time_secs=300, max_query_retries=3,
+                 rate_limiter=None, logger=None):
         """Initializes the thread.
 
+        @param name:  The name of this warmer, used when it logs its stats.
         @param max_failure_count: If this number of temporary failures is experienced when fetching a single
             pod's information, that pod will be placed on the blacklist.  Note, any permanent error will result
             in the pod being moved to the blacklist.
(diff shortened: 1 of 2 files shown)


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -464,7 +469,23 @@ def __init__(self, max_failure_count=5, blacklist_time_secs=300, max_query_retri
         self.__last_reported_warming_attempts = dict()
 
     def start(self):
-        super(ControlledCacheWarmer, self).start()
+        self.__lock.acquire()
+        try:
+            super(ControlledCacheWarmer, self).start()
+            self.__is_running = True
```
echee2: nit: 

Wondering can we just use `StoppableThread._run_state.is_running` instead?

Maybe expose a is_running property on `StoppableThread`

This should avoid duplicating this state here

echee2: I'm also suspect this new non-deterministic exception occuring in CI and locally may be caused by a race condition because we use a separate is_running flag (instead of relying on StoppableThread runstate's flag.
 
```
======================================================================
ERROR: test_already_warm (scalyr_agent.builtin_monitors.tests.kubernetes_monitor_test.ControlledCacheWarmerTest)
  File "/home/circleci/scalyr-agent-2/scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py", line 209, in test_already_warm
    warmer.block_until_idle( self.timeout )
  File "/home/circleci/scalyr-agent-2/scalyr_agent/builtin_monitors/kubernetes_monitor.py", line 706, in block_until_idle
    raise Exception( "block_until_idle timed out" )
Exception: block_until_idle timed out
```

czerwingithub: Yup, it's a good suggestion to re-use run_state.  I should have thought of that.  I'll take a look at the non-deterministic failure.  I'll have a revision in the next hour for this PR.

czerwingithub: After digging into this more (and discussing with you), since run_state does not know when the thread is actually started, we are not going to re-use it.  It would require too risky of a refactor to change run_state to track start times.

czerwingithub:
@echee2 I don't know if you want to take one last look at this.  As we talked about, I didn't do your suggestion of re-using run_state because it was it's not possible with the current run_state approach.

echee2:
Nope, all good. I did a "rebase and merge" to keep the commit history linear.

PR #249, opened 2019-06-20
Fix non-deterministic already_warm test failure

czerwingithub:
Testing the `already_warm` case is tricky because the test
thread is racing against the ControlledCacheWarmer thread.

In particular, once the ControlledCacheWarmer thread notices
there is a pod to warm (because it was returned from the
`pick_next_pod` method) it (A) first checks if it is already
in the cache and (B) requests it from the API.  For the
`already_warm` case, the entry must in the cache before (A).
Also, if the thread gets to (B), it will block indefinitely
until we simulate an API response.. which we don't do in this
test.  So, we need to make sure it is in the cache before (A).
The easiest way to do this is not invoke `end_marking` until
we simulate placing it in the cache, since `end_marking` is what
adds the pod to the warming list and makes it eligible to
be returned from `pick_next_pod`.

This explains the test failures where `already_warm` failed
while blocking for the ControlledCacheWarmer to be idle.

### DIFF
--- a/scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
+++ b/scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
@@ -196,12 +196,18 @@ class ControlledCacheWarmerTest(ScalyrTestCase):
 
         warmer.begin_marking()
         warmer.mark_to_warm(self.CONTAINER_1, self.NAMESPACE_1, self.POD_1)
+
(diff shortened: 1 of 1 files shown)


czerwingithub:
@imron I ran across this test failure while I was doing another PR.  I had asked you to remove the code that simulated an API request when you were simulating adding something to the cache.  I think that simulated API request masked this bug in the testing.

czerwingithub:
@imron ping.. I'd like to get this merged since it's holding up another PR.

PR #251, opened 2019-06-21
AGENT-121: Change default docker mode to raw-logs

echee2:
The current docker agent defaults to the legacy `docker-api` mode whereby raw logs mounted on the agent container are not directly monitored, rather stdout/err logs are gotten via docker api, written to disk once more, and those files monitored.

We originally kept the legacy mode as default in order not to impact existing customers. However, that approach is not only suboptimal but the docker api is not 100% reliable. 

The better approach is to directly monitor mounted other-container logs (raw logs mode). 

This PR changes the default mode to that and also emits a warning whenever a mounted container log file is missing (e.g. in the case where customer fails to mount the docker storage directory per instructions).

# How tested?

Manually: 

1. ensured `Starting docker monitor (raw_logs=True)` is emitted in agent.log.
2. purposefully omit the mounting of /var/lib/docker/containers, then verified the warning is 
emitted.

### DIFF
--- a/docker/docker-json-config/agent.d/docker.json
+++ b/docker/docker-json-config/agent.d/docker.json
@@ -3,10 +3,9 @@
 // docker container log directory (which must be mounted into the
 // agent's container).
 {
- monitors: [
-   {
-    module: "scalyr_agent.builtin_monitors.docker_monitor",
-    docker_raw_logs: true
-   }
+  monitors: [
+    {
+      module: "scalyr_agent.builtin_monitors.docker_monitor"
+    }
   ]
 }
--- a/scalyr_agent/builtin_monitors/docker_monitor.py
+++ b/scalyr_agent/builtin_monitors/docker_monitor.py
@@ -98,10 +98,10 @@ define_config_option( __monitor__, 'log_mode',
                      convert_to=str, default="docker_api", env_aware=True, env_name='SCALYR_DOCKER_LOG_MODE')
 
 define_config_option( __monitor__, 'docker_raw_logs',
-                     'Optional (defaults to False). If True, the docker monitor will use the raw log files on disk to read logs.'
+                     'Optional (defaults to True). If True, the docker monitor will use the raw log files on disk to read logs.'
                      'The location of the raw log file is obtained by querying the path from the Docker API. '
                      'If false, the logs will be streamed over the Docker API.',
-                     convert_to=bool, default=False, env_aware=True)
+                     convert_to=bool, default=True, env_aware=True)
 
 define_config_option( __monitor__, 'metrics_only',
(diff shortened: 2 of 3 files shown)


docker/docker-json-config/agent.d/docker.json
```diff
@@ -3,10 +3,9 @@
 // docker container log directory (which must be mounted into the
 // agent's container).
 {
- monitors: [
-   {
-    module: "scalyr_agent.builtin_monitors.docker_monitor",
-    docker_raw_logs: true
```
echee2: Please pay attention to the removal of this line -- specifically whether this will impact existing customers?

echee2: Safe to remove? Does it impact existing customers?

imron: It won't impact existing customers.  Only customers who use the new version will get this config file, and for them the default will be true.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -508,6 +508,7 @@ def start( self ):
         #create and start the DockerLoggers
         self.__start_docker_logs( self.docker_logs )
         self._logger.log(scalyr_logging.DEBUG_LEVEL_1, "Initialization complete.  Starting docker monitor for Scalyr" )
```
imron: I think you can remove this DEBUG_LEVEL_1 log now that the following line prints almost the same message.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -101,7 +101,7 @@
                      'Optional (defaults to False). If True, the docker monitor will use the raw log files on disk to read logs.'
```
imron: Need to update the comment that says it defaults to `False`.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -707,6 +708,10 @@ def __start_docker_logs( self, docker_logs ):
                     global_log.info( "Error adding log '%s' to log watcher - %s" % (log['log_config']['path'], e) )
 
             if self._use_raw_logs:
+                log_path = log['log_config']['path']
+                if not os.path.exists(log_path):
+                    global_log.warn("Missing file detected for container log path '%s'. Please ensure that "
+                                    "/var/lib/docker/containers (or custom storage directory) is mounted." % log_path)
```
imron: 'custom storage directory' is not so clear.  I would reword the second sentence to something like:

"Please ensure that the host's Docker container directory (by default `/var/lib/docker/containers`) has been mounted in the Scalyr Agent container."

PR #252, opened 2019-06-24
Feature/improved track warming times

czerwingithub:


czerwingithub:
@echee2 I'll look into the test failures.

czerwingithub:
@echee2 Ok, this is ready for review with all of the changes we discussed this morning.  I rebased so I had to do a force push.  If you look at the files changed, it should give you an accurate picture of what has been changed.

PR #253, opened 2019-06-25
AGENT-55: Negative int config variables are incorrectly parsed as longs.

echee2:
While attempting to remove rate limits on linux system metrics monitor for AGENT-175, I discovered that setting -1 for int config variables raises a type conversion error.   Apparently, -1s are parsed into longs and subsequently fail type-conversion to int.

This PR fixes the json parser library to handle negative ints.

### DIFF
--- a/scalyr_agent/json_lib/parser.py
+++ b/scalyr_agent/json_lib/parser.py
@@ -526,6 +526,7 @@ class JsonParser(object):
         """Parse a numeric literal."""
         literal_buffer = []
         all_digits = True
+        sign = 1
     
         while not self.__scanner.at_end:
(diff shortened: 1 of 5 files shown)


echee2:
@czerwingithub @imron apparently AGENT-55 is not windows specific.  I believe it is actually a bug in our parser (see main description above).

As you can see, the [new unit test](https://github.com/scalyr/scalyr-agent-2/pull/253/files#diff-67cbb8c7497483dab268b80c92e815dfR45) I added fails with the long->int conversion error:-

https://circleci.com/gh/scalyr/scalyr-agent-2/3632?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #254, opened 2019-06-25
Fix 2 bugs and add a unit test to feature/improvedTrackWarmingTimes

echee2:
This PR fixes 2 bugs and add a unit test as follows:

# Bugs

1. Logic for stale entries should also return never-accessed objects
2. Recording of already-warmed entries should also be done for the case where a WarmingEntry does not exist.

# Unit test

Seems to me there are 2 `already-warmed` situations:
a) Another thread already-warms a container before `mark_to_warm` (the container will not be scheduled for warming). 
b) Another thread already-warms a container after `mark_to_warm` (the container is scheduled for warming)

The existing `test_already_warm` tests case (b).  Case (a) isn't tested.  I'm adding a test for case (a) and renaming `test_already_warm` for clarity.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -467,6 +467,7 @@ class ControlledCacheWarmer(StoppableThread):
         # `total`, `success`, `already_warm`, `temp_error`, `perm_error`, `unknown_error`.
         self.__warming_attempts = dict()
         self.__last_reported_warming_attempts = dict()
+        # Gathers statistics about how long it took to warm cache entries.  The values are expressed in seconds.
         self.__warming_times = HistogramTracker([1, 5, 30, 60, 120, 300, 600])
 
     def start(self):
@@ -534,17 +535,24 @@ class ControlledCacheWarmer(StoppableThread):
         @type pod_namespace: str
         @type pod_name: str
         """
+        # This will be set to True if this entry was previously marked to be warmed but the cache indicated it
+        # has already been warmed.
+        was_warmed = False
         start_time = time.time()
         self.__lock.acquire()
         try:
             if not container_id in self.__active_pods:
-                self.__active_pods[container_id] = ControlledCacheWarmer.WarmingEntry(pod_namespace, pod_name, start_time)
+                self.__active_pods[container_id] = ControlledCacheWarmer.WarmingEntry(pod_namespace, pod_name,
+                                                                                      start_time)
             else:
                 warming_entry = self.__active_pods[container_id]
-                warming_entry.is_warm = self.is_warm(pod_namespace, pod_name)
+                was_warmed = self.__update_warming_state(warming_entry, start_time)
             self.__active_pods[container_id].is_recently_marked = True
         finally:
             self.__lock.release()
+
+            if was_warmed:
+                self.__record_warming_result(container_id, already_warm=True)
 
     def end_marking(self):
         """Indicates the end of marking all active containers has finished.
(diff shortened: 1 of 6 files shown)


scalyr_agent/monitor_utils/k8s.py
```diff
@@ -361,6 +361,9 @@ def __get_stale_objects(self, access_time):
                 if hasattr(obj, 'access_time'):
                     if obj.access_time is None or obj.access_time < access_time:
                         stale.append((namespace, obj_name))
+                else:
```
echee2: @imron pls review this part of the PR.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -542,11 +542,11 @@ def mark_to_warm(self, container_id, pod_namespace, pod_name):
...
         try:
             if not container_id in self.__active_pods:
-                self.__active_pods[container_id] = ControlledCacheWarmer.WarmingEntry(pod_namespace, pod_name,
-                                                                                      start_time)
+                warming_entry = self.__active_pods[container_id] = ControlledCacheWarmer.WarmingEntry(
+                    pod_namespace, pod_name, start_time)
             else:
                 warming_entry = self.__active_pods[container_id]
-                was_warmed = self.__update_warming_state(warming_entry, start_time)
+            was_warmed = self.__update_warming_state(warming_entry, start_time)
```
czerwingithub: IMO, it's a false positive to consider something as `already_warmed` if it comes through the new entry branch.  When we create that entry, we set the `is_warm` field to `False` without even checking the cache.  If we then look at updating the warming state, then if it is in the cache, we will mark it as `already_warmed`.

In my mind, `already_warm` should only count those cases where a container/pod is active and we, at some point, observe it is not in the cache.. so it is added to the list to warm... and then before this entry is retrieved via `pick_next_pod`, the pod is added to the cache by something else.

Now, I think part of this is going to come down to what we define what `already_warm` should mean.

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -194,18 +194,61 @@ def test_basic_case(self):
...
         warmer.end_marking()
 
+        fake_cache.set_response(self.NAMESPACE_1, self.POD_1, success=True)
+        warmer.block_until_idle(self.timeout)
+
+        stats = warmer.get_report_stats()
+        success_count = stats.get('success', (1, 1))
+        self.assertEqual(success_count[0], 0)
+        already_warm_count = stats.get('already_warm', (0, 0))
+        self.assertEqual(already_warm_count[0], 1)
```
czerwingithub: What's the point of round 1?  In my mind, you do not need to simulate the entry being retrieved from the API to test the `already_warm` case.

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -194,18 +194,61 @@ def test_basic_case(self):
...
+
+        # Finally confirm that mark_to_warm records the already_warm stat
+        stats = warmer.get_report_stats()
+        success_count = stats.get('success', (1, 1))
+        self.assertEqual(success_count[0], 0)
+        already_warm_count = stats.get('already_warm', (0, 0))
+        self.assertEqual(already_warm_count[0], 2)
+
+
+    def test_already_warm_after_marking(self):
```
czerwingithub: Did you see my other PR where I added in the case for the second type of `already_warmed`?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
echee2: iiuc for new-entry brach, if the pod is concurrently warmed then it will still get scheduled for warming, but upon `__pick_next_pod`, it will then register as `already_warm` anyway right? 

So does this become an issue of when the `already_warm` recording is done (either here in `mark_to_warm` versus later)?

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
echee2: Round 1 was for testing the above "new entry" code branch

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
echee2: Oops I pulled your branch earlier than https://github.com/scalyr/scalyr-agent-2/pull/252/commits/90acc90fdc7365a2a2bcaa565d4fabf210109dec

PR #256, opened 2019-06-27
Option b/no batch updates

imron:
These commits make the agent always use the controlled warmer for the cache, and also remove batch updates (and other related functionality) so that the cache only has one main path of execution, rather than several conditional paths of execution.


imron:
Note: the test failure in `test_already_warm_two` seems to be non-deterministic.

PR #257, opened 2019-06-27
AGENT-183: Changed needed to merge option-b into master

echee2:
This is part of the option-b -> master integration effort.

In this PR, I do the following:
1. Productize the saving of JSON responses, including various configuration settings as discussed
2. Restrict k8s /version check to every hour


scalyr_agent/configuration.py
```diff
@@ -1109,13 +1125,30 @@ def __verify_main_config_and_apply_defaults(self, config, file_path, apply_defau
...
+        # Minimum latency of responses to be logged.  Responses faster than this limit are not logged.
+        self.__verify_or_set_optional_float(
+            config, 'k8s_log_api_min_latency', 0.0, description, apply_defaults, env_aware=True
+        )
+        # If positive, api calls with the same path will be rate-limited to a message every interval seconds
+        self.__verify_or_set_optional_int(
+            config, 'k8s_log_api_ratelimit_interval', 0, description, apply_defaults, env_aware=True
+        )
+
+        # TODO-163 : make other settings more aggressive
```
imron: Presumably this is coming in another commit?

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1060,9 +1087,11 @@ def update_cache( self, run_state ):
                 self._pods_cache.mark_as_expired(current_time)
 
                 self._update_cluster_name( local_state.k8s )
-                # TODO-163: Don't call this every time through the loop
-                # Maybe once an hour or once a day
-                self._update_api_server_version(local_state.k8s)
+
+                # Update the api server version every hour
+                if last_version_check_time is None or current_time - last_version_check_time > 3600:
```
imron: We could do away with the `is None` check if `last_version_check_time` was initially set to `0` (the beginning of the epoch) because `current_time - <00:00:00 1 January 1970>` is always going to be greater than 3600.

I know we had a discussion previously about not confusing semantics, but I feel this semantic is not so mixed, and it also makes the code easier to read.

I'll leave it up to you to decide.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1451,21 +1490,26 @@ def query_api( self, path, pretty=0, return_temp_errors=False, rate_limited=Fals
...
+                limited_txt = ''
+                if rate_limited:
+                    limited_txt = ' (rate limited)'
+                logged_response.append('k8s.query_api%s: %s' % (limited_txt, path))
+                stack_trace = ''.join(traceback.format_stack())
+                logged_response.append('\\n%s' % stack_trace.replace('\n', '\\n'))
 
             # Make actual API call
+            latency = float('inf')
+            if log_responses and self.log_api_min_latency > 0:
```
imron: Can you make a new boolean `log_latency` or similar that equals the value of this condition, and then check against this variable in this if statement, and also in the following one on line 1523.  This will ensure that if the condition ever changes, we don't need to update multiple places.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1474,6 +1518,16 @@ def query_api( self, path, pretty=0, return_temp_errors=False, rate_limited=Fals
                     raise K8sApiTemporaryError('Temporary error seen while accessing api: %s' % str(e))
                 else:
                     raise
+            finally:
+                # conditionally record latency regardless of exception
+                if log_responses:
```
imron: See previous comment regarding using a single boolean variable to represent these conditions.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1488,15 +1542,22 @@ def query_api( self, path, pretty=0, return_temp_errors=False, rate_limited=Fals
...
+                logged_response.append(response.text.replace('\n', '\\n'))
 
             return util.json_decode( response.text )
 
         finally:
-            if logged_response_file:
-                logged_response_file.close()
+            # Only debug-log the response if all criteria (status code, latency, response len) are satisfied
+            if log_responses:
+                if (response_status_code//100) != 2 or not self.log_api_exclude_200s:
```
imron: Could you add a comment explaining what this is doing.  This is the first time I'd ever seen `//`

scalyr_agent/configuration.py
echee2: Yep, a different PR

scalyr_agent/monitor_utils/k8s.py
echee2: Good suggestion!

echee2:
@imron , ready for next review.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -710,9 +714,15 @@ class _CacheConfig( object ):
     Internal configuration options for the Kubernetes cache
     """
 
-    def __init__( self, api_url="https://kubernetes.default", verify_api_queries=True, cache_expiry_secs=30,
-                  cache_purge_secs=300, cache_expiry_fuzz_secs=0, cache_start_fuzz_secs=0,
-                  query_timeout=20, log_api_responses_to_disk=False, agent_log_path=None):
+    def __init__(self, api_url="https://kubernetes.default", verify_api_queries=True, cache_expiry_secs=30,
+                 cache_purge_secs=300, cache_expiry_fuzz_secs=0, cache_start_fuzz_secs=0,
+                 query_timeout=20,
+                 log_api_responses=False,
```
imron: As an idea for another day, I wonder if it might be worth refactoring all of logging of api calls to files in to a separate class, and then passing a single object down, rather than an extra 5 params.  Not necessary for this commit, but something to think about.

PR #259, opened 2019-06-27
AGENT-64: Fix several skipped tests

czerwingithub:
1.  Fix issue with the scalyr_client where the testing code was
not properly clearing out some global state.

2.  Fix the CopyingManager end-to-end tests.  The testing code has
several issues in its use of two condition variables.  Simplified
code so that there were no longer any race conditions.  Also
added in deadlines for waiting on the condition variable to
prevent the tests hanging indefinitely.

### DIFF
--- a/scalyr_agent/tests/copying_manager_test.py
+++ b/scalyr_agent/tests/copying_manager_test.py
@@ -628,7 +628,7 @@ class TestableCopyingManager(CopyingManager):
             # This notifies any threads waiting in the `run_and_stop_at` method.  They would be blocking waiting for
             # the CopyingManager thread to stop at this point.
             self.__test_state_cv.notifyAll()
-            # We need to wait until some other state is set as the stop state.  The `notifyAll` in `run_and_stop_at_at`
+            # We need to wait until some other state is set as the stop state.  The `notifyAll` in `run_and_stop_at`
             # method will wake us up.
             self.__test_state_cv_wait_with_timeout(start_time)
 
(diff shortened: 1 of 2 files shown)


czerwingithub:
@imron @echee2 I put you both as reviewers but really only one needs to do it.  Unfortunately, this is multi-threaded test code so a little complicated.  I'm happy to give any more details.. but no real production code needed to be changed to get these tests to pass.

czerwingithub:
@imron @echee2 This is not a priority.. can wait until you guys are back from vacation.

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -626,26 +602,94 @@ def set_response(self, status_message):
...
+        start_time = time.time()
+        while self.__test_stop_state == 'all' or current_point == self.__test_stop_state:
+            self.__test_is_stopped = True
+            if self.__test_required_transition is not None:
+                raise AssertionError('Stopped at %s state but did not transition through %s' % (
+                    current_point, self.__test_required_transition))
+            # This notifies any threads waiting in the `run_and_stop_at` method.  They would be blocking waiting for
+            # the CopyingManager thread to stop at this point.
+            self.__test_state_cv.notifyAll()
+            # We need to wait until some other state is set as the stop state.  The `notifyAll` in `run_and_stop_at_at`
```
echee2: nit: `at_at` in docstring

czerwingithub: Done

PR #261, opened 2019-06-28
Added test feature to help investigate hung threads

czerwingithub:
This is meant to investigate the case where the test process does
not exit because it is waiting for some thread to stop that will
never stop.

To implement this, I added the ability to set a name prefix for
all StoppableThreads created.  The test framework uses this to set
a prefix including the test case, so that all threads can be tied
back to the test case that created it.

Also created a thread in the testing framework which will wait
60s for the tests to finish.  If they do not (and hence hung),
it will print out all active thread names.

Hopefully this will help give information about which test case
is creating threads that hang the tests.

### DIFF
--- a/scalyr_agent/test_base.py
+++ b/scalyr_agent/test_base.py
@@ -19,8 +19,11 @@
 __author__ = 'czerwin@scalyr.com'
 
 import sys
+import threading
+import time
 import unittest
 
+from scalyr_agent.util import StoppableThread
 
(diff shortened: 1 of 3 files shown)


czerwingithub:
@echee2 Not a high priority, take a look when you get back.

echee2:
Code LGTM.  I ran the unit tests locally in an infinite loop and it finally hung, producing these results:
image

PR #263, opened 2019-07-07
AGENT-187: Move standalone smoketest code into repo.  Unify unit/smoke tests into single image.

echee2:
This PR is the result of many iterations trying to unify unit/smoke tests.  Supporting older python versions is a real pain but I think this PR finally succeeds in doing so with a single docker image.  (Note the docker file for creating the test image will be in a separate repo). 

# Summary

- Modify circleCI yaml to use new unified image

- Checkin standalone smoketest code into this repo

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
       - setup_remote_docker:
           docker_layer_caching: true
       - run:
-          command: docker run -it -e TEST_BRANCH=${CIRCLE_BRANCH} -e PYTHON_VERSION=2.6 scalyr/scalyr-agent-ci-unittest:1 /tmp/unittest.sh
+          command: docker run -it -e TEST_BRANCH=${CIRCLE_BRANCH} -e PYTHON_VERSION=2.6 scalyr/scalyr-agent-ci-unittest:2 /tmp/unittest.sh
 
   unittest-25:
     docker:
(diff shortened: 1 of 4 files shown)


echee2:
@imron I don't need you to review the smoke/unit test code here (and I don't recommend you try as it's better to wait for all changes, including the docker image Dockerfile to be reviewed at the same time).

However, what I do need you to review here are 2 one-liner changes pertaining to AGENT-186.  I believe this is a real bug in python 2.5

scalyr_agent/connection.py
```diff
@@ -258,7 +258,7 @@ def connect(self):
             self.sock = socket.create_connection((self.host, self.port), self.__timeout)
         else:
             self.sock = create_connection_helper(self.host, self.port, timeout=self.__timeout)
-        if self._tunnel_host:
+        if hasattr(self, '_tunnel_host') and self._tunnel_host:
```
echee2: here's the first one liner

scalyr_agent/connection.py
```diff
@@ -315,7 +315,7 @@ def connect(self):
         else:
             self.sock = create_connection_helper(self.host, self.port, timeout=self.__timeout)
 
-        if self._tunnel_host:
+        if hasattr(self, '_tunnel_host') and self._tunnel_host:
```
echee2: here's the second one-liner

scalyr_agent/connection.py
imron: Curious.  I'm still investing to see why this might be happening.

PR #269, opened 2019-07-10
AGENT-183b: Ratelimit all k8s api access & k8s events

echee2:
# Default rate limiter composed within KubernetesAPI

This PR is pretty tricky.  I examined various approaches and after a day of thinking about pros and cons, I think the cleanest approach is to push the rate limiter down into `KubernetesApi`.  The main problem if we do not do this is that there just too many entry points that will need to be continually managed in order to figure out which rate limiter to pass in. 

This solution creates a RateLimiter when the KubernetesApi is created.  This then becomes the "default" rate limiter.  All existing operations may simply not specify any `query_options` and hence will end up using the default query_options/rate limiter.  However, an operation may override `query_options` (passed to `query_api_with_retries()`), and thus override the rate limiter.  If no rate limiting is desired, simply pass in explicit `None`. 

I think this is a clean solution from a maintenance perspective. 

# Static sync method for getting rate limiter by name

We need to share rate limiters between k8s and k8s_events monitors.  (In the future maybe even elsewhere.)  To support this, I've created a static method on BlockingRateLimiter that lazily creates a named rate limiter.  Currently, the created name limiter will derive its settings from the passed-in global configuration object.  (We may modify this in the future to support flexible/different configurations, but for now not needed).

# _CacheConfig now has a global configuration sub-object

Having to keep copying global config fields is a pain and creates unnecessary maintenance complexity, so I also include as part of this PR, passing `global_config` deeper down, including making it substate of `_CacheConfig`.

Note `need_new_k8s_object` has been modified to account for relevant fields within the global Configuration. 

# update_object no longer dangerously swallows exceptions

Swallowing exceptions in lower layers is fraught with danger especially over time when devs forget / staff changes.  To support the recent cachewarmer effort, we had made a provision to propagate the exception given a certain flag embedded within the query_options.  I feel this is (a) not visible enough and (b) too easy to mess up.  Instead, `update_object` should never swallow exceptions and it's higher-level invokers may choose to do so given the application context. Furthermore, swallowing should be explicitly declarative and eminently clear in the code, and therefore should *not* be the default behavior.  Rather, at every code point, devs should be forced to ask themselves whether swallowing is the right thing to do.  As such, I've moved the conditional swallowing into `lookup` and declared an explicit `swallow_k8s_api_exception` kwarg that defaults `False`.


scalyr_agent/monitor_utils/k8s.py
```diff
@@ -426,42 +406,8 @@ def _add_to_cache( self, obj ):
             finally:
                 self._lock.release()
 
-    def _process_objects( self, k8s, kind, objects ):
```
echee2: Seems this function wasn't used anywhere.  Ok to delete?

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -340,12 +321,11 @@ def _check_if_alive( self, k8s, node ):
...
         result = {}
         try:
-            options = ApiQueryOptions(max_retries=self.__max_query_retries, rate_limiter=self.__rate_limiter)
             # this call will throw an exception on failure
-            result = k8s.query_api_with_retries('/api/v1/nodes/%s' % node, options,
+            result = k8s.query_api_with_retries('/api/v1/nodes/%s' % node,
                                                 retry_error_context=node, retry_error_limit_key='k8se_check_if_alive')
         except Exception, e:
-            # TODO-163: add debug statement
+            global_log.log(scalyr_logging.DEBUG_LEVEL_1, "_check_if_alive False for node %s" % node)
```
imron: Can we print the exception here (not necessarily the traceback, but at least the exception message).

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -526,6 +506,25 @@ def run(self):
             global_log.info('kubernetes_events_monitor exiting because it has been disabled.')
             return
 
+        def _create_k8s_api(rate_limiter_key):
```
imron: This function seems to rely on several variables being declared in the outer scope, e.g. `k8s_verify_api_queries` and `k8s_api_url`.

I wonder if these should be passed as parameters instead.  Otherwise non-obvious problems will likely occur if this is ever refactored out in to a class-level function rather than a nested function.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -526,6 +506,25 @@ def run(self):
...
+                'log_api_exclude_200s': self._global_config.log_api_exclude_200s,
+                'log_api_min_response_len': self._global_config.log_api_min_response_len,
+                'log_api_min_latency': self._global_config.log_api_min_latency,
+                'log_api_ratelimit_interval': self._global_config.log_api_ratelimit_interval,
+                'agent_log_path': self._global_config.agent_log_path,
+                'query_options_max_retries': self._global_config.k8s_controlled_warmer_max_query_retries,
+                'rate_limiter': BlockingRateLimiter.get_instance(rate_limiter_key, self._global_config,
+                                                                 logger=global_log),
+            }
+            if not k8s_verify_api_queries:
```
imron: This variable is not defined until later, and from a different scope.

scalyr_agent/monitor_utils/k8s.py
imron: Yes.  This is ok to delete.  It's no longer needed now that we are doing individual queries and no batch queries.

PR #270, opened 2019-07-10
Option b/master

echee2:


echee2:
@imron, you don't need to work on fixing `DynamicLogPathTest`.   I managed to fix it.

echee2:
@imron, 

here are the 2 commits needing review:

https://github.com/scalyr/scalyr-agent-2/pull/270/commits/ab7bec00d97a3d509da128ba938fcda1f1bc8d23

https://github.com/scalyr/scalyr-agent-2/pull/270/commits/1171cdf67f6e3377929167e9b0bbc5715eecc719

Also, could you perform your own manual validation?

[ Note: The smoketest breaks and I'm looking into it.  However, manual testing / visual checks were OK ]

.circleci/smoketest_k8s.sh
```diff
@@ -53,7 +69,9 @@ kubectl create configmap scalyr-config \
 # Build local image (add .ci.k8s to version)
 TEMP_DIRECTORY=~/temp_directory
 mkdir $TEMP_DIRECTORY
+/bin/rm -rf $TEMP_DIRECTORY/*
```
imron: Code smell:  Never `rm -rf` a variable without checking whether its empty, and surrounding it in quotes during the `rm -rf`

It's not a problem in this instance because `$TEMP_DIRECTORY` is guaranteed to be `~/temp_directory`, *but* if for whatever reason some future update changes this or introduces a bug such that `$TEMP_DIRECTORY` is empty, the side effect shouldn't be `rm -rf /*`.  See for example [here](https://www.extremetech.com/extreme/197686-nasty-steam-for-linux-bug-can-wipe-all-your-user-files)

imron:
Looking in to them.

echee2:
@imron here's one more commit (for the daemonset YAML)
https://github.com/scalyr/scalyr-agent-2/pull/270/commits/7cb984c744c85209a04c3101f0347132a6caed9a

.circleci/smoketest_k8s.sh
echee2: Good catch.  Removed as it was unnecessary.

PR #274, opened 2019-07-25
Fix some bugs that resulted in tests hanging.

czerwingithub:
The primary issue was that some tests would check the current
time on a fake clock, if it wasn't great enough, wait until
the clock advanced by waiting for notifications on it.  However,
that first check was not atomic wrt to the waiting... so it was
possible for advances/notifications to be missed between the two
steps.

To fix this, we had to introduce a way to atomically check the
clock value while waiting for advances on it.

### DIFF
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -530,8 +530,10 @@ class RunState(object):
         """
         deadline = self.__fake_clock.time() + timeout
 
-        while deadline > self.__fake_clock.time() and self.is_running():
-            self.__fake_clock.simulate_waiting()
+        def deadline_exceeded_or_not_running(current_time):
+            return current_time > deadline or not self.is_running()
+
+        self.__fake_clock.simulate_waiting(exit_when=deadline_exceeded_or_not_running)
 
         return not self.is_running()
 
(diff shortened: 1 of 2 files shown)


czerwingithub:
@imron I must have made some unanticipated change.  I'll fix these tests and ping you when it is ready for review.

czerwingithub:
@imron Ok, this is ready for review now.

scalyr_agent/util.py
```diff
@@ -1315,25 +1353,60 @@ def _is_running(self):
...
-        elif timeout > RedirectorClient.BUSY_LOOP_POLL_INTERVAL:
-            timeout = RedirectorClient.BUSY_LOOP_POLL_INTERVAL
-        self._run_state.sleep_but_awaken_if_stopped(timeout)
+        elif deadline > poll_deadline:
+            deadline = poll_deadline
+
+        if self.__fake_clock is not None:
+            self._run_state.sleep_but_awaken_if_stopped(deadline - current_time)
+        else:
+            self.__simulate_busy_loop(deadline)
```
echee2: I might be reading this wrong but it seems to me this line only executes if `self.__fake_clock` is `None`, in which case line 1406 always raises exception?

scalyr_agent/util.py
```diff
@@ -1315,25 +1353,60 @@ def _is_running(self):
...
+        if deadline - current_time < 0:
             raise RedirectorError('Deadline exceeded while waiting for %s' % description)
-        elif timeout > RedirectorClient.BUSY_LOOP_POLL_INTERVAL:
-            timeout = RedirectorClient.BUSY_LOOP_POLL_INTERVAL
-        self._run_state.sleep_but_awaken_if_stopped(timeout)
+        elif deadline > poll_deadline:
+            deadline = poll_deadline
+
+        if self.__fake_clock is not None:
+            self._run_state.sleep_but_awaken_if_stopped(deadline - current_time)
```
echee2: iiuc, we can call `self._run_state.sleep_but_awaken_if_stopped` directly here instead of going through `sleep_but_awaken_if_stopped`

scalyr_agent/util.py
echee2: The greater question I have is why are we simulating a busy loop if the fake clock is `None`?  Shouldn't w be sleeping for a real duration of time?

PR #275, opened 2019-07-26
Make fixes to enable json slow encoding

czerwingithub:
I deleted one test because it relied on UTF-32 decoding to 32bit
characters, which is not a guaranteed in Python 2.7.  Unichrs
can be either 16bit or 32bit.  Since this is such a small case
and I couldn't find a way to determine if unichr where 32 or 16,
I just had to delete the test.

Medium term, we might just want to delete the slow encoding code
paths.  This JSON library should really only be used to talk to
the Scalyr server... and the Scalyr server relies on the fast
encoding code paths.  I see there are a few places in the code
where we still use this to serialize for places other than
/addEvents.  We should just move those to the standard json
libraries.

### DIFF
--- a/scalyr_agent/json_lib/serializer.py
+++ b/scalyr_agent/json_lib/serializer.py
@@ -131,7 +131,7 @@ HAS_UTF8 = re.compile(r'[\x80-\xff]')
 
 # Used for an optimization when escaping a string. This regexp matches a continuous sequence of regular ascii
 # characters (from char 32 to 127).
-SIMPLE_MATCHER = re.compile('^[ -~]*')
+SIMPLE_MATCHER = re.compile('^[ -~]+')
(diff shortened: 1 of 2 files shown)


echee2:
This LGTM but maybe @imron would like to take a quick look as well

imron:
> and the Scalyr server relies on the fast encoding code paths.

The server can also work without fast encoding code paths can't it?  At least it does for the fluentd plugin, so presumably it *could* for the agent too.  I wonder if the customer "fast encoding" is faster or slower than the ujson regular encoding?

imron:
> I see there are a few places in the code where we still use this to serialize for places other than /addEvents. We should just move those to the standard json libraries.

I made a ticket (AGENT-254)

PR #277, opened 2019-07-31
create dummy objects if not on accepted endpoints

imron:


echee2:
How was this tested?  How do I test this?

imron:
This was tested manually after installing argo (see [here](https://github.com/argoproj/argo/blob/master/demo.md#2-install-the-controller-and-ui)) and then running the `hello-world` example (see [here](https://github.com/argoproj/argo/blob/master/demo.md#4-run-simple-example-workflows) under the section about creating workflows through kubectl rather than the argo cli tool).

Without this change, any workflows created don't show up on the log overview page.  With this change they do.

I realize manual testing isn't ideal, but this was just a small snippet from a larger PR I'm working on for full customer resource support.

PR #278, opened 2019-08-01
AGENT-186: Check for _tunnel_host

echee2:
I don't fully understand why it didn't manifest in the existing smoketest image (most likely due to the pre-built python 2.5 binary that is used in that image was compiled such that the bug did not manifest).  

I think what I've done in this PR is the correct thing to do though as it codes defensively against missing `_tunnel_host` (which we already do in other areas of the code but apparently miss 2 spots)


echee2:
I separated this out from AGENT-187 ci work.

PR #280, opened 2019-08-05
AGENT-213: Change logger error to exception

echee2:
Display stack trace if `_get_containers` raises exception.
Temporary preview version.


scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -357,8 +357,10 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
                 result[cid] = {'name': cid, 'log_path': None, 'labels': None }
 
     except Exception, e:  # container querying failed
-        logger.error("Error querying running containers", limit_once_per_x_secs=300,
-                     limit_key='docker-api-running-containers' )
+        logger.exception("Error querying running containers: %s, filters=%s, only_running_containers=%s"
+                         % (str(e), filters, only_running_containers),
```
czerwingithub: `filters` looks like a dict above.. so this won't work here.  You can serialize a dict by `json.dumps(filters)`.  Well, `filters` can also be `None`, but I just tested it and `json.dumps` works ok for None.

Also, is `only_running_containers` a boolean?  If so, you need to do `str(only_running_containers)`

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -524,7 +524,11 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
                                             logger.log( scalyr_logging.DEBUG_LEVEL_2, "Including container '%s' based on pod annotations, %s" % (short_cid, str(pod.annotations)) )
 
                         except Exception, e:
-                          logger.error("Error inspecting container '%s' - %s, %s" % (cid, e, traceback.format_exc()), limit_once_per_x_secs=300,limit_key="docker-api-inspect")
+                          logger.exception(
+                              "Error inspecting container '%s' - %s, %s" % (cid, e, traceback.format_exc()),
```
czerwingithub: Be sure to put `str(e)` to serialize it to a string.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -539,8 +543,10 @@ def _get_containers(client, ignore_container=None, restrict_to_container=None, l
                 result[cid] = {'name': cid, 'log_path': None}
 
     except Exception, e:  # container querying failed
-        logger.error("Error querying running containers", limit_once_per_x_secs=300,
-                     limit_key='docker-api-running-containers' )
+        logger.exception("Error querying running containers: %s, filters=%s, only_running_containers=%s"
+                         % (str(e), filters, only_running_containers),
```
czerwingithub: Same here about `filters` and `only_running_containers`.

scalyr_agent/builtin_monitors/docker_monitor.py
echee2: I'm pretty sure `%s` calls `str()` on whatever object is passed . 
https://docs.python.org/2/library/stdtypes.html (search for `String (converts any Python object using str()`)
So actually the `str(e)` is superfluous.

Earlier I had inserted a `raise ValueError('test string'))` inside the main loop and verified the string interpolation worked with the plain obj.

imron: `%s` does always call `str()` if using `%` arguments after the string e.g. `log.info( "hello %s" % e )`.

Will call `str(e)`.

It won't do it if passing the values as arguments, e.g. `log.info( "hello %s", e )`

In this case str(e) is not called.

The code base seems to use a mix of these different styles, but inside a `%` argument list, `str()` will be called, so I think this is ok.  Likewise for dicts, so we shouldn't need `json.dumps` either.

I've tested this by causing a dummy exception in the relevant code block, and `filters` and the execption are both printed correctly as strings when just passing in the variable.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: See above about `%` arguments.  I don't think this is needed.

scalyr_agent/builtin_monitors/docker_monitor.py
czerwingithub: TIL that those cases are treated differently.  Thanks, I did not realize that.

PR #281, opened 2019-08-08
AGENT-223: Fix total_bytes_skipped recording bug

echee2:
total_bytes_skipped is reported incorrectly for k8s. Cause seems to be log_proccessing.perform_processing incorrectly calculates uploaded bytes from the log message without taking into account other json attributes.

Manifestation is very scary-looking agent_status lines that keep increasing in total_bytes_skipped.

### DIFF
--- a/scalyr_agent/log_processing.py
+++ b/scalyr_agent/log_processing.py
@@ -1733,8 +1733,6 @@ class LogFileProcessor(object):
                         self.__total_bytes_copied += bytes_copied
                         bytes_between_positions = self.__log_file_iterator.bytes_between_positions( original_position, final_position)
                         self.__total_bytes_skipped += bytes_between_positions - bytes_read
-                        # log.info('Bytes skipped in completion callback: bytes_between_positions=%s, bytes_read=%s, bytes_skipped=%s'
-                        #          % (bytes_between_positions, bytes_read, bytes_skipped))
 
                         self.__total_bytes_dropped_by_sampling += bytes_dropped_by_sampling
                         self.__total_bytes_pending = self.__log_file_iterator.available
--- a/scalyr_agent/log_processing.py
+++ b/scalyr_agent/log_processing.py
@@ -135,6 +135,10 @@ class LogLine(object):
         # attrs is a dict of optional attributes to attach to the log message
         self.attrs = None
 
+        # length of raw line (in contrast
+        self.raw_line_len = len(self.line)
+
+
 class LogFileIterator(object):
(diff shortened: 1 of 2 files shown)


scalyr_agent/log_processing.py
```diff
@@ -135,6 +135,10 @@ def __init__(self, line):
         # attrs is a dict of optional attributes to attach to the log message
         self.attrs = None
 
+        # length of raw line (in contrast
+        self.init_line_len = len(self.line)
```
imron: I think calling this `raw_line_len` is clearer.

scalyr_agent/log_processing.py
```diff
@@ -1646,6 +1651,7 @@ def perform_processing(self, add_events_request, current_time=None):
                     if not add_events_request.add_event(event, timestamp=line_object.timestamp, sequence_id=sequence_id, sequence_number=sequence_number):
                         #time_spent_serializing -= fast_get_time()
 
+                        log.info('Resetting file position to %d (not add_events_request)' % position.mark_offset)
```
imron: Don't need to include this log line as it was only for testing and we don't this filling real logs (this path will happen almost every time for high volume logs), plus it was for the negative bytes issue anyway.

scalyr_agent/log_processing.py
```diff
@@ -1661,6 +1667,7 @@ def perform_processing(self, add_events_request, current_time=None):
                             add_events_request.set_position(original_events_position)
                             self.__log_file_iterator.seek(position)
                             buffer_filled = True
+                            log.info('Resetting file position to %d (not added_thread_id)' % position.mark_offset)
```
imron: Same with this log line.

scalyr_agent/log_processing.py
```diff
@@ -1717,7 +1724,10 @@ def completion_callback(result):
                     if result == LogFileProcessor.SUCCESS:
                         self.__total_bytes_copied += bytes_copied
                         bytes_between_positions = self.__log_file_iterator.bytes_between_positions( original_position, final_position)
-                        self.__total_bytes_skipped +=  bytes_between_positions - bytes_read
+                        bytes_skipped = bytes_between_positions - bytes_read
```
imron: This entire change is not needed (also related to negative skipped bytes)

imron:
I wrote a failing unit test (that passes with this fix).

If you rebase your branch on top of the `imron/skipped-bytes-test` branch, you'll get the failing test before this commit (thus allowing for easy testing/verifying that this branch fixes the issue)

PR #283, opened 2019-08-15
Temporarily disable k8s_ver

echee2:


echee2:
@imron here's a separate PR based off `option-b/master` that' temporarily disables k8s_ver.  

I have manually verified that it doesn't show up in UI and that the dropdowns work again.

PR #284, opened 2019-08-15
rm .DS_Store

Ongman:
I presume `.DS_Store` is vestigial.


echee2:
Thanks for catching this!

PR #285, opened 2019-08-19
AGENT-196/211: Pure python TLS 1.2 + cert validation

echee2:
Implements a python 2.6+ compatible pure-python TLS 1.2 solution as follows:

## Background

Python ssl is based off the OS `libssl`. Older versions of `libssl` such as those running on most python 2.6 and pre-2.7.9 versions so not support TLS 1.2.  Making customers upgrade python is certain a bigger ask than upgrading just the agent.  So can we implement Agent changes that support this?  For most customers, we believe the answer is yes.

Strategy:  When establishing a connection to Scalyr backend, if TLS 1.2 is not supported by agent, attempt to reconnect but with a pure python implementation.  I found the `tlslite-ng` lib that is 2.6 compatible.  So, this approach works for majority customers except those running 2.5 or earlier.

## Caveat

During testing I discovered a limitation with the chosen approach:

On `alpine` based systems (and perhaps others like Windows), the `certvalidator` library (which I use for server cert chain validation) fails to validate. Failure is a result of a dependent lib (`oscrypto`) that calls `ctypes` `find_library`.  `find_library` depends on certain standard O/S conventions or programs, which apparently are not implemented the same in `alpine` (and perhaps windows).  See https://github.com/docker-library/python/issues/111 for more info.

Workaround:  if `certvalidator` fails, make a system call to `openssl` directly.  This requires customers to install `openssl` if they have not already done so.  However, `openssl` is ubiquitous and well known so it's the best I can think of for now.

## Changes 

1. Fallback to TLSLite connection if Httplib + openssl fails
2. Include tlslite & dependencies as static third_party files
3. Ship the 2 root certs & 2 more intermediate certs as individual pem files.  These are needed by tlslite for server certificate chain validation.
4. Server cert validation first uses `certvalidator` library.  If it fails (e.g. on alpine systems where ctypes `find_library` doesn't work), fallback to `openssl` binary.  
5. README updated to reflect additional 3rd party libs.

## How was this tested

1.  Added smoketest test where openssl-based connections fail to establish, thus falling back to pure-python TLS 1.2 implementation. (for python 2.6 and 2.7)

2. User agent was visually confirmed
image


steve-scalyr:
Do you see anyrisks in moving to tlslite-ng? For instance, does it use more CPU at high data volumes?

echee2:
I did an analysis a while back of CPU and memory stability.  Findings are recorded in https://scalyr.myjetbrains.com/youtrack/issue/AGENT-196 .
Summary is that CPU increase seems tolerable (17% instead of 15%, stable over 3 day test). 

I thought I had run the test at 0.1MB/s but looking at the Cloudwatch network out it seems to be ~250KB/min or 4KB/sec. (Maybe I was generating 0.1MB/s but after compression it's much less).

I agree with @czerwingithub that it's risky to rely on anything else than the widely used standard openssl. 

@imron @czerwingithub and myself have been discussing this at recent standups.  Two thoughts are:
1. We should really try to minimize number of users using `tlslite`
2. Python 3 migration is only a few months out.  If this drags on any much longer an alternative route is simply require an upgrade to python 2.7.9 or above, where TLS 1.2 will be guaranteed.

steve-scalyr:
So – is all of this tlslite-ng business only for python versions prior to 2.7.9? If it's for a minority of customers (and they can get around it by upgrading Python) that seems much less concerning.

docker/Dockerfile.k8s
```diff
@@ -5,7 +5,7 @@ FROM python:2.7-alpine3.9 as scalyr-dependencies
 MAINTAINER Scalyr Inc <support@scalyr.com>
 
 RUN apk --update add build-base python-dev gcc
-RUN pip --no-cache-dir install --root /tmp/dependencies ujson yappi
+RUN pip --no-cache-dir install --root /tmp/dependencies ujson yappi tlslite-ng
```
imron: > FROM **python:2.7**-alpine3.9
> ...
> RUN pip --no-cache-dir install --root /tmp/dependencies ujson yappi **tlslite-ng**

Is this required given that we know we'll be running on 2.7?

scalyr_agent/third_party/README.md
```diff
@@ -27,6 +27,12 @@ The following libraries are included:
   * [six](#six)
   * [backports.ssl-match-hostname](#ssl-match-hostname)
   * [PySocks](#PySocks)
+  * [tlslite-ng](#tlslite-ng)
```
imron: For consistency with other third_party libs, we should put the licenses for each of these new libraries in the `scalyr_agent/third_party/licenses` directory.

docker/Dockerfile.k8s
echee2: Oh, this should have been deleted (since we are including src code in 3rd party libs).

PR #286, opened 2019-08-19
Make k8s_ignore_namespaces an ArrayOfStrings

echee2:
# Background

`k8s_ignore_namespaces` is inconsistently handled.  It should be typed as an `ArrayOfStrings`, which is comma-delimited.  This enables user to specify a single comma-separated string environment variable, consistent with some other parameters.

 However, to preserve legacy behavior (splittable by whitespace), I define a new type `SpaceAndCommaSeparatedArrayOfStrings` which is a subclass of `ArrayOfStrings`.  (And necessary plumbing to make it work).

# Changes

1. Make `k8s_ignore_namespaces` an SpaceAndCommaSeparatedArrayOfStrings.
2. Remove duplicate `k8s_ignore_namespaces` from `kubernetes_monitor.py`

# Testing
- Augmented `_test_environment_aware_global_params` to handle ArrayOfString types.
- manual testing: Entered the following configmap entry:
`SCALYR_K8S_IGNORE_NAMESPACES: "kube-system a , b , c"`
- verified agent.log shows correct ignored namespaces:
image

PR #291, opened 2019-09-04
AGENT-241: allow empty config params

echee2:
Empty string config params are currently treated identically as if they were not defined at all. 

This fails to allow customers to override config params that have default values (e.g. `k8s_ignore_namespaces` is currently set to `kube-system`). 

For example, a customer may want to include `kube-system` namespace and should be able to define `SCALYR_K8S_IGNORE_NAMESPACES`: ""` (empty string), thus allowing `kube-system` namespaced pods to be logged.

# CHANGES

This PR changes the config param framework code to allow empty strings to be converted where it makes sense (for example if a config param is a `str` or `ArrayOfStrings` or `bool`).

Please see the new unit tests for clarity

# NOTE

This PR is based off [AGENT-236-k8s-ignore-namespaces](https://github.com/scalyr/scalyr-agent-2/pull/286) branch because it is closely related, and in fact, fixes some bugs for the `k8s_ignore_namespaces` config param introduced by this PR.

# TESTS

- Add unit tests to assert behavior of util functions.
- Fixes unit tests from the [AGENT-236-k8s-ignore-namespaces](https://github.com/scalyr/scalyr-agent-2/pull/286) branch.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -1531,13 +1531,13 @@ class KubernetesMonitor( ScalyrMonitor ):
         # The namespace whose logs we should not collect.
         global_namespaces_to_ignore = self._global_config.k8s_ignore_namespaces
         default_val = Configuration.DEFAULT_K8S_IGNORE_NAMESPACES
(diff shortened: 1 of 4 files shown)


echee2:
cc @czerwingithub since this is a fundamental config behavior change

PR #292, opened 2019-09-04
AGENT-242: Move exception into k8s_events_monitor main loop

echee2:
While validating the option-b/master branch, I noticed a lot of pod-not-found exceptions. However, the exception is caught outside the for line in lines loop. This means any lines that come after the offending line will be skipped.

# CHANGES
- Move K8sApiException and general exception inside the `for line in lines` loop
- Keep `ConnectionError` outside the main loop because it is very often triggered by `for line in lines` iterator.
- Add another catch-all exception outside the loop to handle generic exceptions (e.g. those triggered at the `for line in lines` iterator itself.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
@@ -583,87 +583,99 @@ class KubernetesEventsMonitor( ScalyrMonitor ):
                             global_log.warning( "Error parsing event json: %s, %s, %s" % (line, str(e), traceback.format_exc() ) )
                             continue
 
-                        # check to see if the resource version we are using has expired
-                        if self._is_resource_expired( json ):
-                            last_event = None
-                            global_log.log( scalyr_logging.DEBUG_LEVEL_1, "K8S resource expired" )
-                            continue
+                        try:
+                            # check to see if the resource version we are using has expired
+                            if self._is_resource_expired( json ):
+                                last_event = None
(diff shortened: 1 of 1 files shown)


scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -583,87 +583,99 @@ def run(self):
...
-                                       "Scalyr Service Account does not have permission to watch available events.  "
-                                       "Please recreate the role with the latest definition which can be found "
-                                       "at https://raw.githubusercontent.com/scalyr/scalyr-agent-2/release/k8s/scalyr-service-account.yaml "
-                                       "K8s event collection will be disabled until this is resolved.  See the K8s install "
-                                       "directions for instructions on how to create the role "
-                                       "https://www.scalyr.com/help/install-agent-kubernetes", limit_once_per_x_secs=300,
-                                       limit_key='k8s-stream-events-no-permission')
-                except ConnectionError, e:
-                    # ignore these, and just carry on querying in the next iteration
+                        except K8sApiAuthorizationException:
```
imron: This can't belong inside the `for line in lines` loop, because if there is an authorization problem (i.e. a misconfigured service account) then the exception will be raised by line 576 `lines = k8s_api_events.stream_events( last_event=last_event )` which will now only print a generic exception message rather than a detailed message about how to fix authorization issues.

Do you mean to be catching other generic API errors here?

echee2: Good catch . I'll relocate it back to original position and just keep a single generic exception catcher inside the loop.

PR #298, opened 2019-09-06
Imron/agent 239 mysql

imron:


scalyr_agent/builtin_monitors/mysql_monitor.py
```diff
@@ -584,7 +584,7 @@ def _derived_stat_read_percentage(self, globalVars, globalStatusMap):
     def _derived_stat_query_cache_efficiency(self, globalVars, globalStatusMap):
         """How efficiently is the query cache being used?"""
         pct = 0.0
-        if globalStatusMap['global.com_select'] + globalStatusMap['global.qcache_hits'] > 0:
+        if 'global.qcache_hits' in globalStatusMap and (globalStatusMap['global.com_select'] + globalStatusMap['global.qcache_hits'] > 0):
```
echee2: @imron are these 2 edits in this file the only changes (i.e. all other files are mere copy of latest `pmysql` code)?

echee2: Also could you describe the testing you've done to verify this PR works?

imron: > all other files are mere copy of latest pmysql code

Yes.  The first commit contains only the changes for the latest first.

The other changes are required to ensure the python version and also to fix a bug that was then occurring when connecting to the test RDS instance.

> Also could you describe the testing you've done to verify this PR works?

Good point.

I've now added a test class for the mysql monitor that checks for these changes.  I reordered history so these tests come before the fix, meaning that if you checkout the commit with the tests, then the tests will fail (except the 2.7 min requirement test when running on 2.7), then if you checkout the commit with the changes the tests will all pass.

The connection error is more difficult to add an automated test for, but I manually configured the agent to connect to the test RDS instance and it all works, whereas the exact same configuration fails using the previous version of pymysql.

imron: P.S. the 2.7 CI unit test failure is unrelated to this change.

PR #300, opened 2019-09-10
AGENT-207: TLS proxy smoke test (mitmproxy)

echee2:
This PR introduces an additional smoke test to validate the proxy-based solution for python 2.4/2.5 customers that will not be able to use the [python-based TLS 1.2 agent upgrade](https://github.com/scalyr/scalyr-agent-2/pull/285).

1. The proxy is configured to use TLS 1.0 agent-side and TLS 1.2 server-side.
2. The proxy is configured in [**reverse proxy mode**](https://docs.mitmproxy.org/stable/concepts-modes/#reverse-proxy) .  This is a key step that enables pre-2.6 customers who cannot instruct scalyr-agent to use a proxy (because the Requests library that supports a proxy only works with python 2.6 and 2.5/2.4 `httplib` library does not allow setting of a proxy.  
3. The agent's SCALYR_SERVER is configured to the mitmproxy address/port
4. mitmproxy will rewrite the host header to qatesting (or whatever is in the actual CI/CD SCALYR_SERVER environment)
5. Because the agent smoketest runs in one container and the mitmproxy runs in another, these containers must be configured to access each other's networks. Furthermore, the mitmproxy's port must be exposed (i.e. port 8080)
6. Interestingly, the python 2.4 environment currently isn't ssl-enabled.  Therefore, the agent talks `http` instead of `https` to the mitmproxy.  (The mitmproxy is capable of handling both http and https traffic, and translating it to tls 1.2)

# TODO

The current smoke tests should in theory suffice as the proxy-to-backend connection is limited to TLS 1.2.
 
However, to be thorough, we should modify the `TLS12_SCALYR_API_KEY`, `TLS12_SCALYR_SERVER` and `TLS12_READ_API_KEY` values in CircleCI to use a backend that has been configured to only accept TLS 1.2 connections.

# Instructions for customers (on running mitmproxy with Scalyr Agent)

1. Run the `mitmproxy` docker image
   - expose it's port
   - run in reverse proxy mode, pointing to the appropriate scalyr_server (e.g. `https://agent.scalyr.com`, `https://upload.eu.scalyr.com`)
   - e.g. command `docker run --net mynet --rm -d --name mitmproxy --expose=8080 mitmproxy/mitmproxy mitmdump --mode reverse:${SCALYR_SERVER}`

2. Configure all scalyr agents (that do not support TLS 1.2) to point to the `mitmproxy` as follows:
   - set `scalyr_server` to `https://<mitmproxy_hostname>:8080`
   - set `verify_server_certificate` to `false` (which disables verification of server certificates by the scalyr agent which the `mitmproxy` breaks)

## Note : this section probably does not apply (only for customers running their own docker with older python versions

Additional instructions if running scalyr agent in docker:

Ensure that the scalyr agent docker container can "see" the `mitmproxy` docker container over the network.  For example, you may need to:

1. Create a docker network via `docker network create mynet` 
2. Make sure to set the agent's `scalyr_server` to the correct url for the `mitmproxy` within the created docker network.  A useful command to determine the `mitmproxy` ip address is:
    - `export MITM_IPADDRESS=$(docker inspect -f "{{ .NetworkSettings.Networks.mynet.IPAddress }}" mitmproxy)` 
    - After which you'd set the agent's `scalyr_server` to `https://${MITM_IPADDRESS}:8080`
3. Launch the `mitmproxy` and scalyr agent docker containers with `--net mynet` command line option.


echee2:
cc @czerwingithub

PR #306, opened 2019-09-16
AGENT-248 - put percpu metrics behind an option

imron:
Testing for this has been done manually, both for the kubernetes and docker monitors.

One interesting thing when listing the individual cpus is that the kubernetes monitor specified an extra field of `dev`, while the docker monitor specifies `cpu`.

There was a bug with the docker monitor that prevented it from actually outputting that extra field, which I've now fixed, but perhaps we want to standardize on a single key for this field?

Also, it turns out the other cpu metrics *were* getting logged, just in a separate place (which is why I didn't notice previously).

# Notes

Here's a snapshot of relevant entries in `kubernetes_monitor.log`:

# k8s monitor: SCALYR_DOCKER_PERCPU_METRICS=True

image

# k8s monitor: SCALYR_DOCKER_PERCPU_METRICS=False

image

# Docker monitor: SCALYR_DOCKER_PERCPU_METRICS=True

image

# Docker monitor: SCALYR_DOCKER_PERCPU_METRICS=False

image

### DIFF
--- a/scalyr_agent/builtin_monitors/docker_monitor.py
+++ b/scalyr_agent/builtin_monitors/docker_monitor.py
@@ -77,6 +77,11 @@ define_config_option( __monitor__, 'docker_log_prefix',
                      'Optional (defaults to docker). Prefix added to the start of all docker logs. ',
                      convert_to=str, default='docker', env_aware=True)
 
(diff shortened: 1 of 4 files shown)


imron:
I'm not sure about how the front end deals with this either, so would also like @czerwingithub to weigh in.

PR #314, opened 2019-09-24
Imron/deleted files

imron:
This should fix up the problem where paths sometimes weren't being removed from the list of log matchers for short-lived logs.

### DIFF
--- a/scalyr_agent/copying_manager.py
+++ b/scalyr_agent/copying_manager.py
@@ -303,6 +303,55 @@ class CopyingManager(StoppableThread, LogWatcher):
         """
         return self.__log_matchers
 
+    def is_path_active( self, path ):
+        """ Returns a list of data structures that still contain a reference to the path
+
+        Only used for testing purposes
+
+        @param path: The path to check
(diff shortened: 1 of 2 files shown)


scalyr_agent/copying_manager.py
```diff
@@ -1029,8 +1033,17 @@ def __remove_logs_scheduled_for_deletion( self ):
                         self.__logs_pending_removal.pop( path, None )
                     finally:
                         self.__lock.release()
+                    removed_paths[path] = True
                     log.warn( "Log scheduled for removal is not being monitored: %s" % path )
 
+        # update log matchers
+        self.__lock.acquire()
+        try:
+            self.__log_matchers[:] = [m for m in self.__log_matchers if m.log_path not in removed_paths]
```
czerwingithub: The call to `remove_paths` updates the `__log_matchers` variable already, and it does it only if the number of paths has reached zero.  Should we really be doing the update here?  And why don't we have to do the same checks?

czerwingithub:
@imron I think this PR is obsolete now.  Can you close it if you agree.

imron:
Closed

PR #318, opened 2019-09-27
Fix unittest to capture output generated by loggers

czerwingithub:
This fixes the tests so that all output generated by our use
of the Python loggers will be captured during unittest execution.

To do this, needed to set the use of stdout for our logging whenever
we are about to run a test case.

Suggest review path:  Look at test_base.py first.  Almost all other changes are just to added in calling the inherited setUp method.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -22,8 +22,10 @@
 __author__ = 'czerwin@scalyr.com'
 
 from scalyr_agent.__scalyr__ import scalyr_init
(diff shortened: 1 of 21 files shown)


czerwingithub:
You also needed another PR I had pending to reset the root logger properly.  I've merged all three PRs in now to master.  If you run `python run_tests.py` under python 2.7, you should see the tests are quiet.

PR #319, opened 2019-10-08
Imron/third party tls

imron:


scalyr_agent/connection.py
```diff
@@ -81,18 +84,26 @@ def connection(server, request_deadline, ca_file, intermediate_certs_file,
 
         """
 
+        if use_tlslite:
+            if use_requests:
+                log.warn( "Both `use_requests_lib` and `use_tlslite` are set to True.  `use_tlslite` is ignored if `use_requests_lib` is True" )
+
+            if sys.version_info[:3] >= (2, 7, 9):
+                log.warn( "`use_tlslite` is only valid when running on Python 2.7.8 or beloew" )
```
czerwingithub: s/beloew/below

scalyr_agent/connection.py
```diff
@@ -269,15 +289,27 @@ def get_cert_bytes(file_names):
...
+            extra_trust_names = [
                 'scalyr_agent_ca_root.pem',
                 'addtrust_external_ca_root.pem'
-            ])
+            ]
+
+            all_cert_names = intermediate_cert_names + extra_trust_names
+            cert_dir = os.path.dirname(self._ca_file)
+            if not cert_files_exist(cert_dir, all_cert_names):
+                path = os.path.dirname(os.path.abspath(__file__))
```
czerwingithub: Add a comment here to explain what this is doing.

scalyr_agent/__scalyr__.py
```diff
@@ -139,9 +139,14 @@ def get_install_root():
 def __add_scalyr_package_to_path():
     """Adds the path for the scalyr package and embedded third party packages to the PYTHONPATH.
     """
-    # prepend the third party directory first so it appears after the package root
+    # prepend the third party directory first so it appears after the package root and
+    # third_party_tls directories
     sys.path.insert(0, os.path.join(get_package_root(), 'third_party'))
 
+    # prepend the third party tls directory first so it appears after the package root
```
czerwingithub: Should we disable this for windows?  It doesn't get included in the py2exe translation.. it would seem like we could get a weird bug if we somehow reference one of these files while running in Windows.

PR #320, opened 2019-10-11
Imron/refactor

imron:
This is the start of the refactoring for the logmatchers, log processors and the copying manager.

This PR only has the changes to the log matchers and log file processors, adding the concept of a `finished` matcher/processor (I opted for `finished` over `scheduled_to_close` for naming).

Calling `finish` on a log processor marks the processor as preparing to finish.  A processor will consider itself fully finished if it is marked as finished, and there are no bytes pending, and there were no bytes read last time the file was processed.

Calling `finish` on a log matcher marks it as preparing to finish.  A log matcher will consider itself fully finished if it is marked as finished, and it has been processed at least once, and all of its open processors are also fully finished.

### DIFF
--- a/scalyr_agent/copying_manager.py
+++ b/scalyr_agent/copying_manager.py
@@ -217,7 +217,14 @@ class CopyingManager(StoppableThread, LogWatcher):
                 else:
                     self.__expanded_server_attributes[key] = value
 
+        # a dict of paths -> log matchers for log matchers that have been dynamically added
+        # this dict should only be touched by the main thread
         self.__dynamic_matchers = {}
+
+        # a dict of paths -> monitor names that have been dynamically added to the copying manager
+        # the copying manager ensures that operations on a dynamically added path can only be performed
+        # by the same monitor (i.e. only the monitor that dynamically added a path can remove or update
+        # that monitor).
         self.__dynamic_paths = {}
 
         # a dict of log paths pending removal once their bytes pending count reaches 0
@@ -386,8 +393,12 @@ class CopyingManager(StoppableThread, LogWatcher):
             matchers = []
             for m in self.__log_matchers:
                 if m.log_path == log_path:
-                    # make sure the matcher is always finished if called from a non scheduled deletion
-                    m.finish()
+                    # Make sure the matcher is always finished if called from a non scheduled deletion (e.g. on shutdown/config reload).
+                    # This ensures that the __dynamic_matchers dict on the main thread will also clean
+                    # itself up when it notices the matcher is finished.
+                    # We set the matcher to finish immediately, because we want the matcher to finish now, not when it's finished
+                    # any existing processing
+                    m.finish( immediately=True )
                 else:
                     matchers.append( m )
 
@@ -1049,12 +1060,12 @@ class CopyingManager(StoppableThread, LogWatcher):
 
             matcher.finish()
 
-            # remove from list of logs pending removal
-            self.__lock.acquire()
-            try:
-                self.__logs_pending_removal.pop( path, None )
-            finally:
-                self.__lock.release()
+        # remove from list of logs pending removal
+        self.__lock.acquire()
+        try:
+            self.__logs_pending_removal = {}
+        finally:
+            self.__lock.release()
 
     def __purge_finished_log_matchers( self ):
         """
@@ -1103,25 +1114,15 @@ class CopyingManager(StoppableThread, LogWatcher):
                 log.log( scalyr_logging.DEBUG_LEVEL_0, "Log matcher not found for %s" % path )
                 continue
 
-            processor_paths = matcher.update_log_entry_config( log_config )
-            for processor_path in processor_paths:
-                self.__log_paths_being_processed.pop( processor_path, None )
+            # update the log config of the matcher, which closes any open processors, and returns
+            # their checkpoints
+            closed_processors = matcher.update_log_entry_config( log_config )
+            for processor_path, checkpoint in closed_processors.iteritems():
+                checkpoints[processor_path] = checkpoint
 
-                # iterate over full list of processors to see if we need to remove any.
-                # We manually iterate so that we can close the processors if necessary
-                # and also so we can get checkpoints so files aren't copied from the beginning
-                # note:  manually calculate the len each loop iteration as it changes if items
-                # are removed
-                i = 0
-                while i < len( self.__log_processors ):
-                    p = self.__log_processors[i]
-                    if p.log_path == processor_path:
-                        checkpoints[p.log_path] = p.get_checkpoint()
-                        p.close()
-                        del self.__log_processors[i]
-                    else:
-                        i += 1
             reloaded.append( matcher )
+
+        self.__remove_closed_processors()
 
         self.__create_log_processors_for_log_matchers( log_matchers, checkpoints=checkpoints, copy_at_index_zero=True )
         self.__create_log_processors_for_log_matchers( reloaded, checkpoints=checkpoints, copy_at_index_zero=True )
@@ -1132,6 +1133,23 @@ class CopyingManager(StoppableThread, LogWatcher):
             self.__pending_log_matchers = [lm for lm in self.__pending_log_matchers if lm not in log_matchers]
         finally:
             self.__lock.release()
+
+    def __remove_closed_processors( self ):
+        """
+        Removes any closed log processors from the __log_processors and __log_paths_being_processed lists
+        """
+        # shallow copy the processor list for iteration
+        processor_list = self.__log_processors[:]
+
+        # set processors to empty
+        self.__log_processors = []
+        self.__log_paths_being_processed = {}
+
+        # add back any processors that haven't been closed
+        for p in processor_list:
+            if not p.is_closed():
+                self.__log_processors.append(p)
+                self.__log_paths_being_processed[p.log_path] = p
 
     def __create_log_processors_for_log_matchers( self, log_matchers, checkpoints=None, copy_at_index_zero=False ):
         """
--- a/scalyr_agent/copying_manager.py
+++ b/scalyr_agent/copying_manager.py
@@ -37,6 +37,7 @@ from scalyr_agent.scalyr_client import AddEventsRequest
 
 log = scalyr_logging.getLogger(__name__)
 
+SCHEDULED_DELETION = "scheduled-deletion"
 
 class CopyingParameters( object ):
     """Tracks the copying parameters that should be used for sending requests to Scalyr and adjusts them over time
@@ -216,24 +217,22 @@ class CopyingManager(StoppableThread, LogWatcher):
                 else:
                     self.__expanded_server_attributes[key] = value
 
-        # We keep track of which paths we have configs for so that when we add in the configuration for the monitor
-        # log files we don't re-add in the same path.  This can easily happen if a monitor is used multiple times
-        # but they are all just writing to the same monitor file.
-        self.__all_paths = {}
+        self.__dynamic_matchers = {}
+        self.__dynamic_paths = {}
 
         # a dict of log paths pending removal once their bytes pending count reaches 0
         # keyed on the log path, with a value of True or False depending on whether the
         # log file has been processed yet
         self.__logs_pending_removal = {}
 
-        # a list of log_matches for logs with configs that need to be reloaded.
+        # a dict of log_configs keyed by log_path for logs with configs that need to be reloaded.
         # Logs need to be reloaded if their configuration changes at runtime
         # e.g. with the k8s monitor if an annotation attribute changes such as the
         # sampling rules or the parser, and this is outside the usual configuration reload mechanism.
         # By keeping logs that need reloading in a separate 'pending' list
         # we can avoid locking around log_processors and log_paths_being_processed containers,
         # and simply process the contents of this list on the main loop
-        self.__logs_pending_reload = []
+        self.__logs_pending_reload = {}
 
         # a list of dynamically added log_matchers that have not been processed yet
         self.__pending_log_matchers = []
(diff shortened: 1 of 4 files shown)


scalyr_agent/log_processing.py
```diff
@@ -1399,6 +1399,12 @@ def __init__(self, file_path, config, log_config, close_when_staleness_exceeds=N
         # Tracks whether the processor has recently logged data
         self.__is_active = False
 
+        # Tracks whether or not the processor has been marked as finished
+        self.__is_finished = False
```
czerwingithub: This means we have two pieces of state `__is_closed` and `__is_finished` that feels pretty similar to one another.  In effect, I don't think they mean anything different... when that state is `true`, the `LogMatcher` that owns it should remove it from its list of processors.  Is that correct, or am I missing something?

I have a strong preference for just re-using the concept of `close` for this feature.  Currently, a LogProcessor is considered `closed` when either (1) the file it was copying has been deleted or (2) the log file is considered stale because it has not had bytes written to it in X seconds and we are configured to close stale log files.  This would just introduce a new condition for that `close` case.

We do still need some additional state to record when a `LogProcessor` has been requested to close.  But, I think that's the only state we need.  We could call it `__close_when_caught_up`.  I'll add more down below to describe this.

scalyr_agent/log_processing.py
```diff
@@ -1399,6 +1399,12 @@ def __init__(self, file_path, config, log_config, close_when_staleness_exceeds=N
...
+
+        # Tracks the number of bytes that was last read from the processor
+        self.__prev_bytes_read = 0
```
czerwingithub: I think we can remove this if we move the check for when to close a processor based on the `__close_when_caught_up` to the completion callback in the `perform_processing` method.

scalyr_agent/log_processing.py
```diff
@@ -1447,6 +1453,36 @@ def __init__(self, file_path, config, log_config, close_when_staleness_exceeds=N
 
         self.__disable_processing_new_bytes = config.disable_processing_new_bytes
 
+    def finish( self ):
```
czerwingithub: Change this to `close_when_caught_up`.  I hate that name, but conveys the idea.  Other alternatives are `close(when_caught_up=False)` (so the idea of being caught is just optional.. but that would imply we support the case `when_caught_up=False` which would require more coding.

scalyr_agent/log_processing.py
```diff
@@ -1447,6 +1453,36 @@ def __init__(self, file_path, config, log_config, close_when_staleness_exceeds=N
...
+        When a log processor is `finished`, then it can be scheduled for removal from the
+        main copy manager once it no longer has any bytes available for reading.
+        """
+        self.__lock.acquire()
+        try:
+            self.__is_finished = True
+        finally:
+            self.__lock.release()
+
+    def is_finished( self ):
```
czerwingithub: I think this can be removed and just rely on the `is_closed` method.

scalyr_agent/log_processing.py
```diff
@@ -1447,6 +1453,36 @@ def __init__(self, file_path, config, log_config, close_when_staleness_exceeds=N
...
+
+        If this method returns True, then the log processor can be removed from the copy manager
+        """
+        # We only lock for __is_finished.  The other fields are only modified on the main thread
+        self.__lock.acquire()
+        try:
+            # __is_finished means the log has been marked as finished
+            # __prev_bytes_read == 0 means the last call to perform_processing didn't have any bytes for this processor
+            # __total_bytes_pending == 0 means there are no more bytes available from the iterator
+            return self.__is_finished and self.__prev_bytes_read == 0 and self.__log_file_iterator.available == 0
```
czerwingithub: Note, I believe there is a race condition here.  It could be that there is a request in-flight that has the last bytes from the file in it... The `available` would still return `0` even though the request is in flight.

Now, this method will only be called (presumably) by the `CopyManager` so it could be that the `CopyManager` only invokes it when it knows there are no requests in flight... but that means we'd have to document this method as such and make sure we don't violate it.  With the changes I propose below, I think we can avoid having to do all of that.

scalyr_agent/log_processing.py
```diff
@@ -1701,6 +1737,9 @@ def perform_processing(self, add_events_request, current_time=None):
             self.__lock.acquire()
             self.__total_bytes_being_processed = bytes_copied
             self.__total_bytes_pending = self.__log_file_iterator.available
+
+            # also keep track of the amount of bytes read this time
+            self.__prev_bytes_read = bytes_read
```
czerwingithub: I dislike that github prevents me from add comments to lines that you didn't change.  Argh.  So, I'll add my comments here for this method.

You should update `_is_closed` in this method as part of the completion callback.  You can see we already update `_is_closed` there for the other conditions.  I always view that the completion back is when _commit_ changes to the `LogProcessor`.  So, that's when we should update `_is_closed`.

For the logic, I think we should set `_is_closed` to `True` if `_close_when_caught_up` was true BEFORE the call above to line 1598 when we invoke `self.__log_file_iterator.scan_for_new_bytes`.  This prevents a subtle race condition.  We need to only consider us being able to close if someone has asked us to close when caught up before we noticed there are no new bytes to be read.  So, that means we should just capture a local copy of `_close_when_caught_up` up there and then refer to it in the completion callback.

scalyr_agent/log_processing.py
```diff
@@ -2282,6 +2322,46 @@ def config(self):
         """
         return self.__log_entry_config
 
+    def finish( self ):
```
czerwingithub: If we are moving the names of this to `close` in `LogProcessor` we should probably call it `close` in the `LogMatcher` as well.  So, maybe `close_when_processing_completed` (ok, don't hit me).  I hate to just use `close` here since the name to me implies the `LogProcessor` will be closed when the method finishes (it is synchronous).  But, I can't up with a better name then `close_when_processing_completed`

scalyr_agent/log_processing.py
```diff
@@ -2282,6 +2322,46 @@ def config(self):
...
+    def is_finished( self ):
+        """
+        Returns true if the log matcher and all of its processors are finished, and the matcher has
+        been processed at least once
+        """
+        result = False
+        self.__lock.acquire()
+        try:
+            # check if we are finished
+            result = self.__is_finished and self.__last_check is not None
```
czerwingithub: Add a comment as to why you check `__last_check`.  I think that is the right thing to do and understand why, but you should document it.

czerwingithub: Why not just return `False` here?  You already have an early exit down below.

scalyr_agent/log_processing.py
```diff
@@ -2282,6 +2322,46 @@ def config(self):
...
+
+            # check if all the processors are finished
+            for processor in self.__processors:
+                # exit early if the matcher or a previous processor is finished
+                if not result:
```
czerwingithub: If you do the early exit above, this one can be written in a more obvious manner:
```
 if not processor.is_finished:
     return False
```

scalyr_agent/log_processing.py
```diff
@@ -2402,6 +2491,8 @@ def find_matches(self, existing_processors, previous_state, copy_at_index_zero=F
             self.__lock.acquire()
             for new_processor in result:
                 self.__processors.append(new_processor)
+                if is_finished:
```
czerwingithub: Add a comment as to why you do this.

scalyr_agent/log_processing.py
imron: As discussed previously, LogMatcher still uses `finish`, but relies on LogProcesser being closed.

scalyr_agent/log_processing.py
imron: Done.  Though I changed `__close_when_caught_up`  to `__close_at_eof`

scalyr_agent/copying_manager.py
```diff
@@ -1038,39 +1075,56 @@ def __scan_for_pending_log_files( self ):
...
+        # reload the config of any matchers/processors that need reloading
+        reloaded = []
+        for path, log_config in pending_reload.iteritems():
+            log.log( scalyr_logging.DEBUG_LEVEL_1, "Pending reload for %s" % path )
+
+            # only reload matchers that have been dynamically added
+            matcher = self.__dynamic_matchers.get( path, None )
+            if matcher is None:
+                log.log( scalyr_logging.DEBUG_LEVEL_0, "Log matcher not found for %s" % path )
+                continue
```
czerwingithub: Isn't this a bug if we hit this?  What invariant is being broken?  that `__pending_log_matchers` did not have an item from `__log_pending_reload`?

scalyr_agent/copying_manager.py
```diff
@@ -1038,39 +1075,56 @@ def __scan_for_pending_log_files( self ):
...
+
+            processor_paths = matcher.update_log_entry_config( log_config )
```
czerwingithub: I'm not a big fan of this interface for `update_log_entry`.  It expects the caller to do too much -- get checkpoints and remember to close the processors.  What if we changed the interface to this:  Inside `update_log_entry_config`, it closes any processors it has... and removes them from its internal list.  It also gets the checkpoints for all processors it is closing and writes that to a passed in `dict` that's an argument on `update_log_entry_config`.

We then use those checkpoints down below when we create the processors.

That leaves how we clean out `__log_paths_being_processed` and `__log_processors`.  Maybe we have a method that iterates over `__log_processors` and removes any processors that are closed, while also removing them from `_log_paths_being_processed`?  It feels like there must be other places where we need this same idea of "removed closed processors and update `_log_paths_being_processed`.

I'm not in love with my suggestion, but I just dislike how much work is being spread out from the `update_log_entry_config` call.

scalyr_agent/log_processing.py
```diff
@@ -2282,6 +2337,46 @@ def config(self):
...
+    def is_finished( self ):
+        """
+        Returns true if the log matcher is finished and all of its processors are closed, and the matcher has
+        been processed at least once
+        """
+        self.__lock.acquire()
+        try:
+            # we are only finished if `is_finished` is set and we have tried to check for a match
+            # on the matcher at least once
+            if self.__is_finished and self.__last_check is None:
```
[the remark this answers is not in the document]
imron: No.  This is catching the case where the matcher has been marked as `finished`, but we have never tried to match it.  In those cases we need to return false, otherwise we won't pick up short lived files.  I'll reword the comment to make this more clear.

scalyr_agent/copying_manager.py
```diff
@@ -381,25 +371,31 @@ def remove_log_path( self, monitor_name, log_path ):
...
+                log.log(scalyr_logging.DEBUG_LEVEL_0, 'Tried removing a log file \'%s\' for monitor \'%s\', but it is not being monitored' % (log_path, monitor_name) )
+                return
+
+            # If we are not a scheduled deletion, make sure only the monitor that added this path can remove it
+            if monitor_name != SCHEDULED_DELETION and log_path in self.__dynamic_paths and self.__dynamic_paths[log_path] != monitor_name:
+                log.log(scalyr_logging.DEBUG_LEVEL_0, 'Tried removing a log file \'%s\' for monitor \'%s\', but it is currently being monitored by \'%s\'' % (log_path, monitor_name, self.__dynamic_paths[log_path] ) )
+                return
+
+            log.log(scalyr_logging.DEBUG_LEVEL_0, 'Removing log file \'%s\' for \'%s\'' % (log_path, monitor_name ) )
+            #do the removals
```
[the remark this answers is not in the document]
imron: No, it's called by other monitors - usually in the case of shutdown.

I can't recall the details but when reloading config, there was some issue where the copying manager was keeping around previous log matchers after the monitors had been shut down.

scalyr_agent/copying_manager.py
```diff
@@ -1007,29 +1040,33 @@ def __remove_logs_scheduled_for_deletion( self ):
...
-                        self.__logs_pending_removal.pop( path, None )
-                    finally:
-                        self.__lock.release()
-                    log.warn( "Log scheduled for removal is not being monitored: %s" % path )
+    def __purge_finished_log_matchers( self ):
+        """
+        Removes from the list of log matchers any log matchers that are finished
+        """
+        # make a shallow copy for iteration
+        matchers = self.__dynamic_matchers.copy()
```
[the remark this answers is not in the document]
imron: `dynamic_matchers` is only touched by the main thread and shouldn't need the lock.  `dynamic_paths` is touched by multiple threads and does need the lock.

scalyr_agent/copying_manager.py
```diff
@@ -1007,29 +1040,33 @@ def __remove_logs_scheduled_for_deletion( self ):
...
+
+        for path, m in matchers.iteritems():
```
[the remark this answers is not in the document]
imron: My main thinking was to simplify the locking code.  Access to `log_matchers` needs to be protected by a lock.  `remove_log_path` also acquires and releases the same lock.  So we'd need to lock over iteration, but then when we call `remove_log_path` inside iteration it'll double lock, causing a deadlock.

We could get around it by locking and making a shallow copy of matchers and iterating over that, but then there's still the middle man.

scalyr_agent/copying_manager.py
imron: Yes it is a bug.  The invariant broken is that we are trying to change the configuration of log file, but we are no longer tracking that file.  It should never happen, but the warning will let us know if it ever gets tripped.

scalyr_agent/copying_manager.py
```diff
@@ -381,25 +371,31 @@ def remove_log_path( self, monitor_name, log_path ):
...
+                log.log(scalyr_logging.DEBUG_LEVEL_0, 'Tried removing a log file \'%s\' for monitor \'%s\', but it is currently being monitored by \'%s\'' % (log_path, monitor_name, self.__dynamic_paths[log_path] ) )
+                return
+
+            log.log(scalyr_logging.DEBUG_LEVEL_0, 'Removing log file \'%s\' for \'%s\'' % (log_path, monitor_name ) )
+            #do the removals
+            matchers = []
+            for m in self.__log_matchers:
+                if m.log_path == log_path:
+                    # make sure the matcher is always finished if called from a non scheduled deletion
+                    m.finish()
```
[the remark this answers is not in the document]
imron: I think you are right.  The point of calling finish here is that the main thread still has a `dynamic_matchers` dict, which can't be touched by other threads, and we need some way of signalling to the main thread to clean that up.  

If finish is set then the matcher will get purged on the main thread, but as you mentioned it won't happen if the log processors aren't caught up.  Perhaps we should have something like `def finish( immediately=False )` which when true will close all the processors straight away.

imron:
This is now ready for further review.

scalyr_agent/log_processing.py
```diff
@@ -2286,19 +2286,27 @@ def __init__(self, overall_config, log_entry_config, file_system=None):
...
         # The time in seconds past epoch when we last checked for new files that match the glob.
         self.__last_check = None
 
     def update_log_entry_config( self, log_entry_config ):
+        """
+        Updates the log config for the log matcher and any of its processors.
+        In order to update the processors, the processors are simply closed and
+        removed from the list of processors.  These processors will be recreated
+        the next time the file matcher is matched.
+
```
czerwingithub: I would just add some documentation that the caller must retain the checkpoints and be sure they are used when running the matcher the next time.

PR #321, opened 2019-10-17
Bug: Fix support for recursive ** globs by switching to glob2 library

BrianChristie:
Python's builtin glob module does not support recursive globbing with `**` until [Python 3.5](https://docs.python.org/3.5/library/glob.html#glob.glob). 

As we presently support Python 2.7, this change moves to using the third-party library _glob2_ which supports `**`


BrianChristie:
Replacing with https://github.com/scalyr/scalyr-agent-2/pull/322 for CircleCI permissions.

PR #322, opened 2019-10-18
Bug: Fix Support for recursive ** globs by switching to glob2 library

BrianChristie:
Python's builtin glob module does not support recursive globbing with ** until Python 3.5.

As we presently support Python 2.7, this change moves to using the third-party library glob2 which supports **

(Resubmitting to fix CircleCI)


BrianChristie:
On hold pending testing of Python 2.6 compatibility.

PR #326, opened 2019-11-02
#AGENT-271 Do not fail to start agent when cannot contact server

yanscalyr:


scalyr_agent/agent_main.py
```diff
@@ -404,9 +404,10 @@ def __perform_config_checks(self, no_check_remote):
                                         'Please visit https://www.scalyr.com/keys and copy a Write Logs key into the '
                                         '\'api_key\' field in the configuration file' % self.__config.file_path)
                     else:
-                        raise Exception('Failed to send request to the server.  The server address could be wrong, '
-                                        'there maybe a network connectivity issue, or the provided api_token could be '
-                                        'incorrect.  You can disable this check with --no-check-remote-server.')
+                        print >> sys.stderr, ('Failed to send request to the server.  The server address could be '
+                                              'wrong, there maybe a network connectivity issue, or the provided '
+                                              'api_token could be incorrect.  You can disable this check with'
+                                              ' --no-check-remote-server. Scalyr agent will keep trying to connect.')
```
czerwingithub: Let's change that last sentence to `Scalyr agent will keep trying to connect in the background.` 

Also, can you make the same change to line 397?  Theoretically, clock skew could be fixed automatically on their end by something like ntpd, so it's better not to fail the agent at startup (in keeping with the spirit of this change).

PR #328, opened 2019-11-04
#AGENT-268 Agent: Log error messages to the log file during startup

yanscalyr:


yanscalyr:
I didn't do my own testing well enough, this doesn't actually log anything to the log file. Seems like telling it to do so would require the config to be loaded, which doesn't happen until after the error messages would be thrown. Needs some thinking.

yanscalyr:
Logs as early as possible, which is after the config check. Automated restarts (which are the usecase here) shouldn't cause valid configuration to change so we are likely to pass that check. Will log if the API key is wrong or the server is just unreachable.

scalyr_agent/agent_main.py
```diff
@@ -404,6 +413,11 @@ def __perform_config_checks(self, no_check_remote):
                                         'Please visit https://www.scalyr.com/keys and copy a Write Logs key into the '
                                         '\'api_key\' field in the configuration file' % self.__config.file_path)
                     else:
+                        log.error('Failed to send request to the server.  The server address could be '
+                                  'wrong, there maybe a network connectivity issue, or the provided '
```
czerwingithub: Let's rewrite that second and subsequent sentences to be "The server address could be wrong, there could be a network connectivity issue, or the provided_token could be incorrect.  The agent will keep trying to connect in the background.  You can disable this check with --no-check-remote-server."

scalyr_agent/agent_main.py
```diff
@@ -394,6 +400,9 @@ def __perform_config_checks(self, no_check_remote):
                 if ping_result != 'success':
                     if 'badClientClockSkew' in ping_result:
                         # TODO:  The server does not yet send this error message, but it will in the future.
+                        log.error('Sending request to the server failed due to bad clock skew.  The system '
+                                  'clock on this host is too off from actual time.  Scalyr agent will keep '
```
czerwingithub: "too off" -> "too far off"

czerwingithub: Instead of "Scalyr agent will" -> "The agent will"

scalyr_agent/agent_main.py
yanscalyr: Where you wrote "provided_token" did you mean that to be a space, like "provided token"?

PR #330, opened 2019-11-07
Reformat with black code formatter and add automatic pre-commit hooks

BrianChristie:
This draft shows reformats the entire codebase with the [black auto formatter](https://black.readthedocs.io/en/stable/). would look like.

Black designed in the same spirit as `go fmt`, it has essentially no configurable options, rigorously follows [PEP8](https://pep8.org), and where there is not a standard to follow, attempts to make a reasonable choice and enforce it.

The idea is to remove all mental overhead of ever having to consider formatting, because it always comes out the same way, automatically.

I've also included [pre-commit](https://pre-commit.com) hooks, which automatically wire up black to run prior to committing. 

Additionally, this enables [additional pre-commit hooks](https://github.com/pre-commit/pre-commit-hooks#hooks-available)
- `end-of-file-fixer`: Makes sure files end in a newline and only a newline.
- `trailing-whitespace`: Trims trailing whitespace.

Once this change is merged, all currently open PR's need to run black manually to avoid merge conflicts.
The easiest way to do this is
- copy the `.pre-commit-config.yaml` and `pyproject.toml` from here
- execute `pre-commit run --all-files`, then push the changes to the open branch.


BrianChristie:
This is ready for merge. 
Let me know if there's any objections, otherwise I'll squash and merge it in.

PR #331, opened 2019-11-13
Feature: Kubernetes Sidecar Mode for High Volume Logging

BrianChristie:
This adds a sidecar logging mode for use in Kubernetes, which can be necessary when sending a higher volume of log data than a single agent is able to keep up with.

This PR adds the environment variable `K8S_SIDECAR_MODE=` `true|false`.

When sidecar mode is enabled, the Kubernetes monitor uses additional logic in ContainerEnumerator.get_containers() to filter the container list to only those which belong to the Pod which the agent is running within. 

Additionally, the logic to avoid capturing the agent's own Kubernetes log file is disabled. This is because presently that is excluded by excluding the agent's own pod name, but in this case with sidecar mode, capturing the agent's own pod is desired. This was discussed and impact is seeing the agent container in the Scalyr UI, if the Agent emits any logs on STDOUT, which is considered acceptable presently. In the future this could be refactored to exclude the agent's container by ContainerID, however this information is not presently available in the Kubernetes downward API so another technique will have to be used.

This has been tested locally and in a GKE cluster (using Docker mode). I'd like to discuss some additional testing prior to merge.

### ToDo

- [x] Test on MiniKube
- [x] Test on GKE

### DIFF
--- a/.circleci/smoketest_k8s.sh
+++ b/.circleci/smoketest_k8s.sh
@@ -76,9 +76,6 @@ kubectl create configmap scalyr-config \
 --from-literal=SCALYR_K8S_CLUSTER_NAME=ci-agent-k8s-${CIRCLE_BUILD_NUM} \
 --from-literal=SCALYR_SERVER=https://app-qatesting.scalyr.com
 
-# TODO: Debugging end to end tests
-kubectl get configmap scalyr-config --output yaml
-
 # The following line should be commented out for CircleCI, but it necessary for local debugging
 # eval $(minikube docker-env)
 
--- a/.circleci/smoketest_k8s.sh
+++ b/.circleci/smoketest_k8s.sh
@@ -76,6 +76,9 @@ kubectl create configmap scalyr-config \
 --from-literal=SCALYR_K8S_CLUSTER_NAME=ci-agent-k8s-${CIRCLE_BUILD_NUM} \
 --from-literal=SCALYR_SERVER=https://app-qatesting.scalyr.com
 
+# TODO: Debugging end to end tests
+kubectl get configmap scalyr-config --output yaml
+
 # The following line should be commented out for CircleCI, but it necessary for local debugging
 # eval $(minikube docker-env)
 
--- a/.circleci/smoketest_k8s.sh
+++ b/.circleci/smoketest_k8s.sh
@@ -74,7 +74,7 @@ kubectl create secret generic scalyr-api-key --from-literal=scalyr-api-key=${SCA
 # Create configmap
 kubectl create configmap scalyr-config \
 --from-literal=SCALYR_K8S_CLUSTER_NAME=ci-agent-k8s-${CIRCLE_BUILD_NUM} \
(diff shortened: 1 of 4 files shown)


.circleci/smoketest_k8s.sh
```diff
@@ -74,7 +74,10 @@ kubectl create secret generic scalyr-api-key --from-literal=scalyr-api-key=${SCA
 # Create configmap
 kubectl create configmap scalyr-config \
 --from-literal=SCALYR_K8S_CLUSTER_NAME=ci-agent-k8s-${CIRCLE_BUILD_NUM} \
---from-literal=SCALYR_SERVER=qatesting.scalyr.com
+--from-literal=SCALYR_SERVER=https://app-qatesting.scalyr.com
+
+# TODO: Debugging end to end tests
```
imron: We can delete this now

scalyr_agent/tests/configuration_k8s_test.py
```diff
@@ -372,3 +374,16 @@ def _test_k8s_ignore_namespaces_supersedes(env_str, local_str, expected):
         _test_k8s_ignore_namespaces_supersedes(None, None, kube_system)
         _test_k8s_ignore_namespaces_supersedes(None, r'""', [])
         _test_k8s_ignore_namespaces_supersedes(r'', r'""', [])
+
+
+class TestK8SUtils(TestConfigurationBase):
+    """
+    Tests for monitor_utils/k8s.py
+    """
+    def test_k8s_utils_qualified_name(self):
```
imron: Can we also have tests for checking when is_valid() is false - testing all combinations, e.g.  `namespace` None but valid `podname`, and `podname` None but valid `namespace`, and both `None`.

Likewise tests for == that test a valid Qualified name against a non-valid Qualified name.

scalyr_agent/tests/configuration_k8s_test.py
```diff
@@ -372,3 +375,36 @@ def _test_k8s_ignore_namespaces_supersedes(env_str, local_str, expected):
...
+    def setUp(self):
+        super(TestK8SUtils, self).setUp()
+        self.__pod_valid_a = QualifiedName("default", "scalyr-agent-2-75d69db5cc-fcl2s")
+        self.__pod_valid_b = QualifiedName("default", "scalyr-agent-2-75d69db5cc-fvhmw")
+        self.__pod_no_namespace = QualifiedName(None, "scalyr-agent-2-75d69db5cc-fcl2s")
+        self.__pod_no_podname = QualifiedName("default", None)
+        self.__pod_no_namespace_no_podname = QualifiedName(None, None)
+
+    def test_k8s_utils_qualified_name_is_valid(self):
+        assert (self.__pod_valid_a.is_valid())
```
imron: Missing this in my comments yesterday, but for consistency with other tests, can you use `self.assertTrue` and `self.assertFalse` for these tests rather than just a plan `assert`. and `assert( not`.

PR #332, opened 2019-11-15
AGENT-263 Improve addEvents format

yanscalyr:


yanscalyr:
One thing to consider possibly for the future, we are sending both the log id and the thread id in the same way, and the values are identical, if the server used the thread id as the log id thats one less field to send.

scalyr_agent/scalyr_client.py
```diff
@@ -715,26 +715,28 @@ def current_size(self):
...
-        @return: True if there was the allowed bytes to send were not exceeded by adding this thread to the
+        @return: True if there was the allowed bytes to send were not exceeded by adding this log to the
             request.
         @rtype: bool
         """
         # Have to account for the extra space this will use when serialized.  See how much space we can allow for
         # the post fix right now.
         available_size_for_post_fix = self.__max_size - self.__buffer.tell()
-        return self.__post_fix_buffer.add_thread_entry(thread_id, thread_name,
+        return self.__post_fix_buffer.add_log_and_thread_entry(thread_id, thread_name, log_attrs,
```
czerwingithub: Fix indentation on the line below.

scalyr_agent/scalyr_client.py
```diff
@@ -986,6 +988,45 @@ def _calculate_per_thread_extra_bytes():
...
+    @return: An array of two int entries.  The first entry is how many extra bytes are added when adding the
+        first thread to the threads JSON array and the second is how many extra bytes are added for all subsequent
+        threads.  (The number differences by at least one due to the need for a comma to be inserted).
+    @rtype: [int]
+    """
+    # An array of the number of bytes used to serialize the array when there are N threads in it (where N is the
+    # index into size_by_entries).
+    sizes_by_entries = []
+
+    # Calculate sizes_by_entries by actually serialzing each case.
```
czerwingithub: s/serialzing/serializing

scalyr_agent/scalyr_client.py
```diff
@@ -986,6 +988,45 @@ def _calculate_per_thread_extra_bytes():
...
+    """Calculates how many extra bytes are added to the serialized form of the threads JSON array
+    when adding a new thread, excluding the bytes for serializing the thread id and name themselves.
+
+    This is used below by the PostFixBuffer abstraction to help calculate the number of bytes the serialized form
+    of the PostFixBuffer will take, without having to actually serialize it.  It was found that doing the heavy
+    weight process of serializing it over and over again to just get the size was eating too much CPU.
+
+    @return: An array of two int entries.  The first entry is how many extra bytes are added when adding the
+        first thread to the threads JSON array and the second is how many extra bytes are added for all subsequent
+        threads.  (The number differences by at least one due to the need for a comma to be inserted).
```
czerwingithub: s/The number differences/The number differs/

scalyr_agent/scalyr_client.py
```diff
@@ -1131,12 +1182,13 @@ def set_position(self, position):
         """
         # The position value should by an array with three entries: the size, the client timestamp, and the number
         # of threads.  Since threads are always added one after another, it is sufficient just to truncate back to that
-        # previous length.
+        # previous length. Same with logs, and number of logs and threads should be the same so piggyback on the check.
```
czerwingithub: Sorry, this comment is not quite clear.  Can you rephrase it?

scalyr_agent/scalyr_client.py
```diff
@@ -1246,16 +1299,19 @@ def __set_attributes( self, thread_id, attributes ):
...
+            tmp_buffer.write(', ')
+        #if attributes is not None:
             # Serialize the attributes object, but we have to remove the closing brace because we want to
             # insert more fields.
-            tmp_buffer.write('attrs:')
-            tmp_buffer.write( scalyr_util.json_encode(attributes) )
-            _rewind_past_close_curly(tmp_buffer)
-            tmp_buffer.write(',')
-        else:
+        #    tmp_buffer.write('attrs:')
```
czerwingithub: Should this really be deleted?

yanscalyr: That is the part that inserts all the log attributes into the single event, so if we want to reduce the message size this is what we remove. Though I shouldn't have left it commented out, should have just removed.

imron: I was thinking we should probably put the entire feature behind a configuration parameter - on by default, but can be turned off in case there are problems.

scalyr_agent/scalyr_client.py
yanscalyr: Fixed where I copied it from as well.

scalyr_agent/scalyr_client.py
yanscalyr: Fixed where I copied it from as well.

scalyr_agent/scalyr_client.py
yanscalyr: That's probably a good idea.

czerwingithub: I debated this in my head a bit (and just had a discussion with Yan).  I would say no, we don't add in this flag.  The main reason I could see us doing it is if we wanted to disable the support in the servers and have customers be able to react to that without upgrading the agent.

However, once this API is public, we cannot control if a customer is using it or not.  So, the server must always support it going forward.

The only thing the flag would give us is a way to turn it off if there is a bug in the client code.  Well, I don't think we should have flags just for that.. if there was a bug (and we do have testing already) then we would either quickly roll out a new version or tell customers to roll back.

czerwingithub:
@yanscalyr Your other changes look good to me.  Let's just get agreement about the feature flag.

scalyr_agent/scalyr_client.py
imron: > The only thing the flag would give us is a way to turn it off if there is a bug in the client code.

Or a bug in the server code.

czerwingithub: Since the server hasn't rolled this out everywhere, ok, let's go ahead and add in the configuration option with a comment saying this will be deleted at some point in the future.

scalyr_agent/scalyr_client.py
```diff
@@ -465,7 +469,7 @@ def close(self, current_time=None):
             self.__connection = None
             self.__last_connection_close = current_time
 
-    def add_events_request(self, session_info=None, max_size=1*1024*1024*1024):
+    def add_events_request(self, session_info=None, max_size=1*1024*1024*1024, disable_new_addevents_format=False):
```
czerwingithub: You already have a property to store `__disable_new_addevents_format` on the `ScalyrClientSession` object.  You shouldn't pass it in here.  In general, we should just specify that once when we create the client session object.

scalyr_agent/scalyr_client.py
```diff
@@ -194,7 +198,7 @@ def ping(self):
         @return:  The status message returned by the server.
         @rtype:
         """
-        return self.send(self.add_events_request())[0]
+        return self.send(self.add_events_request(disable_new_addevents_format=self.__disable_new_addevents_format))[0]
```
czerwingithub: You shouldn't need to pass this argument to `add_events_request` since this object already has that value.  See comment below.

scalyr_agent/configuration.py
```diff
@@ -417,6 +417,10 @@ def profile_clock( self ):
     def profile_log_name( self ):
         return self.__get_config().get_string( 'profile_log_name' )
 
+    @property
+    def disable_new_addevents_format( self ):
```
czerwingithub: Calling things "new_addvents_format" is not a great practice since, when there is another iteration of the `AddEvents`, we will be confused what "new" version this is referring to.

Let's call this `disable_logfile_addevents_format`.  It's a clunky name, but we shouldn't need to use it.. and we will delete it in a month otherwise.  You could argue that since there will not be another `new` format in the next month, it's not necessary.. but I'd like us to develop good habits.

PR #333, opened 2019-11-18
Agent 276 add code coverage

ArthurKamalov:
Add coverage to:
* unit tests
* docker and k8s smoke tests.

If 'build_package.py' used to build smoketesing images, it gets Dockerfiles located in docker/coverage directory, which are with enabled coverage.


docker/coverage/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# base image that creates all necessary dependencies for
```
czerwingithub: Ok, I have to admit, I'm not a big fan of creating separate Buildfiles for each of these targets especially putting them in a separate directory.  Everytime we added in a new python module dependency, we would have to come down into this directory as well as add them in.  It won't be top of mind to people.

I'm going to suggest something a little more funky.  We add in a `--coverage` option to the `build_package` command.  It only works for docker/k8s builds.  When that is specified, before it uses the build file, it first rewrites it to add in the necessary changes.  From what I can see, you just need to add in `coverage` to the `RUN pip` line.. and then replace the `CMD` line.  Correct?  You could write a little regex that catches those lines and modifies them.  You could also fail with a big error message if you do not find those lines to modify.

That would also mean you could revert your changes to `build_package.py` to take in a different docker file path.

What do you think?  I know it's a little weird... but it will create less maintenance headaches in the future.

build_package.py
```diff
@@ -795,14 +826,41 @@ def build_base_files(base_configs="config"):
     os.chdir("py")
 
     shutil.copytree(make_path(agent_source_root, "scalyr_agent"), "scalyr_agent")
+
+    # patching "agent_main.py" entrypoint to enable coverage.
```
czerwingithub: Hmm.. patching the agent_main.py file seems more risky here.  Why not just patch the Dockerfile to replace the `CMD` line with what you had it in the coverage-specific Dockerfile files?    It was something like `CMD ["coverage", "run", "/usr/share/scalyr-agent-2/py/scalyr_agent/agent_main.py", "--no-fork", "--no-change-user", "start"]`.

After all, you are already patching the Dockerfile.  It seems better to just patch one thing.

ArthurKamalov: It will be great if it works without this when `docker-json` and `k8s` will start working again. I had some problems when I tried to get coverage results after agent termination and I was able to solve this only by enabling coverage from code.

Let's see when those jobs will work, maybe there is no need for this.
I removed `agent_main.py` patching and returned Dockerfile CMD patching. .

PR #334, opened 2019-11-20
AGENT-277:  Improve test coverage for copying_manager.py

ArthurKamalov:
* covered_remove_log_path,
* added stale request test.
* start from checkpoint test
* status generation test.
* logs_initial_position test
* schedule_pending_log_path_for_removal


scalyr_agent/tests/copying_manager_test.py
```diff
@@ -245,6 +245,29 @@ def test_schedule_log_path_for_removal(self):
...
+        self.append_log_lines( path, "line1\n" )
+
+        log_config = {
+            "path": path
+        }
+
+        self._manager.add_log_config('unittest', log_config)
+        self._manager.schedule_log_path_for_removal('unittest', path)
+
+        self.fake_scan()
```
czerwingithub: Is there a reason you need to call `fake_scan` three times?  If there is a reason, you should document it here.  If there isn't, and it works with just one scan, let's remove the other ones.

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -784,6 +821,186 @@ def test_pipelined_requests_with_retry_error(self):
...
+
+            fp = open(os.path.join(self._config.agent_data_path, name), 'w')
+            fp.write(json_lib.serialize(data))
+            fp.close()
+
+        # shift time on checkpoint files in the past, to make them stale.
+        shift_time_in_checkpoint_file('checkpoints.json')
+        shift_time_in_checkpoint_file('active-checkpoints.json')
+
+        # start manager.
```
czerwingithub: Add a comment here that we are specifically looking to make sure that the lines fifth and sixth do not show up since they were skipped over due to staleness.

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -784,6 +821,186 @@ def test_pipelined_requests_with_retry_error(self):
...
+        self.assertEquals('Fourth line', lines[1])
+
+        # stopping one more time, but now emulating that checkpoint files are stale.
+        controller.stop()
+
+        self.__append_log_lines('Fifth line', 'Sixth line')
+
+        def shift_time_in_checkpoint_file(name):
+
+            fp = open(os.path.join(self._config.agent_data_path, name), 'r')
```
czerwingithub: Add a comment that you are rewriting the checkpoint file to make it seem like the checkpoint was written in the past.

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -784,6 +821,186 @@ def test_pipelined_requests_with_retry_error(self):
...
+        (request, responder_callback) = controller.wait_for_rpc()
+        lines = self.__extract_lines(request)
+
+        self.assertEquals(2, len(lines))
+        self.assertEquals('First line', lines[0])
+        self.assertEquals('Second line', lines[1])
+
+        controller.stop()
+
+        fp = open(os.path.join(self._config.agent_data_path, 'active-checkpoints.json'), 'r')
```
czerwingithub: Maybe you should pull out a common method for rewriting the checkpoint file and use it here and `test_start_from_full_checkpoint`.  It would take two arguments.. the path of the checkpoint and the delta to adjust the time field by.

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -784,6 +821,186 @@ def test_pipelined_requests_with_retry_error(self):
...
+        data = json_lib.parse(fp.read())
+        fp.close()
+
+        data['time'] += 100
+
+        fp = open(os.path.join(self._config.agent_data_path, 'active-checkpoints.json'), 'w')
```
czerwingithub: Add comment as to what this is testing.. why does shifting it by 100 test new things.

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -784,6 +821,186 @@ def test_pipelined_requests_with_retry_error(self):
...
+        self.assertEquals(2, len(lines))
+        self.assertEquals('First line', lines[0])
+        self.assertEquals('Second line', lines[1])
+
+        from scalyr_agent import copying_manager
+
+        # backup original 'time' module
+        orig_time = copying_manager.time
+
+        class _time_mock(object):
```
czerwingithub: Wrap this in a try/finally block so that if there is an exception, you will always restore `copying_manager.time` to its original value.

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -784,6 +821,186 @@ def test_pipelined_requests_with_retry_error(self):
...
+        # stop thread on manager to write checkouts to file.
+        controller.stop()
+
+        # write some new lines to log.
+        self.__append_log_lines('Third line', 'Fourth line')
+
+        # create new copying manager, but passing previous directory with same log and checkouts.
+        controller = self.__create_test_instance(root_dir=previous_root_dir, auto_start=False)
+
+        self._manager.start_manager(dict(fake_client=True))
```
czerwingithub: Why do you need the `auto_start=False` but then immediately start it here?

scalyr_agent/tests/copying_manager_test.py
ArthurKamalov: It does not work with fewer scans... I tried to document it in the comments in the new commit.

scalyr_agent/tests/copying_manager_test.py
ArthurKamalov: It’s a pity I didn’t do it myself... Thanks.

scalyr_agent/tests/copying_manager_test.py
ArthurKamalov: This is the case, when we have both "checkpoints.json" and "active-checpoints.json", but "active-checpoints.json" has newer time, so it get all values from active_checpoints.
https://github.com/scalyr/scalyr-agent-2/blob/8726d6c2483c9ceb19a37697f266261c31e00b0b/scalyr_agent/copying_manager.py#L1038

So, I thought that it would be enough just to make "active-checpoints.json" bigger, assuming that it has the same time as "checkpoints.json", but after your comment I decided to make it more reliable by reading  "checkpoints.json" time, and setting bigger value to "active-checpoints.json"

scalyr_agent/tests/copying_manager_test.py
ArthurKamalov: The problem was that when copying manager starts, it stops only at "SLEEPING" state. https://github.com/scalyr/scalyr-agent-2/blob/8726d6c2483c9ceb19a37697f266261c31e00b0b/scalyr_agent/tests/copying_manager_test.py#L1144-L1157. This means, it has to do full iteration before we can stop on needed state, and we can't intercept the first "SENDING".
The solution was to not allow it to do its first iteration, by disabling auto start in constructor and by starting it manually. After, I decided to pull automated start from controller's constructor to test cases. I think it is clearer and there is no need to make additional logic to constructors.

scalyr_agent/tests/copying_manager_test.py
ArthurKamalov: Please correct me if I missed something, but which kinds of exceptions we can face here? During copying manager runtime or during declaration and mocking?  In both cases there are only one standard lib time.time() call and few assignments. Do you think they really can fail?

czerwingithub: I agree with you that the code as it stands can't fail.  However, what happens if someone comes along and adds a new line that can throw an exception?  They might not notice that they need to do a try/finally to protect against it.  I think it's generally a good habit and future proofs the code if you use the try/finally pattern when you acquire resources or need to be sure to undo side effects.

scalyr_agent/tests/copying_manager_test.py
czerwingithub: Ok, that looks good, thanks.

scalyr_agent/tests/copying_manager_test.py
ArthurKamalov: That's a very good point. Thank you.

scalyr_agent/tests/copying_manager_test.py
ArthurKamalov: Done.

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -1048,10 +1048,11 @@ class _time_mock(object):
             # This dummy 'time()' should be called on new copying thread iteration
             # to emulate huge gap between last request.
             def time(_self):
-                res = orig_time.time() + self._manager._CopyingManager__config.max_retry_time
-                # restore original 'time' module
-                copying_manager.time = orig_time
-                return res
+                try:
```
czerwingithub: move the try to line 1059, right after you override `copy_manager.time`.

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -1048,10 +1048,11 @@ class _time_mock(object):
...
+                    return orig_time.time() + self._manager._CopyingManager__config.max_retry_time
+                finally:
```
czerwingithub: Move the finally to after line 1067... essentially, the end of the test case.

Essentially, I like to see the `try/finally` be at the same level as where you are overriding the `time` value.

ArthurKamalov: Oh, how could I forget about this. 
Done.

PR #337, opened 2019-11-21
#AGENT-283 Don't include intermediate certs in win32 build

yanscalyr:
Intermediate certs were causing validation to fail for some reason.

### DIFF
--- a/build_package.py
+++ b/build_package.py
@@ -224,7 +224,8 @@ def build_win32_installer_package(variant, version):
     shutil.copy(make_path(agent_source_root, 'win32/ScalyrShell.cmd'), 'Scalyr/bin/ScalyrShell.cmd')
 
     # Copy the cert files.
-    cat_files(glob_files(make_path(agent_source_root, 'certs/*.pem')), 'Scalyr/certs/ca_certs.crt',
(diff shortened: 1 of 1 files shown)


build_package.py
```diff
@@ -224,7 +224,8 @@ def build_win32_installer_package(variant, version):
     shutil.copy(make_path(agent_source_root, 'win32/ScalyrShell.cmd'), 'Scalyr/bin/ScalyrShell.cmd')
 
     # Copy the cert files.
-    cat_files(glob_files(make_path(agent_source_root, 'certs/*.pem')), 'Scalyr/certs/ca_certs.crt',
+    # AGENT-283: Certificate validation on windows seems to fail when the intermediate certs are present, skipping them
+    cat_files(glob_files(make_path(agent_source_root, 'certs/*_root.pem')), 'Scalyr/certs/ca_certs.crt',
```
yanscalyr: Instead of filtering out `*_intermediate.pem` I went with limiting the scope to `*_root.pem` to match similar code on like 738. Let me know if I should still prefer the filtering out and I'll change it.

czerwingithub: I think that's fair point about being consistent with other code, so that's fine with me.

PR #339, opened 2019-11-22
Attempt to echo out a secret

yanscalyr:
Testing our CI security by trying to echo out a secret.


yanscalyr:
Looks safe, not surprised or else CircleCI secrets wouldn't be secret at all.

Basically execution is tied to whatever Circle account is following the repo, so the Scalyr CircleCI account gets builds with the secrets, my personal CircleCI account got a build without them. This also explains why all my pull requests constantly fail almost every test, no secrets were set up.

PR #343, opened 2019-12-03
Bug: Fix non-deterministic glob2 test

BrianChristie:
The test `test_matches_recursive_glob` in tests/log_processing_test.py was non-deterministic. 

This sorts the returned list of log processors to fix that. I've run it multiple times and seems to always pass now.

### DIFF
--- a/scalyr_agent/tests/log_processing_test.py
+++ b/scalyr_agent/tests/log_processing_test.py
@@ -2156,9 +2156,10 @@ class TestLogMatcher(ScalyrTestCase):
                 self.__config, self._create_log_config(self.__glob_recursive)
             )
             processors = matcher.find_matches(dict(), dict())
+            processors.sort() # required to make this test deterministic.
             self.assertEquals(len(processors), 2)
-            self.assertEquals(processors[0].log_path, self.__path_three)
-            self.assertEquals(processors[1].log_path, self.__path_four)
+            self.assertEquals(processors[1].log_path, self.__path_three)
+            self.assertEquals(processors[0].log_path, self.__path_four)
 
             self._close_processors(processors)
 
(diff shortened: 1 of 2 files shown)


BrianChristie:
The failing CircleCI tests appear to be unrelated to the Glob2 test and changes.

scalyr_agent/tests/log_processing_test.py
```diff
@@ -2031,9 +2031,10 @@ def test_matches_recursive_glob(self):
         if sys.version_info >= (2, 6):
             matcher = LogMatcher(self.__config, self._create_log_config(self.__glob_recursive))
             processors = matcher.find_matches(dict(), dict())
+            processors.sort() # required to make this test deterministic.
```
czerwingithub: Hmm.. I would suggest we do this slightly differently.  This is doing the sorting at the test-level and fixing it for this specific test.  Let's make the `find_matches` call itself deterministic.  Most likely, this means you need to sort the output of whatever is doing the globbing.  That way future test cases won't trip over this as well.  Also, sometimes the order files are listed matter to the agent.. so making everything deterministic can avoid weird bugs in the future as well.

BrianChristie: Makes sense.
Is sorting the result just above the return here what you are thinking?https://github.com/scalyr/scalyr-agent-2/blob/bchristie/fix-flaky-glob2-tests/scalyr_agent/log_processing.py#L2559

czerwingithub: I was actually thinking of sorting the results from `glob.glob` here https://github.com/scalyr/scalyr-agent-2/blob/bchristie/fix-flaky-glob2-tests/scalyr_agent/log_processing.py#L2489

PR #347, opened 2019-12-09
introduce option to overwrite journald parser

slomo:
Since journald might contain different formats of logs, there should be an option to set the parser associated with the data. I can now setup parser like `journald_myapplication` in scalyr.

### DIFF
--- a/scalyr_agent/builtin_monitors/journald_monitor.py
+++ b/scalyr_agent/builtin_monitors/journald_monitor.py
@@ -117,6 +117,14 @@ define_config_option(
     default=10 * 60,
 )
 
+define_config_option(
+    __monitor__,
+    "parser",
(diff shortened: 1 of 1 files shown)


slomo:
I am not sure if that is aligned with your product strategy, but the feature would be supper useful. Maybe this would even make sense on ScalyrMonitor inside https://github.com/scalyr/scalyr-agent-2/blob/master/scalyr_agent/scalyr_monitor.py in order to be available to all monitors (I guess shell monitor would also profit from this change). 

At least for journal it is a super important feature because in my opinion journal is more a log transport than a specific kind of log, which can contain many log formats. In my particular case some 3rd party but also my own applications are logging into journal. I use structured log, but the thirdparty applications have "regular" log formats.

PR #348, opened 2019-12-12
Implement extended config for journald monitor

yanscalyr:
Allow log configuration for the journald monitor using log config lines containing `journald_unit`, which will match on the `unit` system field in the journald log line. a `journald_unit` value equal to `.*` will configure the default logger for the monitor, anything else will create a new log file.


yanscalyr:
Needs more tests, any suggestions on unit tests I could add for this would be appreciated.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +559,169 @@ def gather_sample(self):
...
+        full_path = os.path.join(
+            global_config.agent_log_path,
+            file_template.safe_substitute({"ID": match_hash}),
+        )
+        log_config["path"] = full_path
+
+    return log_config
+
+
+class LogManager:
```
czerwingithub: As discussed, we are going to tighten up this abstraction.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +559,169 @@ def gather_sample(self):
...
+    def __init__(
```
czerwingithub: After our discussion, I believe the only parameters should be `global_config`, `formatter`, `max_log_size`, `max_log_rotations`.  (`formatter` is used when creating a logger`.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +559,169 @@ def gather_sample(self):
...
+        self.default_config = default_config
+        self._max_log_size = max_log_size
+        self._max_log_rotations = max_log_rotations
+        self.module_name = module_name
+        self.global_config = global_config
+        self.log_watcher = None
+        self.all_configs = self.global_config.get_all_log_configs([])
+        self.loggers_created = False
+
+        self.initialize()
```
czerwingithub: Let's have `initialize` return back a list of functions.. and have a `self.__log_config_creators` which is the list of functions.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +559,169 @@ def gather_sample(self):
...
+
+    def initialize(self):
+        """Called at the end of __init__, intended for overriding by subclasses
+        If configuration can be invalid, it should be checked here
```
czerwingithub: Let's have this take in as an argument the `log_configs` variable and have it returns a list of functions as described above.  Also document this method as needing to through invalid configuration exceptions if any of the `log_configs` are not valid.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +559,169 @@ def gather_sample(self):
...
+        """
+        for config in self.all_configs:
+            match = self.match_config_to_logger(config)
+            if match:
+                regex = re.compile(match)
+                if match not in self.loggers:
+                    self.loggers[match] = {}
+                self.loggers[match]["regex"] = regex
+
+    def match_line_to_logger(self, data, extra_fields=None):
```
czerwingithub: Replace this with `get_config(unit)`.  This will return the log configuration to use for that unit.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +559,169 @@ def gather_sample(self):
...
+        """
+        if extra_fields and "unit" in extra_fields:
+            for match in self.loggers.keys():
+                if match != ".*" and self.loggers[match]["regex"].match(
+                    extra_fields["unit"]
+                ):
+                    return match
+        return None
+
+    def match_config_to_logger(self, config):
```
czerwingithub: Replace this with `get_logger(unit)`.  This is implemented using `get_config`, looking at the `path` field in the returned config.  This will create a logger on demand for the path.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +559,169 @@ def gather_sample(self):
...
+        """Get the unique identifier of the logger that should have this config, intended for overriding by subclasses
+
+        @param config: The configuration for a logger
+        @return: A string uniquely identifying the matched logger
+        """
+        if "journald_unit" in config:
+            return config["journald_unit"]
+        return None
+
+    def set_log_watcher(self, log_watcher):
```
czerwingithub: This just needs to set an instance variable to hold the `log_watcher` now.  When we create the loggers on demand, then we will tell the `log_watcher` that it needs to copy the file up to Scalyr.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +559,169 @@ def gather_sample(self):
...
+        """
+        self.log_watcher = log_watcher
+        if not self.loggers_created:
+            for config in self.all_configs:
+                match = self.match_config_to_logger(config)
+                if match:
+                    self.create_logger(match, config)
+            self.loggers_created = True
+
+    def emit(self, data, extra_fields=None):
```
czerwingithub: Remove this `emit` method.  We will now just rely on the user of this abstraction to call `info` on the logger returned from `get_logger`.

scalyr_agent/scalyr_logging.py
```diff
@@ -831,13 +831,26 @@ class MetricLogFormatter(BaseFormatter):
...
 
     def __init__(self):
-        # TODO: It seems on Python 2.4 the filename and line number do not work correctly.  I think we need to
-        # define a custom findCaller method to actually fix the problem.
         BaseFormatter.__init__(
             self, "%(asctime)s [%(monitor_name)s] %(message)s", "metric-formatter"
         )
 
 
+class JournaldLogFormatter(BaseFormatter):
```
czerwingithub: We should just move this class to the the journald_monitor.py file.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -412,18 +461,93 @@ def _process_entries(self):
...
+        )
+
+        string_buffer.write(
+            "%s %s" % (metric_name, scalyr_util.json_encode(metric_value))
+        )
+
+        if extra_fields is not None:
+            for field_name in extra_fields:
+                field_value = extra_fields[field_name]
+                field_name = self._force_valid_metric_or_field_name(
```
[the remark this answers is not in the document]
yanscalyr: Actually it looks like the extra_fields is configurable, so if we want to match the existing logging behavior I think we need this.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -415,7 +473,8 @@ def _process_entries(self):
             try:
                 msg = entry.get("MESSAGE", "")
                 extra = self._get_extra_fields(entry)
-                self._logger.emit_value("details", msg, extra_fields=extra)
+                logger = self.log_manager.get_logger(extra["unit"])
```
czerwingithub: Is `unit` guaranteed to always be present?  I think it would be safe to do a `extra.get("unit", "")` here just in case.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +563,129 @@ def gather_sample(self):
...
+                catchall_defined = True
+        if not catchall_defined:
+            config_matchers.append(self.create_config_matcher({}))
+
+        return config_matchers
+
+    def create_config_matcher(self, conf):
+        """ TODO: document
+        """
+        config = conf.copy()
```
czerwingithub: This only does a shallow copy.. Let's do a deep one just in case.  We don't really need it now.. but I don't want us to get bit by this latter on if the code changes and we do actually need one.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +563,129 @@ def gather_sample(self):
...
+            match_hash = "monitor"
+
+        def config_matcher(unit):
+            if regex.match(unit) is not None:
+                full_path = os.path.join(
+                    self._global_config.agent_log_path,
+                    file_template.safe_substitute({"ID": match_hash}),
+                )
+                matched_config = JsonObject({"parser": "journald", "path": full_path})
+                matched_config.update(config)
```
czerwingithub: Why not move the creation of `matched_config` to above the definition of `config_matcher`.  It doesn't look to me like you rely on anything in here.

scalyr_agent/builtin_monitors/journald_monitor.py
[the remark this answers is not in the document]
czerwingithub: Right, but extra_fields is defined when the monitor starts up in config option `journal_fields`.  So, we could just do the renaming of the metrics there, instead of having to do it on every `emit`.  Also, this is code from `scalyr_logging`, right?  Let's expose that as a static method and invoke it here, rather than coding the code here.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -435,3 +563,129 @@ def gather_sample(self):
...
+
+    def initialize(self):
+        """ TODO: document
+        """
+        config_matchers = []
+        catchall_defined = False
+        for config in self._global_config.journald_log_configs:
+            config_matcher = self.create_config_matcher(config)
+            config_matchers.append(config_matcher)
+            if config["journald_unit"] == ".*":
```
czerwingithub: You don't really need to even bother to see if there is an existing catch all.  You just need to append yours at the end.. if there is one defined earlier in the list, then that will always be hit and yours will have no effect.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -227,6 +233,18 @@ class JournaldMonitor(ScalyrMonitor):
     "Read logs from journalctl and emit to scalyr"
 
     def _initialize(self):
+        self._metric_or_field_name_rule = re.compile("[_a-zA-Z][\w\.\-]*$")
+        self._file_template = Template("journald_${ID}")
```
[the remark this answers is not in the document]
czerwingithub: Ping on this comment.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -284,11 +316,27 @@ def _initialize(self):
...
     def run(self):
+        self.log_manager.set_log_watcher(self._log_watcher)
         self._checkpoint = load_checkpoints(self._checkpoint_file)
         self._reset_journal()
         ScalyrMonitor.run(self)
 
+    def set_log_watcher(self, log_watcher):
+        self._log_watcher = log_watcher
+
+    def _get_log_watcher(self):
```
czerwingithub: Add documentation mention this returns self also... for a getter I wouldn't expect this so the documenation is nice.

scalyr_agent/builtin_monitors/journald_utils.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+
+    def __init__(self):
+        scalyr_logging.BaseFormatter.__init__(
+            self, "%(asctime)s [journald_monitor()] %(message)s", "metric-formatter"
+        )
+
+
+class LogConfigManager:
+    def __init__(
```
czerwingithub: Add some class-level documentation about what this abstraction does.

scalyr_agent/builtin_monitors/journald_utils.py
```diff
@@ -0,0 +1,155 @@
...
+        for config in self._global_config.journald_log_configs:
+            config_matcher = self.create_config_matcher(config)
+            config_matchers.append(config_matcher)
+        # Add a catchall matcher at the end in case one was not configured
+        config_matchers.append(self.create_config_matcher({}))
+
+        return config_matchers
+
+    def create_config_matcher(self, conf):
+        """ Create a function that will return a logger configuration when passed in data that matches that config.
```
czerwingithub: replace `logger` with `log`

scalyr_agent/builtin_monitors/journald_utils.py
```diff
@@ -0,0 +1,155 @@
...
+
+        def config_matcher(unit):
+            if regex.match(unit) is not None:
+                return matched_config
+            return None
+
+        return config_matcher
+
+    def get_config(self, unit):
+        """ Get a logger configuration that matches the passed in data based on the configured config matchers.
```
czerwingithub: Replace `logger` with `log`

PR #350, opened 2019-12-18
Refactor and reduce use of json_lib

czerwingithub:
We will be deprecating the use key parts of json_lib such
as its own serialization to help reduce the amount of tricky
string Python 2 to 3 work we must do.

To help with this, this changes:

1.  Moves all references to the json_lib parsing and serialization
    code to a few central methods in `util.py` so that we can
    very easily see what is using the serialization.  Also added
    more meaningful function names to capture why the parsing or
    serialization is being used.

2.  Eliminates places where json_lib is used to parse or serialization
    JSON objects when it is not really needed.  It really is only
    need for parsing config files (which might have comments) and
    serializing JSON for requests to Scalyr.  All other uses
    have been eliminated.

Note:  Some "risky" changes are that the checkpoint files are now
being parsed using the standard JSON parsers.  So, when users
upgrade, they theorectically could see issues in that the previous
checkpoint was written by json_lib and is being parsed by the standard.
This shouldn't really be an issue since the serialization happened
without any of the Scalyr optimizations turned on.  The only potential
worry I have is if there were weird UTF-8 characters in file names
that were not properly encoded by json_lib.

Other potential areas are:
- Windows upgrade path (API response now parsed using standard JSON)
- K8s event monitor (event objects parsed using standard JSON)

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 # ------------------------------------------------------------------------
 # author:  Imron Alston <imron@scalyr.com>
-import scalyr_agent.json_lib.objects
 
 __author__ = "imron@scalyr.com"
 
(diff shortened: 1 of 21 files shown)


czerwingithub:
@imron Suggested review path:  Take a look at the changes in `util.py`.  Most of the rest of the code changes are done to use the new methods defined there.

PR #351, opened 2019-12-19
Remove json_lib serialization

czerwingithub:
To help with Python 2->3, we are eliminating the full
json_lib serialization support.  Previously, this was only
used for serializing requests to Scalyr so that we could
leverage Scalyr-specific optimizations.  We realized the code
was already set up to just apply that optimization to strings
without needing the full JSON serialization support.

As part of this change, also had to make anything that
is included in the request (such as server attributes) to
encoded as a dict rather than JsonObject since standard libraries
do not know about JsonObject.


czerwingithub:
@yanscalyr There was one bug in my code.  Can you give a LGTM on this?

PR #353, opened 2019-12-20
Move glob2 third party lib to python 2-specific third party lib

czerwingithub:
As part of the Python 2->3 migration, we are refactoring our
third party library set up.  There are some features that
are supported in Python 3 that we used to have to include
third libraries to get.  To that end, we created new version
specific third party libs.

glob2 falls into this category.  Since we are targeting
support for Python 3.7 for our Python 3 release, we can use
the recursive globbing support that was added in Python 3.5


scalyr_agent/log_processing.py
```diff
@@ -62,6 +55,12 @@
 from os import listdir
 from os.path import isfile, join
 
+if sys.version_info[0] < (3, 5):
```
yanscalyr: The `[0]` seems incorrect to me, wouldn't that get you the major version when we want the full tuple?

czerwingithub: Good catch, thanks.

PR #354, opened 2019-12-21
Remove UUID third party lib

czerwingithub:
Python 2.5 and higher include uuid so we can remove it from third party.


czerwingithub:
@ArthurKamalov I'm assigning this to you since most of the rest of the team are going on PTO.  It's a small change.

PR #355, opened 2020-01-03
Checkpoint file tests, and throw exception for invalid UTF-8

yanscalyr:


scalyr_agent/tests/copying_manager_test.py
```diff
@@ -1022,6 +1046,66 @@ def test_start_without_active_checkpoint(self):
...
+        self.__append_log_lines("First line", "Second line")
+        (request, responder_callback) = controller.wait_for_rpc()
+        lines = self.__extract_lines(request)
+
+        self.assertEquals(2, len(lines))
+        self.assertEquals("First line", lines[0])
+        self.assertEquals("Second line", lines[1])
+        controller.stop()
+
+        self.__append_log_lines("Third line", "Fourth line")
```
czerwingithub: Add some comments explaining what you are looking for here.  For example, "In the case of a bad checkpoint file, the agent should just pretend the checkpoint file does not exist and start reading the logfiles from the end.  In this case, that means lines three and four will be skipped."

scalyr_agent/tests/copying_manager_test.py
```diff
@@ -1022,6 +1046,66 @@ def test_start_without_active_checkpoint(self):
...
+        controller = self.__create_test_instance(
+            root_dir=previous_root_dir, auto_start=False
+        )
+        self._manager.start_manager()
+        (request, responder_callback) = controller.wait_for_rpc()
+        lines = self.__extract_lines(request)
+        self.assertEquals(0, len(lines))
+
+    def test_start_with_non_utf8_checkpoint(self):
+        controller = self.__create_test_instance()
```
czerwingithub: Add comment here explaining this is just a different form of a bad checkpoint file.

scalyr_agent/copying_manager.py
```diff
@@ -1031,7 +1033,7 @@ def __read_checkpoint_state(self):
             # if the active checkpoint file is newer, overwrite any checkpoint values with the
             # updated full checkpoint
             active_checkpoints = scalyr_util.read_file_as_json(
-                active_checkpoint_file_path
+                active_checkpoint_file_path, strict_utf8=True
```
czerwingithub: Can you add a little more to the error logged at line 1048.  Have it say: `Could not read checkpoint file due to error.  Ignoring checkpoint file.`

czerwsup:
This is going to seem supremely random, but this was the I could find a way to contact @czerwingithub I was debugging something and came across your user name and it blew my mind a little bit.  Czerw is my maiden name, and I don't see it often across the web so I thought I would reach out! Apologies if this does not interest/amuse you like it does me!

PR #358, opened 2020-01-06
AGENT-311: Resolve race condition in test_already_warm_two

yanscalyr:


scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -272,6 +272,10 @@ def test_already_warm_two(self):
         warmer = self.__warmer_test_instance
         fake_cache = self.__fake_cache
 
+        # Stop the warmer thread since we don't need it for the test, and to avoid a race condition that sometimes
+        # results in finding too many "already_warm" results
+        warmer.stop()
+
```
czerwingithub: Do we need to restart it after this test finishes?  Or is this warmer instance only created for this test?

yanscalyr: A warmer instance is created once per test case so no need to restart.

czerwingithub: Great, thanks.

PR #360, opened 2020-01-10
Write checkpoint file on startup to avoid many unnecessary logs

yanscalyr:


scalyr_agent/copying_manager.py
```diff
@@ -690,7 +690,7 @@ def run(self):
                 # Just initialize the last time we had a success to now.  Make the logic below easier.
                 last_success = current_time
 
-                last_full_checkpoint_write = current_time
+                last_full_checkpoint_write = 0.0
```
czerwingithub: Why is this the right thing to do?

yanscalyr: To decide when to write the checkpoint file we compare `last_full_checkpoint_write` to `current_time`, and if the later is further ahead than `self.__config.full_checkpoint_interval` only then do we write out the file.
Old way had the agent thinking it wrote the file recently on startup, new way tell it that it hasn't written it since 1970 or so, so it will write it as soon as it can and make sure we have some kind of file to attempt to read from.

czerwingithub: Ok, how about you add a comment ahead of this that says `// Force the agent to write a new full checkpoint as soon as it can`

PR #361, opened 2020-01-14
Arthur/migration m2 test branch

ArthurKamalov:


czerwingithub:
Ok, I know this is a work in progress, but let's start creating separate PRs for this.  Can we make a separate PR for upgrading the request library?  And then I would suggest we separate out the changes to make `json_lib` compatible...What else could we break into separate PRs after that?  I don't want you to have to spend a ton of time re-organizing these, but there's a lot going on.

Also, I think we might want to do the `json_lib` differently.  When we parse a string as json, I think why not just convert that byte string into UTF-8 right away, and then the rest of the code (such as the scanner) works at the unicode character level?  Basically, I'd like to take the approach, where we can, of converting to unicode as soon as we can.  I know, the line processing code paths break this.. but that's because we do want those to be as efficient as possible.

PR #364, opened 2020-01-15
Exclude smoke-k8s and smoke -docker-json from coverage report.

ArthurKamalov:
* Exclude smoke-k8s and smoke -docker-json from coverage report.
* Add smoke-standalone-27 to coverage report.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,7 @@ jobs:
       - run:
           command: |
             mkdir -p coverage-standalone-27 && \
+            # this container seems unnecessary.
             docker container create --name dummy -v shared_vol:/app alpine && \
             docker cp $(pwd)/.circleci/smoketest_standalone.sh dummy:/app/ && \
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -59,10 +59,19 @@ jobs:
           docker_layer_caching: true
       - run:
           command: |
+            mkdir -p coverage-standalone-27 && \
             docker container create --name dummy -v shared_vol:/app alpine && \
```
ArthurKamalov: @czerwingithub, I tried to run a job without this `dummy` container lines, and nothing changed.  Maybe we can remove those lines?

czerwingithub: Eh, let's leave them for now.... I don't fully understand these tests right now.  Just add a comment saying you believe they can be removed.

.circleci/config.yml
```diff
@@ -342,6 +350,5 @@ workflows:
       - coverage:
           requires:
             - unittest-27
-            - smoke-docker-json
             - smoke-docker-syslog
-            - smoke-k8s
```
czerwingithub: I would just comment out the lines for `smoke-docker-json` and `smoke-k8s` with a TODO to re-enable them when they are passing again.

czerwingithub:
@ArthurKamalov We don't need this anymore, so I'm going to close it.  Thanks for putting it forward though.

PR #365, opened 2020-01-16
Fix syslog dependency on docker to only be for docker mode

yanscalyr:


scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -41,13 +41,22 @@
...
+from scalyr_agent.json_lib import JsonObject
+from scalyr_agent.scalyr_monitor import BadMonitorConfiguration
+
+docker_module_available = True
+
+try:
+    from scalyr_agent.builtin_monitors.docker_monitor import (
+        get_attributes_and_config_from_labels,
+    )
+    from scalyr_agent.builtin_monitors.docker_monitor import DockerOptions
```
czerwingithub: Can't the two previous statements be combined?

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -41,13 +41,22 @@
...
+except ImportError:
+    docker_module_available = False
```
czerwingithub: Add a comment saying this is not expected to happen when using docker since we add in the docker module when we build the container image.  Sorry not exact text is not good but you can clean it up.

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1320,6 +1329,12 @@ class SyslogMonitor(ScalyrMonitor):
     """
 
     def _initialize(self):
+        if self._config.get("mode") == "docker" and not docker_module_available:
+            raise BadMonitorConfiguration(
+                "The `docker` module was not found, syslog monitor docker mode is not available.",
```
czerwingithub: Replace with "Failing syslog monitor since docker mode was requested but the `docker` module could not be imported.  This may be due to not include `docker` library when building container image."

scalyr_agent/util.py
```diff
@@ -561,6 +561,35 @@ def is_list_of_strings(vals):
     return True
 
 
+def get_parser_from_config(base_config, attributes, default_parser):
```
czerwingithub: Ok, I know you are just refactoring this code, but can you add a unit test for this function?  I'd like us to get in the habit of adding test cases when we can to help keep improving coverage.

PR #367, opened 2020-01-21
Agent 295 m2 manual changes and strings

ArthurKamalov:
* `cString.StringIO` replaced with `io.StringIO` | `io.BytesIO`
* `open` or `file` replaced with `io.open`
* str replaced with `six.text_type `
* add `unicode_literals` for python2.
* some string literals were changed to bytes literals(`b""`) to work with` io.BytesIO` or` io.open("b")`
* also `str` was removed from logic, only `six.text_type` was keeped.


scalyr_agent/util.py
```diff
@@ -337,18 +345,20 @@ def create_unique_id():
         is also encoded so that is safe to be used in a web URL.
     @rtype: str
     """
-    return base64.urlsafe_b64encode(sha1(uuid.uuid1().bytes).digest())
+    # 2->TODO this function should return unicode.
+    base64_id = base64.urlsafe_b64encode(sha1(uuid.uuid1().bytes).digest())
+    return base64_id.decode("utf-8")
```
czerwingithub: Should the documented return type be byates?

scalyr_agent/util.py
```diff
@@ -549,11 +559,11 @@ def get_pid_tid():
...
     try:
         return "(pid=%s) (tid=%s)" % (
-            str(os.getpid()),
-            str(six.moves._thread.get_ident()),
+            six.text_type(os.getpid()),
+            six.text_type(six.moves._thread.get_ident()),
         )
     except:
-        return "(pid=%s) (tid=Unknown)" % (str(os.getpid()))
+        return "(pid=%s) (tid=Unknown)" % (six.text_type(os.getpid()))
```
czerwingithub: The return type should be a text string , right?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2454,17 +2454,27 @@ def get_cluster_name(self, k8s_cache):
...
+            cluster_name = six.ensure_text(cluster_name)
             return cluster_name
 
         return (k8s_cache and k8s_cache.get_cluster_name()) or None
 
     def _get_node_name(self):
         """ Gets the node name of the node running the agent from downward API """
-        return os.environ.get("SCALYR_K8S_NODE_NAME")
+        # 2->TODO in python2 os.getenv returns 'str' type. Convert it to unicode.
+        node_name = os.environ.get("SCALYR_K8S_NODE_NAME")
```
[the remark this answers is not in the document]
ArthurKamalov: Done. Also, I have added custom fixers to our modernize script, now it searches for `os.environ` too, and if it is found, it acts like he found unmodernized code.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1535,11 +1535,18 @@ def _ensure_session(self):
 
     def get_pod_name(self):
         """ Gets the pod name of the pod running the scalyr-agent """
-        return os.environ.get("SCALYR_K8S_POD_NAME") or os.environ.get("HOSTNAME")
+        # 2->TODO in python2 os.environ returns 'str' type. Convert it to unicode.
+        result = os.environ.get("SCALYR_K8S_POD_NAME") or os.environ.get("HOSTNAME")
```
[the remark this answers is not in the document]
ArthurKamalov: Done

scalyr_agent/tests/configuration_docker_test.py
```diff
@@ -92,13 +95,13 @@ def test_environment_aware_module_params(self, mock_docker):
...
                 else custom_name.upper()
             )
-            envar_value = str(value[1])
+            param_value = value[1]
             if value[2] == ArrayOfStrings:
                 # Array of strings should be entered into environment in the user-preferred format
                 # which is without square brackets and quotes around each element
-                envar_value = envar_value[1:-1]  # strip square brackets
-                envar_value = envar_value.replace("'", "")
+                envar_value = "[{}]".format(", ".join(param_value))
```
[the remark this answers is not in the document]
ArthurKamalov: When we get a string from the `list` with `unicode` elements, there is an additional `u` character at the beginning of each element. 
For example for  `str(["s1", "s2", "s3"])` becomes `'[u\'s1\', u\'s2\', u\'s3\']'` so we need to strip `u` too because config parser can not handle this.  I thought that the join will do the same job without stripping single quotes and `u`.  I have removed square brackets. I really don't know why I put them here... thanks.

scalyr_agent/config_util.py
```diff
@@ -306,8 +298,13 @@ def get_config_from_env(
 
     if strval is None:
         env_name = env_name.lower()
+
         strval = os.getenv(env_name)
 
+    # 2->TODO in python2 os.getenv returns 'str' type. Convert it to unicode.
```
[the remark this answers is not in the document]
ArthurKamalov: done

scalyr_agent/configuration.py
```diff
@@ -2832,8 +2835,10 @@ def import_shell_variables():
                         var_name = entry
                         default_value = ""
 
-                    if var_name in os.environ and len(os.environ[var_name]) > 0:
-                        result[var_name] = os.environ[var_name]
+                    var_value = os.environ.get(var_name)
+                    # 2->TODO in python2 os.environ returns 'str' type. Convert it to unicode.
+                    if var_value is not None and len(var_value) > 0:
```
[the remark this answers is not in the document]
ArthurKamalov: done

scalyr_agent/copying_manager.py
```diff
@@ -1157,7 +1159,7 @@ def __get_next_add_events_task(self, bytes_allowed_to_send, for_pipelining=False
         )
 
         if for_pipelining:
-            add_events_request.increment_timing_data(pipelined=1.0)
+            add_events_request.increment_timing_data(**{"pipelined": 1.0})
```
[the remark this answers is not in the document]
ArthurKamalov: In Python2 `"pipelined" parameter name is stored as `str`, not `unicode`, and this causes errors further.
```
>>> def foo(**kwargs): print(type(kwargs.keys()[0]))
>>> foo(bar="Hello")
<type 'str'>
>>> foo(**{u"bar": "Hello"})
<type 'unicode'>
```

scalyr_agent/log_processing.py
```diff
@@ -624,7 +624,7 @@ def readline(self, current_time=None):
         # or see if we need to parse it as json
         elif self.__parse_format == "json":
             try:
-                json = scalyr_util.json_decode(result.line)
+                json = scalyr_util.json_decode(result)
```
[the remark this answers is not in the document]
ArthurKamalov: Oops, yes, bad cleanup, sorry.  I think I missed this part after my attempts to refactor LogLine. I had an idea to cache the decoded version of the line in it and reuse it where it is needed.

scalyr_agent/log_processing.py
```diff
@@ -2442,11 +2443,11 @@ def __replace_groups_with_hashed_content(re_ex, replacement_ex, line):
                     replacement_matches += 1
                     if group_hash_indicator in replacement_ex:
                         # the group needs to be hashed
-                        replaced_group = replaced_group.encode("utf8")
+                        replaced_group = replaced_group
                         replaced_group = replaced_group.replace(
                             group_hash_indicator,
                             scalyr_util.md5_hexdigest(
-                                _group + redaction_rule.hash_salt
+                                (_group + redaction_rule.hash_salt).encode("utf-8")
```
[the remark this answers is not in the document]
ArthurKamalov: Yes. Good Idea.

scalyr_agent/scalyr_client.py
```diff
@@ -1506,28 +1522,32 @@ def __set_attributes(self, thread_id, attributes):
         self.__thread_id = thread_id
         self.__attrs = attributes
         # A new event.  We have to create the serialization base using provided information/
-        # 2->TODO: use BytesIO, make all written data - binary.
-        tmp_buffer = StringIO()
+        # 2->TODO: should it be bytes, or it will be better to leave it  as unicode and just convert it
```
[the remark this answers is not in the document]
ArthurKamalov: In Python3 json libraries return only unicode, I am pretty sure. That's why I actually have done in the opposite way because I wanted to keep python2 as close as possible to python3 behavior.
https://github.com/scalyr/scalyr-agent-2/blob/76d9e328aea6e21b1f0699e5fdcfe5a89be13101/scalyr_agent/util.py#L145-L156

scalyr_agent/scalyr_client.py
```diff
@@ -1506,28 +1522,32 @@ def __set_attributes(self, thread_id, attributes):
...
-                tmp_buffer.write(scalyr_util.json_encode(thread_id))
-                tmp_buffer.write(", ")
+                tmp_buffer.write(b"log:")
+                tmp_buffer.write(six.ensure_binary(scalyr_util.json_encode(thread_id)))
+                tmp_buffer.write(b", ")
         if self.__disable_logfile_addevents_format and attributes is not None:
-            tmp_buffer.write("attrs:")
-            tmp_buffer.write(scalyr_util.json_encode(attributes))
+            tmp_buffer.write(b"attrs:")
+            # 2->TODO should we worry about type of keys and values in "attributes"?
```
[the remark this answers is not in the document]
ArthurKamalov: Got it.

scalyr_agent/util.py
ArthurKamalov: `get_pid_tid` function is mostly used in log messages creation, so, I think yes it should.

scalyr_agent/test_util.py
```diff
@@ -183,10 +186,13 @@ def parse_scalyr_request(payload):
...
         rewritten_payload += payload[last_processed_index + 1 : x.start(0)]
         # Read the 4 bytes that describe the length, which is stored in regex group 1.
-        length = struct.unpack(">i", x.group(1))[0]
+        # 2->TODO struct.pack|unpack in python2.6 does not allow unicode format string.
+        length = struct.unpack(six.ensure_str(">i"), x.group(1))[0]
         # Grab the string content as raw bytes.
         raw_string = payload[x.end(1) : x.end(1) + length]
-        rewritten_payload += scalyr_util.json_encode(raw_string.decode("utf-8"))
+        rewritten_payload += six.ensure_binary(
+            scalyr_util.json_encode(raw_string.decode("utf-8"))
```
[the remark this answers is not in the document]
ArthurKamalov: Actually, there are a lot more places where we need a text string from `scalyr_util.json_encode` rather than binary string.
Considering this and the fact that python3 json libraries return only text strings, always return binary strings seems like extra work. 
Please take a look at my current variant. I've just introduced an additional keyword argument - `binary` and passing it when binary string is needed.

scalyr_agent/platform_posix.py
```diff
@@ -1180,12 +1192,13 @@ def report_status(self, message):
...
             characters.
-        @type message: str
+        @type message: six.text_type
         """
         # We write out the number of bytes in the message followed by the message.  The number of bytes might be
         # different from the message length in the case of higher ascii, but we'll punt on that for now.
-        self.__fp.write("%d\n" % len(message))
-        self.__fp.write("%s" % message)
+        # 2->TODO message needs to be encoded in Python3
+        self.__fp.write(b"%d\n" % len(message))
```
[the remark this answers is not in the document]
ArthurKamalov: True... Done.

scalyr_agent/tests/syslog_monitor_test.py
```diff
@@ -326,15 +330,15 @@ def test_run_tcp_server(self):
...
+        expected = b"TCP TestXX\n"
         self.connect(s, ("localhost", 8514))
         s.sendall(expected)
         time.sleep(1)
 
         self.monitor.stop(wait_on_join=False)
         self.monitor = None
 
-        f = open("agent_syslog.log")
+        f = open("agent_syslog.log", "rb")
```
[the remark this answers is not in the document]
ArthurKamalov: I looked again on this and understood that this is just log reading, it easily can be done with text mode.

scalyr_agent/scalyr_client.py
```diff
@@ -565,7 +565,9 @@ def generate_body():
             return add_events_request.get_payload()
 
         return self.__send_request(
-            "/addEvents", body_func=generate_body, block_on_response=block_on_response
+            six.ensure_str("/addEvents"),
```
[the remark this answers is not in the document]
ArthurKamalov: `Httplib`(if used)  builds request just by concatenation of request path, body, end headers.
When request path or headers are `unicode`, it tries to decode body implicitly before concatenation, but it can't because it has no utf-8 data. That's why I use `six.ensure_str`, as it's name implies, it keeps 'str' on both versions, e.g. binary in python2 and text in python3

scalyr_agent/json_lib/parser.py
```diff
@@ -233,20 +229,22 @@ def parse_value(self):
             elif c == "n":
                 self.__match("null", "unknown identifier")
                 return None
+            elif c is None:
```
czerwingithub: Add a comment to indicate what `c` being `None` represents.  Does it mean we are at the end of the stream?  Or an invalid character?

scalyr_agent/json_lib/parser.py
```diff
@@ -233,20 +229,22 @@ def parse_value(self):
...
+            # this must be after None check because we can not compare with None in python3
             elif c == "-" or "0" <= c <= "9":
                 return self.__parse_number()
             elif c == "}":
                 return self.__error("'}' can only be used to end an object")
-            elif c == "`":
-                return self.__parse_length_prefixed_string()
+            # 2->TODO prefixed string parsing can be removed because json_lib is only used to parse configs.
+            #elif c == "`":
+                #return self.__parse_length_prefixed_string()
```
czerwingithub: Let's throw an error in this case.

scalyr_agent/json_lib/parser.py
```diff
@@ -290,6 +288,9 @@ def __parse_object(self):
             else:
                 return self.__error("Expected string literal for object attribute name")
 
+            if key is not None:
```
czerwingithub: Why do you need this?  Can you add a comment

scalyr_agent/json_lib/parser.py
```diff
@@ -369,7 +371,7 @@ def __parse_length_prefixed_string(self):
         be at the backtick (`).
 
         @return: The string literal
-        @rtype: str
+        @rtype: six.text_type
```
czerwingithub: Can we just delete this method?

scalyr_agent/json_lib/parser.py
```diff
@@ -638,11 +645,14 @@ def __match(self, chars, error_message):
         and consume them.  In case of a mismatch, raise an exception.
         Only supports low-ASCII characters.
 
-        If error_message is None, we generate a default message."""
+        If error_message is None, we generate a default message.
```
czerwingithub: Why type is `chars` here?  Can you add documentation for that.

czerwingithub: Is it still the case this only supports low-ASCII characters?

scalyr_agent/json_lib/parser.py
```diff
@@ -729,17 +739,17 @@ def __peek_next_non_whitespace(self):
 
         while True:
             # TODO: support any Unicode / UTF-8 whitespace sequence.
-            raw_c = self.__scanner.peek_next_ubyte(none_if_bad_index=True)
-            if raw_c is None:
-                return raw_c
-            c = ord(raw_c)
-            if c == 32 or c == 9 or c == 13 or c == 10:
+            c = self.__scanner.peek_next_ubyte(none_if_bad_index=True)
```
czerwingithub: We should probably change the name for `peek_next_ubyte` to `peek_next_uchar`, right?  Same with `read_ubyte`.

scalyr_agent/third_party/tcollector/tcollector.py
```diff
@@ -133,7 +139,8 @@ def read(self):
                           self.name, len(out))
                 for line in out.splitlines():
                     LOG.warning('%s: %s', self.name, line)
-        except IOError, (err, msg):
+        except IOError as xxx_todo_changeme:
+            (err, msg) = xxx_todo_changeme.args
```
czerwingithub: Fix name of variable.

scalyr_agent/third_party/tcollector/tcollector.py
```diff
@@ -143,11 +150,14 @@ def read(self):
...
-            self.buffer += self.proc.stdout.read()
+            stdout_data = self.proc.stdout.read()
+            if stdout_data:
+                self.buffer += stdout_data.decode("utf-8")
             if len(self.buffer):
                 LOG.debug('reading %s, buffer now %d bytes',
                           self.name, len(self.buffer))
-        except IOError, (err, msg):
+        except IOError as xxx_todo_changeme1:
+            (err, msg) = xxx_todo_changeme1.args
```
czerwingithub: Fix name of variable.

czerwingithub:
@ArthurKamalov Ok, those were my last comments going over your original submission.  I'll be going back now and looking at your responses.

scalyr_agent/scalyr_logging.py
```diff
@@ -705,16 +723,24 @@ def findCaller(self):
             if filename == _srcfile or filename == logging._srcfile:
                 f = f.f_back
                 continue
-            rv = (filename, f.f_lineno, co.co_name)
+            # 2->TODO just copied this from 'findCaller' from python3 logging library. Should we keep it?
```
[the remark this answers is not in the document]
ArthurKamalov: Nothing wrong I'm sure. Removed.

scalyr_agent/util.py
ArthurKamalov: No... 
Also, I have made changes to other docstrings.

scalyr_agent/json_lib/parser.py
ArthurKamalov: Done. Yes, it means the end of the stream or an empty stream. Invalid character is handled by the last else clause.

scalyr_agent/json_lib/parser.py
ArthurKamalov: Done.

scalyr_agent/json_lib/parser.py
ArthurKamalov: `key` variable can be `None` after` self.__parse_string()'.

https://github.com/scalyr/scalyr-agent-2/blob/c4bfd11c7e44597b66d81d0d5f2727336a135450/scalyr_agent/json_lib/parser.py#L462-L470

but now I can see that an additional check for a quote does not make much sense and can be removed at all. Do you agree?

scalyr_agent/json_lib/parser.py
ArthurKamalov: Done

scalyr_agent/monitor_utils/tests/annotation_config_test.py
```diff
@@ -192,7 +192,7 @@ def test_list_of_lists(self):
 
     def test_mixed_list_and_dict_key(self):
         annotations = {
-            "log.config.scalyr.com/test.2.match_expression": "third",
+            "log.config.scalyr.com/1.2.match_expression": "third",
```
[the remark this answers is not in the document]
ArthurKamalov: Oops, bad clean up. Thanks.

scalyr_agent/platform_posix.py
```diff
@@ -359,7 +367,8 @@ def logger(message):
             sys.stderr.flush()
             si = open(self.__stdin, "r")
             so = open(self.__stdout, "a+")
-            se = open(self.__stderr, "a+", 0)
+            # 2->TODO io.open does not allow buffering disabling on text files.
```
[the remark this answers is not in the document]
ArthurKamalov: To use binary mode here.

scalyr_agent/json_lib/parser.py
ArthurKamalov: Oh, right, it is not. Thanks.

ArthurKamalov: The type is `six.text_type`(unicode) here. I think it was confusing because I did not write docstring correctly.

scalyr_agent/json_lib/parser.py
ArthurKamalov: Yes, absolutely correct. Thanks.
Maybe we should rename the whole class?

scalyr_agent/third_party/tcollector/tcollector.py
ArthurKamalov: done

scalyr_agent/third_party/tcollector/tcollector.py
ArthurKamalov: Done.

scalyr_agent/builtin_monitors/tomcat_monitor.py
```diff
@@ -690,7 +693,7 @@ def get_value_as_str(value):
                 return "%d" % value
             elif type(value) is float:
                 return "%f" % value
-            elif type(value) is str:
+            elif type(value) is six.text_type:
```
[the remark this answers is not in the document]
ArthurKamalov: Removed it. Yes, `repr` for `unicode` returns a different result, it could cause problems, thanks.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -1279,8 +1279,7 @@ def __create_log_config(
         @param path: the path of the log file being configured
         @param attributes: Any attributes to include as part of the log_config['attributes']
         @param base_config: A base set of configuration options to build the log_config from
-        @type parser: str
-        @type path: str
+        @type path: six.text_type
```
czerwingithub: Add back in `@type` for `default_parser`

scalyr_agent/remote_shell.py
```diff
@@ -78,15 +78,15 @@ def read(self, max_bytes):
     def write(self, input_str):
         """Writes the string to the underlying socket.
         @param input_str: The string to send.
-        @type input_str: str
+        @type input_str: six.binary_Type
```
czerwingithub: lowercase the `T`

scalyr_agent/util.py
```diff
@@ -77,8 +77,8 @@ def ujson_dumps_custom(obj, fp):
             :param obj: The object to serialize
             :param fp: If not None, then a file-like object to which the serialized JSON will be written.
             :type obj: dict
-            :return: If fp is not None, then the str representing the serialization.
-            :rtype: str
+            :return: If fp is not None, then the string representing the serialization.
+            :rtype: Python3 - six.text_type, Python2 - six.text_type | six.binary_type
```
czerwingithub: Add documentation as to under what conditions `text_type` or `binary_type` are returned in Python 2.

scalyr_agent/util.py
```diff
@@ -104,8 +104,8 @@ def json_dumps_custom(obj, fp):
...
```
czerwingithub: Same.

scalyr_agent/util.py
```diff
@@ -1362,7 +1373,8 @@ def _write_stream(self, stream_id, content):
         self.__channel_lock.acquire()
         try:
             if self.__channel_lock is not None:
-                self.__channel.write(struct.pack("i", code) + encoded_content)
+                # 2->TODO struct.pack|unpack in python2.6 does not allow unicode format string.
```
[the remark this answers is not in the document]
czerwingithub: Did you have a question about this comment?  I know you said you had a question about one of my comments about `struct`.

scalyr_agent/test_util.py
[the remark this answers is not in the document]
czerwingithub: Looked this over, this is ok.

scalyr_agent/json_lib/parser.py
czerwingithub: Well, it should raise an error instead.  It's probably a good idea to verify preconditions.

scalyr_agent/json_lib/parser.py
czerwingithub: Yes, let's change the class name.

scalyr_agent/util.py
ArthurKamalov: I have to admit that I have been a bit hasty. Only standard library `json.dumps` with the option `ensure_ascii=False` can produce `unicode`, so, unless we do not use such option, it can return only six.binary_type

scalyr_agent/builtin_monitors/docker_monitor.py
ArthurKamalov: Oops, the `parser` was actually `default_parser`, so I had to rename it, instead of removing. Thanks.

scalyr_agent/util.py
ArthurKamalov: Done

scalyr_agent/remote_shell.py
ArthurKamalov: Done.

scalyr_agent/util.py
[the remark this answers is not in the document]
ArthurKamalov: Yes,  and sorry for the delay.
 
If you are asking particularly about `_write_stream` function and `RedirectorServer`, than yes, they are covered.

The previous approach was to not change the source code and just to monkey patch `struct.pack|unpack` with wrapper functions that convert format string to `unicode`(You can see it in `scalyr_agent.compat` before last changes).
But after that, I started to think that is too implicit, and might not work in scripts without additional monkey patching. 

So I decided to go a proven path and to add new compatible functions in scalyr_sgent.compat and made the modernize script replace original struct functions with them(same as with `os.environ`)

scalyr_agent/json_lib/parser.py
ArthurKamalov: Done. 

I just wanted to say, there is already verification for a double quote before every '__parse_string' function call and it just wouldn’t have been called in the opposite case.  But now I agree with you because this additional check can prevent errors in the future.

.circleci/modernize/custom_fixers/fix_struct_unicode.py
```diff
@@ -0,0 +1,27 @@
+from lib2to3.fixer_base import BaseFix
+from lib2to3.pgen2 import token
+from lib2to3.fixer_util import Name, Node, Leaf, LParen, RParen, Dot, Newline
+import libmodernize
+
+
+class FixStructUnicode(BaseFix):
+    """
+        This fixer searches for places where we calling struct.pack or struct.unpack,
```
czerwingithub: `where we calling` should be `where we are calling`

scalyr_agent/util.py
```diff
@@ -78,7 +79,7 @@ def ujson_dumps_custom(obj, fp):
             :param fp: If not None, then a file-like object to which the serialized JSON will be written.
             :type obj: dict
             :return: If fp is not None, then the string representing the serialization.
-            :rtype: Python3 - six.text_type, Python2 - six.text_type | six.binary_type
+            :rtype: Python3 - six.text_type, Python2 - six.binary_type
```
czerwingithub: Seeing the different return types between Python3 and Python2 still throws me, though I understand, based on how this code is used, it's fine.  Let's add a comment after line 77 to something like this:

`Note, this function returns different types (text vs binary) based on which version of Python you are using.  We leave the type unchanged here because the code that invokes this function will convert it to the final desired return type.  Otherwise, we'd be double converting the result in some cases.`

scalyr_agent/util.py
```diff
@@ -105,7 +106,7 @@ def json_dumps_custom(obj, fp):
...
```
czerwingithub: Add similar comment to above at line 104 here.

.circleci/modernize/custom_fixers/fix_struct_unicode.py
ArthurKamalov: Thanks. Done.

scalyr_agent/util.py
ArthurKamalov: Done.

scalyr_agent/util.py
ArthurKamalov: Done.

czerwingithub:
@ArthurKamalov It looks like some of the tests are failing under Python 3.  One looks like an order/unordering problem.. and then I'm not sure about the other.  Please take a look at fixing the tests and then let's get this merged in.

ArthurKamalov:
@czerwingithub
The first bug is related to the `dict` ordering problem in Python < 3.6. Those versions do not preserve the insertion order in `dict`, so we should not expect it in tests.
There are two affected tests:
    * The first test - `scalyr-client-test.test_timing_data`, here we do assertion of the string that was created from `dict`, The solution - to sort `dict` before string creation.
    * The second test - `url_monitor_test.test_post_request_with_data`, assertion of the `list`. The solution to sort values before comparison.

The second bug is related to the `re.sub` bugfix that was made in Python 3.7.  Now it works as it should. https://docs.python.org/3.7/library/re.html#re.sub (see 3.7 change.)
There is one affected test  - `log_processing_test.test_unicode_redactions`. Here we have `(.*)` pattern and fixed re.sub replaces last empty symbol too because of "zero or more" token

PR #369, opened 2020-01-22
M2: manual changes applied to json_lib.

ArthurKamalov:


scalyr_agent/json_lib/tests/serializer_test.py
```diff
@@ -14,11 +14,12 @@
 # ------------------------------------------------------------------------
 #
 # author:  Steven Czerwinski <czerwin@scalyr.com>
+from __future__ import unicode_literals
 from __future__ import absolute_import
 
 __author__ = "czerwin@scalyr.com"
 
-from cStringIO import StringIO
+from io import BytesIO
```
Kami: Since a lot of those Python 2 and 3 changes are related to types (bytes vs unicode), I think it would be good to add some more unit tests for the JSON serialization abstraction to ensure it works correctly with all the supported inputs (bytes and unicode and either works with it natively or performs unicode <-> bytes conversion where and if needed).

ArthurKamalov: `json_lib` is not used as a full-fledged JSON library anymore, It has been kept only for special cases. For example, parsing is only needed to read Scalyr config file(It has some differences from standard JSON) and serializing is not needed at all, except length prefixed string serialization which is intended for reading by Scalyr servers. In other cases, regular JSON libraries(`json`, `ujson`) are used.


As Python 3 requires to strictly distinguish bytes and unicode, the idea was to treat strings as much as possible as unicode and keep them as bytes only when needed(for example file reading, networking). To be maximally close to the behavior when running on Python 3, for Python 2 we should distinguish bytes and unicode too and minimize implicit conversion as possible. I mean that there are not many places where we should be able to handle unicode and bytes both, so it is better to strictly specify for each function what types it accepts and returns.

Does it make sense for you?:)

Kami: Thanks for clarifying.

This means sense and I agree with - we should always be very explicit and deliberate with types with use.

I just had some bad experience in the past with code which could accept multiple types, but this was in the context of user / developer supplied data.

In this case, where we control all the code and types, we should probably be fine since we can be explicit and deliberate about it.

PR #372, opened 2020-01-22
AGENT-323: Missing `stream` event attribute in Docker and K8s

yanscalyr:
The result of a missed case (log line specific attributes) during the changes to the addEvents format. This PR changes `add_missing_attributes` to act differently depending on if the attributes being added are log line specific or general to the log file overall.

### DIFF
--- a/scalyr_agent/log_processing.py
+++ b/scalyr_agent/log_processing.py
@@ -2162,7 +2162,7 @@ class LogFileProcessor(object):
         """
         result = Event(base=self.__base_event)
         if line_object.attrs:
-            result.add_missing_attributes(line_object.attrs)
+            result.add_missing_attributes(line_object.attrs, log_line_attributes=True)
         result.set_message(line_object.line)
         if sampling_rate != 1.0:
             result.set_sampling_rate(sampling_rate)
--- a/scalyr_agent/log_processing.py
+++ b/scalyr_agent/log_processing.py
@@ -1652,7 +1652,7 @@ class LogFileProcessor(object):
         """ Adds items attributes to the base_event's attributes if the base_event doesn't
         already have those attributes set
         """
-        self.__base_event.add_missing_attributes(attributes)
+        self.__base_event.add_attributes(attributes)
 
     def set_max_log_offset_size(self, max_log_offset_size):
         """Sets the max_log_offset_size.
@@ -2162,7 +2162,7 @@ class LogFileProcessor(object):
         """
         result = Event(base=self.__base_event)
         if line_object.attrs:
-            result.add_missing_attributes(line_object.attrs, log_line_attributes=True)
+            result.add_attributes(line_object.attrs, overwrite_existing=True)
         result.set_message(line_object.line)
         if sampling_rate != 1.0:
             result.set_sampling_rate(sampling_rate)
--- a/scalyr_agent/scalyr_client.py
+++ b/scalyr_agent/scalyr_client.py
@@ -1469,15 +1469,17 @@ class Event(object):
         # We only stash a copy of attrs for debugging/testing purposes.  We really will just serialize it into
         # __serialization_base.
         self.__log_id = None
-        self.__attrs = attrs
+        if attrs is not None:
+            self.__attrs = dict(attrs)
+        else:
+            self.__attrs = dict()
         self.__disable_logfile_addevents_format = disable_logfile_addevents_format
         if (attrs is not None or thread_id is not None) and base is not None:
             raise Exception("Cannot use both attrs/thread_id and base")
 
-        if self.__attrs is None:
-            self.__attrs = dict()
-
         self.__thread_id = None
+
+        # Used to get attributes from the parent (such as the logfile attributes) for this event.
         self.__parent_event = None
         if base is not None:
             # We are creating an event that is a copy of an existing one.  Re-use the serialization base to capture
@@ -1489,7 +1491,7 @@ class Event(object):
                 base.__disable_logfile_addevents_format
             )
         else:
-            self.__set_attributes(thread_id, attrs)
+            self.__set_attributes(thread_id, self.__attrs)
 
         # The typical per-event fields.  Note, all of the fields below are stored as strings, in the serialized
         # forms for their event fields EXCEPT message.  For example, since ``sequence_id`` should be a string on the
@@ -1508,10 +1510,6 @@ class Event(object):
 
     def __set_attributes(self, thread_id, attributes):
         """ Set the attributes and thread id of an Event.
-
-        In the default case of `log_line_attributes` being False this will not add the attributes to the serialization
-        base because those attributes should be included in the attributes of the log. If `log_line_attributes` is True
-        they will be serialized but not set to `__attrs`, so any events based off this one will not inherit them.
         """
         self.__thread_id = thread_id
         self.__attrs = attributes
@@ -1543,22 +1541,33 @@ class Event(object):
         self.__serialization_base = tmp_buffer.getvalue()
 
     def __get_attributes_to_serialize(self):
+        """Return the attributes that should be included in the serialization for this specific event.
+        This does not include attributes from its parent if using the logfile_addevents_format, and does not include
+        attributes from this event that already exist in the parent with the same value.
+        """
+        result = dict()
         if self.__disable_logfile_addevents_format:
-            result = dict()
             if self.__parent_event:
                 result = dict(self.__parent_event.__attrs)
             if self.__attrs:
                 result.update(self.__attrs)
-            return result
         else:
-            return self.__attrs
+            if self.__parent_event:
+                for key in self.__attrs:
+                    if (
+                        key not in self.__parent_event.__attrs
+                        or self.__parent_event.__attrs[key] != self.__attrs[key]
+                    ):
+                        result[key] = self.__attrs[key]
+            else:
+                result = self.__attrs
+        return result
 
     def add_attributes(self, attributes, overwrite_existing=False):
         """ Adds items attributes to __attrs if the __parent_event doesn't
         already have those attributes set.
 
-        If overwrite_existing is True an attribute will be added even if the __parent_event has the same key defined,
-        unless the value also matched to avoid wasting bytes.
+        If overwrite_existing is False an attribute will not be added if the key already exists in __attrs.
         """
         if attributes:
             attributes = dict(attributes)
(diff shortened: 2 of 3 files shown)


scalyr_agent/scalyr_client.py
```diff
@@ -1502,9 +1502,10 @@ def __init__(
         self.__has_non_optimal_fields = False
         self.__num_optimal_fields = 0
 
-    def __set_attributes(self, thread_id, attributes):
+    def __set_attributes(self, thread_id, attributes, log_line_attributes=False):
```
czerwingithub: Add comment for method especially documenting `log_line_attributes`

scalyr_agent/scalyr_client.py
```diff
@@ -1531,22 +1534,28 @@ def __set_attributes(self, thread_id, attributes):
 
         self.__serialization_base = tmp_buffer.getvalue()
 
-    def add_missing_attributes(self, attributes):
+    def add_missing_attributes(self, attributes, log_line_attributes=False):
```
czerwingithub: Documentation for method, especially `log_line_attributes`.

scalyr_agent/log_processing.py
```diff
@@ -2162,7 +2162,7 @@ def __create_events_object(self, line_object, sampling_rate):
         """
         result = Event(base=self.__base_event)
         if line_object.attrs:
-            result.add_missing_attributes(line_object.attrs)
+            result.add_missing_attributes(line_object.attrs, log_line_attributes=True)
```
czerwingithub: Hmm.. I don't know why we only set the attributes if the event doesn't already have a value for the attributes.  If the base event has a value for `stream`, we want the attribute for this event to override it.  Meaning, if the log attributes have a value of `stream=foo` and the line has an attribute of `stream=bar`, we want `bar` to win.  if you look at the implementation of `add_missing_attribute`, we'll only update the attributes if `stream` is not present.

scalyr_agent/scalyr_client.py
```diff
@@ -1531,22 +1534,28 @@ def __set_attributes(self, thread_id, attributes):
...
         """
         if self.__attrs:
             changed = False
-            new_attrs = dict(self.__attrs)
+            new_attrs = dict()
+            if not log_line_attributes:
+                new_attrs = dict(self.__attrs)
             for key, value in six.iteritems(attributes):
-                if not key in new_attrs:
+                if not key in self.__attrs:
```
czerwingithub: As mentioned above, I think we need to also update `self.__attrs` if the value in the `self.__attrs` is not the same as what is in `attributes`.  Maybe we can make that behavior contingent on `log_line_attributes=True` in case this break existing behavior.  though, if we do that, we should have a TODO to investigate if that is really necessary.

scalyr_agent/tests/scalyr_client_test.py
```diff
@@ -161,6 +161,24 @@ def test_set_position_with_log_and_thread(self):
 
         request.close()
 
+    def test_set_log_line_attributes(self):
```
czerwingithub: Also have a test case where the base event has a different an attribute named `source`.

scalyr_agent/scalyr_client.py
yanscalyr: `self.__attrs` is set to `attributes` in the followup `__set_attributes` call, I already put in a case so that in the new format for log line attributes we don't set it but the way we currently use it it shouldn't matter if we do or don't. That field is only used when inheriting things from a base event or checking which attributes are missing during `add_missing_attributes`, neither of which we call more than once on any event as far as I can tell.

scalyr_agent/log_processing.py
yanscalyr: For cases where the log line attributes how does the server interpret both the log having the attribute and the event? Which wins out, and does the server even read event attributes if it has log attributes?

If log ones win out anyway we may not have a good way to do things on our end to make the event ones win out.

yanscalyr: For now I hid overwriting behind `log_line_attributes=True` in case somewhere in the depths the code relies on it not being overwritten, though I couldn't even begin to guess where or why.

czerwingithub:
As discussed, let's take a different approach.  There's a few readability drawbacks.  Here's the list of changes as I see it:

1.  Add a new instance variable called `_parent_event` to `Event`.  Do not copy its attributes at the time of the event creation.. we'll just use `_parent_event._attrs` when we go to serialize the event.
2. Rename `add_missing_attributes` to `add_attributes(overwrite_existing=False)`.  That should only modify the local `_attrs` (not the ones in `_parent_event`).
3.  Create a new private method with some name like `_get_attributes_to_serialize` where it returns the attributes to serialize as part of the event.  If we are using the new format, it will not include the parent's attributes.  If it is not using the new format, it combines `_attrs` and `_parent_event.attrs`.  We have to do some logic to make sure the lower event's `_attrs` values always win for duplicate keys.

Let me know if you have any questions.

scalyr_agent/scalyr_client.py
```diff
@@ -1474,13 +1474,17 @@ def __init__(
         if (attrs is not None or thread_id is not None) and base is not None:
             raise Exception("Cannot use both attrs/thread_id and base")
 
+        if self.__attrs is None:
```
czerwingithub: At line 1472, let's do a cast to `dict` here because I think we might get `JsonObjects`.

scalyr_agent/scalyr_client.py
```diff
@@ -1474,13 +1474,17 @@ def __init__(
...
+            self.__attrs = dict()
+
         self.__thread_id = None
+        self.__parent_event = None
```
czerwingithub: Add comment: `Used to get attributes from the parent (such as the logfile attributes) for this event.`

scalyr_agent/scalyr_client.py
```diff
@@ -1502,16 +1506,16 @@ def __init__(
...
         """ Set the attributes and thread id of an Event.
 
         In the default case of `log_line_attributes` being False this will not add the attributes to the serialization
         base because those attributes should be included in the attributes of the log. If `log_line_attributes` is True
         they will be serialized but not set to `__attrs`, so any events based off this one will not inherit them.
         """
         self.__thread_id = thread_id
-        if self.__disable_logfile_addevents_format or not log_line_attributes:
-            self.__attrs = attributes
+        self.__attrs = attributes
```
czerwingithub: Relook at the comments for this method.. I think those are implementation details not needed to be covered here.

scalyr_agent/scalyr_client.py
```diff
@@ -1540,40 +1542,54 @@ def __set_attributes(self, thread_id, attributes, log_line_attributes=False):
 
         self.__serialization_base = tmp_buffer.getvalue()
 
-    def add_missing_attributes(self, attributes, log_line_attributes=False):
-        """ Adds items attributes to the base_event's attributes if the base_event doesn't
+    def __get_attributes_to_serialize(self):
```
czerwingithub: Add comment for this method.. something like `Return the attributes that should be included in the serialization for this specific event.  This does not include attributes from its parent if using the logfile_addevents_format.`

scalyr_agent/scalyr_client.py
```diff
@@ -1540,40 +1542,54 @@ def __set_attributes(self, thread_id, attributes, log_line_attributes=False):
...
             if changed:
-                self.__set_attributes(
-                    self.__thread_id, new_attrs, log_line_attributes=log_line_attributes
-                )
+                self.__set_attributes(self.__thread_id, self.__attrs)
         else:
-            self.__set_attributes(
-                self.__thread_id, attributes, log_line_attributes=log_line_attributes
-            )
+            self.__set_attributes(self.__thread_id, attributes)
```
czerwingithub: In this case, we should be adding the new attributes to the ones existing in `__attrs`, but you are just overwriting `__attrs` with `attributes`.

scalyr_agent/scalyr_client.py
```diff
@@ -1540,40 +1542,54 @@ def __set_attributes(self, thread_id, attributes, log_line_attributes=False):
...
+            attributes = dict(attributes)
+
+        if self.__parent_event:
             changed = False
-            new_attrs = dict()
-            if not log_line_attributes:
-                new_attrs = dict(self.__attrs)
             for key, value in six.iteritems(attributes):
-                if log_line_attributes or not key in self.__attrs:
+                if key not in self.__parent_event.__attrs or (
```
czerwingithub: I don't think you should be looking at `__parent` here at all.    You should only be adding them based on the contents of `__attrs`.  So, see if `__attrs` has the key or not.. if it does, only add to `__attrs` if the values are different.

The part where you try not to include attributes that are duplicated in `__parent` should be handled in `__get_attributes_to_serialize`.

scalyr_agent/scalyr_client.py
```diff
@@ -1540,40 +1542,54 @@ def __set_attributes(self, thread_id, attributes, log_line_attributes=False):
...
+    def __get_attributes_to_serialize(self):
+        if self.__disable_logfile_addevents_format:
+            result = dict()
+            if self.__parent_event:
+                result = dict(self.__parent_event.__attrs)
+            if self.__attrs:
+                result.update(self.__attrs)
+            return result
+        else:
+            return self.__attrs
```
czerwingithub: Here you should only return the attributes that are not duplicates from `__parent_event.__attrs`.  So, you'll want to iterate over `__attrs`, checking contents of `__parent_event.__attrs`.

scalyr_agent/scalyr_client.py
```diff
@@ -1543,39 +1541,49 @@ def __set_attributes(self, thread_id, attributes):
...
         already have those attributes set.
 
-        If overwrite_existing is True an attribute will be added even if the __parent_event has the same key defined,
-        unless the value also matched to avoid wasting bytes.
+        If overwrite_existing is False an attribute will not be added if the key already exists in __attrs.
         """
         if attributes:
             attributes = dict(attributes)
 
         if self.__parent_event:
```
czerwingithub: Remove this.. no need to check `__parent_event`.  This method should operate the same way either way.

scalyr_agent/scalyr_client.py
```diff
@@ -1543,39 +1541,49 @@ def __set_attributes(self, thread_id, attributes):
...
-                ):
+                if key not in self.__attrs or overwrite_existing:
                     changed = True
                     self.__attrs[key] = value
 
             if changed:
                 self.__set_attributes(self.__thread_id, self.__attrs)
         else:
-            self.__set_attributes(self.__thread_id, attributes)
+            self.__attrs.update(attributes)
```
czerwingithub: Remove this `else` case, you don't need it.

PR #377, opened 2020-01-28
Use py.test test runner in combination with tox

Kami:
This pull request updates the repository to use ``pytest`` (https://docs.pytest.org/en/latest/) Python test runner in combination with ``tox``  (https://tox.readthedocs.io/en/latest/) for running tests under multiple Python versions locally and on CI.

## Background, Context

Right now the code uses standard Python test runner and discovery.

This works fine for simple setups, but using tox and py.test has multiple advantages:

1. ``tox`` allows user to run a specific target (e.g. tests, lint checks, etc.) in an isolated (virtual environment) fashion using a single command (e.g. ``tox -epy2.7-unit-tests``). It takes care of creating the virtual environment with all the dependencies, etc.
2. ``pytest`` test runner is much more powerful than standard Python test runner and offers multiple advantages:
   - More user-friendly output which makes it easier to spot failures with related tracebacks, log messages, etc.
   - More powerful assertions by default (e.g. fail if some test produced ERROR level log output, etc.)
   - Ability to use various plugins such as the one for printing top slowest tests, benchmark one (which allows us to ensure some unit test doesn't regress in timing / gets slower over time, etc.).
    - Ability to run tests in parallel

If others agree that this change is a good idea, I can continue with this PR and finish it based on the TODO items described below.

On a related note, it's probably best to wait on other Python 3 related PRs to get finished and merged first to avoid merge conflicts (that's also one of the reason why I didn't touch existing Circle CI jobs in this PR - to avoid merge conflicts).

# Example Output

Screenshot from 2020-01-28 11-14-59

## TODO

- [x] Decide if we want this change or not and proceed based on that
- [x] Decide where coverage, modernize and other target "implementation details" should live (e.g. tox.ini vs circle ci config). I think putting that logic inside tox.ini is better and then we can just call that tox target locally and on CI
- [x] Finish it
  - [x] Add any other missing targets (e.g. lint, modernize)
  - [x] Update existing jobs in circle config
  - [x] Document how to run the tests locally
  - [x] Make sure CI passes
  - [x] Verify  that py test collection / discovery output matches the output ``find_all_tests()``
  - [x] Submit coverage data to codecov.io

## TODO (future PRs)

- Agree on the line length limit and enforce it as part of black and lint checks
- Update other unittest targets to utilize tox on Circle CI
- Add pylint check


Kami:
@czerwingithub Correct, we can rely on two levels of CircleCI directory level caching:

1. Standard Circle CI directory cache for Python dependencies (wheels)
2. Caching the whole ``.tox/`` directory with the pre-built virtual environments

With 1), tox virtual environment still needs to be re-created, but that's relatively fast since dependencies will already be cached.

2) has some edge cases, but there are ways around it with tox extensions (https://venv-update.readthedocs.io/en/master/).

czerwingithub:
@Kami One other question.  In the documentation for `pytest`, they mention you case use the regular old `assert` call instead of all of the `assert*` calls due to their use of introspection.  Just so I can understand expectations, do you think we should make an effort to write tests using `assert` going forward?

Kami:
Yeah, using regular ``self.assertXYZ`` methods from the base unit test class still works.

Here is an example output when a test fails because assertion failure:

```bash
scalyr_agent/tests/scalyr_logging_test.py::ScalyrLoggingTest::test_rate_limit_no_write_rate FAILED                                                                                    [ 86%]
scalyr_agent/tests/scalyr_logging_test.py::ScalyrLoggingTest::test_rate_limit_throttle_write_rate PASSED                                                                              [ 93%]
scalyr_agent/tests/scalyr_logging_test.py::ScalyrLoggingTest::test_sibling_modules PASSED                                                                                             [100%]

========================================================================================= FAILURES ==========================================================================================
_______________________________________________________________________ ScalyrLoggingTest.test_limit_once_per_x_secs ________________________________________________________________________

self = <scalyr_agent.tests.scalyr_logging_test.ScalyrLoggingTest testMethod=test_limit_once_per_x_secs>

    def test_limit_once_per_x_secs(self):
        log = scalyr_logging.getLogger("scalyr_agent.foo")
        log.info(
            "First record",
            limit_once_per_x_secs=60.0,
            limit_key="foo",
            current_time=0.0,
        )
        log.info(
            "Second record",
            limit_once_per_x_secs=60.0,
            limit_key="foo",
            current_time=1.0,
        )
        log.info(
            "Third record",
            limit_once_per_x_secs=60.0,
            limit_key="foo",
            current_time=61.0,
        )
    
        self.assertTrue(self.__log_contains("First record"))
        self.assertFalse(self.__log_contains("Second record"))
        self.assertTrue(self.__log_contains("Third record"))
    
        # Now test with different keys.
        log.info(
            "First record",
            limit_once_per_x_secs=30.0,
            limit_key="foo",
            current_time=0.0,
        )
        log.info(
            "Second record",
            limit_once_per_x_secs=60.0,
            limit_key="bar",
            current_time=1.0,
        )
        log.info(
            "Third record",
            limit_once_per_x_secs=30.0,
            limit_key="foo",
            current_time=31.0,
        )
        log.info(
            "Fourth record",
            limit_once_per_x_secs=60.0,
            limit_key="bar",
            current_time=31.0,
        )
    
>       self.assertTrue(self.__log_contains("First recordx"))
E       AssertionError: False is not true

scalyr_agent/tests/scalyr_logging_test.py:324: AssertionError
```

Kami:
I noticed @ArthurKamalov made some changes to run all the unit tests inside a Docker container which is built as part of the Circle CI job using ``pyenv``.

@ArthurKamalov What do you think about using ``tox`` instead of ``pyenv`` inside ``.circleci/docker_unified_smoke_unit/unittest/unittest.sh``?

The idea is that the developer can just run a command such as ``tox -epy2.7-unit-tests`` locally (either inside or outside a docker contain, depending on their preferences and development environment setup).

That's also somewhat related to what @czerwingithub mentioned yesterday on docker layer caching and pricing.

I'm personally fine with an overall Docker container approach and I think it's a good idea for "reproducability and portability" reasons, but if I'm understanding it correctly, we now utilize docker layer caching for all the ``unittest_python3_executor`` jobs.

I assume this will increase our costs (IIRC, previously, we only used it for a couple of Circle CI job containers).

Probably the best approach is to wait a couple of weeks and gather some hard data on our costs and usage so we don't speculate too much. Perhaps the "pricing overhead" won't be that bad.

ArthurKamalov:
@Kami, yes, I totally agree with you. It was a quick but very heavy solution to use docker for new python3 versions, and I want to change it so mush...  I have never used tox, but I really want to start, and I will be glad to help with it :)

Kami:
@ArthurKamalov Great.

I will try to finish this so we can merge the first iteration which includes the following functionality:

- We run unit tests under python 2.7 using py.test test runner
- new lint target which runs black, flake8 and modernize

All of that will be powered by tox.

And once that's merged, we can both start updating existing (currently Docker based) unit test Circle CI jobs to follow the same pattern.

Kami:
I also added new ``lint`` target to tox in this PR which includes a new ``flake8`` check.

flake8 primarily detects styling issues, but it also detected some actual issues (e.g. missing imports).

All the changes should be purely styling, so it shouldn't affect the functionality in any case, but it's still a big change so we should keep an eye on any "weird" issues after the merge.

The plan also is to add pylint and mypy lint checks in the future, but those changes are more involved so it's better to do it separately to avoid this PR from growing any larger.

Kami:
@ArthurKamalov While working on this PR, I noticed a couple of small Python 3 related issues - bff6aa510cad8cd7a405af5dbc93b8c5eabcb062.

Would appreciate if you can confirm those changes are in-line with what you did in other PRs.

Kami:
As part of this PR, I also hooked up MyPy check.

This means I also needed to fix various issues which have been detected by MyPy and add some type annotations -  eb03d92ad1ea3ef170ee52d3f7810febadd64f5a.

Right now, we don't have any type annotations for the existing code yet, which means those issues are only related to lack of type information which couldn't be automatically inferred by MyPy.

Having said that, the check did already detect a couple of minor issues, but it will be a lot more helpful once we start annotating the existing and new code.

Going forward, the plan should be to add type annotations for any new code or the existing code which is being modified.

As far as ``typing`` import goes - I used ``if False`` workaround. This means it doesn't affect existing code in any way since that line is only executed when running under MyPy and it means it also works correctly on our CI where tests run under Python 3.6 where that module is available.

There are some other solutions possible for that, like adding ``typing`` install_dependencies dependency for Python versions where it's not available (2.7, < 3.4), but this doesn't really bring any value and just introduces additional complexity (and on top of that, right now this program vendors / bundles all teh dependencies and doesn't utilize setup.py requires statements).

---

On that note, this PR should now also be more or least finished (at least for the first iteration) and ready for a review.

Kami:
I also added pylint check in - 2fba33003d91ae5ad091cc8cdc5ca823cb562818.

It already identified some real issues (not passing enough arguments to a function, invalid argument type, etc.), but there were also a lot of false positives and places which would require bigger changes / refactoring to fix the violations. In those places I used "ignore" pragmas.

Now that this check is in place, it will make it much easier and faster to detect such issues early on during the development (and perhaps also makes us re-think some decisions which would result in violations).

Retrofing such checks later on is always much harder (and in a lot of those places we currently don't have much choice but adding ignore pragma since "real fix" would require bigger changes and introduce bigger risk, especially in places where test coverage is not that great).

----

@ArthurKamalov There were a couple of more small Python 3 related issues which have been detected by those new lint check which I addressed.

One of them was related to ``cStringIO`` which I replaced with ``six.moves.StringIO`` and the other one was related to ``message`` attribute on the ``Exception`` class instances which is not available in Python 3 anymore.

In that scenario, I used the following pattern which I used in the past in other projects during Python 3 migration:

```python
message = getattr(error, 'message', str(e))
```

Kami:
I missed a couple of more spots - 61b29a8527a71a8dd76a49c9b973a064d6f82776, 59530a6cdd23249a04bdbdba637acaef1a416ba7. Now lint checks also run on Python files inside the repo root.

@ArthurKamalov @czerwingithub if you encounter any issues / wrong behavior during package builds after my PR has been merged, please let me know.

Adding r"" to those regex strings should be correct (59530a6cdd23249a04bdbdba637acaef1a416ba7), otherwise it will try to handle \ as escape sequence (and using r is also preferred over escaping \ since escaping gets more messy), but it's still possible I missed a spot (aka r should only be used in places where \ are not escaped already).

To give a concrete example, either of those is fine and has the same end result:

```python
foo = r"\.idea"
```

```python
foo = "\\.idea"
```
But the first approach (raw strings) is preferred since it results in less mess with ``\`` characters.

---

> Regular expressions use the backslash character ('\') to indicate special forms or to allow special characters to be used without invoking their special meaning. This collides with Python’s usage of the same character for the same purpose in string literals; for example, to match a literal backslash, one might have to write '\\\\' as the pattern string, because the regular expression must be \\, and each backslash must be expressed as \\ inside a regular Python string literal. Also, please note that any invalid escape sequences in Python’s usage of the backslash in string literals now generate a DeprecationWarning and in the future this will become a SyntaxError. This behaviour will happen even if it is a valid escape sequence for a regular expression.
> 
> The solution is to use Python’s raw string notation for regular expression patterns; backslashes are not handled in any special way in a string literal prefixed with 'r'. So r"\n" is a two-character string containing '\' and 'n', while "\n" is a one-character string containing a newline. Usually patterns will be expressed in Python code using this raw string notation.

https://docs.python.org/3/library/re.html#module-re

Kami:
On a related note, I also needed to push some changes to make ``build_package.py`` mostly work under Python 2 and 3 to get new lint checks to pass.

It will likely need more work (and testing) in the future, to verify it works correctly under Python 2 and Python 3.

I assume that's also part of our longer term plan (automated tests for the build package script and the generated files).

@ArthurKamalov ^ just a heads up

Kami:
(as per my Slack comment)

This pull request also improved / fixed code coverage reporting.

Previously we didn't include modules which didn't have any tests and weren't imported by any code in the overall coverage report (those are mostly monitors).

Now we also include those files in the report which means that the total coverage number decreased.

https://codecov.io/gh/scalyr/scalyr-agent-2/tree/fix_sigint_handler_on_no_fork/scalyr_agent/builtin_monitors vs https://codecov.io/gh/scalyr/scalyr-agent-2/tree/use_pytest_test_runner/scalyr_agent/builtin_monitors

PR #378, opened 2020-01-28
Submit code coverage reports to codecov

Kami:
Small change which updates existing ``coverage`` Circle CI job to submit combined / aggregated coverage data to codecov.io service.

This means that in addition to the raw plain text code coverage output which is currently displayed in the Circle CI job output we now also get nicer visualization such as this one - https://codecov.io/gh/scalyr/scalyr-agent-2/branch/submit_code_coverage_to_codecov.

In addition to the nice web based visualization, we also get two other things:

1. Pull request comments with information about coverage for a specific PR
2. Ability to define a coverage threshold (either absolute or relative) and fail a PR via Github status check if that threshold is not reached

### DIFF
--- a/.circleci/.coveragerc_ci
+++ b/.circleci/.coveragerc_ci
@@ -2,10 +2,10 @@
 source =
     /home/circleci/scalyr-agent-2/
     /usr/share/scalyr-agent-2/py/
-    
+
 [report]
 skip_covered=True
 omit =
+    */.virtualenvs/*
     */venv/*
     */third_party*/*
-    */scalyr-agent/tests/*
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -300,7 +300,7 @@ jobs:
           command: |
             virtualenv venv
             source venv/bin/activate
-            pip install coverage
+            pip install coverage "codecov==2.0.15"
             mkdir -p cov
             cp .circleci/.coveragerc_ci cov/.coveragerc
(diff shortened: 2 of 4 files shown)


Kami:
I believe the path remapping and PR comments should be enabled once codecov.yml file is added / merged into master branch.

Kami:
All right, this is now all hooked up and ready for a review.

The PR comments are now also working, but they are not much use until this is merged into master and the coverage can be compared against the base coverage from master.

Kami:
I pushed a small change to also include coverage for the actual test files.

This affect whole project coverage number, but it's useful since it allows us to make sure that all the tests are indeed correctly covered and executed.

If other people think this adds some confusion, coverage data for actual tests can be removed (codecov.io also allows you to partition / organize coverage reports per arbitrary flags (https://docs.codecov.io/docs/flags) so that's another option to organize the coverage reports into).

Kami:
I noticed the pull request comments don't appear to be working.

I believe this may be related to the current merging strategy (squash / rebase instead of merge commits).

I believe this confuses codecov since it means it can't find the base commit for a particular PR change once other PRs are merged into master because those merges change and rewrite git history.

Codecov docs (https://docs.codecov.io/docs/comparing-commits) claim rebase approach is indeed supported, but I need to verify that.

It may well be that rebase approach is supported, but not the rebase + squash.

PR #379, opened 2020-01-28
Fix intermediately failing logging rate limit test

Kami:
While working on #377, I noticed ``ScalyrLoggingTest.test_rate_limit_no_write_rate`` tests fails very often.

After digging in, I believe I found an issue in the test itself and a potential fix / workaround.

## Context, Background

Right now we set ``max_write_burst`` aka ``bucket_size`` to 250 bytes.

The problem is that the actual message string value we use when determining the message size is the fully formatted string and that string is much longer than the original string we pass to the logger method in the tests. 

Formatted string value also contains module name, timestamp and for the first successfully logged message after the rate limit being hit, it's also prefixed with a warning that a previous message has been rate limited.

For example:

```bash
.... Warning, skipped writing 1 log lines due to limit set by `monitor_log_write_rate` option...
2020-01-28 14:43:34.401Z INFO [core] [scalyr_logging_test.py:269] Second message
```

This means that depending on the test timing and bucket fill rate the actual length of the first and last formatted message which should be written successfully may exceed this 250 bytes limit we originally used and the last message ("second message") will be rate limited instead of being successfully logged and the test will fail.

## Proposed Solution

There are a couple of possible solutions.

I went with the simplest one which simply increases ``max_write_burst`` value so it will always be greater than ``len(formatted first message) + len(formatted last message with rate limited previous message warning)``.

---

On some somewhat related note - part of a problem also is that we prefix first successfully logged message after rate limit being hit with a warning that we skipped writing x log lines and this affects the actual message length we use when determining if there is still space left in the bucket.

A perhaps better approach would be to not take that warning string into account when determining message size for rate limit purposes (but that's out of scope here and to be discussed).

### DIFF
--- a/scalyr_agent/tests/scalyr_logging_test.py
+++ b/scalyr_agent/tests/scalyr_logging_test.py
@@ -38,6 +38,7 @@ class ScalyrLoggingTest(ScalyrTestCase):
             agent_log_file_path=self.__log_path,
         )
         self.__logger = scalyr_logging.getLogger("scalyr_agent.agent_main")
+        self.__logger.set_keep_last_record(False)
 
     def test_output_to_file(self):
         self.__logger.info("Hello world")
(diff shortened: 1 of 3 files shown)


scalyr_agent/tests/scalyr_logging_test.py
```diff
@@ -240,14 +240,20 @@ def test_rate_limit_no_write_rate(self):
             use_disk=True,
             logs_directory=os.path.dirname(self.__log_path),
             agent_log_file_path=self.__log_path,
-            max_write_burst=250,
+            max_write_burst=500,
```
Kami: NOTE: Using this static value here still isn't the best / most robust approach since test may start to fail if we change the default format string and / or the warning to be longer.

An approach to make this more robust and future proof is to instantiate two ``LogRecord`` objects with the corresponding messages, obtain the length of the fully formatted message (for the second log message, this includes the warning) and then make sure that ``max_write_burst = (len(message_one_formatted) + len(message_two_formatted_with_warning)) + 1`` or something along those lines.

Kami:
The approach I used to test this change is it so to run ``scalyr_agent/tests/scalyr_logging_test.py:ScalyrLoggingTest`` and ``scalyr_agent/tests/scalyr_logging_test.py:ScalyrLoggingTest.test_rate_limit_no_write_rate`` test in a loop.

From my ``~/.bashrc``:

```bash
function multif {
    n=$1
    shift
    for ((i=0;i<n;i++)) do
        $@
        if [ $? -ne 0 ]; then
            echo " == Command failed == "
            break
        fi
    done
}
```

And then:

```bash
multif 5000 nosetests -s -v --nocapture scalyr_agent/tests/scalyr_logging_test.py:ScalyrLoggingTest
```

Keep in mind though, that even without this change, running the test in isolation (by itself) was not that likely to fail compared to running it in combination with other tests. I assume the reason for that is timing.

scalyr_agent/tests/scalyr_logging_test.py
czerwingithub: I'll throw out a few alternatives as well.  We could add a parameter in `AgentLogManager.set_log_destination` allowing us to disable using a formatter.  The charged bytes would be equal to just the length of the original string, making it easier to reason about the tests.  This would require a little bit of plumbing and adds in a mode that we would never really use in production.. so not perfect.

Another alternative:  Record the len of the string in the LogRecord (regardless of whether or not it keeps filtered out) and use the `last_record` to get the last record and do a check (before we do a second that the log contains `Second message`) that the length of the string is less than what the static value was.  Add a comment there to say if the length has changed, we probably need to up the static `max_write_burst`.

I wasn't clear from your approach how you would instantiate the `LogRecord` objects such that we are sure they are using the same formatter as what is used by the filter.

scalyr_agent/tests/scalyr_logging_test.py
```diff
@@ -240,14 +240,20 @@ def test_rate_limit_no_write_rate(self):
...
         self.__logger.set_keep_last_record(True)
         try:
-            string_300 = "a" * 300
+            # NOTE: Actual value which is being used for the rate limitting is the formatted value
+            # and that value contains much more information than the string we generate here.
+            # This means that 450 + common formatted string data will aways be > 500 and we need to
+            # make sure that "max_write_burst" value we use is large enough so formatted "First
+            # message" and "Second message" (with a warning) fit in that value, otherwise depending
+            # on the timing and fill rate, the test may fail.
+            string_300 = "a" * 450
```
czerwingithub: Let's change the name to `string_450`.

scalyr_agent/tests/scalyr_logging_test.py
Kami: I agree about the first approach - it's similar to mocking.

You need to be careful and find a good balance, because if you end up mocking too much, you may just end up testing mocks itself and not the code which is supposed to be tested.

Because of that I went with the variation of the second approach you suggested - 58ba694d1431b37830466261cf4bb2e42db77037.

I think it's reasonable enough and will make sure tests now fail fast, consistently and loudly in case a format string changes, test contains invalid value for ``max_write_burst`` or similar which is much better than intermediate failures which are hard to track down.

Storing those two values on the original ``LogRecord`` object does mean the object itself will be slightly larger, but I don't think that's a big deal since those values are very small compared to the actual log messages :)

And If I added a check to do this only during the debug / test mode, this would introduce additional code complexity for a little good reason (if that ever becomes an issue, which I doubt, we can change / improve that in the future).

scalyr_agent/tests/scalyr_logging_test.py
Kami: Fixed in 58ba694d1431b37830466261cf4bb2e42db77037.

PR #380, opened 2020-01-28
Log raw uncompressed request body in the Scalyr HTTP client debug log

Kami:
## Problem Description

Right now the Scalyr API HTTP client logs compressed request body in the debug log if compression is available / enabled (which should be true in most scenarios).

This makes things harder to troubleshoot / debug since it means you need to manually decompress that data or disable compression before you can introspect it.

## Proposed Solution

I propose updating the code to log raw uncompressed string payload instead.

The purpose of the debug log is to making troubleshooting / debugging easier, so I think it makes sense to log raw uncompressed request body string by default.

I don't think it should have any negative impact on security or similar.

First of all, this is only logged under DEBUG log level where users can expect some credentials will be logged (e.g. API token). In addition to that, this information is already available in the debug log today, the only difference is that it's in a compressed format.

Before:

```bash
2020-01-28 16:41:24.703Z Level 4 [core] [scalyr_client.py:343] Sending POST /addEvents with body "BZh91AY&SYq\xdbM;\x00\x00S\x1f\x80P\x07wp\x00`\x00\n\xbf\xef\x9d\x1a \x00d4\xd4M\x18\x93@\xcdL\x8fD\x1e\x9a\x9eA\xaam&\x9e\xa3M\x1a\x00\r\x00\xd1\xa6\x0b$\x94\x04$\xe5)(\xcfY\xa3\x1c\x10P\x16\xf9\xa9\xc0bk~m\xc6\x0b\x86\x99\xe8\xc7\xa0\x19d\xf4\x8ar\x87{}\xd5\xc9\x1f\x8be(\xd2b\x9b\x87t.\xd7\x9dH\xb8.v\xa9\xfdT7\x99!>\xad`\xc2^f>\x81\x13edT\x85\x06\x13Db\x08\x04\xa1\x04\\\xcc\xdeo\xfc]\xc9\x14\xe1BA\xc7m4\xec"
```

After:

```bash
2020-01-28 16:41:24.703Z Level 4 [core] [scalyr_client.py:343] Sending POST /addEvents with body "{"session":"foo","sessionInfo":{"serverHost":"tomaz-vagrant.dev","tier":"dev"},"threads":[],"token":"bar", events: [], logs: [], threads: [], client_time: 1580229684 }"
```

(NOTE: I removed private / sensitive info from both examples)

## TODO

- [x] Test case


scalyr_agent/tests/scalyr_client_test.py
```diff
@@ -905,3 +908,43 @@ def get_user_agent():
...
+        body (payload) should be logged under DEBUG log level.
+        """
+        self.assertEqual(mock_log.info.call_count, 0)
+
+        session = ScalyrClientSession(
+            "https://dummserver.com", "DUMMY API KEY", SCALYR_VERSION
+        )
+        # One for "using server" and other for "using session" message
+        self.assertEqual(mock_log.info.call_count, 2)
+        self.assertEqual(mock_log.log.call_count, 0)
```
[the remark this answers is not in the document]
Kami: I agree - I'm usually not a big fan of mocking myself (there needs to be a balance) since it's usually testing the implementation details which may change and that can make tests less robust / code more coupled to the tests.

Another approach I could go with is to configure logging to a file and simply check that file contains a specific log line.

In fact, that's what I've seen in a couple of other places so perhaps also a small TestCaseWithLogCapture or similar base class refactor is also a good idea.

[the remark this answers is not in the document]
Kami: Here is the approach I end up going with - 7da48802944a23cc5e25a6b459b7597628d9372b, 8086b58b7d514fdf85d10dbbc43f01368719d3a7.

It should be more robust since it doesn't rely on mocking the internals / implementation details. 

I still left one mock check in place though, because I still want to check that body was indeed compressed (aka compression function was called) since that's the edge case that this test is supposed to test.

---

First I wanted to go with a more generic approach where the log output could either be stored to a file or to a class variable (developer would either specify a file handle or StringIO object), but because of the way ``set_log_destination`` function is currently structured, this would either require:

1. Major refactoring of that function
2. "Hacking" approach which relies on ``use_stdout=True`` behavior and mocking ``sys.stdout`` (less code than 1), but on the hack-ish side)

I'm not a huge fan of either of those approaches and I wanted to keep this PR as simple and small as possible, so I went with the "write to files" approach.

As mentioned above, this brings those tests closer to "integration" tests spectrum, but I don't think that's a big deal and in fact, that's already done by various existing unit test cases today.

---

@czerwingithub let me know what you think.

In the future, I would also be OK with implementing either 1) or 2), especially if multiple tests utilize that functionality since it would allow us to avoid disk writes inside the unit tests.

If you are OK with this approach, I can update some of the existing test cases to use that class (but if you think I should leave this out here and keep the scope down, I can also do it in a separate add-on PR once this one is merged or not do it at all at this point).

scalyr_agent/tests/test_util_test.py
```diff
@@ -0,0 +1,59 @@
...
+    Test class which verified BaseScalyrLogCaptureTestCase works correctly.
+    """
+    def test_assertLogFileContainsRegex(self):
+        # Log files should have been created in setUp()
+        self.assertTrue(os.path.isfile(self.agent_log_path))
+        self.assertTrue(os.path.isfile(self.agent_debug_log_path))
+
+        # Write some data to the file and verify assertLogFileContainsRegex works correctly
+        with open(self.agent_log_path, 'w') as fp:
+            fp.write(r'line 1\line 2\line 3')
```
czerwingithub: Did you mean to put newline characters in here so that they are actually separate lines?

Also, any particular reason you are using raw strings?  With the 2->3 migration, we are trying to be careful with our string usage.

scalyr_agent/test_base.py
```diff
@@ -149,6 +153,83 @@ def verify_setup_invoked(self):
...
+            agent_debug_log_file_suffix='_debug'
+        )
+        scalyr_logging.__log_manager__.set_log_level(scalyr_logging.DEBUG_LEVEL_5)
+
+        self.agent_log_path = os.path.join(self.logs_directory, 'agent.log')
+        self.agent_debug_log_path = os.path.join(self.logs_directory, 'agent_debug.log')
+
+        # Print the paths to which we store the output to so they can be introspected by the
+        # developer
+        # TODO: Should we only do this on assertion failure?
```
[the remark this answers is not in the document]
Kami: Agreed, will make a change and move this to ``tearDown`` where we can detect if there was an assertion failure.

As described in another comment in that class, we currently only detect if ``assertLogFileContainsRegex`` assertion method has failed.

If any other assertion method fails, we don't detect that.

The reason is that's it's not trivial to detect if the test has failed in a cross Python-version and test runner manner (it would result in a bunch of boiler plate code I wanted to avoid in this PR, but if needed, we can add / change this later on).

scalyr_agent/tests/test_util_test.py
Kami: Yeah, that's a mistake on my part - there is no need to use a raw string here and there should indeed be a new line characters.

---

On a somewhat related note, I implemented ``assertLogFileContainsRegex`` so it mimics implementation from an existing test so it's line oriented.

To support also checks against the whole file content I also added ``assertLogFileContainsRegex``.

It's possible to also do line checks with a second more "generic" method so if people think it's too complicated or confusing to have two methods, we can just have one which performs checks against the content from the whole file.

Kami:
@czerwingithub I just went ahead and updating the existing tests to use this new base class + assertion methods - 59bb3572444570f2a6af97b9adb691f8a0d1d7b5.

scalyr_agent/test_base.py
```diff
@@ -322,3 +248,142 @@ def assertGreater(self, a, b, msg=None):
...
+        self.agent_debug_log_path = os.path.join(self.logs_directory, 'agent_debug.log')
+
+    def tearDown(self):
+        super(BaseScalyrLogCaptureTestCase, self).tearDown()
+
+        if self.__assertion_failed:
+            # Print the paths to which we store the output to so they can be introspected by the
+            # developer
+            test_name = self._testMethodName
+            print('Storing agent log file for test "%s" to: %s' % (test_name, self.agent_log_path))
```
czerwingithub: Ok, maybe minor nit.. but shouldn't this be in the past tense now?  `Stored agent log file for test...`?  Same below.

scalyr_agent/tests/scalyr_logging_test.py
```diff
@@ -21,30 +21,25 @@
...
-            agent_log_file_path=self.__log_path,
-        )
+        self.__log_path = self.agent_log_path
         self.__logger = scalyr_logging.getLogger("scalyr_agent.agent_main")
         self.__logger.set_keep_last_record(False)
 
     def test_output_to_file(self):
         self.__logger.info("Hello world")
-        self.assertTrue(self.__log_contains("Hello world"))
+        self.assertLogFileContainsLineRegex(file_path=self.agent_log_path,
```
czerwingithub: Optional:  What do you think of having `file_path` default to `agent_log_path`?  So, if you do not specify `file_path`, we just check the agent log.  I can see the argument it's not as declarative about what's going on..but also, most of the times, we are checking the agent log so reduces typing.

czerwingithub:
@Kami Thanks, I added a few minor comments.  Ack about the issues with handling different assertion failures across Python versions.  I think this seems like a pretty good balance.

scalyr_agent/tests/scalyr_logging_test.py
Kami: Yeah, I can make that change.

In fact that's what I did at first since that's what the tests have done originally, but then I changed my mind and went with the more "explicit" approach.

Kami: I made a change with that argument being optional in ea484e0642d0bed5cf39d6a66c3fd79bfdb5e4dd.

PR #381, opened 2020-01-29
Update README, add NOTICE file

Kami:

1. Fix formatting in README.md so it's consistent and it's not mixing rST and Markdown style headings
2. Add new "Copyright, License, and Contributors Agreement" section to README.md
3. Add new NOTICE.md file which contains a notice about included third party libraries. This is not mandatory when using ASF 2.0 license, but it's a common practice and used by most ASF projects (in ASF case NOTICE file usually also contains all the copyright notices from all the bundled libraries)
4. Update scalyr_agent/third_party/README.md and add a corresponding license name next to the library name for a quick reference

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -1,0 +1,15 @@
+Scalyr Agent
+Copyright 2014-2020 Scalyr, Inc.
+
+This repository contains various third party libraries which have been included from other
+projects.
+
+Those libraries are included in the following directories and availble under their corresponding
+licenses:
+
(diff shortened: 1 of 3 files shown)


NOTICE.md
```diff
@@ -0,0 +1,15 @@
+Scalyr Agent
+Copyright 2014-2020 Scalyr, Inc.
+
+This repository contains various third party libraries which have been included from other
+projects.
+
+Those libraries are included in the following directories and availble under their corresponding
```
czerwingithub: s/availble/available

Kami: Fixed, thanks.

PR #382, opened 2020-01-29
Fix flaky smoke-k8s test

czerwingithub:
Force retry of querying the agent's container's id if the API
server indicated the container is still not ready.  This would
result in the CRI determination failing and the K8s tests to
fail.

### DIFF
--- a/scalyr_agent/monitor_utils/k8s.py
+++ b/scalyr_agent/monitor_utils/k8s.py
@@ -1005,6 +1005,12 @@ class _CacheConfigState(object):
             self._lock.release()
 
 
+class AgentPodNotReadyException(Exception):
+    """Raised when the agent pod is not fully ready according to the K8s API server."""
+
+    pass
+
+
 class KubernetesCache(object):
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1106,15 +1113,33 @@ def _update_cluster_name(self, k8s):
...
         self._lock.acquire()
         try:
-            self._api_server_version = gitver
+            is_version_set = self._api_server_version is not None
+            last_check_time = self._last_api_server_version_update
         finally:
             self._lock.release()
 
+        # We only update if we haven't updated it in the last hour.
+        if is_version_set or current_time - last_check_time > 3600:
```
yanscalyr: Should it be `if not is_version_set` here?

czerwingithub: Argh, good catch.  This is why we need unit tests for this code.  The smoke test passes because this just results in us constantly resetting the version which does not impact correctness.

PR #384, opened 2020-01-30
Replace docker-py dependency with last version of docker.

ArthurKamalov:
docker.Client changed to docker.APIClient.
dependencies changed.


Kami:
Good catch :+1: 

I've actually encountered a similar issue here - https://github.com/scalyr/scalyr-agent-2/pull/377/files#diff-913fedaed3cb05fe362c0ceb35e2d827R16

My issue was related to two of the ``docker-py`` dependencies depending on a different and conflicting version of ``requests`` which would cause newer versions of ``pip install`` to fail.

ArthurKamalov:
@czerwingithub

I added `requests` with docker in `pip install` in `Dockerfile's` and in `dev-requirements.txt`, because docker installs its own version.

> @ArthurKamalov Have you done any testing for this? I guess the docker smoke test is probably enough?

Honestly, smoke tests and some unit tests are only things that we can rely on, and they pass. It seems like the basic API remains the same.

Do you think we should add more tests before upgrade this library?

czerwingithub:
@ArthurKamalov We already have requests `2.15.1` installed in the `third_party` directory.  I don't think we need to reinstall it using `pip`, do we?

ArthurKamalov:
@czerwingithub 
If we do not have already installed requests library it will be installed as a requirement for docker, and its version probably will not be the same as we use in the agent. I have tried to force pip to not install request as docker requirement, but pip does not support such a feature.

Kami:
I believe part of that^ is related to how we manage dependencies.

"Production" dependencies are bundled with the repo, and ``dev-requirements.txt`` is only used for development and testing. We need to make sure those are kept in sync for scenarios where bundled dependencies are not used.

One thing which is not clear to me is how ``docker-py`` dependency is installed on the host where the agent with docker monitor runs.

Do we require the end user to install that dependency manually? Or we assume it's already available in PYTHONPATH where the agent with docker monitor runs?

---

On that note, that's the reason why we special sandboxing and isolation model for "packs" in StackStorm world (think of a pack as a monitor or a set of related monitors).

In that model, every pack has a dedicated virtual environment which contains Python dependencies it relies on. That's important when user wants to run different packs on the same host - in such scenario it's highly likely that some packs will have conflicting versions of dependencies.

That's not necessary the problem we have right now, but something we should keep in mind for the future.

And on that note, that sandboxing model is not ideal either - it works well, but we don't share system packages in pack virtual environments which means directories with virtual environments can get large.

docker/Dockerfile.syslog
```diff
@@ -5,7 +5,7 @@ FROM python:2.7-slim as scalyr-dependencies
 MAINTAINER Scalyr Inc <support@scalyr.com>
 
 RUN apt-get update && apt-get install -y build-essential
-RUN pip --no-cache-dir install --root /tmp/dependencies ujson yappi docker-py==1.10.3
+RUN pip --no-cache-dir install --root /tmp/dependencies ujson yappi requests==2.15.1 docker==4.1.0
```
Kami: Eventually, to remove confusion, I think we should have multiple requirements.txt files:

- ``requirements.txt`` - file which contains dependencies the agent depends on which are not bundled with the agent directory. Think ujson, yappi, docker, etc. To be good citizen, we should also declare those in ``setup.py`` under requirements section. Granted, things get a bit more complicated with "optional" dependencies such as docker one which is only required when docker monitor is running...
- ``dev/test-requirements.txt`` - file which only contains dev / testing dependencies (right now we are mixing those and production one in the same file)

Those files should then also be used by those Dockerfiles.

czerwingithub:
@ArthurKamalov Ok, that makes to me now why you want to install `requests` how you are.  I'm fine with that approach.

@Kami The `docker-py` dependency is a new issue we are still trying to resolve.  Previously, we had bundled it and all of its dependencies into the `third_party` library.  However, just a few weeks ago, we decided we should remove it because, we thought, they only time it was needed was when folks were running the agent in docker mode.. in which case they were using our docker image.  Our docker image was changed to automatically install `docker-py` and its dependencies.

The benefit of this is it removed a whole ton of dependencies from `third_party` and we didn't need to worry about them in the Python 2 to 3 migration.

HOWEVER, we've gotten reports from customers indicating we made a bad assumption.  There are customers who run the docker monitor even though they aren't running the agent as a docker container (typically because they want to grab the stats and log files from the host, in addition to all logs runninger under docker.)  For those customers, we are telling them right now to manually install docker-py on their host systems... but if there are a lot of folks in this situation, we are going to have to revisit.  It's not an ideal solution for them.

Maybe we will go back to including all of `docker-py`'s dependencies in `third_party`.  Unfortunately, those have grown a fair bit.

czerwingithub:
@ArthurKamalov Just checking in on this PR to make sure it's still on your radar.  I think it still needs some work to address the right comments. If that's not the case, please just say so.

docker/Dockerfile.syslog
ArthurKamalov: Maybe I  should create separate task/PR for this?

ArthurKamalov:
I made small changes related to docker dependencies and requirements files.

- Created separate requirements file for all docker images. Now we need to modify only this file, and changes will affect all docker images.
- Created `extra-requirements` with all dependencies which may be needed in some cases, such as ujson, docker, in standalone mode.

czerwingithub:
@Kami @ArthurKamalov Right now it looks like we have `docker` declared in both `requirements.txt` and `extra_requirements.txt`.  What's the thinking behind having `docker` in the main `requirements.txt`?  I know this one is a bit weird since we are saying it is only needed when running the docker monitor.

Kami:
I'm fine with having multiple requirements.txt files. In fact, I also made a similar change in another PR.

Ideally, we could use something like Pipefile in the future (https://pipenv-fork.readthedocs.io/en/latest/basics.html), but in the mean time, multiple requirements.txt files should work.

We just need to make sure we organize and name them accordingly - right now it's a bit of a wild west  since we mix different type of dependencies together (aka test vs prod vs monitor dependencym etc.).

ArthurKamalov:
> Right now it looks like we have `docker` declared in both `requirements.txt` and `extra_requirements.txt`. What's the thinking behind having `docker` in the main `requirements.txt`? I know this one is a bit weird since we are saying it is only needed when running the docker monitor.

@czerwingithub 
Actually, the `requirements.txt`file is not the main, it is only for docker builds and it is located in `/docker` directory. The `extra-requirements.txt` is only one new file that was added to the project root and its purpose to help to install some additional stuff, like `ujson`, or `docker` for standalone versions.

PR #387, opened 2020-01-30
AGENT-331 Remove sleep from syslog_monitor test to make it less flaky.

ArthurKamalov:
Ticket: https://scalyr.myjetbrains.com/youtrack/issue/AGENT-331.


scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1380,7 +1380,7 @@ def _initialize(self):
         self.__server_list = self.__build_server_list(self._config.get("protocols"))
 
         # our disk logger and handler
-        self.__disk_logger = None
+        self._disk_logger = None
```
czerwingithub: Add a comment that we are making this protected for testing purposes.

czerwingithub:
@ArthurKamalov Let me know if you agree to the change to mock the `increment_lines` method.. I think it might be the better approach

scalyr_agent/tests/syslog_monitor_test.py
```diff
@@ -257,6 +259,22 @@ def test_config_port_too_high(self):
...
 
 
+class TestAutoFlushingRotatingFileHandler(AutoFlushingRotatingFileHandler):
+    """subclass of AutoFlushingRotatingFileHandler with overridden '_internal_flush'
+    to notify test thread that stream was flushed.
+    AutoFlushingRotatingFileHandler is inherited from old-style class in python2.6, that's why we can not use 'super'.
+    """
+
+    def __init__(self, *args, **kwargs):
+        AutoFlushingRotatingFileHandler.__init__(self, *args, **kwargs)
```
Kami: ``super(AutoFlushingRotatingFileHandler, self).__init__(..)`` is usually more robust  and preferred in case the inheritance hierarchy changes.

ArthurKamalov:
@czerwingithub 

> @ArthurKamalov Let me know if you agree to the change to mock the `increment_lines` method.. I think it might be the better approach

I agree with you. My approach has relied on internal implementation, this is not good. Thank you for pushing me in the right direction.

scalyr_agent/tests/syslog_monitor_test.py
ArthurKamalov: Unfortunately in python2.6 `AutoFlushingRotatingFileHandler` belongs to a class hierarchy that begins from an old-style class :( 
Sorry, I had to put this information in constructor docstring, not a class.

scalyr_agent/builtin_monitors/syslog_monitor.py
ArthurKamalov: Returned back the mangled version.

czerwingithub:
@ArthurKamalov I hit approve and then later remembered a point I had brought up.  I believe you want to set the autoflush delay for zero for the tests since there still could be a race condition between when the bytes actually hit the disk.

scalyr_agent/tests/syslog_monitor_test.py
```diff
@@ -315,45 +353,87 @@ def connect(self, socket, addr, max_tries=3):
 
         return connected
 
+    def wait_for_flush(self):
```
czerwingithub: Is this still used / valid?

czerwingithub:
@ArthurKamalov I think there's just one unaddressed comment left.  Let me know when this is ready for review again.

scalyr_agent/tests/syslog_monitor_test.py
ArthurKamalov: Removed, thank you for pointing out.

ArthurKamalov:
@czerwingithub
I've done as you said - set flush delay to zero. It still does flushing in a separate thread, but I hope that time delay is negligible and can't be a reason for the race condition.

Kami:
I've resolved merge conflict, fixed lint issues and merged this into master since I still saw this issue every now and then locally.

PR #390, opened 2020-02-03
[RFC] Correctly handle SIGINT (CTRL+C) when running in non-daemon / non-fork mode

Kami:
## Background

While testing some functionality using ``--no-fork`` mode, I noticed that CTRL+C aka ``SIGINT`` signal doesn't result in an expected behavior.

When running an application in foreground mode you would usually expect it to start the clean up process and exit when CTRL+C is used.

Right now that's not the case because the agent involves "write status file" handler on SIGINT.

This means you need to explicitly send the foreground process SIGKILL / SIGTERM signal to make it exit which, you could argue, kinda defeats the purposes of no fork / no daemon / foreground mode.

## Proposed Change

I believe that when running in foreground mode, CTRL+C should result in the program invoking the cleanup procedures and exiting.

This pull request implements that change.

---

Keep in mind that this change is not fully backward compatible since now when running in foreground mode, agent won't write status file on CTRL+C, but it will exit instead.

There are a couple of options here:

1) Do nothing
2) First write agent status file and then invoke termination handler + exit
3) Change which signal is used. In the past, for such use cases I used either ``SIGUSR1`` or ``SIGUSR2``

Keep in mind that this change is just a proposal.

If we indeed agree on it and go with either 1) or 3), this will likely also require changes in other places, at the very minimum the documentation will need to be updated.


czerwingithub:
There was some reason, back in the dawn of time, why I didn't use `SIGUSR1` for the status command.  It would have been nice if I had left that in comments.  Sigh.  I want to say it was because of some issue with old systems and Python 2.4.  I vaguely remember that.  But, doing a bit of Google searching didn't turn anything up.

Anyway, I would vote for #3.  I do want to reduce friction with development... and I don't think there is much risk if we change it over to `SIGUSR1`.  There will be a corner case during upgrades... when a user upgrades to the new version of the agent using `SIGUSR1`, they won't be able to run `status -v` (at least with success) until the agent process itself is restarted.  I think this is such a corner case we probably don't need to worry about it.  After all, the normal RPM and Debian install process handles restarting the agent for you.

Kami:
@czerwingithub I agree with you with 3) being the most reasonable.

I made that change in f71c00e8c4a54cc166e1a43cbd7f45237d917da0, but I still need to add some tests.

scalyr_agent/platform_posix.py
```diff
@@ -599,14 +600,50 @@ def start_agent_service(self, agent_run_method, quiet, fork=True):
         """
         self._log_init_debug("Starting agent %s" % scalyr_util.get_pid_tid())
 
+        logger = scalyr_logging.getLogger("scalyr_agent")
```
Kami: NOTE: I know other code uses ``_log_init()`` function to make sure log lines are correctly handled before initialize phase is finished, but I don't think there is a need to use that in the signal handlers.

Here we can use logger directly, because by the time signal handlers are set up and registered, initialization phase should already be finished.

Kami:
I've added some basic mocking based unit tests here - 214b8dc.

It's not 100% ideal since we don't test the actual handler functionality, but doing that would be more involved and until we don't have a good consistent story for end to end tests, I don't want to introduce code which may change or get obsolete in the very near future.

Once we come up with good and consistent patterns for handling end to end tests, I can add additional tests for this functionality in a future PR.

scalyr_agent/platform_posix.py
czerwingithub: I think that is fair.  Maybe just add a comment here to call that out?

Kami: Agreed, added in dda3ed8.

czerwingithub: Thanks.

PR #392, opened 2020-02-04
Fix a couple of small Python 2 and 3 compatibility issues

Kami:
I tried to run latest version of master under Python 2 and Python 3 while testing the signals change and I encountered some exceptions in the log related to mixing bytes and unicode.

This pull request fixes two small issues I've encountered.

There are likely many more which will only be uncovered only while running the agent and that's where a more systematic approach + end to end tests come into play.

Ideally as part of end to end tests, we would spin up an agent process under Python 2 Python 3 with some real life configuration and DEBUG level set and then we would grep agent.log and agent_debug.log for exceptions / errors and work from there.

### DIFF
--- a/scalyr_agent/scalyr_client.py
+++ b/scalyr_agent/scalyr_client.py
@@ -312,7 +312,9 @@ class ScalyrClientSession(object):
             else:
                 body_str = ""
 
-            body_str_raw = body_str
+            # Workaround to fix issue with logging non utf-8 characters. We simply ignore
+            # non utf-8 characters
+            body_str_raw = body_str.decode("utf-8", "ignore")
 
             self.total_request_bytes_sent += len(body_str) + len(request_path)
 
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -18,6 +18,10 @@ from __future__ import unicode_literals
 from __future__ import division
 from __future__ import absolute_import
 from __future__ import print_function
+
+if False:
+    from typing import Union
+
 import codecs
 import sys
 import struct
@@ -391,6 +395,7 @@ def md5_hexdigest(data):
 
 
 def remove_newlines_and_truncate(input_string, char_limit):
+    # type (Union[str, bytes], int) -> str
     """Returns the input string but with all newlines removed and truncated.
 
     The newlines are replaced with spaces.  This is done both for carriage return and newline.
@@ -400,7 +405,7 @@ def remove_newlines_and_truncate(input_string, char_limit):
     @param input_string: The string to transform
     @param char_limit: The maximum number of characters the resulting string should be
 
-    @type input_string: str
+    @type input_string: str or bytes
     @type char_limit: int
 
     @return:  The string with all newlines replaced with spaces and truncated.
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -406,6 +406,7 @@ def remove_newlines_and_truncate(input_string, char_limit):
     @return:  The string with all newlines replaced with spaces and truncated.
     @rtype: str
     """
+    input_string = six.ensure_text(input_string)
     return input_string.replace("\n", " ").replace("\r", " ")[0:char_limit]
 
 
--- a/scalyr_agent/tests/configuration_test.py
+++ b/scalyr_agent/tests/configuration_test.py
@@ -1774,15 +1774,21 @@ class TestJournaldLogConfigManager(TestConfigurationBase):
 
         lcm = LogConfigManager(config, None)
         matched_config = lcm.get_config("test")
-        self.assertEqual("TestParser", matched_config["parser"])
-        self.assertEqual(
-            six.text_type([{"match_expression": "a", "replacement": "yes"}]),
-            six.text_type(matched_config["redaction_rules"]),
-        )
-        self.assertEqual(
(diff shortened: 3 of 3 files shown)


scalyr_agent/util.py
```diff
@@ -406,6 +406,7 @@ def remove_newlines_and_truncate(input_string, char_limit):
     @return:  The string with all newlines replaced with spaces and truncated.
     @rtype: str
     """
+    input_string = six.ensure_text(input_string)
```
Kami: FYI:

```bash
python3 /usr/sbin/scalyr-agent-2 --no-fork start

  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/scalyr_client.py", line 464, in __receive_response
    response_as_json = scalyr_util.json_decode(response)
  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 179, in json_decode
    return _json_decode(text)
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'

During handling of the above exception, another exception occurred:

  File "/usr/sbin/scalyr-agent-2", line 526, in __start
    self.__perform_config_checks(no_check_remote)
  File "/usr/sbin/scalyr-agent-2", line 450, in __perform_config_checks
    ping_result = client.ping()
  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/scalyr_client.py", line 237, in ping
    return self.send(self.add_events_request())[0]
  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/scalyr_client.py", line 572, in send
    "/addEvents", body_func=generate_body, block_on_response=block_on_response,
  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/scalyr_client.py", line 381, in __send_request
    return receive_response()
  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/scalyr_client.py", line 376, in receive_response
    return self.__receive_response(body_str, current_time)
  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/scalyr_client.py", line 470, in __receive_response
    scalyr_util.remove_newlines_and_truncate(response, 1000),
  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 409, in remove_newlines_and_truncate
    return input_string.replace("\n", " ").replace("\r", " ")[0:char_limit]
TypeError: a bytes-like object is required, not 'str'
Terminating agent, please fix the error and restart the agent.
```

scalyr_agent/scalyr_client.py
```diff
@@ -312,7 +312,9 @@ def __send_request(
             else:
                 body_str = ""
 
-            body_str_raw = body_str
+            # Workaround to fix issue with logging non utf-8 characters. We simply ignore
+            # non utf-8 characters
+            body_str_raw = body_str.decode("utf-8", "ignore")
```
Kami: The issue is related to ``LogRecord.getMessage`` throwing because the actual request body can also contain non utf-8 characters:

```bash
  File "/usr/lib/python2.7/logging/handlers.py", line 76, in emit
    if self.shouldRollover(record):
  File "/usr/lib/python2.7/logging/handlers.py", line 156, in shouldRollover
    msg = "%s\n" % self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 734, in format
    return fmt.format(record)
  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/scalyr_logging.py", line 852, in format
    return BaseFormatter.format(self, record)
  File "/home/vagrant/w/scalyr/scalyr-agent-2/scalyr_agent/scalyr_logging.py", line 802, in format
    result = logging.Formatter.format(self, record)
  File "/usr/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc0 in position 316: ordinal not in range(128)
```

Kami:
Fixed intermediate test failure issue (https://app.circleci.com/jobs/github/scalyr/scalyr-agent-2/15192) in e7a86fb.

scalyr_agent/util.py
czerwingithub: Can you update the type hint for `input_str`?  I would suggest we explicitly say it can be either `six.text_type` or `six.binary_type`.

Kami: Yeah, I will update the docstring and add type annotation.

On that note, that's also where MyPy would and will come very handy.

If we had type annotations for that function and most of the other code, MyPy would detect such issues so we could probably enforce and stick with a single strict type (bytes).

But without that, this change (aka accepting both types) is probably the most reasonable right now.

Kami:
> It looks like we'd like the smoke tests to set a debug level and exercise the status -v command as well... and then look for errors in the logs. What do you think?

Yeah that would definitely help with uncovering issues like this one.

scalyr_agent/util.py
Kami: I've updated docstring and added type annotation in e529d8b.

This may require some small adjustments once #377 is merged in and MyPy check is actually hooked up.

PR #394, opened 2020-02-05
Convert server response to text before python3.5 json.loads.

ArthurKamalov:
One suggestion about https://github.com/scalyr/scalyr-agent-2/pull/392


ArthurKamalov:
As I understand, except block for this part is needed when `json_decode` can't deserialize JSON string due to some unexpected reason. But in python 3.5 this can raise an exception too, just because native JSON library does not accept bytes in `loads` function, even if they are absolutely valid for deserialization.
https://github.com/scalyr/scalyr-agent-2/blob/fb7fb38481063673d32c8c32c6b97cebbe09a8d7/scalyr_agent/scalyr_client.py#L465-L473

To be sure that in exception block we deal with truly invalid JSON string, we need to make 3.5 `json.loads` work with bytes. One of the solutions is to wrap it and pass the converted string.

https://github.com/scalyr/scalyr-agent-2/blob/892da57211c98aea4458156efff1bdf3c34cb283/scalyr_agent/util.py#L129-L137

scalyr_agent/scalyr_client.py
```diff
@@ -310,7 +310,7 @@ def __send_request(
                 else:
                     body_str = body
             else:
-                body_str = ""
+                body_str = b""
```
Kami: Good catch :+1: 

The decode below would indeed fail if we don't ensure this variable contains bytes.

Kami:
@ArthurKamalov 

> As I understand, except block for this part is needed when json_decode can't deserialize JSON string due to some unexpected reason. But in python 3.5 this can raise an exception too just because native JSON library does not accept bytes in loads function, even if they are absolutely valid for deserialization.

Correct. And yeah, ideally in that block we should try to catch a more specific exception type.

scalyr_agent/util.py
```diff
@@ -127,7 +127,15 @@ def json_dumps_custom(obj, fp):
             else:
                 return json.dumps(obj, sort_keys=True, separators=(",", ":"))
 
-        return lib_name, json_dumps_custom, json.loads
+        if sys.version_info[0] == 3 and sys.version_info[1] < 6:
```
Kami: LGTM, but can you please also add a test case for this function?

Just a simple call to this function which pass a bytes and string object to it and asserts the result.

ArthurKamalov: Forgot about it... Thank for reminding me.

I was thinking of where this test should be located and I chose `json_lib/encode_decode_test.py`

Also, it looks like the `json_lib/encode_decode_test.py` is no longer associated with `json_lib`, Now it is about testing the other JSON libraries for having the same behavior and results. I thought it will be more correct to move it to from `json_lib` to the other tests.

On the other hand, I am not sure should we keep these tests at all.  In many ways, it looks like extra testing of the standard and third-party libraries.

@Kami @czerwingithub, I would be glad to hear your opinion.

czerwingithub: @ArthurKamalov I think these tests were originally meant to uncover differences between how the third party libs acted.  I'm fine with moving them out, but I think we should keep them, at least the test cases that ensure the libraries behave the same.  We could move them out to as a separate `util_json_test.py` file since it is testing the json shims we have in `util.py`... or just add it as a next test case in `util_test.py` itself.

scalyr_agent/tests/util/json_util_test.py
```diff
@@ -100,6 +104,10 @@ def __runtest(library):
         if sys.version_info[:2] > (2, 5):
             __runtest(JSON)
 
+        # do the same check but now with binary string.
+        if type(text) is six.text_type:
```
Kami: ``isistance`` check is usually preferred, but this also works.

ArthurKamalov: Yes, you are right, I should take this as a habit.

PR #395, opened 2020-02-05
Move Circle CI unit test targets to tox, reduce usage of Circle CI DLC feature

Kami:
This pull request is based on changes from #377. It updates Circle CI unit test jobs to run directly inside a Circle CI job container using tox.

This way we can avoid another layer of indirection which previously involved building and running a dedicated Docker container for those tests. Now those tests run directly inside a regular Circle CI job container using tox.

This also allow us to reduce reliance on (expensive) Docker layer caching. Those new jobs now only rely on "standard" Circle CI directory based caching.

Only exception are Python 2.6 unit tests. Here we don't have much choice because recent tox versions don't support Python 2.6 anymore and the tox notation we use relies on features / fixes in the more recent versions of tox.

In the near future, we can hopefully drop support for this Python version all together.

## Other Changes

I included two other small changes:

1. Increased a sleep in the existing test to make occasional failure less likely (yes, using sleep is not ideal, but alternative approaches would be more involved).
2. Use regular Circle CI Docker image and not the ones with browsers bundled in. Images with browsers bundled are larger which results in longer pulls. Jobs we use have no need for browsers.


Kami:
As an experiment, I also disabled DLC for all the builds.

This should save us 4 x 200 (unit test jobs) + 8 x 200 (smoke tests) credits for each build.

I posted a slightly more detailed analysis on Slack, but at this point this change seems worth it. 

Overall (the build as a whole) doesn't take much longer and currently we pay 10 credits per minute of CPU time for Docker and VM executor builds. At the worst case, this change adds another 1-4 minutes to the job build time, which still is only 40 credits.

So having said that, this analysis isn't really exhaustive and numbers may / will change in the future (once Docker images we use get larger and have more layers, once we add more tests, etc.).

So just another thing to track over time and continuously re-evaluate :)

scalyr_agent/tests/syslog_monitor_test.py
```diff
@@ -332,7 +332,10 @@ def test_run_tcp_server(self):
         expected = "TCP TestXX\n"
         self.connect(s, ("localhost", 8514))
         s.sendall(expected.encode("utf-8"))
-        time.sleep(1)
+
+        # NOTE: Code below relies on sleep. Ideally we should reduce reliance on sleep to make it
+        # more robust and less likely to fail depending on timing.
+        time.sleep(2)
```
czerwingithub: Note, @ArthurKamalov has a PR under review (which I just approved) which should fix this issue the "right way" without needing to increase the sleep.  I think you might want to revert your changes here to avoid conflicting with him.

Kami: Yep, that was just a quick hack / workaround in the mean time :)

PR #396, opened 2020-02-06
Move Python 2.6 unit tests to tox, Circle CI config clean up

Kami:
This pull request updates Circle CI config so it gets rid of some of the unused attributes.

In addition to that, it updates ``unittest-26`` job so it uses the same tox + py.test approach as we use for other unit test jobs. This way all the unit test jobs now follow the same consistent approach.

Getting this to work under Python 2.6 was a bit more involved since a lot of newer versions of the libraries we depend on don't support Python 2.6 anymore, so I need to use older versions in some places.

---

Now that those changes are in place, I believe we can also remove some of the now unused / deprecated code.

This means files such as ``all_tests.py`` and ``run_tests.py``.

I know currently there is some logic in the agent and build package script to determine if a dev version if being ran based on the presence of ``run_tests.py`` file.

We can update that logic to use ``conftest.py`` or some other file for that purpose. Right now we ship tests with distribution artifact so in this case, I also believe we should leave all the files which are needed to run those tests in place aka also included in the distribution tarball.

So I think it's better to use some other "dummy_file" for that purpose.


.circleci/config.yml
```diff
@@ -37,8 +37,6 @@ jobs:
     working_directory: ~/scalyr-agent-2
     docker:
       - image: circleci/python:3.8
-        environment:
-          PYTHONPATH=./scalyr_agent/third_party
```
Kami: Those lines are not used since the tests and code rely on ``scalyr_init`` adding that directory to PYTHONPATH so it's better to remove it to reduce confusion.

.circleci/config.yml
```diff
@@ -106,13 +96,30 @@ jobs:
             .coverage.unittest-27
 
   unittest-26:
+    working_directory: ~/scalyr-agent-2
     docker:
-      - image: circleci/python:2.7-jessie
+      - image: dockershelf/python:2.6-dev
+        environment:
+          PYTHONPATH=./scalyr_agent/third_party
```
czerwingithub: Since scalyr_init modifies the path, do you need this?  (Just going based on your other comments)

tox.ini
```diff
@@ -20,6 +19,28 @@ whitelist_externals =
...
 
+# NOTE: Older version of tox which still supports Python 2.6 doesn't support installing
+# requirements from a file so we need to declare them inline
+# In addition to that, we need to use older versions of some of the libraries
+# (importlib, pytest, etc.)
+# We can get rid of this once we drop Python 2.6 support.
+[testenv:py2.6-unit-tests]
+basepython = python2.6
+install_command = pip install -U --force-reinstall {opts} {packages}
+deps =
```
czerwingithub: We need to make sure this stays in sync with the requirements file, right?  If so, how about adding documentation to it that if you modify the requirements, this should be updated as well.

.circleci/config.yml
Kami: You are correct, I will remove it.

tox.ini
Kami: Yeah, my plan for a future PR was to improve that since right now the approach we use is not ideal (we shove test + dev requirements in a same file), in addition to that, those requirements are also copy and pasted in various places (Dockerfile, etc. and they are not even pinned to a specific / same version everywhere).

So my plan was to split ``dev-requirements.txt`` into multiple requirements file and then use those files as applicable (to avoid duplicating this data in multiple places).

If you think it belongs here, I can also add those changes here instead of working on them in a separate future PR.

Kami:
And to provide some actual data (with the disclaimer that those values vary based on a lot of factors such as test timing, was cache used, etc.):

With my optimizations / changes from other PR which has already been merged into master:

```bash
Usage data for recent workflow runs for workflow "unittest" with status="success" and branch "master"

Workflow ID: 9c8f5283-1793-4ddd-b157-a38369271c26 (https://circleci.com/workflow-run/9c8f5283-1793-4ddd-b157-a38369271c26)
Start time: 2020-02-05T22:47:15.390Z
Duration: 361 seconds
Status: success
Total duration: 361 seconds
Credits used: 352 (~0.21 USD)

```

With additional optimization / changes from this PR:

```bash
Usage data for recent workflow runs for workflow "unittest" with status="success" and branch "test_runner_cleanup"

Workflow ID: 1c022a39-de70-4ad6-851f-6c3e98450c7a (https://circleci.com/workflow-run/1c022a39-de70-4ad6-851f-6c3e98450c7a)
Start time: 2020-02-06T15:34:17.317Z
Duration: 375 seconds
Status: success
Total duration: 375 seconds
Credits used: 313 (~0.19 USD)
```

In short, it should save a couple of more credits.

.circleci/config.yml
Kami: Removed in e15c2eed6d488a938a6954322b733d225aa51b30.

tox.ini
Kami: In the mean time, I went with a change so we read all dependencies for py2.6-unit-tests target from a file.

Kami:
I pushed some more Circle CI config cleanup - there is now less duplicated code since we rely on parameters in more places.

PR #397, opened 2020-02-06
Email us weekly Circle CI credit usage reports

Kami:
This pull request hooks up a script I used yesterday to retrieve Circle CI per workflow credit usage and cost data (with some minor polish) to CI so we (I) receive weekly usage report for each successful workflow run.

This will allow us to identify any outliers / regressions in the credit usage. Right now it runs once a week. If it gets too noisy for me, I'll change it to every two weeks or similar.


scripts/circleci_usage_data.py
```diff
@@ -0,0 +1,267 @@
+#!/usr/bin/env python
+# Copyright 2014 Scalyr Inc.
```
czerwingithub: 2020?

scripts/circleci_usage_data.py
```diff
@@ -0,0 +1,267 @@
...
+    buff, project_slug, workflow, status="success", branch="master", limit=10
+):
+    # type: (StringIO, str, str, str, str, int) -> None
+    """
+    Get usage data for the provided project and write it to the provided buffer.
+    """
+    url = CIRCLE_CI_API_INSIGHTS_URL.format(
+        project_slug=project_slug, workflow=workflow
+    )
+    params = {"limit": "100", "branch": branch}
```
czerwingithub: Why not use the value from the `limit` argument here?

scripts/circleci_usage_data.py
```diff
@@ -0,0 +1,267 @@
...
+
+    return branches
+
+
+def get_usage_data_for_branch(
+    buff, project_slug, workflow, status="success", branch="master", limit=10
+):
+    # type: (StringIO, str, str, str, str, int) -> None
+    """
+    Get usage data for the provided project and write it to the provided buffer.
```
czerwingithub: I would add a comment about how `limit` factors into this.  Reading this, it wasn't obvious to me what limit applied to.

scripts/circleci_usage_data.py
```diff
@@ -0,0 +1,267 @@
...
+        count += 1
+        if count >= limit:
+            break
+
+
+def print_usage_data(
+    project_slug, workflow, status="success", branch="master", limit=10, emails=None
+):
+    # type: (str, str, str, str, int, Optional[List[str]]) -> None
+    if branch == "all":
```
czerwingithub: Add documentation that branch can be `all`.

scripts/circleci_usage_data.py
Kami: I just copied the header from other file, but yeah, we should update all the files to say 2014-2020 (can be done in a separate PR :))

scripts/circleci_usage_data.py
Kami: The limit as in the command line option refers to the maximum number of workflow run we want to see for each branch and the API endpoint returns runs for all the workflows.

Right now that API endpoint doesn't support per-workflow filtering so we just request some large number of items and do that late filtering in the Python code (and we also do some additional late filtering based on the workflow status, etc.).

So yeah, that approach is not 100% ideal and robust since we may actually miss some data (depending on the filters and actual data returned by the API), but that's good enough for this use case, imo (and that API endpoint doesn't support pagination either).

Kami:
I pushed a couple of commits which I believe should address all the review feedback.

PR #400, opened 2020-02-10
Enable flake8 lint check for scalyr_agent/builtin_monitors/ directory

Kami:
This pull request enables ``flake8`` checks on the ``scalyr_agent/builtin_monitors/`` directory and fixes lint violations and false positives in that directory.

This was not done as part of the previous main lint PR to avoid that PR and diff from growing too large hard to review.


scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -1936,11 +1935,11 @@ def configure_from_monitor(self, monitor):
                 % six.text_type(e)
             )
             # if there was an error, reset all values back to defaults
-            label_exclude_globs = self.label_exclude_globs
-            label_include_globs = self.label_include_globs
-            use_labels_for_log_config = self.use_labels_for_log_config
-            label_prefix = self.label_prefix
-            labels_as_attributes = self.labels_as_attributes
+            # TODO(Tomaz): This code seems wrong and unused?
```
czerwingithub: Yeah.. you could add a TODO for imron to fix this.  I suspect it should look something like:
```
    try:
       label_exclude_globs = label_exclude_globs
       label_include_globs = monitor.label_include_globs
       use_labels_for_log_config = monitor.use_labels_for_log_config
       label_prefix = monitor.label_prefix
       labels_as_attributes = monitor.labels_as_attributes

       self.label_exclude_globs = label_exclude_globs
       self.label_include_globs = label_include_globs
       self.use_labels_for_log_config = use_labels_for_log_config
       self.label_prefix = label_prefix
       self.labels_as_attributes = labels_as_attributes
   except Exception as e:
       #report the exception
       # but then do not need to do anything else because we didn't update self until all values extracted from monitor.
```

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2732,10 +2730,6 @@ def check_containers(self, run_state):
                 for cid, info in six.iteritems(running_containers):
                     pod = None
                     if "k8s_info" in info:
-                        pod_name = info["k8s_info"].get("pod_name", "invalid_pod")
-                        pod_namespace = info["k8s_info"].get(
-                            "pod_namespace", "invalid_namespace"
-                        )
```
czerwingithub: These values are used below in the warning that is now commented out (line 2738).  It's not clear to me if we should just remove that line or not, but just in the interest of kicking the can down the road, how about you inline the `info["k8s_info"].get("pod_name", "invalid_pod")` and `info["k8s_info"].get("pod_namespace", "invalid_namespace")` into the log line (i.e., don't use the variable names).

scalyr_agent/builtin_monitors/docker_monitor.py
Kami: Yeah, I assumed so, but I wasn't sure I rather didn't touch it.

/cc @imron

scalyr_agent/builtin_monitors/kubernetes_monitor.py
Kami: Yeah, agreed, will make that change.

Kami: Agreed, added it back in f6e2f6d and commented it out.

scalyr_agent/builtin_monitors/docker_monitor.py
imron: Ok, for context, this code is used so that the syslog monitor can be configured from options set in the docker monitor, so we didn't need to redefine all of the same config options for the syslog monitor when it runs in docker mode.  It only gets run when the syslog monitor is running in docker mode and the docker monitor is also running.

The exception handler is there, to account for the fact that the monitor passed in might not have those attributes set - which probably happened in development and was left in to handle cases where the attribute name was changed or was removed, or to take care of passing an erroneous parameter (i.e. not a DockerMonitor), with the premise that if one of the options didn't exist, then none of them should be set and a warning should be logged.

The bug here is that in the exception handler, the assignment is around the wrong way, i.e. instead of `label_include_globs = self.label_include_globs` they all should have been `self.label_include_globs = label_include_globs` 😬

Steven's suggestion also works, but note the first line `label_exclude_globs = label_exclude_globs` should be `label_exclude_globs = monitor.label_exclude_globs`

Kami: Thanks for the context and additional clarification.

I believe this should do it then - e318349c2cedff89c6191f52968bd7cf46fd6223.

Please confirm :)

imron: That seems right to me now.  I don't know if @czerwingithub has any other comments on the rest of the PR, so I'll leave it to him to approve.

czerwingithub: I'm good with this as well.

PR #401, opened 2020-02-10
Add packages smoke tests.

ArthurKamalov:


tests/config.ini
```diff
@@ -0,0 +1,5 @@
+[agent_settings]
+SCALYR_API_KEY=0cFzhZxkn4HVSOsXkb0Zsj_N8IRq1j_26NqllaXzn0xA-
```
Kami: It would be better to provide those are Circle CI environment variables which are handled are secrets by default (aka are not leaked in the build logs, etc.).

I believe we also already do that for some of the end to end tests.

ArthurKamalov: Oops, I forgot to remove it:)

This path should be in `.gittignore`, and it can be created to configure local tests.
The idea was to use it only locally when there are no environment variables.

Kami: No problem, I also assume you already rotated those credentials :)

tests/conftest.py
```diff
@@ -0,0 +1,31 @@
+import configparser
```
Kami: Please make sure all the new Python files contain Apache license header.

tests/tools/agent_file.py
```diff
@@ -0,0 +1,78 @@
...
+            consistency="consistent"
+        )
+
+    def write_content(self, data: str):
+        self.host_path.write_text(data)
+
+    def read_content(self):
+        return self.host_path.read_text()
+
+    def read(self, size: Optional[int] = ...) -> str:
```
Kami: We still support Python 2.x at the moment, so for the time being, we still need to utilize Python 2 compatible syntax for type annotations (aka type annotations inside comments).

But yeah, in the future once we drop support for Python 2, we can start to utilize native type annotation syntax which is much nicer.

Another option is we only run those tests under Python 3, but then it will still be a bit confusing / messy since different code may contain different type annotations notation.

Kami: But overall, it's good that any new code contains type annotations :+1:

tests/tools/agent_runner.py
```diff
@@ -0,0 +1,144 @@
...
+import abc
+import json
+import tempfile
+
+from .agent_file import MountedFile
+from .utils import create_temp_dir
+
+
+class BaseContainerRunner(abc.ABC):
+    def __init__(self, name, image, docker_client, command=None):
```
Kami: A comment / docstring which describes on a high level what that class does would be nice.

tox.ini
```diff
@@ -74,3 +74,10 @@ commands =
 commands =
     rm -f .coverage
     py.test --cov=scalyr_agent/ --cov=tests/
+
+
+[testenv:rpm]
+deps =
```
Kami: Just a heads up - if those tests will eventually rely on some environment variables being set, you will need to specify ``passenv`` option - https://tox.readthedocs.io/en/latest/config.html#conf-passenv

By default, tox tries to ensure a "clean" execution environment for each tox run, so it doesn't preserve most of the environment variables (kinda like sudo vs sudo -E).

scalyr_agent/agent_main.py
```diff
@@ -865,14 +865,17 @@ def __run(self, controller):
                     logs_initial_positions = None
 
                 # 2->TODO it was very helpful to see what python version does agent run on. Maybe we can keep it?
+
+                python_version_str = ".".join(map(six.text_type, sys.version_info[:3]))
```
Kami: I'm fine with only including the version part (aka the change you made), since it makes parsing, etc. easier, but it also doesn't hurt to include the full version string (current behavior).

dev-requirements.txt
```diff
@@ -9,6 +9,10 @@ pytest==4.6.9
 pytest-coverage
 coverage==4.5.4
 codecov==2.0.15
+PyYAML==5.3
+docker==4.1.0
+six==1.13.0
+pathlib2==2.3.5; python_version <= '2.7'
```
Kami: I believe this conditional dependency notation only works with more recent version of setuptools so we need to ensure more recent / latest version is used everywhere where dev-requirements.txt is used.

This may or may not be an issue where the package builds run. Ideally we would eventually also run periodic builds on our Jenkins to ensure that pipeline is always working with the latest changes as well.

(Somewhat related - https://github.com/apache/libcloud/pull/1393/files#diff-2eeaed663bd0d25b7e608891384b7298R177)

scalyr_agent/tests/smoke_tests/rpm_smoke_test.py
```diff
@@ -0,0 +1,105 @@
...
+import six
+
+from scalyr_agent.tests.smoke_tests.tools.utils import create_temp_dir_with_constant_name
+
+from .tools.package.base_builder import Builder
+from .tools.package.distribution_package_builder.build_image import DistributionPackageBuilder
+from .tools.package.rpm.build_image import RpmBaseDistributionImageBuilder
+from .tools.utils import copy_agent_source
+
+dockerfile = \
```
Kami: Minor style / personal preference things - I prefer to store such larger template values in fixture files on disk instead of in-line.

ArthurKamalov: Yeah, It looks more readable. Done.

scalyr_agent/agent_main.py
ArthurKamalov: Agree.

.circleci/circleci_cli.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
+# This script is convenient to run circlici job on lical macine.
```
Kami: small type - lical -> local

.circleci/circleci_cli.py
```diff
@@ -0,0 +1,83 @@
...
+# What this script does:
+# - converts config to 2.0 version because CircleCi CLI does not support config version higher than 2.0.
+# - set root user for all jobs with 'setup_remote_docker',
+#   because 'circleci' does not have permisions to run docker commands.
```
Kami: Which user does the command run under by default?

One option would also be to fail and print out an error message and say that that user needs to be added to local ``docker`` or similar.

.circleci/config.yml
```diff
@@ -178,6 +186,113 @@ commands:
...
+            pip install "tox==<< parameters.tox_version >>"
+      - run:
+          name: Run Unit Tests under Python << parameters.python_version >>
+          command: |
+            export AGENT_HOST_NAME=agent-smoke-standalone-${CIRCLE_BUILD_NUM}
+            tox -e<< parameters.tox_target >>
+      - save_cache:
+          key: deps2-tox-{{ .Branch }}-<< parameters.python_version >>-venv-{{ checksum "dev-requirements.txt" }}
+          paths:
+            - "~/.cache/pip."
```
Kami: Small typo, should remove final ``.`` from the path.

.circleci/config.yml
```diff
@@ -178,6 +186,113 @@ commands:
...
+
+      - run:
+          name: "Show agent.log"
```
Kami: Another option would be to persist it as an artifact, similar as we do for benchmarking jobs.

.circleci/config.yml
```diff
@@ -178,6 +186,113 @@ commands:
...
+          command: |
+            cat ~/scalyr-agent-dev/log/agent.log
+
+  smoke-package:
+    parameters:
```
Kami: It would be good to add ``description`` attribute for the whole job and the parameters.

.circleci/config.yml
```diff
@@ -178,6 +186,113 @@ commands:
...
+
+      - save_cache:
+          key: deps2-tox-{{ .Branch }}-3.6-venv-{{ checksum "dev-requirements.txt" }}
+          paths:
+            - "~/.cache/pip."
+
+      - run:
+          name: "Show agent.log"
+          command: |
+            docker cp scalyr_agent_smoke_<< parameters.distribution >>_<< parameters.python_version >>:/var/log/scalyr-agent-2/agent.log agent.log
```
Kami: Same here, should we persist this file as an artifact?

.circleci/config.yml
```diff
@@ -692,20 +862,29 @@ workflows:
       - unittest-35
       - unittest-27
       - unittest-26
+      - smoke-rpm-package-py3
```
Kami: We have tons of job now so we will need to keep an eye on our Circle CI usage and potentially adjust which jobs run on every push and which only on merge to master or periodically.

smoke_tests/standalone_smoke_test.py
```diff
@@ -0,0 +1,342 @@
...
+
+    return pattern
+
+
+NORMAL = 0
+ITERATIVE = 1
+
+
+class AgentVerifier(object):
+    NAME = None  # type: Optional[six.text_type]
```
Kami: Please add a comment on what this class does.

Same for normal and iterative constants.

smoke_tests/tools/package/base_builder.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        pass
+
+    def build(self):
+        """
+        Build docker image.
+        """
+        build_context_path = create_temp_dir_with_constant_name(".scalyr_agent_testing")
+
+        dockerfile_path = build_context_path / "Dockerfile"
```
Kami: Should use os.path.join

smoke_tests/tools/package/base_builder.py
```diff
@@ -0,0 +1,133 @@
...
+        dockerfile_path.write_text(self.get_dockerfile_content())
+        if self._copy_agent_source:
+            agent_source_path = build_context_path / "agent_source"
```
Kami: Same here

smoke_tests/tools/agent_runner.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def write_to_file(self, path, data):
+        # type: (Path, Any[six.text_type, six.binary_type]) -> None
+        """
+        Write data to the file located in 'path'
+        :return:
+        """
+        data = six.ensure_text(data)
+
+        file_object = self._file_objects.get(six.text_type(path))
```
Kami: Any particular reason why you are storing open file handles (for performance reasons or similar)?

To be on the safe side, we should make sure we always close those file handles (e.g. inside ``stop()`` or similar).

And if performance is not an issue, we could perhaps just use ``with`` context manager and open a file each time this method is called.

Kami: EDIT: I see you use ``write_line`` in a loop here - https://github.com/scalyr/scalyr-agent-2/pull/401/files#diff-5eabc0cf126ed78d936ed2bafb25674eR212

So I assume that's the reason to avoid potentially man open system calls in that scenario?

dev-requirements.txt
ArthurKamalov: Maybe we just keep it until something shows up?)

.circleci/config.yml
ArthurKamalov: Oops. Thanks

.circleci/circleci_cli.py
ArthurKamalov: I am afraid it does not allow us to add a user or to add user to a group without root permissions too. Also if we use `sudo`, it works only locally.

.circleci/config.yml
ArthurKamalov: I thought that `agent.log`  is not so big, so it would be faster just to look on output, but saving it as an artifact is a better approach. Done.

.circleci/config.yml
ArthurKamalov: Done.

smoke_tests/standalone_smoke_test.py
ArthurKamalov: Done. 
I actually removed those constants because they were not used.

smoke_tests/tools/package/base_builder.py
ArthurKamalov: I thought this operator does the same thing https://docs.python.org/3/library/pathlib.html#operators

smoke_tests/tools/agent_runner.py
ArthurKamalov: Yep, I thought it would be better to keep file handlers opened, but now I think 1000 opens and writes should be fine too. Also, this can be additional stress test for agent log readers. What do you think?

.circleci/circleci_cli.py
ArthurKamalov: Oops, Thanks.

smoke_tests/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+import pytest  # type: ignore
+import six
+import yaml
+
+from smoke_tests.tools.compat import Path
+from scalyr_agent import compat
+
+
+def pytest_addoption(parser):
+    parser.addoption("--config", action="store")
```
Kami: I assume this parameter refers to the circle ci config file, or?

In any case, please document it using help string.

smoke_tests/standalone_smoke_test.py
```diff
@@ -0,0 +1,324 @@
...
+from scalyr_agent import compat
+
+from smoke_tests.tools.agent_runner import AgentRunner
+from smoke_tests.tools.request import ScalyrRequest
+
+from six.moves import range
+import six
+
+_TIMESTAMP_PATTERN = r"\d{4}-\d{2}-\d{2}\s+\d{2}:\d{2}:\d{2}.\d+Z"
+_LEVEL_PATTERN = r"INFO|WARNING|ERROR|DEBUG"
```
Kami: Do we enable debug log level for agent during those runs?

It appears we don't, right? If we do, we should also introspect / follow agent_debug.log log file for messages.

smoke_tests/standalone_smoke_test.py
```diff
@@ -0,0 +1,324 @@
...
+            server_address=self._server_address,
+            read_api_key=compat.os_environ_unicode["READ_API_KEY"],
+            max_count=5000,
+            start_time=self._start_time,
+        )
+
+    def prepare(self):
+        pass
+
+    def _verify(self):
```
Kami: Would be good to throw NotImplemented error here to signal this method needs to be implemented by the sub class.

smoke_tests/standalone_smoke_test.py
```diff
@@ -0,0 +1,324 @@
...
+    if message:
+        pattern_str = "{0}{1}".format(base_pattern, message)
+    else:
+        pattern_str = base_pattern
+    pattern = re.compile(pattern_str)
+
+    return pattern
+
+
+class AgentVerifier(object):
```
Kami: We could also signal that this is a base class by using ``@six.add_metaclass(abc.ABCMeta)`` decorator which works under Python 2 and Python 3.

We just need to confirm if it works under Python 2.6 (I know that works under 2.7, but not sure about 2.6.).

smoke_tests/tools/package/base_builder.py
Kami: Ah, sorry, I didn't know you are working with PathLib objects, I assume you were working with plain strings :)

Kami:
I was curious how much additional credits this burns for every build, and it doesn't seem to be too bad.

```bash
Workflow ID: 3506adea-28dd-412d-96aa-cf7f0d1d180b (https://circleci.com/workflow-run/3506adea-28dd-412d-96aa-cf7f0d1d180b)
Start time: 2020-02-26T08:44:55.552Z
Duration: 374 seconds
Status: success
Total duration: 374 seconds
Credits used: 429 (~0.26 USD)
```

Vs around ~350 credits per unittests workflow run on master branch right now.

So around 70 or so additional credits for each branch / PR push build run (and + ~230 or so credits for master push / merge because of the benchmarking jobs).

---

That seems fine for now, if it becomes an issue or we need to optimize it, we can make subset of jobs periodic cron jobs (e.g. only run job under one Python version on each push and run jobs for all Python versions nightly / weekly + on merge to master).

@czerwingithub ^ Just a heads up.

czerwingithub:
Ack on the heads up about cost.

ArthurKamalov:
@Kami I agree with your idea to not run integration/end-to-end tests on every push. That's why one of my goals was to make those tests easily run locally (from pytest, tox or CircleCi CLI).
Also, I hope that the increased usage of credits is caused by the introduction of the new 4 jobs (2 rpm and 2 deb) and the cost of each job is not increased.

smoke_tests/standalone_smoke_test.py
ArthurKamalov: Yes, we don't, but It is a good idea to enable it. But I think this should be added in a separate PR. Debug log can contain binary length prefixed strings and to read those we need to return some of the `jsonlib` features.

Kami:
@ArthurKamalov 

> @Kami I agree that we do not need to run integration/end-to-end tests on every push. That's why one of my goals was to make those tests easily run locally (from pytest, tox or CircleCi CLI).
Also, I hope that the increased usage of credits is caused by the introduction of the new 4 jobs (2 rpm and 2 deb) and the cost of each job is not increased.

Yeah, I had a look and new non DLC based smoke test jobs, for the most part, finish faster than the old ones.

But as you said, we did add 3 + 4 new jobs. Out of those 4 run only on master and other 3 finish relatively fast so the overhead shouldn't be large (and we can adjust in the future, if needed).

Kami:
Let's get this merged and make any improvements needed in future PRs :)

PR #402, opened 2020-02-10
Add two new metrics to linux process metrics monitor and other small improvements

Kami:
This pull request includes a couple of small improvements in the Linux Process Metrics monitor:

1. Add two new metrics:

* ``app.io.wait`` - number of I/O delays
* ``app.mem.majfalt`` - number of page faults required physical I/O

Both of those metrics come handy when troubleshooting various I/O (noisy neighbor, caching) and other related issues.

I know each metrics is something we need to store (and costs us and our users money), so if others are against it, I'll remove it.

2. Add ``__repr__`` to a couple of classes to make troubleshooting and development easier.

This doesn't affect actual functionality in anyway. It's simply used when ``print()`` / ``__repr__()`` is called in the object in question.

3. Small improvement to the error log message since that error likely indicates that process with that PID doesn't exist.


scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -243,8 +255,12 @@ def is_cumulative(self):
             "app.uptime",
             "app.disk.bytes",
             "app.disk.requests",
+            "app.io.wait",
```
czerwingithub: I believe the major page fault count is cumulative as well, correct?  It is the number since the process began.

czerwingithub:
@Kami I spoke with Mark O'Brien and we decided it would be better to have some way turn off these new metrics if folks do have an issue with the extra bytes of log ingestion.  We've often talked about needing some way of turning off certain metrics.  I wonder if we should just invest in making a general purpose mechanism.  Maybe have a configuration option that is a list of match expressions.  If any one of them matches a metric, it is not included in the output?  It requires a little work to use but it does provide versatility.  I'm open to other suggestions as well.

Kami:
@czerwingithub I some how missed your last comment.

I think a more generic approach allowing user to specify a blacklist of metric names to disable would be reasonable.

For complexity and performance reasons, I do think we need to keep it simple - e.g. just a simple static black list of blacklisted metrics names (vs something more complex such as supporting both black and white list + globs).

scalyr_agent/builtin_monitors/linux_process_metrics.py
Kami: Correct, pushed a fix - 43038a5c3a50827741a21501d98f100ed72a7c1b.

Kami:
@czerwingithub Now that #458 has been merged, is there anything else blocking this PR from getting merged?

I added a changelog entry and instructed users to utilize this new monitor config option if they wish to exclude those metrics for their process monitors.

PR #404, opened 2020-02-12
Improved parsing of JSON and CRI long log lines

czerwingithub:
Introduced concept of "extended lines" in the LogFileIterator.
These are used when we are using an internal parser (such as json)
but the underlying line is longer than the max line size.

At the cost of using extra memory, customers can now change settings
to read the entire JSON object and then break the resulting log
line extracted from the JSON object into smaller fragments less
than the max line size.

Had to introduce a buffer to hold the extended line content as well
as a secondary position object to point at which fragment in the
extended line content should be returned next.


czerwingithub:
### Problem

Kubernetes and Dockers use log file formats that essentially wraps the customer's log lines.  For example, K8s emits a JSON object for each line emitted by the container, with a special field named `log` holding the original log line written by the container.  This adds in an extra level of parsing that must occur.  Once to unwrap the JSON objects, and then another to parse the actual customer log line.  To ease the burden on the customer, the Scalyr Agent automatically unwraps this first layer by parsing the JSON objects and just sending the actual log line content up to Scalyr.

The problem is this interacts badly with the agent's idea of maximum line length.  If the serialized JSON object is longer than the max line length, the agent will chop it off and still try to parse the resulting line as JSON.  This fails miserably and the agent ends up sending the non-parsed line up to Scalyr.  This results in customer's data being polluted with half-valid JSON objects.

### Solution

This commit solves this by allowing the agent to read lines longer than the max line size at the cost of extra memory consumption (while we are reading such lines).  This is only enabled for internal parsers (like the K8s JSON and CRI parsers).  With this, the entire JSON object can be parsed and the real log line content recovered.  The agent still breaks that line up into fragments controlled by the max line size (otherwise, the Scalyr Server would enforce this by just truncating the line and thus losing customer data).

### Implementation design

This is done in the `LogFileIterator`.  I introduce the idea of "extended lines" which are lines that can span buffer pages.  In the case of internal parsers, we will attempt to read an entire line into an extra buffer called `ExtendLineBuffer`, if this means reading many extra pages of data.  There is still a limit but the customer can set that however they see fit at the cost of the agent consuming more memory while an extended line is being parsed.

I made changes to introduce a separate position variable `__extended_line_position` which tracks where in the extended line we will next read a fragment (if we are in the middle of reading an extended line).  This position is also included in the checkpoint state so that we can pick up reading from the middle of an extended line as appropriate.

scalyr_agent/log_processing.py
```diff
@@ -670,8 +763,91 @@ def readline(self, current_time=None):
...
+                result, self.__position, raw_line_length
+            )
+            # As part of the invariant for iterating over the extended log line, we always leave the
+            # buffer position just at the start of the extended log line.
+            self.__buffer.seek(original_buffer_position)
+
+            # If we have a previous fragment position but just reread the extended line (which may happen during
+            # seeks or restoring from a checkpoint), check to make sure we that looks like it is the same one as
+            # we got last time we parsed the line. Otherwise, invalidate the fragment position.  This will cause us to
+            # just copy the extended line again, It may be different if they changed configuration parameters for the
```
imron: If we change the config, don't all existing log iterators get closed down and new ones created? In which case we should never see them change?

I think it's still worth checking for this, but it should never happen because of a config change.

scalyr_agent/log_processing.py
```diff
@@ -1490,6 +1723,255 @@ def get_offset_into_buffer(self, position):
...
+
+            self.__raw_size = raw_size
+            self.__mark_position_start = mark_position_start
+
+        def get_next_fragment(self, fragment_position, max_line_length):
+            """
+            Returns the next fragment from the extend log line pointed at by the fragment position.
+
+            :param fragment_position: The fragment position pointing at the next fragment to read.
+            :param max_line_length: The maximim line lenght, used to control the maximum size of the
```
imron: `maximim` -> `maximum`
`lenght` -> `length`

scalyr_agent/log_processing.py
```diff
@@ -392,7 +427,16 @@ def get_sequence(self):
         @return: A tuple containing a (sequence_id, sequence_number)
         @rtype: (six.text_type, int)
         """
-        return self.__sequence_id, self.__sequence_number_at_mark + self.__position
+        # Add in to the position for thingy
```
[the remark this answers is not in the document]
czerwingithub: Ha, sorry about that.  Clearly I meant to go back and spruce that up.

scalyr_agent/log_processing.py
```diff
@@ -569,9 +649,13 @@ def readline(self, current_time=None):
...
-        if self.__buffer is None or (
-            available_buffer_bytes < self.__max_line_length
-            and more_file_bytes_available
+        if (
+            self.__buffer is None
+            or more_file_bytes_available
+            and (
+                available_buffer_bytes < self.__max_line_length
+                or not self.__have_buffer_for_fragment_position(available_buffer_bytes)
+            )
```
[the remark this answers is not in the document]
czerwingithub: Thanks for that.  Ok, I've redid it a little bit and I think it is more readable now.  I also realized that the `self.__buffer is None` check is superfluous because the second condition (whether or not we have at least `max_line_size`  bytes available in the buffer) will cover that case as well.

scalyr_agent/log_processing.py
czerwingithub: I updated the comments here to be more explicit.  This can happen when we restore from a checkpoint due to either a agent soft restart or a full restart.  The previous LogFileIterators were shut down, but they still persisted a checkpoint reflecting how they read the line (which is based on the
value of max_line_length and the max_internal_parse_line values).

scalyr_agent/log_processing.py
czerwingithub: Thanks.

imron:
All those changes look good to me 👍

PR #406, opened 2020-02-13
AGENT-335 Fix kubelet auth on newer k8s versions

yanscalyr:


Kami:
I believe we already have end to end / integration tests for this code path as part of our smoke tests, right?

yanscalyr:
I believe smoke k8s should cover it.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2142,34 +2147,63 @@ def __init__(self, k8s, port=10255, host_ip=None):
...
         }
         self._session.headers.update(headers)
 
-        global_log.info("KubeletApi host ip = %s" % host_ip)
-        if host_ip:
-            self._http_host = "http://%s:%d" % (host_ip, port)
-        else:
-            self._http_host = None
+        global_log.info("KubeletApi host ip = %s" % self._host_ip)
+        self._build_host()
```
czerwingithub: Let's declare two variables here, called `_kubelet_url` and `_fallback_kubelet_url`.  Let's set them by doing something like:
```
    self._kubelet_url = self._build_kubelet_url(url_template, host_ip)
    self._fallback_kubelet_url = self._build_kubelet_url(FALLBACK_KUBELET_URL_TEMPLATE, host_ip)
```

(where `FALLBACK_KUBELET_URL_TEMPLATE` is a constant variable defined in this file.)

The `_kubelet_url` will represent the URL we are currently using to contact the kubelet api.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2142,34 +2147,63 @@ def __init__(self, k8s, port=10255, host_ip=None):
...
         self._timeout = 20.0
 
+    def _build_host(self):
```
czerwingithub: Replace this with `_build_kubelet_url` mentioned above.  Also make it static and add a test case to verify the string.Template substitution works correctly.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2142,34 +2147,63 @@ def __init__(self, k8s, port=10255, host_ip=None):
...
     def query_api(self, path):
         """ Queries the kubelet API at 'path', and converts OK responses to JSON objects
         """
-        url = self._http_host + path
-        response = self._session.get(url, timeout=self._timeout)
+        url = self._host + path
+        try:
+            response = self._session.get(url, timeout=self._timeout, verify=False)
+        except ConnectionError as e:
+            if not self._fallback:
```
czerwingithub: Instead of having `self._fallback`, do a check of `if self._kubelet_url != self._kubelet_fallback_url`.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2142,34 +2147,63 @@ def __init__(self, k8s, port=10255, host_ip=None):
...
+                global_log.warning(
```
czerwingithub: There's duplicate code below.  Let's consolidate that.  Just set a local variable called `fallback = True` here and a `fallback_error_message` to the string you are logging here.  We will actually log it down below.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2142,34 +2147,63 @@ def __init__(self, k8s, port=10255, host_ip=None):
...
+        self._protocol = "http"
+        self._build_host()
+
     def query_api(self, path):
         """ Queries the kubelet API at 'path', and converts OK responses to JSON objects
         """
-        url = self._http_host + path
-        response = self._session.get(url, timeout=self._timeout)
+        url = self._host + path
+        try:
```
czerwingithub: I'm not a big fan of recursive calls like down below.  I'd rather we do something where this is a loop, and we keep retrying until we exit the loop.  You could just do a `while True` and then clearly always break out of the loop down below unless a local variable named `fallback` is `True`.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2142,34 +2147,63 @@ def __init__(self, k8s, port=10255, host_ip=None):
...
-                "Invalid response from Kubelet API.\n\turl: %s\n\tstatus: %d\n\tresponse length: %d"
-                % (url, response.status_code, len(response.text)),
-                limit_once_per_x_secs=300,
-                limit_key="kubelet_api_query",
-            )
-            raise KubeletApiException(
-                "Invalid response from Kubelet API when querying '%s': %s"
-                % (path, six.text_type(response))
-            )
+            if not self._fallback:
```
czerwingithub: Is there a particular status code we can key off of to trigger the fallback?  It seems here that any error (such as us issuing a poorly formatted request) will trigger us to fallback.  That worries me a bit.

czerwingithub: Let's move this code out of here.  Just set `fallback` to True and set the `fallback_error_message` to what you were going to log.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2142,34 +2147,63 @@ def __init__(self, k8s, port=10255, host_ip=None):
...
+                    scalyr_logging.DEBUG_LEVEL_3,
+                    "Invalid response from Kubelet API.\n\turl: %s\n\tstatus: %d\n\tresponse length: %d"
+                    % (url, response.status_code, len(response.text)),
+                    limit_once_per_x_secs=300,
+                    limit_key="kubelet_api_query",
+                )
+                raise KubeletApiException(
+                    "Invalid response from Kubelet API when querying '%s': %s"
+                    % (path, six.text_type(response))
+                )
```
czerwingithub: At the bottom of the loop, just check if `fallback` is `True`, and if so then log the error message and update `self._kubelet_url`.

scalyr_agent/monitor_utils/k8s.py
yanscalyr: I realized there was no need for the exception check since thats not the response using the wrong port gives, I didn't follow the suggestions exactly as a result since I think I can make it cleaner now.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2122,16 +2126,22 @@ class KubeletApi(object):
         A class for querying the kubelet API
     """
 
-    def __init__(self, k8s, port=10255, host_ip=None):
+    def __init__(
+        self,
+        k8s,
+        host_ip=None,
+        kubelet_url_template=Template("https://${host_ip}:10250"),
```
czerwingithub: Let's add this as a top-level configuration option in `configuration.py`.

yanscalyr: Since `KubeletAPI` doesn't currently have the config, would it be better to pass in the config, or to pass in the already constructed template?

PR #407, opened 2020-02-13
Fail early and loudly and throw a more user-friendly exception if ca_cert_path / intermediate_certs_path file doesn't exist

Kami:
This pull request updates agent code to fail loudly and fast if either file specified by ``ca_cert_path`` and ``intermediate_certs_path`` config option doesn't exist.

Previously a somewhat cryptic / non clear exception was logged in the logs and agent process continued to run (without being able to ship any data to the API):

```bash
2020-02-13 20:43:07.330Z ERROR [core] [connection.py:519] [error="client/connectionFailed"] Failed to connect to "https://app-qatesting.scalyr.com" due to errno=2.  Exception was "[Errno 2] No such file or directory".  Closing connection, will re-attempt
```

Now the agent fails early and doesn't start:

```bash
$ scalyr_agent/agent_main.py start --no-fork --no-change-user 

  File "scalyr_agent/agent_main.py", line 526, in __start
    self.__perform_config_checks(no_check_remote)
  File "scalyr_agent/agent_main.py", line 448, in __perform_config_checks
    client = self.__create_client(quiet=True)
  File "scalyr_agent/agent_main.py", line 1214, in __create_client
    raise ValueError("Invalid path \"%s\" specified for the ca_certs_path config option: file doesn't exist" % (ca_file))
ValueError: Invalid path "/tmp/fooo" specified for the ca_certs_path config option: file doesn't exist
```

Keep in mind that this issue would only arise if user manually installed the agent directly from a git repo or similar. If they use the packages, those filles should be included and installed by the packages.

(I encountered this issue while hooking benchmark runs to Circle CI)

## TODO

- [x] Tests


scalyr_agent/agent_main.py
```diff
@@ -1207,6 +1207,17 @@ def __create_client(self, quiet=False):
         if self.__config.verify_server_certificate:
             ca_file = self.__config.ca_cert_path
             intermediate_certs_file = self.__config.intermediate_certs_path
+
+            # Validate provided CA cert file and intermediate cert file exists. If they don't
+            # exist, throw and fail early and loudly
+            if not os.path.isfile(ca_file):
+                raise ValueError(("Invalid path \"%s\" specified for the \"ca_cert_path\" config "
+                                  "option: file doesn't exist" % (ca_file)))
```
czerwingithub: There seems to be extra parens around the argument to `ValueError` and around `ca_file` as well.  Are those necessary?  Same down below.

Kami: It's technically not required since a string which is split across multiple lines is already inside parenthesis, but it appears black formatting tool is not too opinionated about that and leaves it in place.

Kami:
Removed redundant parenthesis and added a test case in e5db7f7.

PR #409, opened 2020-02-14
Fix metric types for "app.disk.requests" metrics

Kami:
This pull request fixes ``app.disk.requests`` (``type={read,write}``) metric type.

Those two metrics represent a total number of read / write requests (counter) so bytes is not a correct unit.

## TODO

- [x] Tests


Kami:
I've added a very basic test case which verifies those two metrics have no unit associated with them - 85ad571ba0edc1fdd2d413db043253d554617f95.

PR #411, opened 2020-02-17
Continuous Tracking of Agent Resource Utilization and Performance

Kami:
This pull request includes code and related changes which are needed so we can continuously track agent resource utilization and performance over time (commits).

It relies on sending this data to a CodeSpeed (https://github.com/tobami/codespeed) instance (https://scalyr-agent-codespeed.herokuapp.com/).

## Benchmarks

This pull request includes 4 type of agent process level resource utilization benchmarks which should serve as a good starting point.

* Idle agent with default config - Idle agent process with no logs configured to watch and default agent Linux process and system level metrics monitors enabled (aka default configuration).
* Idle agent with default built-in monitors disabled - Idle agent process with no logs configured to watch and also default agent Linux process and system level metrics monitors disabled.
* Agent under load with an existing 50 MB log file - This agent is configured to consume existing 50 MB Apache access log file from disk (that log is also configured with accessLog parser).
* Agent under load which consumes dynamically growing log file. This agent is configured to consume log from a file which is written to during the benchmark. As noted on Slack, this benchmark is not ideal (it relies on generated random data which is not compressible and doesn't represent real life log data well), but it's a start.

All those benchmarks run under 2 Python versions (Python 2.7 and Python 3.5) so we can compare and track performance across different Python versions.

In addition to that, we also track number of log lines for a particular log level (INFO, WARNING, ERROR, DEBUG) emitted during each of those benchmark runs. This will allow us to track regression and issues (e.g. number of error log lines has increased). In fact, those values already provided me with valuable info while hooking up those benchmarks.

## Implementation

Implementation relies on a couple of Python and shell scripts. The most important one being the one which starts the agent with a specific log file and captures metrics of the running agent process and submits them to CodeSpeed instance.

And as you can see, majority of the current implementation is also "abstracted away" by the Circle CI job configuration - it's relatively easy to add a new simple process level benchmark - new job needs to be added to Circle CI and a couple of parameters configured.

At first I was thinking about writing a Python abstraction / classes instead, but then I decided this would be over-engineering and over-complicating in this case.

Python level abstraction would make more sense when we starting adding synthetic benchmarks which measure / track performance of specific pieces of the agent code. In that case, a base class which encapsulates tracking of duration (+ other metrics, multiple runs + 99 percentile, etc.) of benchmark test methods + submitting those metrics to CodeSpeed makes more sense.

# Note on Execution Environment

As discussed before, currently we run those benchmarks directly inside a Circle CI job container. 

This is done for simplicity and consistency reasons (so we have a single place to go to for all the build related stuff).

Over time, we will see how well this works out. It we will have issues with inconsistent CPU related metrics (because of noisy neighbors, server over-subscription, etc.), we might need to look into alternatives.

Some options include:

* Try using Circle CI VM executor
* Perform the run multiple times and use average / 99 percentile value of those runs (to account for noise during runs).
* Run those benchmarks on self-hosted Jenkins (ideally in that case, we would also use dedicated VMs for those tests otherwise we don't really gain that much)

## TODO

- [x] Finish the implementation
- [x] Document available benchmarks and metrics in the ``benchmarks/README.md`` file
- [ ] Cherry pick those changes in a new branch based on a release branch so we can compare resource utilization to code before Python 3 changes were introduced
- [x] Comment out testing / development Circle CI config changes


Kami:
@czerwingithub 

> One question I have to ask, though I suspect I know the answer. You have some moderate amount of code here to run the Linux process monitor's abstraction on the Scalyr Agent to gather perf data. Why not just let the monitor run from within the agent, and parse the resulting log file?

One of the reasons for that was so we can also track those metrics for agent process which has those monitors disabled.

If we went with the build in monitors approach, we would still need to do some processing to correctly handle gauge metrics.

Having said that, I'm all for simplicity, the less code we need to maintain, the better. In fact, that's one of the reasons I started a change in #413 (so we could eventually utilize that functionality in the future).

.circleci/config.yml
```diff
@@ -61,6 +66,116 @@ commands:
...
+      run_time:
+        description: "How long to run the capture for (in seconds)."
+        type: integer
+        default: 120
+      capture_interval:
+        description: "How often to capture agent process level metrics during the process run time (in seconds)."
+        type: integer
+        default: 10
+      agent_pre_run_command:
+        description: "Optional bash command / script to run before staring the agent and the metrics capture script."
```
czerwingithub: s/staring/starting/

.circleci/config.yml
```diff
@@ -61,6 +66,116 @@ commands:
...
+        type: string
+        default: ""
+      agent_post_run_command:
+        description: "Optional bash command / script to run after staring the agent and the metrics capture script."
```
czerwingithub: s/staring/starting/

.circleci/config.yml
```diff
@@ -458,29 +682,48 @@ jobs:
 
 workflows:
   version: 2
-  unittest:
+#  unittest:
```
czerwingithub: Any reason this is commented out?  You could either add a TODO describing when to uncomment it, or uncomment it now.

benchmarks/scripts/send_usage_data_to_codespeed.py
```diff
@@ -0,0 +1,305 @@
...
+Dependencies:
+
+    - scalyr_agent Python package
+    - requests
+    - numpy
+"""
+
+from __future__ import absolute_import
+
+if False:
```
czerwingithub: Add a comment explaining why this is here.

benchmarks/scripts/send_usage_data_to_codespeed.py
```diff
@@ -0,0 +1,305 @@
...
+            "result_value": metric_result["value"],
+        }
+
+        if commit_date:
+            item["revision_date"] = commit_date.strftime("%Y-%m-%d %H:%M:%S")
+
+        # Include optional pre-computed data for gauge metrics
+        for key in ["min", "max", "std_dev"]:
+            if metric_result.get(key, None) is not None:
+                item["min"] = metric_result[key]
```
czerwingithub: Should this be `item[key]` here?

benchmarks/scripts/start-agent-and-capture-metrics.sh
```diff
@@ -0,0 +1,109 @@
...
+# run and use a computed percentile value or similar.
+
+# Every command failure (aka non zero exit) in the script should be treated as a fatal error
+set -e
+
+SCRIPT_DIR=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
+
+# Verify mandatory environment variables are set
+if [ -z "${CODESPEED_URL}" ]; then
+    exit 2
```
czerwingithub: Add error message indicating missing env var.

benchmarks/scripts/utils.py
```diff
@@ -0,0 +1,197 @@
...
+"""
+This module contains various utility functions for sending data to CodeSpeed.
+"""
+
+from __future__ import absolute_import
+
+from datetime import datetime
+from argparse import ArgumentParser  # NOQA
+
+if False:
```
czerwingithub: Add a comment describing why this is here.

benchmarks/scripts/start-agent-and-capture-metrics.sh
Kami: Yep, I also planned to refactor this in a common.sh file so we can re-use this check between both of the scripts.

.circleci/config.yml
Kami: Yeah, that was a testing change (so it didn't trigger all the builds on push to WIP pr).

I will uncomment it before merging into master :)

benchmarks/scripts/start-agent-and-capture-metrics.sh
Kami: Pushed a change where common functionality is refactored in common.sh file.

PR #414, opened 2020-02-18
Update pre-commit configuration so it matches tox lint configuration

Kami:
This pull request updates pre-commit config file so it matches our lint configuration from ``tox`` config.

This means pre-commit now runs the same checks (with the same configuration) as running ``tox -e lint`` does.

By default, ``pre-commit`` creates virtual environment for each check / hook, but I wanted to avoid that since we already have a virtual environment with all the dependencies for out tox targets (not to mention that there are also dependency version and configuration issues there).


Kami:
@lakshmi-kannan ^ You might be interested in how I utilized "local" hooks which don't rely on pre-commit creating virtual environments for each one. In places where tox is already used, it makes sense to use local checks.

PR #417, opened 2020-02-19
Update the `envfrom` yaml to use `apps/v1` apiVersion

yanscalyr:
Mostly done through running `kubectl convert -f k8s/scalyr-agent-2-envfrom.yaml`, with some minor manual reordering.

From my testing this change should work just fine for Kubernetes versions 1.9 and higher, and our stats show no one using versions older than that.

### DIFF
--- a/k8s/scalyr-agent-2-envfrom.yaml
+++ b/k8s/scalyr-agent-2-envfrom.yaml
@@ -1,77 +1,104 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: DaemonSet
 metadata:
+  annotations:
+    deprecated.daemonset.template.generation: "0"
+  creationTimestamp: null
+  labels:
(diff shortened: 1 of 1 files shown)


Kami:
Would it be possible / make sense to hook up this command as a check to our CI to ensure this file is always up to date with the API version we want?

Aka something along the lines of:

```bash
kubectl convert -f k8s/scalyr-agent-2-envfrom.yaml > converted.yaml
diff k8s/scalyr-agent-2-envfrom.yaml converted.yaml || (echo "Generated k8s/scalyr-agent-2-envfrom.yaml file is not up to date. Run kubectl convert and add and commit the generated file" ; exit 1)
```

yanscalyr:
The command is actually deprecated so I don't think it will work for any updates to the API. Besides that the converted output lacks comments that we include so there would be differences every time.

PR #418, opened 2020-02-19
Small pre-commit config fixes and improvements

Kami:
This pull request contains three fixes / improvements to the pre-commit config:

1. Add missing ``types`` attribute to the pylint check. Without that, it will try to run pylint on all the files and not just Python ones.

2. Update ``black`` hook and make sure that also automatically formats the files and doesn't just perform a check.

Performing just a check is desired behavior for the CI, but pre-commit hook should also apply auto formatting fixes.

3. Make sure we run MyPy in incremental mode - this way it will use local cache and the runs will be faster.

Keep in mind that's it's desired that on the CI (such as ``tox -elint`` target) we don't use cache to avoid any stale cache related issues, but during local development where pre-commit runs, using cache is desired, otherwise MyPy check will be slow.

Results (with a single changed Python file).

Before:

```bash
$ time pre-commit run
black....................................................................Passed
flake8...................................................................Passed
pylint...................................................................Passed
mypy.....................................................................Passed

real	0m13.221s
user	0m6.796s
sys	0m1.568s
```

After:

```bash
$ time pre-commit run
black....................................................................Passed
flake8...................................................................Passed
pylint...................................................................Passed
mypy.....................................................................Passed

real	0m5.460s
user	0m4.440s
sys	0m0.612s
```

Keep in mind that a lot of those checks (pylint, MyPy) are "cross Python package" and need to process multiple files even if a single file has changed (to track the types, called functions, etc.).

### DIFF
--- a/.circleci/modernize/modernize.py
+++ b/.circleci/modernize/modernize.py
@@ -387,7 +387,7 @@ if __name__ == "__main__":
         "files",
         metavar="FILES",
         type=str,
-        nargs="+",
+        nargs="?",
         help="File paths to run the script on. If not specified it runs on all the files.",
     )
     parser.add_argument(
--- a/.circleci/modernize/modernize.py
+++ b/.circleci/modernize/modernize.py
@@ -24,6 +24,11 @@
 # Note: This is only python3 script.
 
 if False:
+    # NOTE: This is a workaround for old Python versions where typing module is not available
+    # We should eventually improve that once we start producing distributions with Python
+    # interpreter and dependencies bundled in.
+    # Adding conditional "typing" dependency would require too much boiler plate code at this point
+
     from typing import Dict
     from typing import Any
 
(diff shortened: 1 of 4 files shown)


Kami:
I pushed another change so we now also run ``modernize`` as part of our pre-commit hook. Now it's fully in sync with tox.ini.

To be able to do that, I also needed to make a change to modernize.py so it supports running modernize checks on file paths which are passed to it as a command line argument.

On top of that, I also updated our lint checks so we run them on the ``.circle/modernize/`` directory.

.circleci/modernize/modernize.py
```diff
@@ -23,6 +23,10 @@
 #   -j --processes <jobs count> run concurrently.
 # Note: This is only python3 script.
 
+if False:
```
czerwingithub: Do you need to leave this in? If so, please add a comment explain what it is for.

Kami: Yeah, sorry, we have a comment in other places where type annotations are used, but not here (will also add it there).

In short, that's a workaround for Python 2 where ``typing`` package is not available by default. 

When MyPy runs under Python 3, those lines will still be executed / parsed, but not during a regular program run.

PR #419, opened 2020-02-20
Cherry pick performance tracking changes into release branch

Kami:
This pull request cherry pick performance tracking changes from #411 on top of release branch so we can compare performance to pre Python 3 compatibility changes.

(Luckily the changes were mostly contained inside benchmarks/ directory so cherry picking wasn't too painful)


Kami:
NOTE: Test failure is nothing related to my change. That's one of the races we've fixed in master.

Kami:
@czerwingithub Those changes don't in any way affect functionality of the agent so unless you have anything against it, we can also merge it into release branch.

benchmarks/README.md
```diff
@@ -0,0 +1,117 @@
...
+random data is not compresible, etc).
+
+#### Benchmarks
+
+You can think of a benchmark as a metric which is captured during the benchmark run. We have
+two types of metrics:
+
+* Gauges - those represent values which can change during the agent run time (think memory usage).
+  We capture / sample those values multiple time during the run-time of the benchmark and at the
+  end, calculate 101.9 percentile and send that value as long as min, max and std_dev to CodeSpeed.
```
czerwingithub: 101.9 percentile?  I'm guessing that is a typo since that doesn't make mathematical sense.

benchmarks/README.md
```diff
@@ -0,0 +1,117 @@
...
+* Counters - those represent values which monotonically increase during the agent run time (think
+  total number of disk write requests).
+
+Each benchmark can also has "less is better" flag is associated with it. This flag represents a
```
czerwingithub: First sentence is a bit munged.  How about: "Each benchmark can also have a "less is better" flag associated with it.

czerwingithub:
@Kami Normally I'd be against updating release since I do want it to be able to reproduce what was released for that particular version.  However, I think doing it in this case is fine since it does not update any files that are included in the release.

benchmarks/README.md
Kami: Yeah, no idea how this typo ended up there - will fix it :)

Kami:
There are still a bunch of intermediate test failures which have been fixed in master so I needed to re-run the build a couple of times to get it to pass.

PR #420, opened 2020-02-20
Performance tracking script fixes and improvements

Kami:
This pull request includes a couple of small fixes and improvements on top of #411:

1. Fix debug level logging (make sure we set debug level for all the loggers)

2. Add two new additional RSS memory usage related metrics / benchmarks - ``memory_usage_rss_shared`` and ``memory_usage_rss_private``.

Those two new metrics might help us with troubleshooting various memory usage related issues. If nothing wonky is going on, "shared" RSS memory metric should stay more or less stable and the private part should be changing.

I used ``psutil`` because that's she simplest and least invasive change. In theory, I could also add those two new metrics to LinuxProcessMetrics monitor, but this would be much more involved and require parsing ``/proc/<pid>/smaps``, etc.

In addition to that, I then couldn't use that change in #419, unless I would also cherry pick the monitoring changes there (which is something we don't want at this point).


Kami:
Also pushed a workaround for a race which occurs every now and then - e4ee55aee3933461b798f007325643c35d39dc93.

Sleep is now ideal, but hopefully it's good enough for now.

PR #422, opened 2020-02-22
fix struct.pack|unpack wrapper for python < 2.7.7

ArthurKamalov:
Fix struct wrapper in `compat` module, because the problem with unicode incompatible strung remains until python2.7.7.


Kami:
LGTM :+1: 

I assume we already have tests for this scenario?

I tried to hook up tests running under Python 2.7.5 on Circle CI, but I couldn't find a good working 2.7.5 Docker image since 2.7.5 is very old now and has many known security vulnerabilities.

One option would be to use Ubuntu or similar base image and install Python 2.7.5 on it as part of the Circle CI job step, but I'm not sure if it's worth adding another builder to our CI pipeline just because of that (we already have many now because of the new benchmark + packaging ones).

@czerwingithub WDYT?

PR #424, opened 2020-02-24
Add memory profiler

Kami:
This pull request adds memory profiler which is based on the pympler library (https://pythonhosted.org/Pympler/).

I followed the existing Profiler abstraction and made it "pluggable" so we now have two types of profilers - CPU profiler and Memory profiler.

To avoid introducing more configuration options (we already have many), I decided to use existing configuration options which are already used for the CPU profiler (enable_profiling, profile_duration_minutes, max_profile_interval_minutes, etc.).

In the future, if we want more flexibility (e.g. different profiling duration and interval for CPU and memory profiler) we can introduce additional configuration options for the memory profiler.

Memory profiling data is written to ``<data dir>/agent.meminfo`` file. Here is the example output - https://gist.github.com/Kami/f746f7fdad98842e3c721f9493e142ca.

 Right now that file is just written to disk and it's not yet consumed by the agent. Initially we can probably just consume it raw / as is without writing a custom parser for it since the output is mostly line oriented anyway.

### DIFF
--- a/scalyr_agent/profiler.py
+++ b/scalyr_agent/profiler.py
@@ -14,16 +14,26 @@
 # ------------------------------------------------------------------------
 # author:  Imron Alston <imron@scalyr.com>
 
+"""
+Module which contains classes and abstractions for CPU and memory profiling.
+"""
+
 from __future__ import unicode_literals
 from __future__ import absolute_import
 
 __author__ = "imron@scalyr.com"
+
+if False:
+    from typing import Optional
+    from typing import Dict
+    from typing import Any
 
 import os
 import random
 import time
 import traceback
 from io import open
+from threading import Thread
 
 import six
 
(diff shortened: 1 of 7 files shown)


imron:
It seems with this, that if profiling is enabled both memory and CPU profiling are always done at the same time and there's no way to only turn on CPU profiling or only turn on memory profiling?

Both of them have random start times, so perhaps there won't be any overlap in the profiling, but if there is, I wonder if there will be any interference, and what effect (if any) the memory profiling will have on CPU usage, and what effect (if any) the CPU profiling will have on memory usage.

imron:
I was thinking some more about this, and one way to do it without introducing too many more options would just have a single option `mode` (or some other better name) that could be `cpu` or `memory`, and defaults to `cpu` (to replicate existing behaviour).

Then in the `__init__` method of the `ScalyrProfiler` it would just create a single profiler based on the value of the `mode` option.

If people want cpu profiling they set `mode` to `cpu`, if they want memory profiling, they set it to `memory`, and if they want both, they have one monitor set to `cpu` and a second monitor set to `memory` (the agent is designed to allow multiple monitors of the same type).

Thoughts?  (cc: @czerwingithub )

Kami:
@imron 

> It seems with this, that if profiling is enabled both memory and CPU profiling are always done at the same time and there's no way to only turn on CPU profiling or only turn on memory profiling?

That's correct (I wanted to avoid too many configuration options and additional complexity which goes along with that)

> Both of them have random start times, so perhaps there won't be any overlap in the profiling, but if there is, I wonder if there will be any interference, and what effect (if any) the memory profiling will have on CPU usage, and what effect (if any) the CPU profiling will have on memory usage.

Yes, it's not ideal, but both profilers run in a thread so there shouldn't be much overhead and "observer effect" (that's also what I observed during testing it), but I think adding additional configuration option, as you suggested, and allowing user to decide which profiler to run should work fine as well.

I'll wait on @czerwingithub take before making any code changes.

czerwingithub:
@Kami @imron Let's just have the one mode for now (CPU and memory) and we can add more later if we see there's an impact.  Tomaz says he hasn't seen any interference in his tests.

scalyr_agent/profiler.py
```diff
@@ -106,86 +236,215 @@ def _start(self, config, current_time):
...
 
-    def update(self, config, current_time=None):
+        global_log.log(
+            scalyr_logging.DEBUG_LEVEL_0,
+            "CPU profiling data written to %s",
+            self._data_file_path,
+        )
+
+
+class PeriodicMemorySummaryCaptureThread(Thread):
```
czerwingithub: You might want to use our `StoppableThread` here.  The advantage is that it has a sleep method that will exit early if you stop the thread.  Down below, this thread could take up to 5 seconds to stop.

It's not clear to me if we have folks that are blocking on that stop or not.  For example, if the agent is doing a soft-restart due to a configuration change, do we tell any of the profiling code to stop before we loading the new configuration?

Kami:
This reminds me - there will be a small change of behavior if we go with a single profiler (which profiles both CPU and memory) approach (aka current implementation in this PR).

It will require user to install 2 3rd party libraries - ``yappi`` and ``pympler``. Not sure how big of an issue that is though - both libraries support Python 2 and Python 3 and pympler is pure Python (unlike yappi) so there are not additional libraries or headers needed.

If others are OK with that, we can go with that we have in this PR at this point, otherwise I can make ``profiler_mode`` config change (which should be trivial).

scalyr_agent/profiler.py
```diff
@@ -106,86 +236,215 @@ def _start(self, config, current_time):
...
+                    "Performing periodic memory usage capture",
+                )
+                self._capture_snapshot()
+                last_capture_time = int(time.time())
+
+            time.sleep(self._sleep_interval)
+
+    def stop(self):
+        # type: () -> None
+        self._running = False
```
czerwingithub: Just to capture this... you aren't protected `_running` even though it will be accessed via different threads because `_running` is always updated atomically and there is no requirements of setting multiple variables together in a protected way?

scalyr_agent/profiler.py
Kami: Ah, cool, I missed the existing ``StoppableThread`` abstraction.

I will update the code to use it. I don't think we will save any lines of code, but it's still a good idea to use the existing abstraction which pretty much does the same thing.

Kami: Pushed a change to use existing StoppableThread abstraction - 06fd81d.

Kami:
I pushed another small change so we try to exclude memory profiler data itself from the profiler output to reduce the observer effect.

PR #429, opened 2020-02-24
Fix six import in config_main.py

Kami:
This pull request fixes an issue with system wide six library being used instead of the one we bundle with the package.

The reason for that was that we imported six before calling ``scalyr_init()`` which set ups ``PYTHONPATH`` and adds ``third_party/`` directory for it.

---

Catching such issues with tests is a bit hard since we would need to cover many different environments and it still wouldn't cover everything.

To avoid such issues in the future, I would recommend writing a lint / check script which checks all the Python files and fails with an error if any of the bundled library is imported anywhere before ``scalyr_init()`` is called.

I think it would be sufficient to just check the file header aka the top global part with imports.

### DIFF
--- a/scalyr_agent/config_main.py
+++ b/scalyr_agent/config_main.py
@@ -37,7 +37,6 @@ import tempfile
 import traceback
 from io import open
 
-import six
 from distutils import spawn
 from optparse import OptionParser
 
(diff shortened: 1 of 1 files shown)


czerwingithub:
I merged this myself since I am hoping I can get a chance to build a new release candidate this afternoon.

PR #430, opened 2020-02-24
pylint plugin for checking / enforcing ordering of bundled third party module imports

Kami:
This pull request adds a pylint plugin which checks for the issues like the one fixed in #429.

It works by inspecting any file which calls ``scalyr_init()`` and ensures any import of a bundled module comes after that line.

Right now it's limited to the modules which call ``scalyr_init()`` directly and doesn't support indirect imports (e.g. module A calls scalyr_init() and imports module B and module B imports a bundled import).

This means it won't catch 100% of the edge cases, but it's better than nothing (implementing indirect import tracking would be much more involved).

Here is the error message pylint / build would fail with if we ran it on commit before #429 was merged:

```bash
************* Module scalyr_agent.config_main
scalyr_agent/config_main.py:40:0: E8002: Import "six" is on line 40, but it should come after scalyr_init() function call which is on line 57. (bundled-import-after-scalyr-init)
```

### DIFF
--- a/pylint_plugins/bundled_imports_checker.py
+++ b/pylint_plugins/bundled_imports_checker.py
@@ -29,9 +29,11 @@ from __future__ import print_function
 
 if False:
     from typing import List
+    from typing import Tuple
     from typing import Dict
     from typing import Set
     from typing import Any
+    from typing import Optional
 
 import os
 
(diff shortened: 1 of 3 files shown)


pylint_plugins/bundled_imports_checker.py
```diff
@@ -0,0 +1,229 @@
...
+        if os.path.isdir(file_path) and os.path.isfile(init_file_path):
+            # Package
+            return (True, file_name)
+
+        return (False, None)
+
+    def is_python_module(directory_path, file_path):
+        # type: (str, str) -> Tuple[bool, Optional[str]]
+        """
+        Return package name if the provided file path is a Python package, None otherwise.
```
czerwingithub: The comment is the same as the method above.

Kami: Pushed a fix - d0ae301cdd0fed2b9324a9e98ea3853808e53b3a.

PR #434, opened 2020-02-26
AGENT-342: Add python check in rpm/deb package pre-install script.

ArthurKamalov:
Include additional logic to preinstall script in `deb` and `rpm` packages. Before installation, the script searches for python by `/usr/bin/env python` command. If a python interpreter is found, the script also checks its version. If any of those checks fail, the script exits with error code and error message.

Selection_027
Selection_028
Selection_029

### DIFF
--- a/build_package.py
+++ b/build_package.py
@@ -736,7 +736,6 @@ def build_rpm_or_deb_package(is_rpm, variant, version):
         "  --maintainer czerwin@scalyr.com "
         "  --provides scalyr-agent-2 "
         '  --description "%s" '
-        '  --depends "python >= 2.4" '
         '  --depends "bash >= 3.2" '
         "  --url https://www.scalyr.com "
         "  --deb-user root "
@@ -1233,6 +1232,31 @@ def create_scriptlets():
     fp = open("preinstall.sh", "w")
     fp.write(
         r"""#!/bin/bash
+
+echo "Check for python."
+# look for 'python' by '/usr/bin/env', exit with error if it is not found.
+/usr/bin/env python --version >/dev/null 2>&1 || \
+{ echo -e >&2 "\e[31mThe suitable python interpreter is not found. Aborting.\e[0m"; \
+echo -e >&2 "The command - '/usr/bin/env python' has to point to the python interpreter with version \e[31m>=2.6 or >=3.5\e[0m."; exit 1; }
+
+# get first two digits from python version.
+version_string=$(/usr/bin/env python --version 2>&1 | grep -Po "\d.\d")
+
+echo "Check python version."
+
+if [[ "$version_string" < "2.6" ]]; then
+        echo -e >&2 "\e[31mThe python interpreter with version '>=2.6 or >=3.5' is required. Current version: ${version_string}.\e[0m"
+        exit 1
+fi
+
+if [[ "$version_string" > "3.0" ]]; then
+    if [[ "$version_string" < "3.5" ]]; then
+        echo -e >&2 "\e[31mThe python interpreter with version '>=2.6 or >=3.5' is required. Current version: ${version_string}.\e[0m"
+        exit 1
+    fi
+fi
+
+echo -e "\e[36mPython interpreter found. Version: ${version_string}\e[0m"
 
 # Always remove the .pyc files and __pycache__ directories.  This covers problems for old packages that didn't have the remove in the
 # preuninstall.sh script.
--- a/build_package.py
+++ b/build_package.py
@@ -1245,18 +1245,18 @@ version_string=$(/usr/bin/env python --version 2>&1 | grep -Po "\d.\d")
 echo "Check python version."
 
 if [[ "$version_string" < "2.6" ]]; then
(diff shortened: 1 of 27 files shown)


build_package.py
```diff
@@ -1234,6 +1233,31 @@ def create_scriptlets():
     fp.write(
         r"""#!/bin/bash
```
Kami: Let's add a docstring on why we need to do that (for compatibility with RHEL 8.1 and other distros where package name is python2 until we have per distro packages).

build_package.py
```diff
@@ -1234,6 +1233,31 @@ def create_scriptlets():
...
 
+echo "Check for python."
+# look for 'python' by '/usr/bin/env', exit with error if it is not found.
+/usr/bin/env python --version >/dev/null 2>&1 || \
```
Kami: I believe I had this comment on another PR or Slack - I would prefer to store all those scripts inside files (e.g. ``installer/scripts/{preinstall.sh,preuninstall.sh}`` and re-use them here instead of storing the script content inline.

Imo, this makes it more organized, easier to test in a standalone fashion and also gives us flexibility to re-use those scripts if we ever move away from build_package.py and fpm tp some other mechanism

build_package.py
```diff
@@ -1234,6 +1233,31 @@ def create_scriptlets():
...
+{ echo -e >&2 "\e[31mThe suitable python interpreter is not found. Aborting.\e[0m"; \
+echo -e >&2 "The command - '/usr/bin/env python' has to point to the python interpreter with version \e[31m>=2.6 or >=3.5\e[0m."; exit 1; }
+
+# get first two digits from python version.
+version_string=$(/usr/bin/env python --version 2>&1 | grep -Po "\d.\d")
```
Kami: Since this needs to run on multiple distros, ideally we would be on the safe side and use grep notation which is available in both - gnu grep and bsd grep.

Kami:
Overall, this approach looks good to me :+1:

build_package.py
```diff
@@ -1234,6 +1233,31 @@ def create_scriptlets():
     fp.write(
         r"""#!/bin/bash
 
+echo "Check for python."
+# look for 'python' by '/usr/bin/env', exit with error if it is not found.
+/usr/bin/env python --version >/dev/null 2>&1 || \
+{ echo -e >&2 "\e[31mThe suitable python interpreter is not found. Aborting.\e[0m"; \
+echo -e >&2 "The command - '/usr/bin/env python' has to point to the python interpreter with version \e[31m>=2.6 or >=3.5\e[0m."; exit 1; }
```
Kami: Perhaps we could provide a bit more user-friendly error message here and instruct user to install ``python`` / ``python2`` package if that binary is not available.

We don't need to go overboard with the distro detection, just a simple detection if it's Debian or RHEL based system and display either apt-get or yum command show work.

build_package.py
ArthurKamalov: Done. No Perl patterns any more :)

build_package.py
ArthurKamalov: Yes, I agree, and I think there is more that can be changed in `build_package.py` script.

installer/scripts/postinstall.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
Kami: Let's please just add ASF license header here.

installer/scripts/preinstall.sh
```diff
@@ -0,0 +1,55 @@
...
+
+
+echo "Check for python."
+# if command was not successful, print hint message and exit.
+if ! /usr/bin/env python --version ; then
+    echo -e "\e[31mCommand 'python' is not found.\e[0m"
+    echo "You can install it by running command:"
+    # get 'ID_LIKE' and 'ID' fields from '/etc/os-release' file and then search for distributions key words.
+    if [[ -f "/etc/os-release" ]]; then
+        found_distros=$(grep -qE  "^ID_LIKE=|^ID=" /etc/os-release)
```
Kami: Hm, are you sure ``-q`` flag needs to be there? Since if it is, it will run in quite mode and won't output anything.

```bash
$ grep -qE  "^ID_LIKE=|^ID=" /etc/os-release
$ grep -E  "^ID_LIKE=|^ID=" /etc/os-release
ID=fedora
```

Kami: On that note, let's please add some test cases for that in the future once other smoke tests PR is merged.

ArthurKamalov: Yes, my bad, thanks for noticing that.

ArthurKamalov: > On that note, let's please add some test cases for that in the future once other smoke tests PR is merged.

Sure thing.

Kami: And as discussed before, I'm fine with just mocking ``/etc/os-release`` file in those tests.

Eventually we do want tests to run on more of the supported distros, but this would be too big of a task for now.

Kami:
I will merge this so I can finish #439.

PR #436, opened 2020-02-26
Fix string issue when building Windows package

czerwingithub:
Fixed issue related to 2->3 strings and unicode.  We were not
properly using a binary string when concat'ing it with a UUID.

### DIFF
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -390,11 +390,7 @@ def create_uuid3(namespace, name):
     :return:
     :rtype: uuid.UUID
     """
-    if six.PY2:
-        return uuid.uuid3(namespace, six.ensure_binary(name))
-    else:
-        # noinspection PyTypeChecker
-        return uuid.uuid3(namespace, six.ensure_text(name))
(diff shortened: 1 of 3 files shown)


scalyr_agent/util.py
```diff
@@ -380,6 +380,23 @@ def create_unique_id():
...
+    :type namespace: uuid.UUID
+    :type name: six.text
+    :return:
+    :rtype: uuid.UUID
+    """
+    if six.PY2:
+        return uuid.uuid3(namespace, six.ensure_binary(name))
+    else:
+        # noinspection PyTypeChecker
+        return uuid.uuid3(namespace, six.ensure_text(name))
```
ArthurKamalov: I think we can avoid using `if` statement by using `six.ensure_str` https://six.readthedocs.io/#six.ensure_str

```suggestion
    return uuid.uuid3(namespace, six.ensure_str(name))
```

Kami: Yep what Arthur said.

Besides that, LGTM.

PR #437, opened 2020-02-27
Use production server for agent benchmark runs, make "server_attributes" config option environment aware

Kami:
This pull request includes a couple of related changes:

1. Update agent config which is used with benchmarks to utilize production server instead of the qa testing one.
2. Make ``server_attributes`` config option environment aware. This allows us to dynamically set this value instead of relying on config rewrites (which is what smoketests currently do)
3. Update benchmark Circle CI jobs to use unique ``server_attributes.serverHost`` for each run to make troubleshooting / debugging easier.

## TODO

- [x] ~~Update secret Circle CI environment variable so it uses prod key for account cloudtech-builds~~ - it appears that was already done by @czerwingithub 
- [x] Verify server change + scalyr_key work correctly

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,6 @@ master_only: &master_only
     branches:
       only:
         - master
-        - benchmarks_use_prod
 
 version: 2.1
 commands:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,6 +96,10 @@ commands:
         description: "Optional bash command / script to run after starting the agent and the metrics capture script."
         type: string
         default: ""
+      agent_server_host:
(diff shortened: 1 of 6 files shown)


Kami:
@czerwingithub I think it's probably also a good idea to set retention policy for ``cloudtech-builds`` account.

We don't generate tons of data since benchmarks don't run on each push, but just on master branch push, but they still exercise some pathological use cases (uncompressable log data) so it's probably still better to have some retention policy set (e.g. 1 week).

Kami:
There was what appears to be a bug / edge case with handling empty strings for JsonObjects which I fixed in b220d8de19e08f43792202b17c2ecf59c113b203.

Now the behavior is consistent with other complex types such as JsonArray.

PR #435, opened 2020-02-26
Fix memory leak in the Configuration.parse() method (aka Configuration object)

Kami:
This pull request fixes memory leak in the ``Configuration.parse()`` method.

---

I was tracking down memory leak in the agent and with various memory and GC instrumentation I got pointers to the agent main loop.

Each time the agent main loop ran and config was reloaded / re-read from disk, number of unreachable objects which can't be collected increased.

I confirmed that's indeed the root cause by adding ``continue`` statement to the main loop before ``self.__verify_config`` method call. With that continue statement in place, number of uncollectable objects stayed constant and didn't increase after each agent main loop run.

After diving in further, I found a memory leak inside ``Configuration.parse()`` method, more specifically, ``__perform_substitutions`` method which is called by ``parse()``.

Those inner functions defined inside ``__perform_substitution`` closure hold reference to the JsonObject config variable and vice versa so they were never released and garage collected.

Here is a simple script I used to verify verify the memory leak and then verify the fix - https://gist.github.com/Kami/bc3c89c4d5d866bfc4d64dd8835d7212

Before the fix:

```bash
$ SCALYR_AGEN_KEY=a python test_memory_leak.py 
Iteration: 1
Unreachable objects: 26

Iteration: 2
Unreachable objects: 52

Iteration: 3
Unreachable objects: 78

Iteration: 4
Unreachable objects: 104

Iteration: 5
Unreachable objects: 130

Iteration: 6
Unreachable objects: 156

Iteration: 7
Unreachable objects: 182

Iteration: 8
Unreachable objects: 208

Iteration: 9
Unreachable objects: 234

Iteration: 10
Unreachable objects: 260
```

To avoid the very large code block, here is the whole output with listed unreachable objects - https://gist.github.com/Kami/6fa4cb07f19a21ff7fde79a5e1d0e06b

After the fix:

```bash
$ SCALYR_AGEN_KEY=a python test_memory_leak.py 
Iteration: 1
Unreachable objects: 0
Unreachable objects:

[]

Iteration: 2
Unreachable objects: 0
Unreachable objects:

[]

Iteration: 3
Unreachable objects: 0
Unreachable objects:

[]

Iteration: 4
Unreachable objects: 0
Unreachable objects:

[]

Iteration: 5
Unreachable objects: 0
Unreachable objects:

[]

Iteration: 6
Unreachable objects: 0
Unreachable objects:

[]

Iteration: 7
Unreachable objects: 0
Unreachable objects:

[]

Iteration: 8
[...]

PR #438, opened 2020-02-27
Add (opt-in) Garbage Collection related metrics to the agent status output

Kami:
This pull request adds new ``gc_stats.garbage`` metric to the agent status output (right now only to the machine readable aka json format).

This value contains a number of uncollectabe objects after a GC run (those are the objects which are not reachable anymore and can't be collected - due to a cyclic reference to itself or similar).

This will allow us to more easily detect regressions and catch memory leaks like 3 big ones which have been fixed in #435 

Do keep in mind though that this only reports uncollectable objects which are unreachable.

If some piece of code which is in use (aka reachable) incorrectly keeps a reference to some other object or similar, this won't be reported in that number.

---

The plan is to hook this metric to CodeSpeed to allow us to catch such memory leak regressions in the future more easily.

I will add new codespeed benchmark which runs agent under load with gc_stats enabled and then report that value to CodeSpeed at the end of the run.

Do keep in mind that I need to use a separate agent process run since gc stats includes overhead which would otherwise have an impact on metrics on other benchmark results.

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -541,6 +541,10 @@ class Configuration(object):
         return self.__get_config().get_bool("disable_leak_update_debug_log_level")
 
     @property
+    def enable_gc_stats(self):
+        return self.__get_config().get_bool("enable_gc_stats")
+
+    @property
     def disable_all_config_updates(self):
         return self.__get_config().get_int(
             "disable_leak_all_config_updates", none_if_missing=True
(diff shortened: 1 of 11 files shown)


scalyr_agent/configuration.py
```diff
@@ -540,6 +540,10 @@ def disable_bandwidth_stats(self):
     def disable_update_debug_log_level(self):
         return self.__get_config().get_bool("disable_leak_update_debug_log_level")
 
+    @property
+    def enable_gc_stats(self):
+        return self.__get_config().get_bool("enable_gc_stats")
```
czerwingithub: Can you add an entry to the test cases in `configuration_test.py` for each top-level configuration option.  We typically add one to `TestConfiguration.test_basic_case` to verify the default is set right.. and then another to  `TestConfiguration.test_overriding_basic_settings` to verify we can override.  I just like to have a double-check on the config names and their default values since once customers use them, it's hard to change them.

Kami: Yep, I plan to add some tests and hook new benchmark job to Circle CI :)

PR #439, opened 2020-02-28
Add lint check for shell scripts

Kami:
This pull request adds lint check for various shell scripts in this repo. It utilizes ShellCheck tool (https://www.shellcheck.net/).

Right now I set severity to error because warning severity reports a bunch of issues in scripts under ``./.circleci/`` (those can be addressed later and separately).

It also addresses various lint issues on the existing shell scripts in this repo.

## TODO

- [x] Add this check to pre-commit hook


Kami:
I also added this check to pre-commit hook, but I made it non-fatal if ``shellcheck`` binary is not found locally.

PR #442, opened 2020-03-03
Add longer running agent benchmarks

Kami:
This pull request updates a list of our CodeSpeed benchmarks and adds another "under load" benchmark which runs the agent under Python 2.7.

This benchmark runs longer than the other loaded ones, so it may give us some insight on the memory usage behavior when running agent under continuous load for a longer period of time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,7 @@ master_only: &master_only
     branches:
       only:
         - master
+        - longer_running_benchmark
 
 version: 2.1
 commands:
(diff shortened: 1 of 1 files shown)


Kami:
On a related note - based on the charts, enabling GC stats doesn't add any significant overhead to the metrics we care about (memory usage, etc.) so perhaps we can just combine GC run with regular one to avoid one extra job / builder.

czerwingithub:
I'm all for reducing the number of builds.

Kami:
I pushed a change to combine the jobs - 8a44fe3.

PR #443, opened 2020-03-03
Add basic tests for the Profiler abstraction

Kami:
This pull request adds some basic mocking based tests for the ScalyrProfiler abstraction.

Keep in mind that currently those tests are very basic and don't verify actual functionality and output format of the profilers.

At this point they only verify that the calling the methods which are part of the base API (update, start, stop) doesn't result in an exception (they would catch the issue such as the one fixed here #440).

### DIFF
--- a/scalyr_agent/tests/test_profilers.py
+++ b/scalyr_agent/tests/test_profilers.py
@@ -179,7 +179,4 @@ def is_file_path_empty(file_path):
     """
     Return true if the provided file path is empty.
     """
-    with open(file_path, "r") as fp:
-        content = fp.read()
-
-    return len(content) == 0
+    return os.path.getsize(file_path) == 0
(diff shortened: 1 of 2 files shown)


scalyr_agent/tests/test_profilers.py
```diff
@@ -0,0 +1,185 @@
...
+        )
+        self.assertFalse(is_file_path_empty(data_file_path))
+
+
+def is_file_path_empty(file_path):
+    # type: (str) -> bool
+    """
+    Return true if the provided file path is empty.
+    """
+    with open(file_path, "r") as fp:
```
czerwingithub: Why not use `os.path.getsize`?

Kami: In the future, I also wanted to assert on the content, but in the mean time using ``os.path.getsize`` should work as well.

Will push that change.

Kami: 6377cd2

PR #444, opened 2020-03-03
Make sure we skip "cert files exist" checks when running in a dev installation

Kami:
This pull request should fix the issue @yanscalyr reported yesterday.

---

I updated the code so under dev installation we skip "cert files exists" checks.

Those bundle certificate files are created as part of the build process inside ``build_package.py`` so they will only be available in package installations.

In addition to that, I updated the code so we also skip "intermediate certs file exist" check on Windows installations since we don't include intermediate certs file there.

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -45,9 +45,17 @@ import time
 from io import open
 
 try:
-    from __scalyr__ import SCALYR_VERSION, scalyr_init
+    from __scalyr__ import SCALYR_VERSION
+    from __scalyr__ import scalyr_init
+    from __scalyr__ import INSTALL_TYPE
+    from __scalyr__ import DEV_INSTALL
+    from __scalyr__ import MSI_INSTALL
 except ImportError:
-    from scalyr_agent.__scalyr__ import SCALYR_VERSION, scalyr_init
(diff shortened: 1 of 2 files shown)


Kami:
I would feel more comfortable if we also had some end to end package tests for Windows.

Hopefully we can add some of those in the future once #401 is merged in.

scalyr_agent/agent_main.py
```diff
@@ -1257,18 +1265,25 @@ def __create_client(self, quiet=False):
...
             # exist, throw and fail early and loudly
-            if not os.path.isfile(ca_file):
+            if not is_dev_install and not os.path.isfile(ca_file):
                 raise ValueError(
                     'Invalid path "%s" specified for the "ca_cert_path" config '
                     "option: file does not exist" % (ca_file)
                 )
 
-            if not os.path.isfile(intermediate_certs_file):
+            # NOTE: We don't include intermediate certs in the Windows binary so we skip taht chek
```
czerwingithub: s/taht/that/

Kami: I wish there was a "steven spell check" plugin available for vim, it seems to be more effective than built in vim spellcheck plugin :D

PR #446, opened 2020-03-05
Run lint checks on tcollector directory

Kami:
In #445, I noticed we don't run lint checks on tcollectors directory since it's located inside ``third_party/`` directory which contains other dependencies we want to ignore.

In ideal world, we would leave any third party dependencies as is and wouldn't touch it (so we can update it from upstream in the future, etc.), but sadly we don't have much choice here - we already made some functionality and other changes and fixes to tcollectors package (ideally, we should also try to get some of those accepted upstream).

### DIFF
--- a/scalyr_agent/builtin_monitors/tests/docker_monitor_test.py
+++ b/scalyr_agent/builtin_monitors/tests/docker_monitor_test.py
@@ -23,7 +23,6 @@ __author__ = "echee@scalyr.com"
 
 import os
 import threading
-import time
 from io import open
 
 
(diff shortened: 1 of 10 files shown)


ArthurKamalov:
>In ideal world, we would leave any third party dependencies as is and wouldn't touch it (so we can update it from upstream in the future, etc.), but sadly we don't have much choice here - we already made some functionality and other changes and fixes to tcollectors package (ideally, we should also try to get some of those accepted upstream).

Yeah, that's why I did not touch it, I thought that we might need to upgrade it in future, now I don't think that we are going to do that:)

Kami:
@ArthurKamalov Yeah, I agree.

Ideally we shouldn't touch any bundled files, but sadly we are where we are at this point :)

PR #447, opened 2020-03-05
agent essential directories creation on DEV_INSTALL.

ArthurKamalov:
If the agent runs from source(`DEV_INSTALL`), it fails in case `log` and `data` directories do not exist.
The small fix makes the agent create those directories before the start.

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -1345,26 +1345,25 @@ class ScalyrAgent(object):
 
         if self.__controller.install_type == DEV_INSTALL:
             # The agent is running from source, make sure that its directories exist.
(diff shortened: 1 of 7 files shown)


scalyr_agent/agent_main.py
```diff
@@ -1343,19 +1343,27 @@ def __verify_can_write_to_logs_and_data(self, config):
         @type config: Configuration
         """
         if not os.path.isdir(config.agent_log_path):
-            raise Exception(
-                "The agent log directory '%s' does not exist." % config.agent_log_path
-            )
+            if self.__controller.install_type != DEV_INSTALL:
```
Kami: This change seems reasonable to me, but to be on the safe side, I would invert the ``if`` condition and call ``os.makedirs`` if it's a dev install.

PR #449, opened 2020-03-06
Test race workaround

Kami:
This pull request updates the test which occasionally fails on Circle CI to use longer max wait.

This should make test failure due to test timing less likely to occur.


Kami:
Re-run uncovered another possible race - https://app.circleci.com/jobs/github/scalyr/scalyr-agent-2/23528.

Bumped another maxwait in the test, should hopefully make the race less likely.

PR #451, opened 2020-03-09
Benchmark k8s

ArthurKamalov:
Created new kubernetes object with kind `Job`  which is located in `benchmarks/k8s/base/codespeed.yaml`.

The Job should be the best choice because the metrics capturing is not a continuous process, and better to consider it as a "task"

Each job spawns pod with 3 containers inside:
first - agent container
second - a container that captures metrics
third - container that writes logs.

Before creating a job, you should create a directory with an overlay where a particular git commit is specified. As an example, you can see such overlays with different release tags.


benchmarks/k8s/docker-images/Dockerfile.agent-2.7
```diff
@@ -0,0 +1,5 @@
+FROM python:2.7
```
Kami: To be on the safe side / for consistency, we should probably also specify patch version, aka ``2.7.17`` or similar.

benchmarks/scripts/start-agent.sh
```diff
@@ -0,0 +1,38 @@
...
+  cd "${AGENT_SOURCE_PATH}"
+  echo "Checkout to target commit."
+  git checkout "${COMMIT_SHA1}"
+else
+  AGENT_SOURCE_PATH=$(realpath "${SCRIPT_DIR}/../..")
+fi
+
+cp /config/config.json /config.json
+
+sed -i "s/{API_KEY}/\"${SCALYR_API_KEY}\"/g" /config.json
```
Kami: For Circle CI benchmark jobs we utilize environment variables, maybe we should also utilize those here?

benchmarks/scripts/send_usage_data_to_codespeed.py
```diff
@@ -299,25 +330,40 @@ def main(
...
+            current_time = time.time()
+            logger.debug("Capturing gauge metrics...")
+            captured = capture_metrics(
+                tracker=tracker,
+                process=process,
+                metrics=METRICS_GAUGES,
+                capture_agent_status_metrics=capture_agent_status_metrics,
+            )
+            # write captured metrics to backup file.
+            backup.write("{0}\n".format(json.dumps(captured)))
```
Kami: Perhaps we should make this writing metrics to a file optional?

This way we can avoid additional interference / disk writes on non-kubernetes Circle CI benchmarks where everything runs inside a single container.

benchmarks/k8s/base/codespeed.yaml
```diff
@@ -0,0 +1,151 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: agent-k8s-codespeed
+spec:
+  template:
+    spec:
+      shareProcessNamespace: true
```
Kami: Would be good to add a comment here on why we need it - because capture script needs access to /proc/<pid> for the agent process...

benchmarks/scripts/write-log-lines.py
```diff
@@ -0,0 +1,47 @@
...
+    pid = wait_for_agent_start_and_get_pid(pid_file_path)
+
+    log_file = os.path.join(agent_data_path, "log", "test_log1.log")
+    log_file2 = os.path.join(agent_data_path, "log", "test_log2.log")
+    log_file3 = os.path.join(agent_data_path, "log", "test_log3.log")
+    log_file4 = os.path.join(agent_data_path, "log", "test_log4.log")
+
+    threads = list()
+    for file, interval in [(log_file, 0.5), (log_file2, 1), (log_file3, 2)]:
+        thread = threading.Thread(target=sender, args=(file, interval))
```
Kami: That works for now.

Eventually, if needed, we can make this script a bit more customizable (how many lines to writte, etc.) and if we want to introduce more load, we can utilize something like flog - https://github.com/mingrammer/flog

benchmarks/scripts/utils.py
```diff
@@ -198,3 +202,38 @@ def parse_commit_date(value):
...
+    return timedelta(**{time_unit: int(number)}).total_seconds()
+
+
+def wait_for_agent_start_and_get_pid(pid_file_path):
+    # type: (str) -> int
+    """IF file 'pid_file_path' exists, we assume that the agent is started. Returns 'pid' of the agent process."""
+    attempts = 0
+    while not os.path.exists(pid_file_path):
+        if attempts > 10:
+            raise RuntimeError("Agent pid file did not appear.")
```
Kami: Perhaps we should also try to cat and include the last 30 lines of the agent.log if available here?

This should help with troubleshooting if there is a typo in a config or similar.

(That's not needed for Circle CI because there we preserve log files as build artifacts, but in may help In Kubernetes and other scenarios)

benchmarks/scripts/send_usage_data_to_codespeed.py
Kami: EDIT: For simplicity I'm fine with keeping it as is (aka always write metrics to a file).

Writing every 10 seconds or so shouldn't add much overhead :)

Kami:
As discussed the other day - can we also set up one agent pod which is long running and never killed (this one should just submit data to scalyr cloudtech-builds account and not send anything to CodeSpeed).

benchmarks/scripts/start-agent.sh
Kami: EDIT: To clarify it - this will only work if we test it with latest master, but not with older versions since you can't specify server_host using environment variable in older versions.

benchmarks/scripts/send_usage_data_to_codespeed.py
ArthurKamalov: I refactored it a little bit and made file write optional. Hope it is easier to understand:)

benchmarks/scripts/start-agent.sh
ArthurKamalov: >EDIT: To clarify it - this will only work if we test it with latest master, but not with older versions since you can't specify server_host using environment variable in older versions.

Yep, I faced the same problem. I think we don't have another option if we want run older versions.

benchmarks/k8s/docker-images/Dockerfile.agent-2.7
ArthurKamalov: Done.

benchmarks/k8s/base/agent-configurations/idle-no-monitors/agent-config-additional-fields
```diff
@@ -0,0 +1 @@
+"implicit_metric_monitor": false, "implicit_agent_process_metrics_monitor": false,
```
Kami: This works for now, although I would prefer us to use config arguments and agent.d directory (aka the same we do for CodeSpeed benchmarks).

Any particular reason why we didn't do this here? Perhaps because we would need to share more files with the Docker container or similar?

PR #453, opened 2020-03-10
Utilize shell and URL monitors in the loaded benchmark #2

Kami:
This pull request makes a small change to the loaded benchmark #2 to also utilize URL and shell monitors with more frequent sample interval.

This way the setup closer resembles our staging setup.

As discussed on slack (/cc @ArthurKamalov), we should also add the same (but longer running) benchmark to Kubernetes. 

This way we will cover both scenarios - short and longer running ones.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,6 @@ master_only: &master_only
     branches:
       only:
         - master
-        - update_benchmarks
 
 version: 2.1
 commands:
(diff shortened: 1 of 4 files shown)


Kami:
This will also help us uncover some Python 3 issues in monitors like the one fixed here - #454.

Screenshot from 2020-03-10 19-17-43

Ideally we would have a benchmark which runs as much monitors as possible. Benchmarks are not replacement for good automated tests, but they may help us identify such Python 3 issues in monitors until we have more testing in place.

PR #454, opened 2020-03-10
Fix shell and URL monitor - make sure we are always working with unicode

Kami:
Small fix to the shell and URL monitor to make sure they work under Python 3.

As discussed before, there are likely more issues like that in the monitors and that's we need some automated tests as soon as possible.


scalyr_agent/builtin_monitors/shell_monitor.py
```diff
@@ -139,6 +141,8 @@ def gather_sample(self):
             output += "\n"
         output += stdout_text
 
+        output = six.ensure_text(output)
```
Kami: I believe it makes more sense for us to convert output from bytes to unicode instead of doing the opposite.

All the config options which are specified by the end user (think extract pattern, etc.) will always be string / unicode and not bytes type so it makes sense we work with that type in those places.

czerwingithub: Agreed.

Kami:
I've added some very basic tests for those two monitors.

It's nothing fancy or extensive, but still better than current status quo (aka no unit tests).

In URLMonitor I rely on actual request to scalyr.com, which is not ideal for unit tests, but we already have a bunch of other integration / end to end tests which rely on access to scalyr.com so I think it's not a big deal (especially since relying on hitting some local mock server would be much more involved and not something I have the time for at this point).

Issue #410, opened 2020-02-14
DaemonSet is not updated for the new version


aksakalli:
[scalyr-agent-2-envfrom.yaml](https://github.com/scalyr/scalyr-agent-2/blob/master/k8s/scalyr-agent-2-envfrom.yaml) is using `extensions/v1beta1` and the default update strategy is `OnDelete` (not `RollingUpdate`). When I run

```
kubectl apply -f https://raw.githubusercontent.com/scalyr/scalyr-agent-2/release/k8s/scalyr-agent-2-envfrom.yaml
```

It is not upgrading the image version. My suggestion is to use `apps/v1` or include 

```
  updateStrategy:
    type: RollingUpdate
```
explicitly or even better create a _well maintained_ helm chart. There was [a PR](https://github.com/helm/charts/pull/1066) but never merged.

## Notes

* [apps/v1](https://github.com/kubernetes/kubernetes/blob/v1.16.7/pkg/apis/apps/v1/defaults.go#L75)
* [extensions/v1beta1](https://github.com/kubernetes/kubernetes/blob/v1.16.7/pkg/apis/extensions/v1beta1/defaults.go#L49)

aksakalli:
solved with the merged PR #455.

PR #455, opened 2020-03-11
K8s improvements

yanscalyr:
https://www.notion.so/scalyr/K8s-Agent-improvements-8b588a7647f64ff09482de9b0659387f

Changes to instructions that would be required:
* Change namespace in the configmap example
* Replace service account and daemonset steps with `kubectl apply -k https://raw.githubusercontent.com/scalyr/scalyr-agent-2/release/k8s`
* Replace upgrading steps with `kubectl apply -k https://raw.githubusercontent.com/scalyr/scalyr-agent-2/release/k8s`
* Add mention of the namespace change for existing deployments, for those wishing to switch they would need to run `kubectl delete -f https://raw.githubusercontent.com/scalyr/scalyr-agent-2/release/k8s.old` first, then follow new install instructions. Otherwise they would need to modify the namespace value in `kustomization.yaml`
* Add `--namespace=scalyr` to example `kubectl` commands
* "Running the DaemonSet on select nodes" section should be modified to use kustomize if possible


k8s/scalyr-agent-2.yaml
```diff
@@ -1,92 +1,110 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
```
aksakalli: that will solve #410 as well

config/agent.json
```diff
@@ -4,14 +4,14 @@
...
-  api_key: "REPLACE_THIS",
+  api_key: "abc123",
 
   // Fields describing this server. These fields are attached to each log message, and
   // can be used to filter data from a particular server or group of servers.
   server_attributes: {
      // Fill in this field if you'd like to override the server's hostname.
      // serverHost: "REPLACE THIS",
-
+u
```
aksakalli: typo

k8s.old/scalyr-agent-2.yaml
```diff
@@ -0,0 +1,92 @@
+apiVersion: extensions/v1beta1
```
czerwingithub: `k8s.old` is too generic... the definition of `old` will change over time.

What's the definition of old here?  Whether or not they used the old agent setup instructions?  If so, then we should maybe call it something like `pre-2.1` instructions or something like that.

k8s/scalyr-agent-2.yaml
```diff
@@ -1,92 +1,110 @@
...
 kind: DaemonSet
 metadata:
+  annotations:
+    deprecated.daemonset.template.generation: "0"
```
czerwingithub: What does this do?

yanscalyr: This was added automatically by the `kubectl` tool that upgraded us up from `extensions/v1beta1`. I don't think we need this.

czerwingithub:
Yeah, let's make sure we aren't including annotations we don't need.  I'm
not sure how they would bite us.

k8s/scalyr-agent-2.yaml
```diff
@@ -1,92 +1,108 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: DaemonSet
 metadata:
+  creationTimestamp: null
```
czerwingithub: Remove `creationTimestamp`?

k8s/scalyr-agent-2.yaml
```diff
@@ -1,92 +1,108 @@
...
+    app: scalyr-agent-2
   name: scalyr-agent-2
 spec:
+  revisionHistoryLimit: 10
+  selector:
+    matchLabels:
+      app: scalyr-agent-2
   template:
     metadata:
+      creationTimestamp: null
```
czerwingithub: Remove `creationTimestamp`?

czerwingithub:
After looking at the outstanding PR you have to update the Scalyr documentation on the website, I think I want to request a few more changes here.  I think we need to come up with a better organization for our manifest and kustomize files.

Right now, you have all of the `k8s` manifest directories at the top-level in the repo, such as `k8s` for the default, then `k8s.pre-2.1.1` for the old version.  I could also see us needing to publish other versions like `k8s.kube-pre.1.14` (for configurations that work with K8s before version 1.14).  In that documentation pre, I suggested we create a `k8s.default-namespace` kustomization to provide a fallback for folks still running the agent in the default namespace.

If we take that approach to its extreme, we could have a ton of k8s related directories at the top-level of the repo.

I propose we just have one `k8s` directory at the top-level.  The manifest files in that directory represent the our default "this is how you should run agent on K8s for modern K8s versions".  Then, within that directory, we create subdirectories like "agent-pre-2.1.1" and "default-namespace", etc.

I think it might seem a little weird to have subdirectories that represent completely separate configuration styles, but it's a tradeoff.  We could create a directory `k8s/base` that represents the blessed way of doing it, and leaving `k8s` empty of manifest files.. but I think that hurts discoverability and hurts people using the old instructions on how to upgrade.

Thoughts?

yanscalyr:
Currently the only thing in our customize file by default is the namespace, if we wan't to give an upgrade path that stays in the default namespace it would be as simple as `kubectl apply -f https://raw.githubusercontent.com/scalyr/scalyr-agent-2/release/k8s/scalyr-agent-2.yaml`, I not sure what we would really include in a "default-namespace" directory.

Moving the `pre-2.1` files is easy enough.

yanscalyr:
I wasn't sure exactly what to put in the new `README.md` files, or what level of detail. Would appreciate feedback on that.

k8s/README.md
```diff
@@ -0,0 +1,13 @@
+## Kubernetes Files for Versions 2.1.1 and Up
+
+Detailed instructions for installing and upgrading the Scalyr agent in kubernetes can be found [here](https://app.scalyr.com/help/install-agent-kubernetes).
```
czerwingithub: I try to be consistent with capitalizing both the S and A in _Scalyr Agent_.  Please take a pass at doing that.  Also, the K in Kubernetes should be capitalized.  Please do that here and in the other README.md file you created.

k8s/README.md
```diff
@@ -1,6 +1,6 @@
 ## Kubernetes Files for Versions 2.1.1 and Up
 
-Detailed instructions for installing and upgrading the Scalyr agent in kubernetes can be found [here](https://app.scalyr.com/help/install-agent-kubernetes).
+Detailed instructions for installing and upgrading the Scalyr Agent in Kubernetes can be found [here](https://app.scalyr.com/help/install-agent-kubernetes).
```
czerwingithub: Right under the first line (the Kubernetes Files header), let's have this text:

This directory contains the manifest files for running the Scalyr Agent on your Kubernetes cluster.  Specifically, it defines the service account and DaemonSet required to run the Scalyr Agent in the `scalyr` namespace.  You must create a secret key and configmap before creating the service account and DaemonSet (see instructions below).

Prior to the Scalyr Agent 2.1 release, the Scalyr Agent K8s instructions created objects in the default namespace.  If you used these instructions prior to Scalyr Agent 2.1, you must either use the manifest files in the <link to default namespace directory> or perform the upgrade instructions documented at <link to where the instructions will be on the Scalyr website>

The follow up with the text you already have here (including the link to detailed instructions and the intended way to use the files)

k8s/agent-pre-2.1.1/README.md
```diff
@@ -1,9 +1,9 @@
 ## Kubernetes Files for Versions Before 2.1.1
 
-Detailed instructions for installing and upgrading the Scalyr agent in kubernetes can be found [here](https://app.scalyr.com/help/install-agent-kubernetes).
+Detailed instructions for installing and upgrading the Scalyr Agent in kubernetes can be found [here](https://app.scalyr.com/help/install-agent-kubernetes).
```
czerwingithub: Insert before the "Detailed instructions" line this text:

This directory contains the manifest files used for installing the Scalyr Agent prior to the 2.1 Scalyr Agent release.  You should only use these manifest files if you are running an instance prior to 2.1.

Here's a question... do we need this directory if we have the `default-namespace` one?  You were just using this as a placeholder to perform the delete... but the `default-namespace` would suffice for that as well?

k8s/default-namespace/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+namespace: default
```
czerwingithub: Let's put the default namespace directly in the manifest files here as well.

k8s/default-namespace/README.md
```diff
@@ -0,0 +1,15 @@
+## Kubernetes Files for Versions 2.1.1 and Up
+
+This directory contains the manifest files used for installing the Scalyr Agent prior to the 2.1 Scalyr Agent release. You should only use these manifest files if you are running an instance prior to 2.1.
```
czerwingithub: Insert this text after the first sentence and delete the second sentence:

Prior to 2.1, the Scalyr Agent K8s instructions created all objects in the default namespace.  This has since changed to create all objects in the scalyr namespace.  You should only use these manifest files if you need to continue running the Scalyr Agent in the default namespace, possibly as a stopgap measure until you can migrate your configuration.

k8s/scalyr-agent-2-configmap.yaml
```diff
@@ -15,4 +15,4 @@ data:
   #
 metadata:
   name: scalyr-config
-  namespace: default
+  namespace: scalyr
```
czerwingithub: Argh, won't this break the current instructions?  It should remain in the default namespace.

k8s/scalyr-agent-2.yaml
czerwingithub: I think you still need to remove this.

k8s/scalyr-agent-2-configmap.yaml
czerwingithub: I think need to comment this out with an instruction above it to uncomment it to create the config map in the scalyr namespace.  Then in your new instructions, make sure you tell them to do that.

yanscalyr: Well the current instructions don't even reference the file here, they tell you to create the config file yourself and give an example right in the docs, so I don't think this should affect anything.

czerwingithub: Ah, good catch.  Thanks for looking at that.  Yep, never mind then.

PR #457, opened 2020-03-12
Enable branch level coverage tracking and reporting

Kami:
This pull request enables branch level coverage tracking and reporting.

This should give a more real picture into code coverage compared to just a simple line based code coverage reporting we currently use.

See more info on branch coverage here - https://coverage.readthedocs.io/en/v4.5.x/branch.html.

Having said that, I'm not actually 100% sure if codecov.io supports branch level coverage tracking for Python projects (will soon see when the build finishes).

### DIFF
--- a/.circleci/.coveragerc_ci
+++ b/.circleci/.coveragerc_ci
@@ -3,6 +3,9 @@ source =
     /home/circleci/scalyr-agent-2/
     /usr/share/scalyr-agent-2/py/
 
+[run]
+branch = True
+
 [report]
 skip_covered=True
 omit =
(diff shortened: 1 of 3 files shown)


Kami:
For completness, here is an example.

1. Report without branch coverage - https://codecov.io/gh/scalyr/scalyr-agent-2/src/490f03f2a3f0a8d075d74df20610ceb23dff6c8c/scalyr_agent/builtin_monitors/shell_monitor.py#L140
2. Report with branch coverage - https://codecov.io/gh/scalyr/scalyr-agent-2/src/842ac5aa17484fb7b3c00456d30594475a019dc7/scalyr_agent/builtin_monitors/shell_monitor.py#L140

In short, it makes it easier to identify which branches we didn't hit.

PR #458, opened 2020-03-16
Allow user to specify a blacklist of metric names which shouldn't be logged and shipped to Scalyr

Kami:
This pull request implements a change which allows user to define a blacklist of metrics name (new ``metric_name_blacklist`` configuration option) which allows user to define which metric names should not be logged and shipped to Scalyr (as per discussion in #402).

## Note on Implementation Details

There are many ways to tackle this problem and I decided to go with a simple "metric name" blacklist based approach.

I think this should work reasonable well for most common use cases and it also introduces the smaller performance penalty compared to a more complex approach which may utilize globs, other fields, etc.

Another thing worth noting is that this is really the "last resort" approach - ideally those metrics should be filtered out at the source - via monitor configuration or similar. But where there is not possible / similar, this new approach can be used.

## Global vs monitor specific configuration option

Right now this configuration is global, but we could also make it monitor specific.

One problem with utilizing monitor specific configuration option is that we would need a bit of a special scenario for handling built in monitors (linux system and linux process metrics).

## Example Configuration

```javascript
...
metric_name_blacklist: ["app.cpu", "app.mem.majflt"]
...

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -334,10 +334,6 @@ class Configuration(object):
 
         return monitor_config
 
-    @property
-    def metric_name_blacklist(self):
-        return self.__get_config().get_json_array("metric_name_blacklist")
-
     # k8s cache options
     @property
     def k8s_ignore_namespaces(self):
@@ -1638,16 +1634,6 @@ class Configuration(object):
         )
         self.__verify_or_set_optional_string(
             config, "https_proxy", None, description, apply_defaults, env_aware=True
-        )
-        # Blacklist for metric names which should be excluded from being sent to Scalyr
-        self.__verify_or_set_optional_array_of_strings(
-            config,
-            "metric_name_blacklist",
-            [],
-            description,
-            apply_defaults,
-            separators=[None, ","],
-            env_aware=True,
         )
         self.__verify_or_set_optional_array_of_strings(
             config,
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -251,9 +251,7 @@ class Configuration(object):
 
             self.__monitor_configs = list(self.__config.get_json_array("monitors"))
 
-            # TODO: Is a good idea to call this on each Configuration object instantiation? We could
-            # probably use a single global singleton instance
-            platform_controller = PlatformController.new_platform()
+            platform_controller = PlatformController.existing_platform()
 
             # We add in special marker for monitor configuration objects which refer to built-in
             # monitors so we can handle those differently inside the MonitorManager
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -334,6 +334,10 @@ class Configuration(object):
 
         return monitor_config
 
+    @property
+    def metric_name_blacklist(self):
+        return self.__get_config().get_json_array("metric_name_blacklist")
+
     # k8s cache options
     @property
     def k8s_ignore_namespaces(self):
@@ -1635,7 +1639,16 @@ class Configuration(object):
         self.__verify_or_set_optional_string(
             config, "https_proxy", None, description, apply_defaults, env_aware=True
         )
-
+        # Blacklist for metric names which should be excluded from being sent to Scalyr
+        self.__verify_or_set_optional_array_of_strings(
+            config,
+            "metric_name_blacklist",
+            [],
+            description,
+            apply_defaults,
+            separators=[None, ","],
+            env_aware=True,
+        )
         self.__verify_or_set_optional_array_of_strings(
             config,
             "k8s_ignore_namespaces",
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -42,7 +42,6 @@ from scalyr_agent.json_lib.objects import (
 from scalyr_agent.monitor_utils.blocking_rate_limiter import BlockingRateLimiter
 from scalyr_agent.util import JsonReadFileException
 from scalyr_agent.config_util import BadConfiguration, get_config_from_env
-from scalyr_agent.platform_controller import PlatformController
 
 from scalyr_agent.__scalyr__ import get_install_root
 from scalyr_agent.compat import os_environ_unicode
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: scalyr_agent/scalyr_logging.py [hunk 1]
```diff
@@ -372,6 +372,16 @@ def emit_value(
                 metric_name=metric_name, metric_value=metric_value
             )
 
+        if metric_name in getattr(monitor, "_metric_name_blacklist", []):
```
czerwingithub: Just for my own education, why use `getattr` here instead of using a `set` to hold the blacklist?

Kami: That was done to make sure existing tests pass - a lot of them use various mock monitor classes without attributes.

Another approach would be to also update all the affected mock test classes.

Kami:
@czerwingithub As far as per monitor configuration goes - my main question is how to configure those metric name blacklists for the implicit linux system and linux process metrics monitor.

For that, we will likely need some special case (likely in the ``platform_linux.py`` where we instantiate those).

Perhaps something along those lines:

```bash
  monitors: [
    {
        module:  "scalyr_agent.builtin_monitors.linux_process_metrics",
        id: "agent",
        pid: "$$",
        metric_name_blacklist: [...]
    },
    {
        module:  "scalyr_agent.builtin_monitors.linux_system_metrics",
        metric_name_blacklist: [...]
    },
```

Linux system metrics one is a bit easier since there can only be a single instance of it and for the process metrics one we can just assume it refers to the implicit agent process metrics monitor if the id is agent.

And we will just need to document this as a special case under those monitor docs.

Kami:
I pushed some additional changes so this config option is now monitor specific and not global (it's still not fully finished though, it needs more tests, refactoring, etc.).

This means users can now do things such as:

```javascript
monitors: [
    {
        module:  "scalyr_agent.builtin_monitors.linux_process_metrics",
        id: "agent",
        pid: "$$",
        metric_name_blacklist: ["app.uptime", "app.disk.iowait"]
    },
    {
        module:  "scalyr_agent.builtin_monitors.linux_system_metrics",
        metric_name_blacklist: ["proc.stat.cpu"],
        sample_interval: 120
    },
    {
        module:  "scalyr_agent.builtin_monitors.linux_process_metrics",
        id: "my-process",
        commandline: ".*my-process.*",
        metric_name_blacklist: ["app.disk.iowait"]
    }
  ]
````

First two entries refer to the default implicit built-in monitors. It means we will apply those config options to the implicit built-in monitor instances instead of spinning two additional monitors instances.

As described in my comment above, we determine if the config entry refers to a built-in monitor based on the value of the module, id and pid config entries.

I think this approach is quite reasonable since it also allows user to configure other options for built-in monitors which previously couldn't be changed directly, such as the ``sample_interval`` one.

Eventually, if needed, we can also introduce support for wildcards - it should be trivial to implement that.

@czerwingithub Let me know what you think.

**Comment by Kami** [comment]
I've added tests and also refactored the code a bit so now the logic / abstraction for determining if a particular monitor configuration refers to a built-in monitor lives on the configuration object.

This means that any configuration consumer can now assume "builtin_monitor" attribute will be present for built-in monitors on the Configuration object.

It does mean we need to instantiate a new platform controller object inside the constructor each time. Probably not a big deal and perhaps we can optimize it to just use a single global singleton instance (we just need to make sure it doesn't introduce a memory leak or similar).

Another, perhaps better approach, would be to refactor ``Configuration`` class constructor to accept ``platform_controller`` argument.

**Comment by czerwingithub** [comment]
@Kami I'm going to take a look at this, but as far as your question about how to handle the implicitly created monitors.  We have a technique that we've communicated with the customers in the past to get around issues like this.  It's not quite ideal, but it works without a lot of complication.

For each of the implicit monitors, there is a global configuration option indicating to not automatically create monitors for them.  For the `linux_system_monitor`, it's `implicit_metric_monitor`.  For the `linux_process_metrics` monitor for the agent, it's `implicit_agent_process_metrics_monitor`.  We tell the customers to set that to false, then they can add in their own definitions for those monitors.

Obviously, the benefit of this is that we don't have to try to match up the monitor definition they are trying to make with the one we are implicitly creating.  The downside is you actually have to find the documentation that describes these config variables.

My apologies, I meant to include this discussion in my review last night, but it slipped my mind.

File: scalyr_agent/configuration.py [hunk 2]
```diff
@@ -250,6 +251,19 @@ def parse(self):
 
             self.__monitor_configs = list(self.__config.get_json_array("monitors"))
 
+            platform_controller = PlatformController.existing_platform()
+
+            # We add in special marker for monitor configuration objects which refer to built-in
+            # monitors so we can handle those differently inside the MonitorManager
```
czerwingithub: Let's resolve if we have to handle the implicit monitors specially.  We may not need this and other changes you made.

scalyr_agent/platform_controller.py
```diff
@@ -81,6 +85,21 @@ def new_platform():
                 return result
         return None
 
+    @staticmethod
+    def existing_platform():
+        """
+        Return an existing instance of the previously registered platform class.
+
+        The instance is instantiatd on first call to the this method. On subsequent calls, existing
```
czerwingithub: s/instantiatd/instantiated

Kami:
@czerwingithub Thanks for the feedback.

I actually thought we may advise our users to do something like that, but I wasn't sure so I implemented matching.

I'm all for simpler and more explicit approach and less code and if we already advise users to do that in the docs, I will go ahead and remove all the code which is related to matching built-in monitors.

If if we ever decide those changes are a good idea, we can always pull them back in :)

Kami:
I removed the code which is related to monitor config matching for built-in monitors. This makes the diff much smaller now.

---

If other people think having some kind of config matching functionality for built-in monitors would still be useful, I can open a separate PR with those changes so we can discuss those changes and approach separately.

The thing worth noting about those changes is is that it would still work with current approach we advise in documentation since we only perform matching if those monitors are not disabled via global config option (basically, it would cover both scenarios).

PR #462, opened 2020-03-17
Hook monitor documentation generation to CI

Kami:
This pull request hooks up monitor documentation generation for all the monitors to CI.

### Background, Context

I noticed some monitor docs are not up to date. It turns out we already have a script (``print_monitor_doc.py``) for generating part of the monitor documentation which can be auto generated, but it doesn't appear to be hooked up anywhere (I checked Jenkins, etc. if I missed some spot, please let me know).

This pull request fixes that and hooks it up to the CI.

First iteration adds new tox target and hooks it up to CI - we just verify that we can generate docs for all the monitor modules, but we don't write those values anywhere.

Second iteration will update generate script so it also writes those auto-generated values to ``docs/monitors/`` and makes sure that directory is up to date as part of every CI run. If it's not up to date, it will indicate developer forgot to run this target locally and commit updates docs.

 This will need a bit more work / doc re-organization though, because right now files in ``docs/monitors/`` contain two types of content - one which is written manually and one which is auto-generated (we will need to separate / re-organize that so we can only update auto generated part of the docs).


Kami:
@czerwingithub If you are fine with those changes, I can also add changes for the second iteration.

There are a couple approaches there, the simplest one is simply splitting documentation file in two parts - one which is manually edited (e.g. top section) and one which is generated automated (e.g. bottom section and everything after ``<!-- auto generated content below -->`` marker or similar).

I checked existing documentation structure and this should work reasonably well to begin with.

Another a bit more involved approach would be to utilize Jinja templates or similar - this way we could put auto generated fragments in arbitrary locations in the document.

Kami:
Here is one example of how we can approach auto generation - 8b61aff1deb639848ca4a243203c1627a5a1513e, 1c527ee72edbae1badc47bb445ac68c62dfe2c27.

It follows a simple marker approach - we simply concatenate auto generated content to the end of the markdown document which contains any non-auto generated content.

The approach is very simple, but that's basically how existing documentation already looked like (manually written content + manually append auto generated content) so it should work for now.

In the future, if we need more flexibility with formatting, etc. we can utilize Jinja templates or similar.

---

I also hooked it up to CI to fail the build if there are any uncommited changes after running the generate monitor docs script.

If there are, this would indicate a developer error - developer likely modified the monitor without running the generate script and committing the generated doc files.

scripts/generate-docs-for-all-monitors.sh
```diff
@@ -0,0 +1,68 @@
...
+# It writes generated documentation files to docs/monitors/<monitor name>.md
+
+# Marker in the Markdown file which marks the line after which automatically generated content
+# should be appended
+AUTO_GENERATED_SECTION_MARKER="<!-- Auto generated content below. DO NOT edit manually, but run tox -egenerate-monitor-docs command instead -->"
+
+set +e
+
+# Verify systemd dependency is available
+python -c "import systemd" 2> /dev/null
```
czerwingithub: In the past, we've made an effort to make sure the `print_monitor_doc` script could be run for any monitor even if it's dependencies could not be loaded.  For example, we could print the documentation for the windows monitor even if we don't have the windows python libraries available.

I would suggest instead of jumping through hoops here to try to import systemd, we just fix the `journald` to try to import `systemd` but only fail when the monitor is actually attempted to be instantiated.

Kami:
@czerwingithub 

> why did you not include the auto-generated description here? You only are including the metrics, config options, and log reference from the Python module?

In some scenarios this monitor description is a bit poor and "clashes" with manually written description.

I can still include it though and going forward this should serve as a motivation for us to improve class comments in the future (and yes, ideally those class comments should service as primary source of truth for anything which can be auto generated - manually written documentation should only refer to higher level concepts and other things which can't be auto generated).

scripts/generate-docs-for-all-monitors.sh
Kami: Yeah, I was thinking about it - this should work as long as the actual auto-generated stuff doesn't reference any constants or similar from imported modules.

I will look into it and if that's indeed not the case, refactor the code so we import that module lazily.

It should also simplify the build config since we ca avoid installing that dependency, but we will still need to install it at some point anyway when we start adding more in depth and end to end tests for those monitors :)

Kami:
@czerwingithub I pushed a change so "is dependency available" check is now lazy and happens during the monitor init phase - 42c28e2340f6c65b6f3d19abbf57d980d1722775, 34196d12d967a8aca5caa3b88829997f4c1bbb76.

As far as the description goes - I made some changes and updated all the docstrings for monitor classes so they contain the latest up to date documentation.

This way we can avoid manual documentation edits and we can rely on most of the monitor documentation being auto generated from monitor files which service as source of truth for that.

This should solve a problem with a single source of truth (aka someone manually updating markdown document, but not the docstring and things getting out of sync).

If you check the overall diff you will see things are more or less in sync now :)

PR #463, opened 2020-03-17
Change python version with config command. Refactor smoketests.

ArthurKamalov:
Most of the changes were guided by this notion page: https://www.notion.so/scalyr/Scalyr-Agent-Switch-for-Python-3-01356f302a5f4679875aeb34448a6a6b

* Added feature that provides the possibility to switch between python2 and python3 by command `scalyr-agent-2-config --set-python` or `scalyr-switch-python. 
* Added more checks to pre/postinstall scripts.
* new test cases to cover all new features. All new test cases implemented for ubuntu, amazonlinux and centos. I have also get centos because it does not have mapped `python` command, only `python2`. To handle this, `build_package.py` script also creates `config_main_py2.py`. This file has `#!/usr/bin env python2` instead of `#!/usr/bin env python`.


.circleci/LOCAL_TESTING.md
```diff
@@ -18,7 +18,7 @@ circleci_cli.py
 
 * On job execution with command - `circleci_cli.py local execute --job <jobname>`,
 it check `<jobname>` and if it starts with `smoke` it assumes that it is a smoke test
-and imports environment variables from `smoke_tests/config.yml`, so you do not have to specify them manually. (see smoke_test/README.md)
+and imports environment variables from `tests/smoke_tests/config.yml`, so you do not have to specify them manually. (see smoke_test/README.md)
```
Kami: Yeah, good idea to move those tests under ``tests/`` directory :+1:

build_package.py
```diff
@@ -879,6 +891,26 @@ def build_base_files(base_configs="config"):
...
+
+    # create copy of the config_main.py with python3 shebang.
+    config_main_path = os.path.join(agent_source_root, "scalyr_agent", "config_main.py")
+    replace_shebang(
+        config_main_path,
+        os.path.join("scalyr_agent", "config_main_py3.py"),
+        "#!/usr/bin/env python3"
+    )
+
+    os.chmod(agent_main_py3_path, stat.S_IREAD | stat.S_IEXEC)
```
Kami: Yeah, ideally we would just copy the permissions of the existing ``agent_main.py`` file.

installer/scripts/scalyr-switch-python.sh
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+if [ $# -lt 1 ]; then
+  echo "Usage: ${0} <python2|python3>"
+  exit 2
+fi
+
+EXECUTABLE_NAME=${1}
+
+/usr/bin/env ${EXECUTABLE_NAME} `which /usr/share/scalyr-agent-2/bin/scalyr-agent-2-config` --set-python=${EXECUTABLE_NAME}
```
Kami: Should we also do additional check here and ensure that the specified python binary exists? Or I guess this line will already throw if it doesn't so it's probably fine.

Kami: Just trying to understand why we need to use ``which`` here. Is that's because ``scalyr-agent-2-config`` is a symlink?

scalyr_agent/config_main.py
```diff
@@ -1248,6 +1293,13 @@ def create_custom_dockerfile(
         "write it to stdout.",
     )
 
+    parser.add_option(
+        "",
```
Kami: Let's also add a help / description string for this option.

scalyr_agent/config_main.py
```diff
@@ -1431,6 +1483,10 @@ def create_custom_dockerfile(
             # do find a need to take some action.
             sys.exit(finish_upgrade_tarball_install(paths[0], paths[1]))
 
+    if options.set_python is not None:
+        set_python_version(options.set_python)
+        print("Agent switched to {0}.".format(options.set_python))
```
Kami: After this line (if the switch succeeded) we should probably exit, aka ``sys.exit(0)``, right?

build_package.py
```diff
@@ -831,6 +831,18 @@ def build_tarball_package(variant, version, no_versioned_file_name):
...
 
+def replace_shebang(path, new_path, new_shebang):
+    with open(path, "r") as f:
+        with open(new_path, "w") as newf:
+            #skip shebang
+            f.readline()
+            newf.write(new_shebang)
+            newf.write("\n")
+            for line in f:
+                newf.write(line)
```
Kami: We could probably simplify that since we know shebang is always on the first line, but this also works.

build_package.py
```diff
@@ -930,6 +962,15 @@ def build_base_files(base_configs="config"):
 
     make_soft_link("../py/scalyr_agent/agent_main.py", "scalyr-agent-2")
     make_soft_link("../py/scalyr_agent/config_main.py", "scalyr-agent-2-config")
+    # add switch python version script.
+    shutil.copy(
+        os.path.join(agent_source_root, "installer", "scripts", "scalyr-switch-python.sh"),
+        "scalyr-switch-python"
```
Kami: I wonder if for consistency with other binaries we should call this one ``scalyr-agent-2-switch-python``.

Kami:
It looks like ``scalyr_agent/run_monitor.py`` can also serve as an entry point, but it appears it's mostly used for testing so probably not something we need to handle here :)

scalyr_agent/config_main.py
Kami: This should be part of our documentation on running agent under Python 2 / 3, but probably also doesn't hurt to it print it here as well.

If there is an agent instance running when this command is ran, it needs to be restarted for it to take an affect.

I don't think we should restart it ourselves, but just print something along the lines of "If you have an existing instance of scalyr-agent-2 process running, you need to restart it for this change to take an affect. You can do that by running /usr/init.d/scalyr-agent-2 restart...".

tests/utils/dockerized.py
```diff
@@ -0,0 +1,44 @@
...
+        root = Path(__file__).parent.parent.parent
+        rel_path = Path("agent_source") / Path(file_path).relative_to(root)
+
+        command = "python -u -m pytest {0}::{1} -s --no-dockerize".format(
+            rel_path,
+            f.__name__
+        )
+
+        def wrapper(*args, **kwargs):
+            no_dockerize = pytest.config.getoption("--no-dockerize")
```
Kami: Just a heads up - https://docs.pytest.org/en/latest/deprecations.html#pytest-config-global

tests/utils/dockerized.py
```diff
@@ -0,0 +1,44 @@
...
+            if no_dockerize:
+                result = f(*args, **kwargs)
+                return result
+
+            builder = builder_cls()
+            builder.build()
```
Kami: Some of those commands (Building docker image), etc. take quite a while so in those scenarios we may want to run py.test with ``-s`` flag to output is not captured and printed to stdout - this way we know why tests are still running and not stuck.

Kami: And ideally we would also periodically print progress inside ``builder.build()`` method, but that's not a blocker for now.

Kami:
Please also make sure we run all the lint check on files inside ``tests/`` directory (if not already).

tests/utils/dockerized.py
Kami: Is there any way we could use image cache here?

If we can't, this will be very slow since we will need to re-build the image from scratch on CI during each job run if DLC is disabled...

tests/utils/agent_runner.py
```diff
@@ -166,16 +166,52 @@ def _create_agent_files(self):
...
+        if self._installation_type == PACKAGE_INSTALL:
+            subprocess.check_call([
+                "/usr/sbin/scalyr-agent-2-config",
+                "--set-python",
+                version
+            ])
+
 
     @staticmethod
     def stop_agent_if_running():
```
Kami: In this method we should probably also check if agent is running in default background mode (aka ``--no-fork`` argument is not used) and if it is, also delete pid file here to avoid potential issue with old and stale pid file laying around.

installer/scripts/scalyr-switch-python.sh
ArthurKamalov: Yes, no need for `which` here. `scalyr-agent-2-config` points on `config_main.py`, since its code is compatible, we can pass it directly to current python.

scalyr_agent/config_main.py
ArthurKamalov: Yes, I think you right.

tests/utils/dockerized.py
ArthurKamalov: > Some of those commands (Building docker image), etc. take quite a while so in those scenarios we may want to run py.test with -s flag to output is not captured and printed to stdout - this way we know why tests are still running and not stuck.
It prints build output, but it looks like we need to wait until the image is built, and only then we can read the build output. I tried to find the way to read in real-time, but without success :(

> Is there any way we could use image cache here?
Yes, now it can get path to cache from py.test option.

tests/utils/agent_runner.py
ArthurKamalov: Actually, I've decided to remove it and keep only `scalyr-agent-2 stop`, because it should work fine too.

scalyr_agent/config_main.py
ArthurKamalov: Good idea. Thanks.

build_package.py
ArthurKamalov: Yeah, at least, the loop can be removed for sure :)

installer/scripts/scalyr-switch-python.sh
ArthurKamalov: I've added additional conditions that allow only `python2` and `python3`

build_package.py
ArthurKamalov: yup, thanks for the good advice.

tests/utils/dockerized.py
ArthurKamalov: Oh, didn't know about that... thanks)

build_package.py
ArthurKamalov: I think I agree with you. 
@czerwingithub what do you think about this name?

tox.ini
```diff
@@ -118,63 +118,85 @@ basepython = python2.6
...
 [testenv:agent_package_smoke_test_ubuntu_python3]
 basepython = python3.6
 passenv =
     SCALYR_API_KEY READ_API_KEY SCALYR_SERVER AGENT_HOST_NAME DOCKER_CERT_PATH DOCKER_HOST DOCKER_TLS_VERIFY
 commands =
-    py.test smoke_tests/package_smoke_tests -s -vv --durations=5 --package-distribution ubuntu --package-python-version python3 {posargs}
+    py.test tests/smoke_tests/package_test.py::test_smoke_package_deb_python3 -s -vv --durations=5 {posargs}
+
+
+[testenv:agent_distributions_tests_ubuntu]
```
Kami: How do we plan to run those on Circle CI?

I didn't see we hooked up those targets in circle ci config yet. Or did I miss it?

EDIT: It seems like we run it as part of the existing package test job?

installer/scripts/postinstall.sh
```diff
@@ -13,6 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+# if there is no python2, switch on python3. Since we passed preinstall, we think that at least python3 is available.
+if ! /usr/bin/env python2 --version >/dev/null; then
+  echo "Python2 was not found. Use python3."
```
Kami: Perhaps should say something like "python2 binary not would, will use python3 binary for running the agent".

Kami: Also, do we have a test case for this scenario?

To make sure that the agent is indeed correctly started after we switch the binary in this post install script.

installer/scripts/scalyr-switch-python.sh
```diff
@@ -0,0 +1,38 @@
...
+# usage: scalyr-switch-python.sh python2|python3
+
+if [ $# -lt 1 ]; then
+  echo "Usage: ${0} <python2|python3>"
+  exit 2
+fi
+
+if [[ ${1} == "--help" || ${1} == "-h" ]]; then
+  echo "Usage: ${0} <python2|python3>"
+  echo "Switch python interpreter varsion for Scalyr agent."
```
Kami: a couple of typos

installer/scripts/scalyr-switch-python.sh
```diff
@@ -0,0 +1,38 @@
...
+  exit 0
+fi
+
+if [  "${1}" != "python2" ] && [ "${1}" != "python3" ]; then
+  echo "Invalid value: ${1}."
+  exit 2
+fi
+
+EXECUTABLE_NAME=${1}
+/usr/bin/env "${EXECUTABLE_NAME}" /usr/share/scalyr-agent-2/bin/scalyr-agent-2-config --set-python="${EXECUTABLE_NAME}"
```
Kami: We should probably do ``set -e`` at the top of the script, this way it will exit with non zero if any of the commands we call fails (more specifically this command).

scalyr_agent/config_main.py
```diff
@@ -1132,6 +1134,53 @@ def create_custom_dockerfile(
...
+PYTHON2 = "python2"
+PYTHON3 = "python3"
+
+
+def set_python_version(version):
+    """Switch agent command main files to another version of python"""
+    controller = PlatformController.new_platform()
+
+    if controller.install_type != PACKAGE_INSTALL:
+        raise Exception("Not a package.")
```
Kami: Should make the error message a bit more verbose and user-friendly (even though this error should never really pop up for the end users, but still)

scalyr_agent/config_main.py
```diff
@@ -1132,6 +1134,53 @@ def create_custom_dockerfile(
...
+                os.remove(target)
+                os.symlink(source, target)
+
+    make_symlink(agent_main_source, scalyr_agent_2_target)
+    make_symlink(config_main_source, scalyr_agent_2_config_target)
+
+    print("Switched agent to python {0}".format(version))
+    print(
+        "If you have an existing instance of scalyr-agent-2 process running, "
+        "you need to restart it for this change to take an affect."
```
Kami: We can also add, "You can do that by running ``/etc/init.d/scalyr-agent-2 restart``" or similar.

Kami:
I've added a couple of more small comments.

Mostly looks good to me, but @czerwingithub  might want to have a quick look at various messages we print (we can probably improve and re-word some, nothing major though).

tests/utils/dockerized.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+        def wrapper(request, *args, **kwargs):
+            no_dockerize = request.config.getoption("--no-dockerize")
+            if no_dockerize:
+                result = f(request, *args, **kwargs)
+                return result
+
+            builder = builder_cls()
+            use_cache_path = request.config.getoption("--image-cache-path", None)
```
Kami: I see you made a change now to use the cache :)

tox.ini
ArthurKamalov: Yes, those tests are not in CircleCI yet. In theory, all those tests can be executed in one job, or we can split them into one job per each OS distro. I think we need to decide what would cost less.

installer/scripts/postinstall.sh
ArthurKamalov: Thanks for pointing on that, I think I haven't paid enough attention to messages.

tox.ini
Kami: We already have tons of separate builders so maybe not a bad idea to combine some of those together (this way we also avoid some overhead of spinning up new Circle CI environment for each job).

It mean mean though that everything will take slightly longer to finish due to those jobs not running concurrently...

But we can also experiment and see, in theory combining it should be slightly cheaper because we avoid overhead of spinning up the environment (which doesn't take that long), but it means the whole build will take longer to complete.

installer/scripts/postinstall.sh
ArthurKamalov: > Also, do we have a test case for this scenario?

I think the test case - `test_<DISTRONAME>_python3` is responsible for that. All those tests call this common function where we check the python version after installation on the machine without python2 
https://github.com/scalyr/scalyr-agent-2/blob/342cb34efd1d320e159b6cefffeb6216ce10abaa/tests/distribution/python_version_change_tests/common.py#L36-L57

tests/conftest.py
```diff
@@ -42,6 +47,26 @@ def pytest_addoption(parser):
         "Fields from this config file will be set as environment variables.",
     )
 
+    parser.addoption(
+        "--no-dockerize",
+        action="store_true",
+        help="Make test cases that were decorated by 'utils.dockerized_case' run on that machine, "
```
Kami: :+1:

tests/distribution/python_version_change_tests/common.py
```diff
@@ -0,0 +1,118 @@
...
+        os.remove(str(Path("/", "usr", "bin", "python")))
+    except IOError:
+        pass
+    os.remove(str(Path("/", "usr", "bin", "python2")))
+
+    install_package_fn()
+
+    runner = AgentRunner(PACKAGE_INSTALL)
+    runner.start()
+    time.sleep(1)
```
Kami: I guess we should just keep an eye on this sleep if we encounter any races because the agent hasn't fully start up yet in the future.

PR #464, opened 2020-03-18
GitHub PR Template

BrianChristie:
# Description

This proposed change adds a GitHub issue template to provide some reminders of what info is useful when creating a new PR.

Thoughts and proposed changes are welcome!

Fixes # N/A

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/pull_request_template.md
+++ b/.github/ISSUE_TEMPLATE/pull_request_template.md
@@ -1,20 +1,11 @@
 # Description
 
-Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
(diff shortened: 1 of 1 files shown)


Kami:
I'm fine with this although I would prefer to get rid of "type" and "how this has been tested" section.

From my experience with other projects, the simplest the template the better it works. I would prefer to just have two sections - Description and Background / Context.

For "type" we should ideally utilize issue / PR labels.

BrianChristie:
SGTM. I've updated to change the headings.

Do you want to merge as-is, or do you think we should add some sample text?

Kami:
I think as-is is fine :)

PR #465, opened 2020-03-19
update apiVersion and add selector to k8 manifest

zak905:
Starting from Kubernetes 1.16 `apps/v1` api version is required, also a selector on a template label is required. I hope this saves time for new adopters.

### DIFF
--- a/k8s/scalyr-agent-2.yaml
+++ b/k8s/scalyr-agent-2.yaml
@@ -1,8 +1,11 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: scalyr-agent-2
 spec:
+  selector:
+    matchLabels:
+      app: scalyr-agent-2
   template:
     metadata:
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the contribution.

I believe @yanscalyr is already working on some of those changes in #455.

---

@yanscalyr How should we proceed here? Is this PR / change safe enough so we can merge it before work in #455 is done, or we should wait for #455 to be finished?

yanscalyr:
I believe this should be safe enough. These changes were already done for the `scalyr-agent-2-envfrom.yaml` which is the one the documentation recommends, and I think the plan is to replace `scalyr-agent-2.yaml` with `scalyr-agent-2-envfrom.yaml` entirely soon.

czerwingithub:
@zak905 Thanks for this PR.  As you probably saw, the same change was being made by a member from the team.  We've now released the Scalyr Agent with this change, so I'm going to close your PR off.

Again, thanks for taking the time to submit this.

PR #466, opened 2020-03-19
Log to stdout when running in `--no-fork` mode

yanscalyr:
~Currently I can't seem to get `scalyr_logging.DEBUG_LEVEL_X` logs to be output to stdout, hence WIP.~
Resolved.


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -4124,6 +4124,14 @@ def run(self):
                 self.__report_container_metrics,
             )
         )
+
+        server_link = self._global_config.scalyr_server
+        if server_link == "https://agent.scalyr.com":
+            server_link = "https://app.scalyr.com"
+        elif server_link == "https://upload.eu.scalyr.com":
+            server_link = "https://app.eu.scalyr.com"
+        global_log.info("To see your logs go to %s" % server_link, force_stdout=True)
```
czerwingithub: I'd actually like to make this even more fancy... we need to create a link that actually shows the logs for this pod on the events page.  I would suggest we move this out of this PR and follow up in another.  I would suggest looking at creating a search that restricts to the log file equaling the agent.log and the server host for this agent.

Also, some folks out there have their `scalyr_server` set to `www.scalyr.com` or `api.scalyr.com`.  I think what we do here is first check if the server host contains `eu.scalyr.com`.  If so, emit the eu version.. otherwise the www.scalyr.com.   It would be good to handle the staging environments as well.

Take a look at the `status -v` output the agent produces.  We do generate a link there as well.

scalyr_agent/configuration.py
```diff
@@ -1593,6 +1599,22 @@ def __verify_main_config_and_apply_defaults(
                 "badDebugLevel",
             )
 
+        self.__verify_or_set_optional_string(
+            config,
+            "stdout_severity",
+            "NOTSET",
```
czerwingithub: Why use `NOTSET`?  Why not just have None or empty string?

scalyr_agent/configuration.py
```diff
@@ -824,6 +825,11 @@ def debug_level(self):
         """Returns the configuration value for 'debug_level'."""
         return self.__get_config().get_int("debug_level")
 
+    @property
+    def stdout_severity(self):
+        """Returns the configuration value for 'stdout_severity'."""
```
czerwingithub: I would add a note here that this is only used if the agent process is not forced.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
yanscalyr: The `status -v` output has exactly what you are asking for as far as I can tell, so just a quick copy over should do it. Its minor enough that another PR might be a bit much.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -4124,6 +4124,16 @@ def run(self):
                 self.__report_container_metrics,
             )
         )
+
+        server = self._global_config.scalyr_server
+        server = server.replace("https://agent.", "https://www.")
+        server = server.replace("https://log.", "https://www.")
```
czerwingithub: This is common code with the `status -v` output right?  We should make a common method in `util.py` that returns the URL for all logs from this `serverHost` given the `scalyr_server` and `server` value.

I don't think this is quite right though, which probably means we didn't update the `status -v` when we should have.  For example, in the K8s instructions for EU, we tell the customer to set the scalyr server to `https://upload.eu.scalyr.com`.  This code here would mean that they would be visiting `https://upload.eu.scalyr.com` to view their events in the Scalyr UI.

I would suggest that anything mapping to (upload,agent,log,app).eu.scalyr.com gets mapped to `https://eu.scalyr.com`.  Anything mapping to (upload,agent,log,app).scalyr.com gets mapped to `https://www.scalyr.com`.  Otherwise, we just leave the server host unchanged.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -4124,6 +4124,16 @@ def run(self):
...
+        global_log.info(
+            "View data from this agent at: %s/events?filter=$serverHost%%3D%%27%s%%27"
```
czerwingithub: I actually would like a URL here that pulls up just the `agent.log`.  A customer is going to see this when they execute `kubectl logs <agent-pod>`.  If they are doing that, I believe they are looking for the logs from this particular application process.. so it would be best to give them a way to do that.

You could do this by adding in a term "$logfile = '/var/log/scalyr-agent-2/agent.log'" to the filter term in the URL.  So, if we are making this common code with `status -v`, you'll have to add in an optional parameter to restrict the logfile to a certain file.

I'm also fine if you want to include both URLs (one for just the agent log and one for all logs output from this server host) but we'd have to make it very clear which is which.  Though, to be honest, looking at all data being gathered by this agent might just to more than they really want here.

scalyr_agent/scalyr_logging.py
```diff
@@ -1010,7 +1019,9 @@ def filter(self, record):
         @return:  True if the record should be logged by this handler.
         @rtype: bool
         """
-        return getattr(record, "force_stdout", False)
+        return getattr(record, "force_stdout", False) or (
+            self.__no_fork and record.levelno >= self.__stdout_severity
+        )
 
 
 class ForceStderrFilter(object):
```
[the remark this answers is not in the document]
yanscalyr: The stderr filter actually still only outputs in cases of `force_stderr` being true, I don't think there is anything to change the doc to here.

docker/Dockerfile.k8s
```diff
@@ -15,6 +15,7 @@ RUN pip --no-cache-dir install --root /tmp/dependencies -r requirements.txt
 FROM python:2.7-slim as scalyr
 MAINTAINER Scalyr Inc <support@scalyr.com>
 
+ENV SCALYR_STDOUT_SEVERITY ERROR
```
czerwingithub, line 18: This could be a little cryptic.  Let's add a comment here saying `Configure agent to only emit log messages with ERROR and above to stdout by default`

scalyr_agent/configuration.py
```diff
@@ -850,6 +851,13 @@ def debug_level(self):
         """Returns the configuration value for 'debug_level'."""
         return self.__get_config().get_int("debug_level")
 
+    @property
+    def stdout_severity(self):
+        """Returns the configuration value for 'stdout_severity'.
+        Only used when running in no-fork mode.
+        """
+        return self.__get_config().get_string("stdout_severity", default_value="NOTSET")
```
czerwingithub, line 859: You do not need to specify the `default_value` here.  The lines you added to invoke `self.__verify_or_set_optional_string` for this configuration value guarantees that `stdout_severity` will have a value by the time this method is invoked.

scalyr_agent/scalyr_logging.py
```diff
@@ -134,6 +136,9 @@ def set_log_destination(
 
     @param use_stdout: True if the logs should be sent to standard out.  If this is False, then use_disk must be true.
     @param use_disk:  True if the logs should be sent to disk.  If this is False, then use_stdout must be true.
+    @param no_fork:  True if we are running in --no-fork mode, logs above a configured severity threshold will be
+    written to stdout.
```
czerwingithub, line 140: Indent four spaces here.

scalyr_agent/util.py
```diff
@@ -698,6 +698,14 @@ def get_parser_from_config(base_config, attributes, default_parser):
     return default_parser
 
 
+def get_web_url_from_upload_url(server):
```
czerwingithub, line 701: Let's add a unit test for this.  I'd like to check all these cases, using both the US and EU URLs.  Also, add in a test for `logstaging.scalyr.com` to verify it comes through unchanged.

scalyr_agent/agent_main.py
```diff
@@ -239,7 +241,7 @@ def main(self, config_file_path, command, command_options):
         quiet = command_options.quiet
         verbose = command_options.verbose
         status_format = command_options.status_format
-        no_fork = command_options.no_fork
+        self.__no_fork = command_options.no_fork
```
czerwingithub, line 244: It's a minor point, but I think it's better to not introduce a state variable like this when it's not really needed.  I think, you just need to add a `no_fork` argument to the `_perform_config_checks` method and then you don't need to create a `self.__no_fork` variable.  I believe it also results in less lines needing to be changed.  Let me know if that's not correct.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -4123,6 +4123,16 @@ def run(self):
                 self.__report_container_metrics,
             )
         )
+
+        server = scalyr_util.get_web_url_from_upload_url(
+            self._global_config.scalyr_server
+        )
+        global_log.info(
+            "View data from this agent at: %s/events?filter=$serverHost%%3D%%27%s%%27&log=%%2Fvar%%2Flog%%2Fscalyr-agent-2%%2Fagent.log"
```
czerwingithub, line 4131: Let's change the leading string to "View the log for this agent at:"

Hmmm... there is a bug in this code in that if `serverHost` has characters that need to be escaped in a string or as  a URL parameter, we won't handle it correctly.  You aren't introducing this bug.. it was already there to begin with.

I don't have a quick fix for this, so maybe add a TODO saying "Handle escaping appropriate characters in serverHost string".

yanscalyr: Sadly the `urllib` way of encoding url params changed between python 2 and 3, or else it would have been trivial.

scalyr_agent/agent_main.py
yanscalyr: It would also need to be passed in to `__run`, and that method is used as a parameter to `__controller.start_agent_service`, we have fork info in there as well but that does add more places to pass parameters, any `platform_controller` implementations. I'm not sure if its better to avoid passing the info around in these places or avoid the extra state. What do you think?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
czerwingithub: Actually, `six` provides a compatibility layer for that, I believe.  Check out this sample code: https://github.com/scalyr/scalyr-agent-2/blob/master/tests/smoke_tests/request.py#L48

scalyr_agent/agent_main.py
czerwingithub: Ok, never mind then.  I would say though is then remove `no_fork` as the parameter for the other methods like `__start`, `__restart`, etc.  We should at least be consistent -- either we are passing it around or relying on it to be there.

PR #467, opened 2020-03-20
Various small optimizations

Kami:
This pull request includes a couple of small optimizations:

1. Make json log parser code more efficient

Instead of iterating over the existing object and creating a new one, we manipulate parsed object in place.

This can be up to 10x or more faster, but it depends on the size of the object in questions - for very small objects (e.g. objects with not many fields or objects with small "line" attribute) there likely won't be any significant difference, but it should show with larger objects.

2. Add ``__slots__`` attribute to LogLine class

We can do that because that class defines a small fixed set of attributes. This should make attribute access a bit more efficient and slightly reduce memory usage.

I did a quick test with object with a short line and it's 1056 vs 888 bytes (instance without slots defined and instance with slots defined).

3. Add support for ``orjson`` json library (sadly not available for older Python versions)

Based on some local testing (and more testing in the past at StackStorm project), for not very large objects, it can be 2-3x faster on serialization and 2-4x faster on deserialization.

---

I also plan to spend some time on it next week (and as discussed in the past), hook some of those things up to CodeSpeed as micro benchmarks for specific pieces of code.

I will make sure to use some representative data (aka actual log lines we may encounter).


scalyr_agent/configuration.py
```diff
@@ -250,6 +252,14 @@ def parse(self):
 
             self.__monitor_configs = list(self.__config.get_json_array("monitors"))
 
+            # Set json library based on the config value. If "auto" is provided this means we use
+            # default behavior which is try to use ujson and if that's not available fall back to
+            # stdlib json
+            json_library = self.__config.get_string("json_library")
```
czerwingithub: Setting this change within the `parse` method isn't quite correct I think with our current model.  Remember, every 30 seconds, we are parsing the configuration state from disk to see if it has changed.  If it is different, then we go ahead and apply those changes by stopping the current monitors and copying manager and starting new ones with the new configuration.  Given that model, I'd be surprised as a developer that the `parse` method has a side effect like this.

What we probably need to do is have a `apply_config` method in something like `agent_main.py` (where the new configuration state is being read) to apply any changes like this to the global state.

Kami: Yeah, I agree about ``parse()`` side affect.

For now, I will move that functionality into ``__create_worker_thread`` and in the future, we can refactor it a bit more and have additional apply config or similar method.

PR #468, opened 2020-03-23
Tests for nginx, apache, tomcat monitor, url, shell, linux process and linux system monitors

Kami:
This pull request adds new base ``ScalyrMockHttpServerTestCase`` test class which is used by various new tests I added for the following monitors:

* Nginx
* Apache
* Tomcat

 In addition to that, I also updated URL monitor tests to utilize this base class + adds some more tests for various scenarios.

Those tests uncovered various failures under Python 3 which I fixed (eventually we should really move away from ``BindableHTTPConnectionFactory`` stuff, but I didn't want to introduce additional changes in this PR).

It also uncovered a bug with invalid extra field for two thread related metrics in the Tomcat monitor.


Kami:
I also added some additional tests for shell monitor (c0a38b372491b8e7a20ba771b3578fc9df518baf) and pushed a fix for Python 3 bug uncovered by those tests (e91f2e2d9250d52912a43dda1e0fe39a1a85d1b7).

PR #469, opened 2020-03-24
Udpate codecov config to ignore top level tests/ directory

Kami:
I noticed codecov started tracking coverage for new top level ``tests/`` directory (introduced in #463 ) for which we currently don't track coverage.

This pull request updates codecov config so we ignore that directory since it otherwise skews the stats and doesn't provide any value until we also enable coverage for those tests.

### DIFF
--- a/.circleci/.coveragerc_ci
+++ b/.circleci/.coveragerc_ci
@@ -13,3 +13,5 @@ omit =
     */venv/*
     */third_party*/*
     tests/*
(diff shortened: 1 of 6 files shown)


Kami:
I also attempted to fix a dead-lock in kubernetes monitor test which occasionally pops up in this PR and pushed some other related improvements.

Will see how this goes after more build runs, locally I can't reproduce the deadlock anymore with those changes.

PR #470, opened 2020-03-26
Bug/merge release to master

czerwingithub:
This commit merges the changes made on the release branch into master.  I performed this merge by just taking the changes from master over the changes in the release.  I believe that's the right approach here but please say something if that's not the case.  I do not believe we need anything currently in the release branch to override the master.


czerwingithub:
Of course, this commit shows nothing is being modified on the master branch but that's what we are going for here.

Kami:
Yeah, that should work fine :+1:

PR #471, opened 2020-03-26
Add release notes document

Kami:
As discussed, this PR adds a new release notes document which describes various changes more in-depth.

First I thought about calling it Upgrade Notes, but I think Release Notes makes more sense here.


czerwingithub:
Argh, I think we had a miscommunication.  I had written up notes late last
night myself and wanted to proof read them this morning before I sent it
out for review.  Ok, I'll take a look at this and maybe combine some of my
content in with yours as it makes sense.

RELEASE_NOTES.md
```diff
@@ -0,0 +1,91 @@
...
+  suitable Python 2 version is not found, it will try to use Python 3 if available (
+  ``/usr/bin/env python3``).
+
+  If you want to select a specific Python version you want to use to run the agent (e.g. you have
+  both Python 2 and Python 3 installed), you can do that using ``scalyr-switch-python`` tool
+  which ships with the agent.
+
+  Selecting Python 2:
+
+  ```bash
```
czerwingithub: We generally remind folks to run commands like this with `sudo`.

RELEASE_NOTES.md
```diff
@@ -0,0 +1,91 @@
...
+  scalyr-switch-python python2
+  /etc/init.d/scalyr-agent-2 restart
+  ```
+
+  Selecting Python 3:
```
czerwingithub: Same here.

RELEASE_NOTES.md
```diff
@@ -0,0 +1,91 @@
...
+  ```
+
+  Keep in mind that you need to restart the agent after running the ``scalyr-switch-python``
+  command for the changes to take an effect.
+
+* This release adds new ``metric_name_blacklist`` monitor configuration option. With this
+  configuration option you can specify which metrics generated by the monitors should not
+  be logged and shipped to Scalyr.
+
+  For example, if you want to exclude ``app.io.wait`` and ``app.uptime`` metrics for the
```
czerwingithub: Are these the new metrics you added in?  I think we should call it out specifically in the notes here that two new metrics were added to the `linux_process_metrics` monitor and that if you do not want these metrics, this is what you should do.

Kami: Yep, agreed, will document that.

PR #475, opened 2020-03-27
Add default python option to  python switch feature

ArthurKamalov:
Add new option to the `scalyr-switch-python` that allows us to chose default system 'python' interpreter which can be accessed through `#!/usr/bin env python`.
Also, new tests were added to cover this new feature, and more tests to cover old python version cases.


tests/common.py
```diff
@@ -18,14 +18,36 @@
...
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
     )
 
+    process.wait()
+
+    stdout = process.stdout.read().decode("utf-8")
+    stderr = process.stderr.read().decode("utf-8")
+
+    if process.returncode != 0:
```
Kami: Yeah, that's a good change, I was thinking about making a change like that myself :+1:

Kami: I see it's using in multiple places so we should probably eventually refactor it in a utility function - e.g. run_command or similar.

installer/scripts/postinstall.sh
```diff
@@ -14,11 +14,65 @@
...
+    return 0
+  fi
+}
+
+
+check_python_version() {
+  # Verify that a suitable Python version is available and set up the agent symlinks
+  if is_python_valid python; then
+    # 'python' command exists
+    if [[ ${config_base_name} == "config_main.py" ]]; then
```
czerwingithub: Shouldn't this be `!=`?  I think you are trying to only run the switch command if the symlink currently doesn't point there.

I actually think we are trying to be too fancy here and should push this logic down into the `config_main.py` file.  I would advocate that the bash script always invokes `scalyr-switch-python` which then invokes `config_main.py`.  The `config_main.py` could make the determination if the symlink is already set correctly or not (and emit output to that effect).

If we take that approach, we could write automated tests to cover those cases for `config_main.py`.  It reduces the number of lines in the bash script (which I'm always fan of) and eliminates the need for the `config_base_name` variables.

PR #476, opened 2020-03-27
Workaround for master build failure (don't use built image cache)

Kami:
I added a new test in #472 and new package tests started to fail on master (they were all passing locally).

It looks like there is an issue where we incorrectly handle Docker image caching so the tests are using old and out of date agent_source/ directory.

This is a short term workaround (don't use cache), but in the future we need to figure out a better approach and solution that since re-building whole image on any change in the agent repo is unacceptable (and wasteful).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,9 +297,9 @@ commands:
             python -m tests.distribution_builders.<< parameters.distribution >>.cmd --dockerfile > ~/Dockerfile
             cat ~/Dockerfile
 
-      - restore_cache:
-          # here the 'tar' file of the docker image is stored. Use it to restore agent image without building it.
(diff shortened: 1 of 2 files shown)


Kami:
Confirmed it's working (although there is another issue which is simpler and I'm working on a fix) - https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/2128/workflows/ba5ef2f7-b098-466b-b1cd-e5596e2a5107/jobs/27236.

As per PR description, this is just a workaround and not a long term solution.

Disabling caching pretty much defeats the whole purpose of the caching framework we have in place.

I haven't had the time to think much yet how we could improve it. Part of the problem is how our Dockerfile is structured - building package depends on the agent_source/ directory which means we need to perform that step early on in the Dockerfile.

Kami:
And here is a fix for other build failure - 162659ace1a29bfc9f5892f1f333c560ee1e20c4.

Luckily that one was easy to fix. Sadly the original issue will need more thought and work put into it.

PR #477, opened 2020-03-27
Add packaging Dockerfiles for various distribution and Python version combos

Kami:
This pull request adds the following new distribution Dockerfiles:

* CentOS 6 with Python 2.6 (default setup)
* CentOS 6 with Python 3.5  (Python 2.6 and  Python 3.5)
* CentOS 7 with Python 2.7 (default setup)
* CentOS 7 with Python 3.6  (Python 2.6 and  Python 3.6)
* Centos 8 Python 2.7 and Python 3.6 - setup we run on the CI
* Ubuntu 14.04 with Python 2.6 (default setup)
* Ubuntu 14.04 with Python 3.5 (Python 2.6 and Python 3.5)
* Ubuntu 16.04 with Python 2.7 (default setup)
* Ubuntu 16.04 with Python 3.5 (Python 2.7 and Python 3.5)
* Ubuntu 18.04 with Python 2.7 and Python 3.7 - setup we run on the CI

Right now we only run package tests under Ubuntu 18.04 and CentOS 8.

I added those images so I can easily test various distro and Python version combinations for ``scalyr-switch-python`` functionality locally.

In addition to that, we can also use those Dockerfiles for other automated tests in the future (and we will need something like this for distro specific packages and tests in the future anyway).

For explicitness sake, I also renamed existing centos and ubuntu directories to centos8 and ubuntu1804 respectively and left symlinks in place. I already made a change to explicitly pin those Docker images to a specific version in #475. Previously we were using no tag aka latest which is not robust and would likely break once new distribution version is released.

# TODO

- [ ] Document available options in README
- [x] Try to get existing packaging tests to pass under all those combos (?)
- [x] Update affected Circle CI config jobs


Kami:
I've pushed couple of more fixes and improvements.

1. We now preserve ``agent.log`` file for each test function run

We run each test function inside a separate and dedicated container. Previous code for copying and preserving agent.log didn't take this into account so we only ended up preserving agent.log file from the last test run.

I updated it so we preserve ``agent.log`` from all the containers. Files are stored in ``<artifacts directory path>/<test function name>`` directory and available on Circle CI under artifacts tab.

2. Update CentOS 8 and Ubuntu 18.04 test case so we also preserve rpm and debian package we build for testing.

This should allow people to easily test those artifacts directly from every change which triggers ``package-test-*`` runs (right now that is merge / push to master).

Basically what you need to do is to navigate to Artifacts tab in Circle CI for ``package-test-{rpm,deb}`` Circle CI job in question - https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/2162/workflows/ccb6638e-2896-47c6-ada9-2c1ccefb6cbf/jobs/27642/artifacts, https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/2162/workflows/ccb6638e-2896-47c6-ada9-2c1ccefb6cbf/jobs/27641/artifacts

3. Added Dockerfile and made some changes so tests now also pass under Ubuntu 14.04.

Kami:
To wrap it up here as well.

That's the bug fix for ```rc*.d`` symlinks not getting created on newer versions of Ubuntu issue - ec8b13b.

It looks like sys-v and upstart are forgiving enough and don't care if BEGIN and END line only contains two ``##``, but that's not the case for systemd which ships on newer versions of Ubuntu.

Kami:
For now, I decided to leave ``package-test-{rpm,deb}`` jobs in place so they run for every PR / change.

This way we will be able to easily test packages produced by each PR.

The approach is not ideal because Docker image caching is currently broken (I needed to disable it last week since it has some bugs) so the build takes a while. The problem should be mitigating in the near future though.

I already talked to @ArthurKamalov last week and his next task will be fixing / improving the Docker image caching mechanism (since it affects all the new Docker based jobs which rely on agent source code and not just this one).

PR #485, opened 2020-03-31
Add double image caching and checksum generation.

ArthurKamalov:
The dependency preparation process for package testing has been split. Now there are two base images for each distribution image.
The first image - `fpm package builder` - image with installed `fpm` tool, the second - image of the particular distribution with preinstalled dependencies. Those two images now can be set as "required"  for the resulting image, so It will not be built without required images. 
How resulting image build works:

1. Import fpm builder image as fist build stage, add agent source and build the desired packages.
2. Import the second image (which is the base image for the current distribution) and just copy packages from the previous (fpm) build stage.  

Also, every image builder now can calculate its own checksum considering everything, including Dockerfile or any other file added to the build context. (For now, this is only Dockerfile and `dev-requirements.txt` file), so we can use it as a key for  CircleCI cache. (see `package-test` command in circleci config)


tests/utils/image_builder.py
```diff
@@ -59,13 +60,40 @@ class AgentImageBuilder(object):
 
     IMAGE_TAG = None  # type: six.text_type
     DOCKERFILE = None  # type: Path
+    REQUIRED_IMAGES = []  # type: List[ClassType[AgentImageBuilder]]
```
Kami: Please add a comment on what this variable is used for.

tests/utils/image_builder.py
```diff
@@ -59,13 +60,40 @@ class AgentImageBuilder(object):
...
+            self.add_to_build_context(
+                root_path,
+                "agent_source",
+                custom_copy_function=_copy_agent_source
+            )
+
+        # this part iterate through all attributes of the current class and get alla that starts with 'INCLUDE_PATH_'.
+        # the value of this attribute is the path to the file to be copied to the image build context.
+        for name, value in type(self).__dict__.items():
+            if name.startswith("INCLUDE_PATH_"):
```
Kami: I think it would probably be better to utilize a single class variable to make it a bit less magical and more explicit.


Perhaps ``INCLUDE_PATHS = []``. And if we need access to the name as well, we can either make item a dictionary or tuple.

Kami:
Please also add some documentation to the readme on how the caching system works and our image layout looks like.

PR #487, opened 2020-04-01
AGENT-364 Don't JSON encode journald messages

yanscalyr:


yanscalyr:
The issue was the code we pulled over from "emit_metrics" did json encoding so we kept it, I believe this kept the existing functionality but I could be wrong. One thing that we might want to add to this PR is also removing the encoding from the "extra fields" which are currently also json encoded, Im not sure about it myself. More changes to existing formats means more theoretical issues for existing users but I people probably don't want their data getting mangled.

Kami:
Let's please add some test cases for it.

czerwingithub:
@yanscalyr Yeah, we should really take this opportunity to make sure we test this and it looks right from what customers would like.  I suspect that yes, we probably don't want to encode the extra fields as well.  I also think that we might have to break the previous format so that we can do something a little more cleanly here.  And yes, we definitely need tests for this.

What I would suggest is that you gather some real examples of journald logs and use that to help write the tests.  We should do it in a unit test where we mock out the part where we fetch content from journald, instead returning examples of your captured journald logs.  Then the tests should compare what is written out to disk with the expected output.  That will also give us a chance to see what the expected output format looks like.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -557,16 +557,12 @@ def format_msg(
...
-        )
+        string_buffer.write("%s %s" % (metric_name, metric_value))
 
         if extra_fields is not None:
             for field_name in extra_fields:
                 field_value = extra_fields[field_name]
-                string_buffer.write(
-                    " %s=%s" % (field_name, scalyr_util.json_encode(field_value))
-                )
+                string_buffer.write(" %s=%s" % (field_name, field_value))
```
czerwingithub, line 565: For this, I think we might want to keep this as json_encode, but surround it with quotes.  For example,

```
" %s=\"%s\"" % (field_name, scalyr_util.json_encode(field_value))
```

I think we should do it that way because it simplifies customers having to write parsers to extract the extra fields values.  In your test cases below, you have an example where the output would be `timestamp=2015-09-05 13:17:04.944355`.  Because of the space, it's hard to tell where that key value pair ends.  Instead, with the change I'm proposing, it would be `timestamp="2015-09-05 13:17:04.944355"`.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -557,16 +557,12 @@ def format_msg(
     ):
         string_buffer = six.StringIO()
 
-        string_buffer.write(
-            "%s %s" % (metric_name, scalyr_util.json_encode(metric_value))
-        )
+        string_buffer.write("%s %s" % (metric_name, metric_value))
```
czerwingithub, line 560: Looking at these changes, a little worried what happens if a customer gets tripped up by these changes.  Unfortunately, I think we need a configuration option that allows them to go back to the original format, just in case.

We could make it configuration option like "format_version" and document this new format as version "2" and the original as "1".  We default to 2 but allow them to switch back.  That applies to this line and then the changes down below to how we emit the extra fields.

scalyr_agent/builtin_monitors/tests/journald_monitor_test.py
```diff
@@ -0,0 +1,254 @@
...
+                null_logger=True,
+                fake_clock=fake_clock,
+            )
+            monitor = manager.monitors[0]
+            monitor.log_manager.set_log_watcher(LogWatcher())
+
+            monitor.gather_sample()
+
+            self.assertLogFileContainsLineRegex(
+                "....\\-..\\-.. ..\\:..\\:..\\.....",
```
czerwingithub, line 147: Why do you need this regex?  Is it because the timestamp is non-deterministic?  It looks like you are trying to use a timestamp.  It would be better if we could just write out the whole line that should be produced by journald and check that explicitly here.

czerwingithub: Testing the exact line also guarantees that all of these extra fields are on the same line and that there are no fields other than the ones listed included in the output.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -557,16 +557,12 @@ def format_msg(
...
 
         if extra_fields is not None:
             for field_name in extra_fields:
```
czerwingithub, line 563: I know this isn't your doing, but we probably should have a deterministic ordering on the fields.. so the keys or something.

yanscalyr: Yeah, that was also the reason for one of the tests checking individual parts of the message rather than the whole thing, we can just sort them I guess.

scalyr_agent/builtin_monitors/tests/journald_monitor_test.py
yanscalyr: Timestamp is generated from the actual time, I can mock that out somehow I'm sure.

scalyr_agent/builtin_monitors/journald_monitor.py
yanscalyr: `json_encode` already adds quotes around it, so I think we might not need any change at all here.

scalyr_agent/builtin_monitors/journald_monitor.py
yanscalyr: One difference I noticed here is removing `json_encode` removes the quotes around the message. I would suggest leaving the surrounding quotes, but that brings up the question of what happens to parsing if the message has quotes in it, and if we `\` escape those we just come back to the issue this was all about...

czerwingithub:
Ok, Yan just record our conversation here.   My apologies because the spec for this has changed in response to what we think will actually solve the customer's problem.

So, we are going to implement two new options for the monitor: `detect_escaped_strings` and `emit_raw_details`.  Both should default to `False` which results in us emitting logs in the same way we are now.  If `detect_escaped_strings` is `True`, then for all emitted fields (details and the extra fields), detect if the value is already an escaped string (by just checking if we have a double quote at the start and end of the string -- I know that could lead to false positives), and if it is, then just emit the field raw without quoting and escaping from us.  If `emit_raw_details` is `True`, then no matter what, just emit the raw details string with no quotes around nor any escaping.

As you pointed out, we should try to make these per-logfile options using your journald extensions.  Yup, let's see if this is easy to do (less than an afternoon of work) and if so do that.  Otherwise, just make them be options on the monitor itself.

Additionally, be sure to provide a documentation example of how to set these values for the default case (i.e., show how to specify the `*.` journald configuration stanza).

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -576,9 +580,16 @@ def format_msg(
         if extra_fields is not None:
             for field_name in sorted(extra_fields):
```
czerwingithub, line 581: Is there an easy way to perform this sort just once, instead of every time we go to write a log message out?  Such as, when we read in the configuration, we perform the sort then?

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -562,12 +556,22 @@ def _process_entries(self):
...
 
-        if self._format_version == 2:
-            string_buffer.write('%s "%s"' % (metric_name, metric_value))
+        detect_escaped_strings = False
+        emit_raw_details = False
+        if log_config:
+            detect_escaped_strings = log_config.get("detect_escaped_strings", False)
+            emit_raw_details = log_config.get("emit_raw_details", False)
+
+        if emit_raw_details or (
```
czerwingithub, line 569: Let's remove some duplicate code here and below.  Let's have a method called `format_key_value(format, key, value, emit_raw, detect_ecaped_strings)`.  It returns a string.

For the details, we would invoke `format_key_value('%s %s', 'details', details_value, emit_raw, detect_escaped_strings)`.  For all other, we would just use the format string of `%s=%s`.  Does that make sense?  It would eliminate having to have two places with those double quotes.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -562,12 +556,22 @@ def _process_entries(self):
                 )
 
     def format_msg(
-        self, metric_name, metric_value, extra_fields=None,
+        self, metric_name, metric_value, log_config, extra_fields=None,
```
czerwingithub, line 559: Let's remove `metric_name` as an argument here and rename `metric_value` to `message`.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -562,12 +556,22 @@ def _process_entries(self):
...
     ):
         string_buffer = six.StringIO()
 
-        if self._format_version == 2:
-            string_buffer.write('%s "%s"' % (metric_name, metric_value))
+        detect_escaped_strings = False
+        emit_raw_details = False
+        if log_config:
+            detect_escaped_strings = log_config.get("detect_escaped_strings", False)
```
czerwingithub, line 566: Add comment here `If True, will attempt to see if the values are already escaped strings, and if so, not add in an additional escaping ourselves`.

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -562,12 +556,22 @@ def _process_entries(self):
...
+            emit_raw_details = log_config.get("emit_raw_details", False)
```
czerwingithub, line 567: Add comment here `If True, will always emit values as their original string value instead of attempting to escaping them.`

scalyr_agent/builtin_monitors/journald_monitor.py
```diff
@@ -135,6 +135,14 @@
     default=20 * 1024 * 1024,
 )
 
+define_config_option(
+    __monitor__,
+    "format_version",
+    "Version of the output format for journal entries.",
```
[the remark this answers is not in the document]
yanscalyr: I think you reviewed an outdated version of the code, since `format_version` is gone now.

scalyr_agent/builtin_monitors/journald_monitor.py
yanscalyr: The extra fields are gotten from each journal message so I don't think we have that option.

scalyr_agent/builtin_monitors/journald_monitor.py
yanscalyr: For the extra fields I think we discussed not taking into account `emit_raw_details`, should this still be the case you think?

scalyr_agent/builtin_monitors/journald_monitor.py
czerwingithub: Ok, never mind then.

scalyr_agent/builtin_monitors/journald_monitor.py
czerwingithub: Yes, sorry, meant to explicitly say that in my comment.  You coded it up right.. always passing in `False` there.

PR #490, opened 2020-04-08
Add monitors tests

ArthurKamalov:
Add test cases for the first tier of the agent monitors.

- mysql
- nginx
- syslog
- url
- shell

Additionally, now fpm builder image has a separate cache in CircleCI.


tests/image_builder/monitors/base/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ubuntu:18.04
```
Kami, line 1: As discussed in the past, would it perhaps make sense to utilize single image for all the monitor tests (we would simply bundle in all the dependencies we need)?

This would probably make it a bit easier to keep up to date. Since maintaining that many Dockerfile and related files can get quite messy.

tests/utils/log_reader.py
```diff
@@ -0,0 +1,166 @@
...
+            while not self._has_data:
+                self._start_cv.wait(0.1)
+        return
+
+    @property
+    def lines(self):
+        with self._lines_lock:
+            return list(self._lines)
+
+    def wait_for_new_line(self):
```
Kami, line 85: We should probably add support for timeout here, right?

In case there is a misconfiguration or test not working correctly or similar so we prevent tests from waiting / hanging util Circle CI job timeout (10 minutes) is reached.

tests/image_builder/monitors/base/Dockerfile
Kami: This way we would also avoid potential issues with single Circle CI job cache growing very large (since I assume we will run all the monitors tests as part of the same job, otherwise it would get to unwindy).

Kami:
For URL, shell and nginx monitor (and some others) we already have somewhat decent unit tests in place so perhaps we should skip those here and just focus on other monitors which we can't easily test as part of our regular unit tests (think monitors which depend on external services which can't be easily mocked - things such as redis, postgresql, mysql, etc/)?

The tests are not 100% the same and in theory end-to-end tests should be more robust, but problem with those is that we already have so many build jobs and builders that things are getting a bit hard to manage. On top of that, unit tests also finish much faster than those end to end ones.

.circleci/config.yml
```diff
@@ -309,12 +309,33 @@ commands:
...
 
       - restore_cache:
           # here the 'tar' files of the docker images is stored. Use it to restore agent image without building it.
           key: smoke-distribution-image-{{ .Branch }}-<< parameters.distribution >>-{{ checksum "~/checksum" }}-<< pipeline.parameters.cache_version >>
 
+
+      - run:
+          name: Copy cached images if they exist.
+          command: |
+            mkdir -p ~/agent_image
```
Kami, line 327: I think it would be better to move this into a script file in scripts/circle/ directory or similar.

I also plan to do this in the near future for some of the benchmark targets, which currently contain tons of in-line commands where a shell script file would be a better fit.

Kami:
@ArthurKamalov I pushed a change which enabled correct code coverage reporting for the agent which are started using subprocess - 045864ce3fd22642365a3ad27547a6bc10429fdf.

I verified it's working correctly.

As per my comment - we will still need to retrieve and combine coverage files for all the test function runs for all the test files.

Each test function runs in a new docker container and results in a new .coverage file.

At the end when we combine all the coverage files we will likely also need to do some processing since coverage data from different tests utilizes different source file paths and this
won't work correctly for combining process where the paths need to be the same.

---

On a relate note - the current log reader implementation either seems to broken or not fully completed yet (please confirm). To test those changes, I needed to comment out a bunch of that functionality since the implementation doesn't appear to be complemented and it's hanging the whole process (I didn't really have time to dig in yet why since it appears to be incomplete).

And as per my other comment, we also need a timeout to prevent very long hanging / blocking.

Kami:
As per discussion between @ArthurKamalov and I, I will also disable some of the Python 3 unit test and smoke standalone jobs to make sure we only run those tests under a single Python 3 version.

Our build matrix is already massive with tons of jobs so we should work on reducing them. And running those tests under a single Python 3 version on each push and master merge should be sufficient.

In the future, if the need arises, we can run some those additional jobs weekly / similar.

Kami:
As discussed with @ArthurKamalov, I updated all the jobs which produce coverage data to submit it directly to codecov.io.

This way we can outsource coverage report merging to codecov.io and simplify our build (get rid of the "coverage" job, etc.).

Sadly the change was quite painful and took a while to get everything working since the whole develop flow and feedback loop was slow (needed to wait for the whole build to finish for every small change I did).

Here you can see a list of jobs which submitted coverage data which was used for final combined report - https://codecov.io/gh/scalyr/scalyr-agent-2/commit/7b8ee2116f262381a40077f783e22cc65be13a47/build:

Screenshot from 2020-04-08 22-54-34

scripts/circleci/verify-circle-ci-config.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
ArthurKamalov, line 1: I guess this file should be named verify-codecov-config.sh?

Kami: Yep, thanks for catching this :)

Kami:
@ArthurKamalov thanks for going over my changes.

I decided to run monitor tests as part of each PR. The reason for that is coverage.

If we want accurate information on how each PR / branch affects coverage, we need to run all the Circle CI jobs which generate coverage as part of every PR.

I think this should be fine though since I got rid a bunch of other Python 3.x jobs so the overall aggregated run time won't be any longer.

ArthurKamalov:
@Kami Everything looks good to me. I appreciate your help, thanks.

PR #494, opened 2020-04-08
Small performance optimization for date parsing functions

Kami:
This pull request includes small performance optimization for ``rfc3339_to_datetime()`` and ``rfc3339_to_nanoseconds_since_epoch()`` function.

Instead of using relatively slow ``time.strptime`` it utilizes regex + datetime.datetime.

Based on the micro-benchmarking tests from WIP ``micro_benchmarking`` branch, this approach is ~2-3x faster.

In absolute terms, it doesn't seem like much, but when running the agent in a loaded scenario using Kubernetes / Docker monitor, it adds up since that function is called for every single log line (and the profiling data @imron provided also shows that we indeed spend non-trivial amount of time in those functions).

## Benchmarking Data / Results

Python 3

Screenshot from 2020-04-08 23-35-37

Python 2

Screenshot from 2020-04-08 23-36-16


scalyr_agent/util.py
```diff
@@ -89,6 +91,8 @@
 Original error: %s
 """.strip()
 
+RFC3339_STR_REGEX = re.compile(r"(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})")
```
czerwingithub, line 94: Can you add a comment above with an example string this matches.  I find that better grounds long regex

scalyr_agent/util.py
```diff
@@ -560,7 +574,7 @@ def rfc3339_to_datetime(string):
     return dt
 
 
-def rfc3339_to_nanoseconds_since_epoch(string):
+def rfc3339_to_nanoseconds_since_epoch(string, use_strptime=False):
```
czerwingithub, line 577: I take it you are using the benchmark test to cover functional testing as well?  Philosophically, I more learn towards also including a test case for this in the `util_tests.py` file mainly for consistency and discoverability.  If I modify this code, I would expect to find its test cases in `util_tests.py`.  I can understand though that it is duplication of effort / code if we went this way.  What are your thoughts?

scalyr_agent/util.py
imron: Also interestingly, for python3,  `\d` will match all Unicode digits (python2 `\d` only matches ascii by default) and I wonder if that will affect performance at all.

Kami: @czerwingithub I've added a comment in fae292682a03b96d5d7884590c0341c983898934.

Based on the benchmark results, it doesn't affect it. In fact, both versions (with and without strptime) are a bit faster under Python 3.

I also just tested it by adding ``re.ASCII`` flag and it appears to be tiny bit faster (~0.1 us, on the side of the statistical error) than the version without it.

scalyr_agent/util.py
Kami: I can also add it in there.

In fact, I'm still deciding how to do it. Pretty much the only reason I added this flag is so we can benchmark and compare the versions (to ensure correctness).

Now that we know that the new version is faster and correct, we could simply remove that flag all togethers.

What's your take on this?

I'm personally fine either way. Perhaps for now I can also add test case to util_tests.py and in the near future, we can remove that flag all together and simplify the code.

scalyr_agent/util.py
Kami: I tried yet another implementation which utilizes ``string.split`` which appears to be another couple of percents faster than regex one.

As per the code comment - I intentionally didn't assign split values to temporary variables (aka unpack it into tuple) which would make it more readable because it's slightly slower (and means we utilize a bunch of temporary very short lived variables).

Python 3 - regex approach:

Screenshot from 2020-04-08 23-35-37

Python 2 - regex approach:

Screenshot from 2020-04-08 23-36-16

Python 3 - string split approach

Screenshot from 2020-04-09 12-16-43


Python 2 - string split approach

Screenshot from 2020-04-09 12-17-18

scalyr_agent/util.py
Kami: I've added additional round-trip correctness test cases - 3a4df848570ce885976786c16b0de1d56de25727.

After some more thinking, I think the best approach is to have three functions - default one which utilizes the fastest approach (aka default public one) and then one which uses regex and one which uses striptime.

The last two would be "private" and only exported and used for testing, benchmarking and auditability purposes.

Kami: @imron suggested another approach - https://github.com/freach/udatetime

Since it's written in C, it's likely that it's faster, but need to verify it since our current implementation doesn't support timezones and udatetime does so there is likely some overhead related to timezone parsing and handling.

In any case, I will go with "multiple functions" approach so we can easily benchmark multiple implementations and compare one to another.

Oh and since that library is posix only, we will still need a fallback for platform and situations where that library is not available (Windows and other platforms where a pre-build wheel / egg is not available).

For out of the box installation, we can try to bundle the pre-built posix egg, but we still need a fallback regardless since that egg is platform specific and may not work in a lot of situations.

Kami:
@czerwingithub @imron I pushed another larger change - 90ba3218c276f830a80ab8048f1852c7a71da2b8.

We now have multiple versions of those functions, but really just two of them are used in real life. Others are there just for benchmark and comparison purposes.

And as per @imron suggestion, I also hooked up implementation based on ``udatetime`` library which appears to be the fastest.

As per my other comment on this PR, this approach depends on a 3rd party library with C extension which is not supported on all the platforms. Because of that, we fall back to the second fastest approach (string split one) if that library is not available.

For better organization, I also moved date parsing functions in a new module. I intentionally didn't refactor some of the duplicated code into additional utility functions to avoid overhead.

So there is some code duplication, but this shouldn't be a problem and we have good test coverage for all those functions.

---

Here are the results from local run:

Screenshot from 2020-04-09 15-36-24

Screenshot from 2020-04-09 15-35-35

Those benchmarks now also run as part of each CI run and the results are available here - https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/2327/workflows/18364813-f1c9-42c5-adfe-3d17a949f0c5/jobs/30243/parallel-runs/0/steps/0-108.

Histograms are also generated and uploaded as Circle CI job artifacts - https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/2327/workflows/18364813-f1c9-42c5-adfe-3d17a949f0c5/jobs/30243/artifacts.

scalyr_agent/date_parsing_utils.py
```diff
@@ -0,0 +1,465 @@
...
+    if parts[0].endswith("Z"):
+        parts[0] = parts[0][:-1]
+
+    # NOTE: We benchmarked multiple versions, including regex one and using string.split
+    # appears to be a bit faster than regex
+    try:
+        # NOTE: I intentionally access values directly in the list and don't assign them
+        # to intermediate variables since it's faster
+        result = parts[0].split("T")
+        result1 = result[0].split("-")
```
imron, line 168: For clarity, could we rename `result1` to something like `date_parts` and `result2` to something like `time_parts`?  This will make it easier to read this in 6 months time.

scalyr_agent/date_parsing_utils.py
```diff
@@ -0,0 +1,465 @@
...
+
+    # create a datetime object
+    # NOTE: We benchmarked multiple versions, including regex one and using string.split
+    # appears to be a bit faster than regex
+    try:
+        # NOTE: I intentionally access values directly in the list and don't assign them
+        # to intermediate variables since it's faster
+        result = parts[0].split("T")
+        result1 = result[0].split("-")
+        result2 = result[1].split(".")[0].split(":")
```
imron, line 378: Same comment above with regards to `split(".")`

scalyr_agent/date_parsing_utils.py
```diff
@@ -0,0 +1,465 @@
...
+        parts[0] = parts[0][:-1]
+
+    # create a datetime object
+    # NOTE: We benchmarked multiple versions, including regex one and using string.split
+    # appears to be a bit faster than regex
+    try:
+        # NOTE: I intentionally access values directly in the list and don't assign them
+        # to intermediate variables since it's faster
+        result = parts[0].split("T")
+        result1 = result[0].split("-")
```
imron, line 377: Same comment as above with respect to naming `result1` and `result2`.

scalyr_agent/date_parsing_utils.py
```diff
@@ -0,0 +1,465 @@
...
+
+
+if udatetime:
+    rfc3339_to_nanoseconds_since_epoch = _rfc3339_to_nanoseconds_since_epoch_udatetime
+    rfc3339_to_datetime = _rfc3339_to_datetime_udatetime
+else:
+    rfc3339_to_nanoseconds_since_epoch = (
+        _rfc3339_to_nanoseconds_since_epoch_string_split
+    )
+    rfc3339_to_datetime = _rfc3339_to_datetime_strptime
```
imron, line 465: Is there a reason we are using `_rfc3339_to_datetime_strptime` instead of `_rfc3339_to_datetime_string_split` here?

scalyr_agent/date_parsing_utils.py
Kami: That's what I originally did (unpacked it in a tuple and used more self explanatory names), but then I did some testing and noticed it had small overhead and results in tons of short-lived variables so I decided to remove it (see the code comment above).

scalyr_agent/date_parsing_utils.py
```diff
@@ -0,0 +1,465 @@
...
+        parts[0] = parts[0][:-1]
+
+    # NOTE: We benchmarked multiple versions, including regex one and using string.split
+    # appears to be a bit faster than regex
+    try:
+        # NOTE: I intentionally access values directly in the list and don't assign them
+        # to intermediate variables since it's faster
+        result = parts[0].split("T")
+        result1 = result[0].split("-")
+        result2 = result[1].split(".")[0].split(":")
```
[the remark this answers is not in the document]
Kami: I believe you are correct - that split is not needed.

scalyr_agent/date_parsing_utils.py
Kami: That was a typo.

Resolved in 2f7970c1a9a92e1ff37c74d8ac09033021884c75.

scalyr_agent/date_parsing_utils.py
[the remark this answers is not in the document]
Kami: Here we go - d2cfd726350d639ea44c7b442e6d039b0fe65dae.

Also covers the other comment.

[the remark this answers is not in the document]
Kami: And with that unnecessary split removed, split string version of ``_to_datetime`` function is now as fast as the udatetime version.

In fact, locally it sometimes finishes faster than udatetime version (although the difference is indeed very small).

Keep in mind this is only true for ``_to_datetime`` version of this function. For ``_to_nanoseconds_seconds_since_epoch`` udatetime is still faster.

The reason for that is that in this second version, we still need to call ``calendar.timegm`` which adds additional overhead which is not needed when using udatetime version.

This also shows that native Python implementation can sometimes also be faster than the C one :)

But this makes sense in this case since ``udatetime`` supports other more complex scenarios (timezones) which we don't need to support.

scalyr_agent/date_parsing_utils.py
Kami: d2cfd72.

Kami:
Just a note to anyone wondering why there is sometimes a big difference between histograms (mostly the max time and confidence bands) when running benchmarks under Python 2 and Python 3.

Under Python 2, ``time.time`` is used for timing and under Python 3 ``time.perf_counters`` is used.

Second approach (only available under Python 3) is more accurate since it usually utilizes ``time.process_time`` which only counts the CPU cycles and doesn't count wall clock time which includes time spend waiting on IO (which could be large on oversubcribed CI like scenarios and this causes large timing variations during tests under Python 2).

scalyr_agent/date_parsing_utils.py
imron: Sorry, I should have been more clear, I don't mean to unpack them, I just mean rename them, something like:

```
...
date_and_time = parts[0].split("T")
date_parts = date_and_time[0].split("-")
time_parts = date_and_time[1].split(":")
...
```

That way it's clear which variable holds which part of the timestamp.  Not to mention, `result1` and `result[1]` look very similar and are easy to confuse.

scalyr_agent/date_parsing_utils.py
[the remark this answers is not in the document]
imron: > And with that unnecessary split removed, split string version of _to_datetime function is now as fast as the udatetime version.

Nice! 👍  How close is the difference?  If it's not going have a significant impact on the hot path, I'd almost be inclined *not* to use udatetime, so as to keep the code simple and not to introduce another dependency.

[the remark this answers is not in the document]
Kami: In ``rfc3339_to_nanoseconds_since_epoch`` method the difference is 30% or so, so it's likely worth it.

In other method case, the difference is much smaller (just a couple of % faster), but I think it's worth to leave it in place because to the "to_nanosecond" method.

It will still work out of the box, if the library is not available we will just fall back to the native approach.

In short, we will follow the same approach as we do with ujson - we won't bundle it, but we will start a documentation page which will instruct advanced users how to squeeze out some additional performance by changing various config option and installing libraries such as ujson and u datetime.

imron:
Oh, I also had to make a small change to get things working on macOS, which doesn't have `librt`.  If we're running on Python < 3.3., then the change detects if we're running on macOS and if so uses `time.time` instead.

Kami:
Yeah, it's actually not a big deal if perf_counters / process_time is not used locally.

time.time should also be relatively OK as long as there are not tons of things running in the background when running micro benchmarks.

scalyr_agent/date_parsing_utils.py
```diff
@@ -0,0 +1,455 @@
...
+        to_micros = 6 - len(fractions)
+        micro = int(int(fractions) * 10 ** to_micros)
+        dt = dt.replace(microsecond=micro)
+
+    return dt
+
+
+def _rfc3339_to_datetime_udatetime(string):
+    """Returns a date time from a rfc3339 formatted timestamp.
+    We have to do some tricksy things to support python 2.4, which doesn't support
```
czerwingithub, line 402: I don't think this comment matches with what the function is doing -- I don't think it relies on `strptime`

scalyr_agent/date_parsing_utils.py
```diff
@@ -0,0 +1,455 @@
...
+        to_micros = 6 - len(fractions)
+        micro = int(int(fractions) * 10 ** to_micros)
+        dt = dt.replace(microsecond=micro)
+
+    return dt
+
+
+def _rfc3339_to_datetime_string_split(string):
+    """Returns a date time from a rfc3339 formatted timestamp.
+    We have to do some tricksy things to support python 2.4, which doesn't support
```
czerwingithub, line 346: I don't think this comment matches with what the function is doing -- I don't think it relies on `strptime`

scalyr_agent/date_parsing_utils.py
```diff
@@ -0,0 +1,455 @@
...
+        to_micros = 6 - len(fractions)
+        micro = int(int(fractions) * 10 ** to_micros)
+        dt = dt.replace(microsecond=micro)
+
+    return dt
+
+
+def _rfc3339_to_datetime_regex(string):
+    """Returns a date time from a rfc3339 formatted timestamp.
+    We have to do some tricksy things to support python 2.4, which doesn't support
```
czerwingithub, line 299: I don't think this comment matches with what the function is doing -- I don't think it relies on `strptime`

scalyr_agent/date_parsing_utils.py
Kami: Addressed in #498 - 8fe40d2426d275315fb0d0ed861b5957e09940fb.

scalyr_agent/date_parsing_utils.py
Kami: Addressed in #498 - 8fe40d2426d275315fb0d0ed861b5957e09940fb.

scalyr_agent/date_parsing_utils.py
Kami: Addressed in #498 - 8fe40d2426d275315fb0d0ed861b5957e09940fb.

PR #495, opened 2020-04-09
[WIP] fixing issues with #466

yanscalyr:


yanscalyr:
Seems like the issue was caused but a lack of filtering on the stdout handler, our other handlers output only if its not a metric and if the record's `agent_logger` value is True. Adding that to the stdout handler causes things to work.

I can't figure out how not having it was causing the issues we saw though...

scalyr_agent/scalyr_logging.py
```diff
@@ -1025,10 +1043,15 @@ def filter(self, record):
         @return:  True if the record should be logged by this handler.
         @rtype: bool
         """
-        return getattr(record, "force_stdout", False)
+        return getattr(record, "force_stdout", False) or (
+            self.__no_fork
+            and record.levelno >= self.__stdout_severity
+            and getattr(record, "agent_logger", False)
+            and record.metric_log_for_monitor is None
```
Kami, line 1050: Would it be possible for ``metric_log_for_monitor`` to ever not be present on the record object?

If so, we should probably also use getattr to be on the safe side.

yanscalyr: This is how its done for our other handlers and we haven't run into an issue yet, so I assume no. That would be a good fix to include regardless, in all the filters.

Kami: It appears this logic is duplicated from ``AgentLogFilter`` class.

Couldn't we just do what we did before (aka always add ``AgentLogFilter`` to every handler - https://github.com/scalyr/scalyr-agent-2/pull/495/files#diff-8c147743752a64d6bd12631abeacf3cbL1622)?

This way we would preserve separation of concerns.

I think it should be safe to do that and that's also the quick fix I tested the other day and it appeared to have worked.

yanscalyr: That filter also filters out debug messages (or non-debug messages depending on construction), and we do want stdout to output both.

yanscalyr: We could split out the debug filtering part to its own filter, separation of concerns seems to imply that would be better.

Kami:
While at it - can you please also add a corresponding changelog and release notes entry?

In the release notes document, we should be more verbose and describe what has changed, what's the new behavior when running with ``--no-fork`` flag (with some examples) and how users can control it using ``SCALYR_STDOUT_SEVERITY`` environment variable.

scalyr_agent/scalyr_logging.py
Kami: One option also is to leave it as is - just add a comment on why we need it here and that the same logic is located in ``AgentLogFilter`` class.

Splitting it into additional DebugLogLevel filter would be ideal, but this may add additional overhead (likely very small one) so if we do that change, it would be good to also add a micro benchmark for it.

PR #497, opened 2020-04-10
Fix copying to cache in circleci

ArthurKamalov:
Fix script that copies image files from the cache and refactors distribution names.


scripts/circleci/copy-docker-images-to-cache.sh
```diff
@@ -16,14 +16,27 @@
 # Script which copies Docker images created during test runs to directory which
 # is cached
 
-if [ -f "~/agent_image/scalyr-agent-testings-fpm_package-builder" ]; then
```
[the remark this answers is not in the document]
Kami: Yeah, that makes sense.

When I was working on fixing it, I was simply too tired and just wanted to get the build to pass.

If I was under less pressure and had more time to work on it, I would likely use a different and more robust approach all together.

Personally, when I work with paths, I pretty much almost always try to resolve from from relative to full absolute paths early on and then work with absolute paths everywhere.

This usually makes it more robust and also prevents issues like this one.

scripts/circleci/copy-docker-images-to-cache.sh
```diff
@@ -16,14 +16,27 @@
...
+    echo "Copy image 'scalyr-agent-testings-monitor-base' to cache."
     cp  ~/agent_image/scalyr-agent-testings-monitor-base ~/monitors_builder_cache/scalyr-agent-testings-monitor-base
     rm ~/agent_image/scalyr-agent-testings-monitor-base
+else
+  echo "Image 'scalyr-agent-testings-monitor-base' is not found."
 fi
 
-cp ~/agent_image/* ~/agent_image_cache/
+# the agent distribution image name should match to this glob.
+if [ "$(ls -A ~/agent_image/scalyr-agent-testings-distribution-*-base)" ]; then
```
Kami, line 37: Yeah this ensures there are no potential left-overs from other images.

I was also thinking on how to improve that, but really didn't have the time to get to it yet.

At some point we should probably also use constants for some of that stuff. And we also need to make sure we document that in our caching docs :)

tests/image_builder/distributions/base/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+BASE_AGENT_DISTRIBUTION_PREFIX = "scalyr-agent-testings-distribution"
+
+
+def create_distribution_image_name(
+    distribution_name,
+):  # type: (six.text_type) -> six.text_type
+    return "{0}-{1}".format(BASE_AGENT_DISTRIBUTION_PREFIX, distribution_name)
+
+
+def create_distribution_base_image_name(
```
Kami, line 30: Both of those functions return the same value for the same input. Is this intentional?

Kami:
@ArthurKamalov I think adding a new base Circle CI command for monitors smoke tests is probably the best approach.

I simply utilized  the existing one because it was the fastest and it reduces some duplication, but yes, that approach is not ideal since we are conflating two different types of tests and caches.

In fact, eventually splitting it out to a separate command was my long term plan, but it looks like we should probably just do it at this point to avoid more surprises :)

ArthurKamalov:
@Kami I added IMAGE_CACHING readme file and also enabled parallel execution of the test cases by `pytest-xdist`.

tests/image_builder/distributions/base/__init__.py
ArthurKamalov: they do not return the same. `create_distribution_base_image_name` returns "*-base" :)

.circleci/config.yml
```diff
@@ -297,115 +297,32 @@ commands:
...
-          paths:
-           - "~/agent_image_cache"
-
-      - save_cache:
-          name: "Save Python Dependencies Dache"
-          key: deps2-tox-{{ .Branch }}-3.6-venv-{{ checksum "dev-requirements.txt" }}
-          paths:
-            - "~/.cache/pip"
-
+            tox -e << parameters.target_name >> -- --no-rebuild --artifacts-path ~/artifacts -n 2
```
Kami, line 325: Can you please provide some data on how running tests in parallel affects the job timing?

.circleci/config.yml
```diff
@@ -435,9 +351,163 @@ commands:
...
 
-      # NOTE: We store the logs to ease with troubleshooting / debugging
-      - store_artifacts:
-          path: ~/artifacts
+            # NOTE: We store the logs to ease with troubleshooting / debugging
+            - store_artifacts:
+                path: ~/artifacts
+
+  init-cached-docker-image:
+    parameters:
```
Kami, line 359: Please add description attribute here which describes what that command does / what kind of purposes it serves.

.circleci/config.yml
```diff
@@ -435,9 +351,163 @@ commands:
...
+          paths:
+            - << parameters.cache-path >>
+
+      - run:
+          name: "Remove image checksum file."
+          command: |
+            rm -f ~/image_checksum
+
+  install-cached-tox:
+    parameters:
```
Kami, line 396: Please add description attribute here which describes what that command does / what kind of purposes it serves.

.circleci/config.yml
```diff
@@ -435,9 +351,163 @@ commands:
...
+          image-builder-path: tests.image_builder.distributions.<< parameters.distribution >>.cmd
+          python_executable_path: ./.tox/<< parameters.tox_target >>/bin/python
+
+      - run-tox-target:
+          agent_host_name: << parameters.agent_host_name_prefix >>-<< parameters.distribution >>-<< parameters.python_version >>
+          target_name:  << parameters.tox_target >>
+          upload_coverage: << parameters.upload_coverage >>
+
+  monitors-tests:
+    description: "A base command for package smoke tests. It invokes tox / shell command with name '<parameters.command>'"
```
Kami, line 472: Please update description attribute so it matches the reality :)

Kami:
Thanks for working on the readmes, etc.

As per my comment the other day - we need to run each job which uploads coverage data as part of every PR so we get accurate coverage numbers for each PR / branch. That's why I made this change - a575aaf42a52b499fbf4f134125251585925710b.

Kami:
@ArthurKamalov when you get a chance, can you please look into the ``package-test-*`` test failures?

On that note - we should also look at improving the output (if possible) on failure. Current output provides no clue why the tests failed (e.g did the container even start? why it didn't? etc.).

It's important we provide as much output and context possible to make troubleshooting failures fast and efficient.

Kami:
Just a heads up - I pushed some changes^ which should make it easier to troubleshoot the failures.

Kami:
Confirmed all the jobs now pass. I also tested it with empty and cached populated scenarios.

Will go ahead and merge it once the new build passes.

PR #498, opened 2020-04-10
Small optimizations for Event serialization

Kami:
This pull request includes two small optimizations which affect performance of the event serialization.

1. First change optimizes ``__set_attributes`` code path by making last curly brace unwinding more efficient.

When serializing a dictionary, I believe we can safely assume that the last value of the serialized string will the the closing curly bracket.

2. Second changes optimizes JSON serialization by getting rid of sorting keys and using custom separator.

Those two things add quite a substantial overhead.

I believe API side doesn't rely on keys being sorted and custom separator so this change should be safe.

I believe the reason we sorted it is to make assertions in the tests easier - if that's the case, we should move that sorting to the test cases itself.

To put things into perspective. Not sorting the keys and using custom separator when using json.dumps, increases the throughput by 3-4 and about 0.5 when using ujson.

As with other micro optimization PR, actual outcome will be very workload specific. The more lines are being parsed per second and the more attributes those lines have, the bigger the impact.

## TODO

- [x] Ask @czerwingithub to confirm we indeed don't rely on key sorting and custom separator
- [x] Add micro benchmarks (already have some locally, but they are base on micro_benchmarks branch so I need to get #494 merged first and then I can add them here)


scalyr_agent/scalyr_client.py
```diff
@@ -1585,8 +1585,14 @@ def __set_attributes(self, thread_id, attributes):
                 tmp_buffer.write(b", ")
         if attributes_to_serialize:
             tmp_buffer.write(b"attrs:")
+            # NOTE: attributes is always a dict so we can simply rewinding logic by assuming }
+            # is always the last character when serializing a dict to json
+            tmp_buffer.write(scalyr_util.json_encode(attributes, binary=True)[:-1])
```
czerwingithub, line 1590: Hmm... yeah, the purist in me wants to verify that that character is a `}` but I imagine doing that check would make this code less readable and add in overhead in a place we don't want it.  You are right though that this should be very deterministic behavior for the JSON libraries.  If it's not the case, we should get test failures.

Maybe one thing we can do to make it more explicit is have a test case that verifies the last character for a serialized dict is always a '}'.  We have tests that verify common behavior across all JSON libraries.  We'd just add the test there commenting why it is important to us.

What do you think?  Overkill?

scalyr_agent/util.py
```diff
@@ -159,9 +159,9 @@ def json_dumps_custom(obj, fp):
 
             if fp is not None:
                 # Eliminate spaces by default. Python 2.4 does not support partials.
-                return json.dump(obj, fp, sort_keys=True, separators=(",", ":"))
+                return json.dump(obj, fp)
```
czerwingithub, line 162: I don't quite remember the spaces by default comment here.. But, I will say, any place where we were doing something to support Python 2.4 or 2.5, we can kill now.

scalyr_agent/scalyr_client.py
Kami: Yep, I agree about the test case. Will add it for just in case to be on the safe side.

Kami:
Here is some data on json.dumps performance with and without sort_keys.

Screenshot from 2020-04-11 20-48-59

Keep in mind that data I use for the benchmark is somewhat artificial and performance difference will be larger with actual event objects.

As mentioned in a comment, I also added ``sort_keys=True`` mode which we can use for tests (without that mode, it would be much more cumbersome to assert on the actual payload values).

On related note, I will also default json serialization library to ``orjson`` under Python 3 (if available), because as discussed above it's substantially faster than ``ujson`` in most scenarios.

scalyr_agent/scalyr_client.py
Kami: I added a test case and verified that the output is the same for both implementations.

Smoke / end to end tests also seem to pass, but we should still keep an eye on it once it's merged.

I think if there are no issues after another load test with this change included, we should be fine :)

scalyr_agent/scalyr_client.py
```diff
@@ -859,10 +859,16 @@ def __init__(
         # 2->TODO use BytesIO, make all data that is going to be written here - binary
         string_buffer = io.BytesIO()
         serialized_base_body = scalyr_util.json_encode(base_body, binary=True)
-        string_buffer.write(serialized_base_body)
 
         # Now go back and find the last '}' and delete it so that we can open up the JSON again.
+        # NOTE: attributes is always a dict so we can simply rewinding logic by assuming }
+        # is always the last character when serializing a dict to json
+        string_buffer.write(serialized_base_body[:-1])
+        # Old implementation
```
imron, line 867: I think we should remove the old implementation from the code.

scalyr_agent/tests/util/json_util_test.py
```diff
@@ -109,6 +116,18 @@ def __runtest(library):
                 )
                 obj3 = util.json_decode(text)
                 self.assertEquals(obj3, obj)
+
+                # Sanity test to ensure curly brace is always the last character when serializing
+                # a dict
```
czerwingithub, line 121: I would explicitly say something like "our rewind logic in the scalyr client relies on this."

scalyr_agent/scalyr_client.py
Kami: Removed.

scalyr_agent/tests/util/json_util_test.py
Kami: Done - ce5154dcd0e316a3ad3803983a332e56c23e2667.

PR #499, opened 2020-04-11
Fix unit test failures on OS X

Kami:
This pull request fixes two potential test failures on OS X.

In first case, the problem was that depending on how tests run, the test would load and read ``.pyc`` instead of ``.py`` file.

This issue would only manifest itself when running tests under different Python versions (e.g. .pyc file was generated under one version and tests run under a different version).

And in other case, we simply skip Linux specific tests on OS X.

### DIFF
--- a/scalyr_agent/builtin_monitors/tests/linux_process_metrics_test.py
+++ b/scalyr_agent/builtin_monitors/tests/linux_process_metrics_test.py
@@ -17,17 +17,20 @@ from __future__ import absolute_import
 
 import os
 import sys
+import platform
+
+import mock
 
 from scalyr_agent.builtin_monitors.linux_process_metrics import ProcessMonitor
 from scalyr_agent.scalyr_monitor import load_monitor_class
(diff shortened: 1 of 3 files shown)


scalyr_agent/builtin_monitors/tests/linux_process_metrics_test.py
```diff
@@ -17,17 +17,20 @@
...
 from scalyr_agent.test_base import ScalyrTestCase
-
-import mock
+from scalyr_agent.test_base import skipIf
 
 __all__ = ["LinuxProcessMetricsMonitorTest"]
 
 
 class LinuxProcessMetricsMonitorTest(ScalyrTestCase):
+    @skipIf(platform.system() == "Darwin", "Skipping Linux Monitor tests on OSX")
```
Kami, line 33: NOTE: The custom decorator we use which supports old Python versions doesn't support decorator the whole class so we need to do it on the method level.

Kami:
Coverage dropped because of the monitors tests issues which will be resolved by #497.

PR #500, opened 2020-04-13
Run unit tests on Mac OSX

Kami:
This pull request updates Circle CI config so we also run unit tests on OS X using Python 3.5.

We don't officially support running agent on the OS X (although it more or less works out of the box), but some developers use OS X for development so we should also run at least unit tests on OS X to detected issues like the one in #499 early.


Kami:
When the cache is populated, the build takes ~1.5 minutes, but if it becomes an issue, we can always only run this job on merge to master.

PR #502, opened 2020-04-14
Fix bug in environment defined monitor config params

czerwingithub:
## Background

This fixes a very specific bug in defining monitor config params via environment variables.  However, the bug really shines a spotlight on the fact that we really need to simplify our monitor configuration support and eliminate providing parameters like `default` and `convert_to` on `get` calls to the `MonitorConfig` object.  It means that you can define those options in two places (in a `define_config_option` call and `MonitorConfig.get`) which creates problems in the code.  I got half-way down the path of moving everything off of `get` but realized it would be too risky of a change without more better monitor testing.

The bug this fixes occurs when:
- You define a configuration parameter using `define_config_option`
- You do not specify the option as required or has a default
- You do specify some other transformation such as a `convert_to`
- You supply the variable through the environment.

The result is the `convert_to` is not applied to the environment variable supplied value.  

I'm going to take it as a TODO to start moving calls off of `get`.


scalyr_agent/tests/scalyr_monitor_test.py
```diff
@@ -278,6 +279,94 @@ def test_list_of_strings(self):
             ["x", "y", "z"],
         )
 
+    def test_environment_based_options(self):
+        # Basic case
+        self.assertEquals(
+            self.define_and_get_from_env(
+                "value in env", default_value="default value", convert_to=six.text_type
```
Kami, line 286: Would also be good to add a test case for scenario where the environment variable value is an empty string.

I don't think empty string is valid value for any option, but just in case.

On that note, in bash is somewhat common to do ``FOO= myscript.sh`` if you want to clear / unset the environment variable for a particular command run (this way you don't need to explicitly call ``unset FOO``).

In our case, right now we still consider variable as set even if it's an empty string, which is probably fine.

czerwingithub: Ok, I added a test case for the empty environment variable value.

I did actually start to implement the change to treat an empty value as not being set.  However, I fortunately tripped over a breaking test case that reminded me there's one "valid" use of an empty string in config options.

We have an option called `k8s_namespaces_to_ignore` which is suppose to be a comma separated list of namespaces for which we shouldn't collect logs.  If you want to not ignore any namespaces, you need to set an empty string which is then treated as an empty list.

It would be nice to have empty string treated as not set, but I think you are agreeing it's also not that big of a deal if we don't in our use cases.

czerwingithub:
@Kami If could take another look when you get a chance please.

Kami:
Probably a good idea to also add a changelog entry. Besides that, LGTM :+1:

PR #510, opened 2020-04-19
Fix PYTHONPATH for agent benchmarks Circle CI job

Kami:
This pull request fixes PYTHONPATH for benchmarks Circle CI job (it was broken accidentally by #489).

In addition to that, it also fixes CodeSpeed executable name (previous one was too long).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -196,7 +196,7 @@ commands:
           # and are handles as secrets: CODESPEED_AUTH
           environment:
             CODESPEED_URL: "https://scalyr-agent-codespeed.herokuapp.com/"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -218,6 +218,7 @@ commands:
             # branch / revision
             export TZ=UTC
             export COMMIT_DATE=$(git show --quiet --date='format-local:%Y-%m-%d %H:%M:%S' --format="%cd" ${CIRCLE_SHA1})
+            export PYTHONPATH=.
```
Kami, line 221: This is needed because I changed from relative to package based imports.

PR #512, opened 2020-04-20
Fix submit_to_codespeed decorator so it works under Python 2

Kami:
This pull request fixes ``submit_to_codespeed`` decorator so it also works correctly under Python 2.

pytest does some weird stuff with detecting original decorators functions so using ``functools.wraps`` doesn't work out of the box under Python 2.7 (pyttest can't detect original wrapped function so the benchmarks doesn't run).

### DIFF
--- a/benchmarks/micro/conftest.py
+++ b/benchmarks/micro/conftest.py
@@ -14,10 +14,10 @@
 
 from __future__ import absolute_import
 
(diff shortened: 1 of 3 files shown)


Kami:
Confirmed that's it's working on CodeSpeed instance.

PR #515, opened 2020-04-22
Remove monotonically increasing timestamps from the agent

imron:
This PR removes code from the agent that ensures all logs have a monotonically increasing timestamp.

Now, if the agent is in standalone mode, timestamps will be set to the current time when reading bytes from disk, or if the agent is in k8s, docker or any other mode that parses lines as json, then the agent will try to extract a timestamp from the json and use that.

This has been tested manually using a script that generates logs with timestamps offset from the current time, and/or timestamps that go backward and forwards in time.

### DIFF
--- a/scalyr_agent/scalyr_client.py
+++ b/scalyr_agent/scalyr_client.py
@@ -39,7 +39,6 @@ import six.moves.http_client
 
 from scalyr_agent.util import verify_and_get_compress_func
 
-
 # noinspection PyBroadException
 try:
     import ssl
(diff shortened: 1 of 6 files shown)


scalyr_agent/scalyr_client.py
```diff
@@ -220,6 +233,9 @@ def __init__(
         # flag to disable new addEvents format, TODO: remove this when we are confident it works
         self.__disable_logfile_addevents_format = disable_logfile_addevents_format
 
+        # whether or not to monotonically increase event timestamps
```
czerwingithub, line 236: "whether or not to enforce monotonically increasing event timestamps within the same session."

PR #516, opened 2020-04-22
Added pylint checkers to enforce necessary py2exe invariants

czerwingithub:
## Overview

This adds a new pylint checker that enforces some necessary invariants for py2exe to work properly when building the Windows binary.  This seemed like the easiest path to have checks for these invariants (these weren't the type of things you can check via unit tests.. at least not easily).

The invariants enforced are:
1.  Any modules imported using `six.moves` must be included in a dict of known imports defined in `setup.py`.  
2.  Any monitor modules must be included in a list of known monitors defined in `setup.py`.

In both cases, we are doing this because `py2exe` cannot properly apply its static dependency analysis to determine what modules should actually be included in the resulting Windows binary.  Both `six.moves` and our monitors use dynamic importing which just confuses `py2exe`.

## Suggested review path

Here's the order I would follow:
* `pylint_plugins/py2exe_checker.py` to see how the checker is implemented and for more details.  * `setup.py` to see how the dict and list are used to generate the Windows binary.
* `.pylintrc` for how I modified the pylint to work more reliably across pre-commit and tox (see below for details of an error I believe I found)
* `tox.ini` and `.pre-commit-config.yaml` for necessary modifications to use new approach
* all other files are just disabling the new checker warnings where appropriate

## Disabled warnings to keep status quo

There are some monitors that we are not including in the Windows binary that maybe we should (such as mysql).  However, for now, I did not add them to the list of monitors to include to just keep the status quo.  I added the appropriate warnings to reflect this.

## Found issue with pylint in pre-commit.

In the `tox.ini` use of `pylint`, we are currently loading the `bundled_imports_checker`.  However, in `pre-commit-config.yaml`, we were not.  I don't know if that was intentional or not, but I figure it wasn't.  To help avoid this problem in the future, I defined what checkers to load in the central `.pylintrc` file.  This created another issue.. in `tox`, we always add the source directory into the PYTHONPATH... but for `pre-commit`, this is not always the case.  So, I added some compensation code in `.pylintrc` to add in the current directory to PYTHONPATH assuming that we always run pylint from the top of the source directory.  This ensures it works across precommit and tox.

## Alternative approaches

I thought a pylint checker was the way to go with this because it would have been hard to check it via unit tests.  I still think this is pretty much true.. but yeah, I could have done something like modify the `six.py` file to throw an exception on any import via `six.moves` that wasn't from our list.  But, that would mean we would have test cases that exercises all possible imports... which, with our low test coverage of plugins, seemed not safe.

In the end, the pylint checker would have been quick to write if I had done one before.. but I spent a moderate amount of time playing with it and figuring out the right way to implement it.. and then also making tox and precommit a little more robust.


lint-configs/python/.pylintrc
```diff
@@ -22,6 +22,11 @@ disable=C0103,C0111,I0011,I0012,W0704,W0142,W0212,W0232,W0613,W0702,R0201,W0614,
 
 [MASTER]
 ignore = third_party,third_party_python2,third_party_tls
+load-plugins=pylint_plugins.bundled_imports_checker,pylint_plugins.py2exe_checker
```
Kami, line 25: That's good adding this here :+1:

tox.ini
```diff
@@ -78,7 +78,7 @@ commands =
     bash -c 'black --check --config pyproject.toml {env:LINT_FILES_TO_CHECK}'
     python .circleci/modernize/modernize.py -j 2
     bash -c 'flake8 --config lint-configs/python/.flake8 {env:LINT_FILES_TO_CHECK}'
-    bash -c 'pylint -E --rcfile=./lint-configs/python/.pylintrc --load-plugins pylint_plugins.bundled_imports_checker {env:LINT_FILES_TO_CHECK}'
```
Kami, line 81: Yeah, it's indeed much better we centralize the settings inside the shared lint configs (where possible) :+1:

setup.py
```diff
@@ -48,213 +48,255 @@
...
-    # },
-    console=[
-        {
-            "script": path.join("source_root", "scalyr_agent", "agent_main.py"),
-            "dest_base": "scalyr-agent-2",
-        },
-        {
-            "script": path.join("source_root", "scalyr_agent", "config_main.py"),
-            "dest_base": "scalyr-agent-2-config",
+    setup(
```
Kami, line 205: Is ``setup()`` function now intentionally nested under ``perform_setup()`` function?

If so, I wonder if this will work under all scenarios since ``setup()`` should be directly importable.

czerwingithub: Yeah, I wasn't quite sure if this was the right way to do this or not.  As you can see from the rest of the PR, I need to be able to import the `WINDOWS_MONITOR_MODULES_TO_INCLUDE` and `WINDOWS_PY2_SIX_MOVES_IMPORTS` in the checker.  However, I was worried about the side effect of loading the setup module with how it was currently written.  For example, we make a copy of VERSION.  And, I didn't quite understand the impact of invoking that `setup` method.

I took this approach because, to the best of my knowledge, we always use `setup.py` as a script -- `python setup.py XXX`.  So, by triggering off of `__name__ == "main"` I thought I would preserve the current behavior.

But, if this isn't the right way to do this, I'm more than happy to change it has well.  You understand this better than I do.

Kami: One option would also be to move those constants to another file and then reading it in setup.py and in the other Python file where they are used (we just need to be careful that the constants file doesn't import any other things from scalyr_agent since that won't be available in setup.py before install phase).

For now, I think we can leave this change as-is and in the future, if we encounter any issues, we can change it.

And as far running it as script goes - yeah that's indeed true for us, but some (packaging) tools rely on importing it (which is less of an issue these days with wheels, but still).

czerwingithub: Ok, thanks for that comment.  I'll go ahead and submit this as-is and we'll change it if we need it.  I suspect we will be taking a hard look at `setup.py` when we move to freezing.

PR #517, opened 2020-04-23
Add support for additional compression algorithms (lz4, zstandard)

Kami:
This pull request adds support for lz4 and zstandard request body compression algorithm to the agent.

To hook this up I also introduced new per-algorithm specific default value for compression_level if that configuration option is not explicitly specified by the end user.

This is important, because different default values make sense for different algorithms.

Keep in mind that this is just the client side part of the changes.

Getting it done end to end will involve changes in multiple places:

- Adding support for those two algorithms to the backend / server side
- Testing it on qatesting first
- ...

Eventually, once those changes are deployed and running in prod for a while we should likely default to a more efficient algorithm (zstandard) in a feature version of the agent.


Kami:
The Java implementation we currently use on the server-side doesn't support block dependence mode for LZ4 which means we need to use block independent mode (compression dictionary is not shared among blocks).

It's not the most ideal thing since it affects compression ratio, but where compression ratio is more important then speed, we should use a different algorithm anyway (e.g. zstandard).

> Block Independence flag
>
> If this flag is set to “1”, blocks are independent. If this flag is set to “0”, each block depends on previous ones (up to LZ4 window size, which is 64 KB). In such case, it’s necessary to decode all blocks in sequence.
> 
> Block dependency improves compression ratio, especially for small blocks. On the other hand, it makes random access or multi-threaded decoding impossible.

https://android.googlesource.com/platform/external/lz4/+/HEAD/doc/lz4_Frame_format.md

Kami:
Here is an example output of some of the new log messages.

Right now we only log them under debug, but it would probably also be useful (help us with troubleshooting things and get better insight on agent / client side compression ratio) by periodically also logging it at the INFO log level.

```bash
2020-04-29 13:30:55.397Z DEBUG [core] [scalyr_client.py:373] Compressed add event request data using "lz4" algorithm and level "9": original_size=1048535 compressed_size=95382 compression_ratio=10.993007066322786 duration=0.0059s
2020-04-29 13:31:00.613Z DEBUG [core] [scalyr_client.py:373] Compressed add event request data using "lz4" algorithm and level "9": original_size=1048341 compressed_size=76075 compression_ratio=13.780361485376273 duration=0.0052s
2020-04-29 13:31:05.811Z DEBUG [core] [scalyr_client.py:373] Compressed add event request data using "lz4" algorithm and level "9": original_size=1048363 compressed_size=83299 compression_ratio=12.585541242992113 duration=0.0056s

2020-04-29 13:31:25.331Z DEBUG [core] [scalyr_client.py:373] Compressed add event request data using "zstandard" algorithm and level "3": original_size=1048306 compressed_size=45119 compression_ratio=23.23424721292582 duration=0.002s
2020-04-29 13:31:30.541Z DEBUG [core] [scalyr_client.py:373] Compressed add event request data using "zstandard" algorithm and level "3": original_size=1048535 compressed_size=54894 compression_ratio=19.101085728859257 duration=0.0012s
2020-04-29 13:31:35.733Z DEBUG [core] [scalyr_client.py:373] Compressed add event request data using "zstandard" algorithm and level "3": original_size=1048440 compressed_size=50295 compression_ratio=20.845809722636446 duration=0.001s


2020-04-29 13:32:31.613Z DEBUG [core] [scalyr_client.py:373] Compressed add event request data using "deflate" algorithm and level "9": original_size=1048540 compressed_size=74190 compression_ratio=14.13317158646718 duration=0.012s
2020-04-29 13:32:36.836Z DEBUG [core] [scalyr_client.py:373] Compressed add event request data using "deflate" algorithm and level "9": original_size=1048509 compressed_size=68492 compression_ratio=15.308488582608188 duration=0.0131s
```

As per benchmarks and as expected, current default deflate algorithm is indeed the slowest.

CHANGELOG.md
```diff
@@ -1,6 +1,16 @@
 Scalyr Agent 2 Changes By Release
 =================================
 
+## 2.1.3 "TBD" - April 30, 2020
```
czerwingithub, line 4: Hmm.. We shouldn't merge this commit until after the 2.1.3 release is cut.  Also, we have to be careful about releasing a version of the agent that supports these compression algorithms until the server-side is deployed to production.

scalyr_agent/scalyr_client.py
```diff
@@ -489,6 +518,11 @@ def __receive_response(self, body_str, send_time):
                     )
                 return "requestFailed", len(body_str), response
 
+            try:
```
czerwingithub, line 521: Can you add a comment why it's ok to ignore the exception.

scalyr_agent/util.py
```diff
@@ -99,6 +104,39 @@
 # serialized values easier.
 SORT_KEYS = False
 
+# Maps user-friendly string we expose in the configuration to the internal Python module name
+SUPPORTED_COMPRESSION_ALGORITHMS = [
+    "deflate",
+    "bz2",
+]
+
+# lz4 and zstandard library is not available for Python 2.5
```
czerwingithub, line 113: I think you mean Python 2.6

scalyr_agent/scalyr_client.py
```diff
@@ -176,15 +176,28 @@ def __init__(
         }
 
         # Configure compression type
+        self.__compression_type = compression_type
+
         self.__compress = None
         encoding = None
 
+        if compression_level < 1 or compression_level > 9:
```
czerwingithub, line 184: Don't lz4 default to zero?  I think this would end up turning that to 9?

CHANGELOG.md
Kami: The idea is to keep changelog always up to date / in sync with all the changes (aka so before the release we just do a quick glance over it and make minor modifications).

The only reason I used 2.3.1 here is because of the build_package.py script which expects a header to be in a specific format.

Otherwise, I would just use ``## in development`` or similar and also skip "packaged by" line until we are ready for an actual release.

Having said that, why shouldn't we merge this commit before then? IIRC, we only build new staging / unstable packages on release branch via manual trigger.

Or am I missing something?

scalyr_agent/scalyr_client.py
Kami: Yep, good catch - I will change this and move it to a configuration layer where we validate those settings.

Kami: Here we go - a09e62521f406da5ec6e1761d515eaf9a31f9b9b.

In ideal world, we would throw (I was never a fan of software silently accepting invalid values), but for backward compatibility reasons, I use a per compression algorithm specific default if an invalid value is specified in the config.

scalyr_agent/util.py
Kami: Yep, fixed - d7f2fada34691dc5010e7839994c2e6474ab1fc6.

scalyr_agent/scalyr_client.py
Kami: Here we go - 61a262fc7ff7846f2a8d7a7038460ed2cfbb9ba2.

And for completeness, here it's how it looked before this change / fix:

```bash
2020-04-30 09:38:46.303Z Level 4 [core] [scalyr_client.py:523] Response was received with body "b'{"bytesCharged":693755,"status":"success"}'"
```

And after it:

```bash
2020-04-30 09:38:46.303Z Level 4 [core] [scalyr_client.py:523] Response was received with body "{"bytesCharged":693755,"status":"success"}"
```

scalyr_agent/scalyr_client.py
```diff
@@ -345,7 +341,31 @@ def __send_request(
...
+
+                start_time = time.time()
+                body_str = self.__compress(body_str)
+                end_time = time.time()
+
+                size_after_compress = len(body_str)
+                compression_ratio = float(size_before_compress) / size_after_compress
+                duration = round((end_time - start_time), 4)
+
+                # TODO: Should we sample this log message aka only log and perform the calculations
```
czerwingithub, line 354: Eh, I think this is probably fine.  You can remove the TODO.

Kami:
For completeness, I also tested this with larger max allowed body size (``max_allowed_request_size: 5900000``).

As expected (the larger the request the bigger the gains should be), zstandard is indeed much faster and more efficient (especially when comparing it against bz2 with level 9 which is our default):

```bash
2020-05-04 12:35:24.480Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "deflate" algorithm and level "9": original_size=5899921 compressed_size=394135 compression_ratio=14.97 duration=0.067s
2020-05-04 12:35:27.704Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "deflate" algorithm and level "9": original_size=5899618 compressed_size=396914 compression_ratio=14.86 duration=0.0581s
2020-05-04 12:35:30.223Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "deflate" algorithm and level "9": original_size=5899646 compressed_size=421543 compression_ratio=14.0 duration=0.0611s
2020-05-04 12:35:32.613Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "deflate" algorithm and level "9": original_size=5899588 compressed_size=408990 compression_ratio=14.42 duration=0.0581s

2020-05-04 12:34:50.363Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "bz2" algorithm and level "9": original_size=5899921 compressed_size=251577 compression_ratio=23.45 duration=0.6114s
2020-05-04 12:34:54.149Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "bz2" algorithm and level "9": original_size=5899994 compressed_size=264431 compression_ratio=22.31 duration=0.6389s
2020-05-04 12:34:57.718Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "bz2" algorithm and level "9": original_size=5899638 compressed_size=281809 compression_ratio=20.93 duration=0.6157s
2020-05-04 12:35:00.698Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "bz2" algorithm and level "9": original_size=5899764 compressed_size=271796 compression_ratio=21.71 duration=0.6154s

2020-05-04 12:36:00.441Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "zstandard" algorithm and level "3": original_size=5899921 compressed_size=339223 compression_ratio=17.39 duration=0.0063s
2020-05-04 12:36:03.424Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "zstandard" algorithm and level "3": original_size=5899627 compressed_size=359995 compression_ratio=16.39 duration=0.0065s
2020-05-04 12:36:05.606Z DEBUG [core] [scalyr_client.py:366] Compressed add event request data using "zstandard" algorithm and level "3": original_size=5899646 compressed_size=388475 compression_ratio=15.19 duration=0.0093s
```

Yes, bz2 does offer a better compression ratio (~23 vs ~17), but this comes at a cost at being 100 more CPU intensive / slower (and around 10x slower for deflate and defalte actually has worse compression ratio - ~15 vs ~17). 

In some scenarios, this cost / trade off (CPU cycles for egress traffic) may make sense, but in most (and especially the default one), it wont.

So with zstandard and moving compression off the sending thread I think we could see some nice performance gains.

PR #518, opened 2020-04-23
Fix submit microbenchmarks result script to support non-UTC timezones

Kami:
This pull request includes the following fixes / improvements to the micro benchmarks job:

1. Fixes the result submission script so it supports non-UTC timestamp. I noticed there was a build failure (https://circleci.com/gh/scalyr/scalyr-agent-2/34236?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification) and it appears the author_time timestamp string contains a local (non-UTC) time string.

2. Update Circle CI job so we also preserve raw benchmark result JSON files. This will make it easier to troubleshoot issues like the one above in the future.

3. Store benchmark results files in ``benchmark_results/`` directory instead of the in the repo root.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -310,9 +310,11 @@ commands:
             if [[ "${CIRCLE_BRANCH}" == "master" ]]; then
               export PYTHONPATH=.
               python benchmarks/scripts/send_microbenchmarks_data_to_codespeed.py \
-                  --data-path="benchmark-results-*.json" \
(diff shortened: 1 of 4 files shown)


Kami:
@czerwingithub Correct (and yes, RFC does allow timezones).

And the last time I was looking into that (when optimizing the parsing function), it did look like Docker always reports in UTC (which is reasonable and makes it easier for us).

And we had that timestamp parsing code running for quite a while now and I don't remember anyone reporting any issues related to it.

PR #526, opened 2020-04-28
Add string compatibility for apache monitor.

ArthurKamalov:
The `apache_monitor.py` file was left unchanged during codebase migration to python3 compatibility. There were some configuration fields with `str`(`six.text_type` is expected), it caused an error during configuration validation.


Kami:
Good catch :+1: 

I was wondering how our tests didn't catch it, then I remembered our tests don't cover the configuration definition part.

I guess eventually, some kind of lint checker for that would be nice :)

PR #527, opened 2020-04-29
Print useful config settings to agent.log

imron:
This commit will cause the agent to print various useful configuration options to
`agent.log` at startup or whenever the configuration changes.

This will be useful in determining the values of the configuration options when
diagnosing customer issues

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -277,6 +277,63 @@ class Configuration(object):
             )
             scalyr_util.set_json_lib(json_library)
 
+    def print_useful_settings(self, other_config=None):
+        """
+        Prints various useful configuration settings to the agent log, so we have a record
+        in the log of the settings that are currently in use.
+
+        @param other_config: Another configuration option.  If not None, this function will
+        only print configuration options that are different between the two objects.
+        """
+
+        options = [
+            "compression_type",
+            "compression_level",
+            "pipeline_threshold",
+            "min_allowed_request_size",
+            "max_allowed_request_size",
+            "min_request_spacing_interval",
+            "max_request_spacing_interval",
+            "read_page_size",
+            "max_line_size",
+            "internal_parse_max_line_size",
+            "line_completion_wait_time",
+            "max_log_offset_size",
+            "max_existing_log_offset_size",
+        ]
+
+        # get options (if any) from the other configuration object
+        other_options = None
+        if other_config is not None:
+            other_options = {}
+            for option in options:
+                other_options[option] = getattr(other_config, option, None)
+
+        first = True
+        for option in options:
+            value = getattr(self, option, None)
+            print_value = False
+
+            # check to see if we should be printing this option which will will
+            # be True if other_config is None or if the other_config had a setting
+            # that was different from our current setting
+            if other_config is None:
+                print_value = True
+            elif (
+                other_options is not None
+                and option in other_options
+                and other_options[option] != value
+            ):
+                print_value = True
+
+            if print_value:
+                # if this is the first option we are printing, output a header
+                if first:
+                    self.__logger.info("Configuration settings")
+                    first = False
+
+                self.__logger.info("\t%s: %s" % (option, value))
+
     def __get_default_hostname(self):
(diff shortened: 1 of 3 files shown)


czerwingithub:
Also, why not just make this a function on the Configuration object itself?  You can then add a test case to make to sure it runs correctly.

imron:
Instead of adding an option to suppress printing, I made it so that the function takes another configuration object as an optional argument. If this argument is specified, then it only prints out the useful config settings that have changed between the two configurations.

I think this is better than suppressing the messages because we want to know what the current settings for these options are, and none of the options being printed should be changing regularly.

If they are changing regularly then that's a problem that needs to be surfaced.

scalyr_agent/configuration.py
```diff
@@ -506,6 +563,12 @@ def include_raw_timestamp_field(self):
         message as recorded by those systems."""
         return self.__get_config().get_bool("include_raw_timestamp_field")
 
+    @property
```
czerwingithub, line 566: I think this is leftover code now?

scalyr_agent/tests/configuration_test.py
```diff
@@ -1557,6 +1558,97 @@ def test_empty_substitution(self):
 
         self.assertEquals(config.api_key, "hi")
 
+    def test_print_config_doesnt_throw(self):
```
czerwingithub, line 1561: I would just call this `test_print_config`.  It's inherent in all tests that we don't want the method to unexpectantly throw an exception.

scalyr_agent/tests/configuration_test.py
```diff
@@ -1557,6 +1558,97 @@ def test_empty_substitution(self):
...
+        don't throw any exceptions
+        """
+
+        self._write_file_with_separator_conversion("""{api_key: "hi there"}""")
+        mock_logger = Mock()
+        config = self._create_test_configuration_instance(logger=mock_logger)
+        config.parse()
+        try:
+            config.print_useful_settings()
+        except Exception as e:
```
czerwingithub, line 1572: Don't bother to catch this here.  Just let the exception bubble up and fail the test.

scalyr_agent/tests/configuration_test.py
```diff
@@ -1557,6 +1558,97 @@ def test_empty_substitution(self):
...
+        mock_logger = Mock()
+        config = self._create_test_configuration_instance(logger=mock_logger)
+        config.parse()
+
+        other_config = self._create_test_configuration_instance(logger=mock_logger)
+        other_config.parse()
+
+        try:
+            config.print_useful_settings(other_config=other_config)
+        except Exception as e:
```
czerwingithub, line 1643: Don't bother to catch.

scalyr_agent/tests/configuration_test.py
```diff
@@ -1557,6 +1558,97 @@ def test_empty_substitution(self):
...
+            max_line_size: 49900,
+            }
+        """
+        )
+        new_config = self._create_test_configuration_instance(logger=mock_logger)
+        new_config.parse()
+
+        try:
+            new_config.print_useful_settings(other_config=config)
+        except Exception as e:
```
czerwingithub, line 1609: Don't bother to catch.

scalyr_agent/configuration.py
imron: Good catch

scalyr_agent/tests/configuration_test.py
imron: Done.

scalyr_agent/tests/configuration_test.py
imron: Done.

scalyr_agent/tests/configuration_test.py
imron: Done.

scalyr_agent/tests/configuration_test.py
imron: Done.

Kami:
This will indeed come handy when troubleshooting various issues :+1:

PR #528, opened 2020-04-29
AGENT-377: Certificate verification when accessing the kubelet 

yanscalyr:
Still needs unit tests and a string conversion case to be covered.

### DIFF
--- a/docs/monitors/kubernetes_monitor.md
+++ b/docs/monitors/kubernetes_monitor.md
@@ -243,7 +243,7 @@ when starting to log a containers \
 |||# ``k8s_kubelet_host_ip``              ||| Optional (defaults to None). Defines the host IP address for the Kubelet \
                                               API. If None, the Kubernetes API will be queried for it
 |||# ``k8s_kubelet_api_url_template``     ||| Optional (defaults to https://${host_ip}:10250). Defines the port and \
-                                              protocol to use when talking to the kubelet API,allowed template \
+                                              protocol to use when talking to the kubelet API. Allowed template \
                                               variables are `node_name` and `host_ip`.
 |||# ``k8s_sidecar_mode``                 ||| Optional, (defaults to False). If true, then logs will only be collected \
                                               for containers running in the same Pod as the agent. This is used in \
--- a/docs/monitors/kubernetes_monitor.md
+++ b/docs/monitors/kubernetes_monitor.md
@@ -243,7 +243,8 @@ when starting to log a containers \
 |||# ``k8s_kubelet_host_ip``              ||| Optional (defaults to None). Defines the host IP address for the Kubelet \
                                               API. If None, the Kubernetes API will be queried for it
 |||# ``k8s_kubelet_api_url_template``     ||| Optional (defaults to https://${host_ip}:10250). Defines the port and \
-                                              protocol to use when talking to the kubelet API
+                                              protocol to use when talking to the kubelet API,allowed template \
+                                              variables are `node_name` and `host_ip`.
 |||# ``k8s_sidecar_mode``                 ||| Optional, (defaults to False). If true, then logs will only be collected \
                                               for containers running in the same Pod as the agent. This is used in \
                                               situations requiring very high throughput.
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -392,7 +392,8 @@ define_config_option(
 define_config_option(
     __monitor__,
     "k8s_kubelet_api_url_template",
-    "Optional (defaults to https://${host_ip}:10250). Defines the port and protocol to use when talking to the kubelet API",
+    "Optional (defaults to https://${host_ip}:10250). Defines the port and protocol to use when talking to the kubelet API,"
+    "allowed template variables are `node_name` and `host_ip`.",
     convert_to=six.text_type,
     default="https://${host_ip}:10250",
     env_aware=True,
@@ -2044,7 +2045,8 @@ class CRIEnumerator(ContainerEnumerator):
             host_ip=kubelet_api_host_ip,
             node_name=node_name,
             kubelet_url_template=Template(kubelet_api_url_template),
-            ca_file=global_config.k8s_service_account_cert,
+            verify_https=global_config.k8s_verify_kubelet_queries,
+            ca_file=global_config.k8s_kubelet_ca_cert,
         )
         self._query_filesystem = query_filesystem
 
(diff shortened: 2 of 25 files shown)


docs/monitors/kubernetes_monitor.md
```diff
@@ -243,7 +243,8 @@ when starting to log a containers \
 |||# ``k8s_kubelet_host_ip``              ||| Optional (defaults to None). Defines the host IP address for the Kubelet \
                                               API. If None, the Kubernetes API will be queried for it
 |||# ``k8s_kubelet_api_url_template``     ||| Optional (defaults to https://${host_ip}:10250). Defines the port and \
-                                              protocol to use when talking to the kubelet API
+                                              protocol to use when talking to the kubelet API,allowed template \
```
czerwingithub, line 246: Replace comma with period to start a new sentence.. Capitalize `allowed`.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -392,7 +392,8 @@
 define_config_option(
     __monitor__,
     "k8s_kubelet_api_url_template",
-    "Optional (defaults to https://${host_ip}:10250). Defines the port and protocol to use when talking to the kubelet API",
+    "Optional (defaults to https://${host_ip}:10250). Defines the port and protocol to use when talking to the kubelet API,"
+    "allowed template variables are `node_name` and `host_ip`.",
```
czerwingithub, line 396: Same comment as above.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2133,12 +2133,17 @@ def __init__(
         self,
         k8s,
         host_ip=None,
+        node_name=None,
+        verify_https=False,
```
czerwingithub, line 2137: Let's default this to True.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2160,21 +2165,30 @@ def __init__(
...
+            FALLBACK_KUBELET_URL_TEMPLATE, host_ip, node_name
         )
         self._timeout = 20.0
 
     @staticmethod
-    def _build_kubelet_url(kubelet_url, host_ip):
-        if host_ip:
-            return kubelet_url.substitute(host_ip=host_ip)
+    def _build_kubelet_url(kubelet_url, host_ip, node_name):
+        if node_name or host_ip:
```
czerwingithub, line 2178: I think this should be an `and`.  I believe the original intention was to refuse to build a URL if you didn't have the host ip.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2183,14 +2197,25 @@ def query_api(self, path):
...
-                    limit_once_per_x_secs=3600,
-                    limit_key="unverified-kubelet-request",
+            verify_connection = self._verify_connection()
+            if not verify_connection:
+                with warnings.catch_warnings():
+                    warnings.simplefilter(
+                        "ignore",
+                        category=requests.packages.urllib3.exceptions.InsecureRequestWarning,
+                    )
+                    response = self._session.get(
```
czerwingithub, line 2207: Let's create a method on this object called `_get(url, verify)` which simply invokes `_session.get`.  Use that method here and below.  Then, for testing, just do a mock path on the object to replace `_get` with a mock.  You can use that to write test cases for both checking `_get` is being invoked with the right parameters, and that the object fallsback appropriately when `_get` returns an error.

scalyr_agent/monitor_utils/k8s.py
yanscalyr: I was thinking of a case where we have one but not the other, we would still be able to build a valid url. But a case where either is missing would be exceptional so and might be better.

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -555,3 +567,145 @@ def run_test():
...
+                sys.stdout, sys.stderr = old_out, old_err
+
+    def test_error_response(self):
+        def fake_get(self, url, verify):
+            resp = FakeResponse()
+            resp.status_code = 404
+            return resp
+
+        with mock.patch.object(KubeletApi, "_get", fake_get):
+            try:
```
czerwingithub, line 672: I think you want to use `self.assertRaises` here.

scalyr_agent/builtin_monitors/tests/kubernetes_monitor_test.py
```diff
@@ -555,3 +567,145 @@ def run_test():
...
+                api = KubeletApi(
+                    FakeKubernetesApi(), host_ip="127.0.0.1", node_name="FakeNode",
+                )
+                api.query_stats()
+                self.assertTrue(False)
+            except KubeletApiException:
+                pass
+
```
czerwingithub, line 680: Can you also add in a test case where you return a `403` to trigger switching to the fallback URL?

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2160,21 +2165,33 @@ def __init__(
...
+            return kubelet_url.substitute(node_name=node_name, host_ip=host_ip)
         return None
 
     def _switch_to_fallback(self):
         self._kubelet_url = self._fallback_kubelet_url
 
+    def _verify_connection(self):
+        """ Return whether or not to use SSL verification
+        """
+        if self._verify_https and self._ca_file:
```
Kami, line 2188: Should we throw inside the constructor where we set the config values if ``verify_https`` is true, but ``ca_file`` doesn't exist (or is None)?

Or will we just outsource that to requests?

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2183,15 +2200,22 @@ def query_api(self, path):
...
-                response = self._session.get(url, timeout=self._timeout, verify=False)
-            if self._kubelet_url.startswith("https://"):
-                global_log.warn(
-                    "Accessing Kubelet with an unverified HTTPS request.",
-                    limit_once_per_x_secs=3600,
-                    limit_key="unverified-kubelet-request",
-                )
+            verify_connection = self._verify_connection()
+            if not verify_connection:
+                with warnings.catch_warnings():
```
Kami, line 2205: We could actually make our own context manager which uses ``warnings.catch_warnings()`` underneath. This would save us a couple of lines of code (and come handy if we want to use the same pattern in multiple places).

Then we could do something like:

```python
with context_managers.ignore_insecure_warnings(verify_connection=verify_connection):
   ....
```

Not a blocker for now, but something we should consider if we use this pattern in multiple places. And it would make "unit level" testing a bit easier as well.

PR #529, opened 2020-04-30
Add namespace whitelist capability for Kubernetes

czerwingithub:
Adds in a namespace whitelist capability to Kubernetes in addition to blacklist.  This allows customers to only collect logs from pods running from a specific set of namespaces.

Note, I have not finished writing all of the test cases but I wanted you to get an early review of this because I'll need to merge this in later tonight.

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -73,6 +73,7 @@ class Configuration(object):
     """
 
     DEFAULT_K8S_IGNORE_NAMESPACES = ["kube-system"]
+    DEFAULT_K8S_INCLUDE_NAMESPACES = ["*"]
 
     def __init__(self, file_path, default_paths, logger):
         # Captures all environment aware variables for testing purposes
@@ -1788,7 +1789,7 @@ class Configuration(object):
         self.__verify_or_set_optional_array_of_strings(
             config,
             "k8s_include_namespaces",
-            ["*"],
+            Configuration.DEFAULT_K8S_INCLUDE_NAMESPACES,
             description,
             apply_defaults,
             separators=[None, ","],
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -420,6 +420,10 @@ class Configuration(object):
         return self.__get_config().get_json_array("k8s_ignore_namespaces")
 
     @property
+    def k8s_include_namespaces(self):
+        return self.__get_config().get_json_array("k8s_include_namespaces")
+
+    @property
     def k8s_api_url(self):
         return self.__get_config().get_string("k8s_api_url")
 
@@ -1776,6 +1780,15 @@ class Configuration(object):
             config,
             "k8s_ignore_namespaces",
             Configuration.DEFAULT_K8S_IGNORE_NAMESPACES,
+            description,
+            apply_defaults,
+            separators=[None, ","],
+            env_aware=True,
+        )
+        self.__verify_or_set_optional_array_of_strings(
+            config,
+            "k8s_include_namespaces",
+            ["*"],
             description,
             apply_defaults,
             separators=[None, ","],
--- a/scalyr_agent/monitor_utils/k8s.py
+++ b/scalyr_agent/monitor_utils/k8s.py
@@ -111,9 +111,6 @@ def cache(global_config):
         @param config: The configuration
         @type config: A Scalyr Configuration object
     """
-    # split comma delimited string of namespaces to ignore in to a list of strings
-    # TODOcz:
-
     cache_config = _CacheConfig(
         api_url=global_config.k8s_api_url,
         verify_api_queries=global_config.k8s_verify_api_queries,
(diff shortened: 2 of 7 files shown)


czerwingithub:
@imron Ok, this PR isn't quite finished but it's close enough for you to take a look.  I believe I just have to finish filling out some of the unit tests (and I've already defined them).  But, I ran out of time before I need to step away.  I'll work with you tonight to get this merged in.

scalyr_agent/configuration.py
```diff
@@ -1781,6 +1785,15 @@ def __verify_main_config_and_apply_defaults(
             separators=[None, ","],
             env_aware=True,
         )
+        self.__verify_or_set_optional_array_of_strings(
+            config,
+            "k8s_include_namespaces",
+            ["*"],
```
imron, line 1791: For consistency with `k8s_ignore_namespaces` maybe we should define `Configuration.DEFAULT_K8S_INCLUDE_NAMESPACES` as a constant value set to `["*"]`

scalyr_agent/monitor_utils/tests/k8s_test.py
```diff
@@ -513,6 +514,68 @@ def test_no_prefetch(self):
...
+        self.assertFalse(test_filter.passes("kube"))
+
+    def test_basic_whitelist_and_blacklist(self):
+        test_filter = K8sNamespaceFilter(
+            global_include=["scalyr", "your-app", "kube"],
+            global_ignore=["kube", "testing"],
+        )
+        self.assertTrue(test_filter.passes("scalyr"))
+        self.assertTrue(test_filter.passes("your-app"))
+        self.assertFalse(test_filter.passes("kube"))
```
imron, line 554: For completeness, do we also want to test `self.assertFalse(test_filter.passes("testing"))`

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -210,6 +209,156 @@ class KubeletApiException(Exception):
...
+        else:
+            return key not in self.__blacklist
+
+    def __str__(self):
+        if self.__whitelist is not None:
+            return "include_only=%s" % ",".join(self.__whitelist)
+        else:
+            return "exclude=%s" % ",".join(self.__blacklist)
+
+    def __eq__(self, other):
```
imron, line 355: In python 2:

> The truth of x==y does not imply that x!=y is false. Accordingly, when defining __eq__(), one should also define __ne__() so that the operators will behave as expected.

In python 3:

> By default, __ne__() delegates to __eq__() and inverts the result unless it is NotImplemented. 

To ensure consistency between 2 and 3, we should define `__ne__`.  [This SO post](https://stackoverflow.com/a/30676267) suggests it's better in python2 to define `__ne__` as `not ==` e.g.

```
def __ne__(self, other):
        return not self == other
```

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -111,9 +112,7 @@ def cache(global_config):
         @type config: A Scalyr Configuration object
     """
     # split comma delimited string of namespaces to ignore in to a list of strings
-    namespaces_to_ignore = []
-    for x in global_config.k8s_ignore_namespaces:
-        namespaces_to_ignore.append(x.strip())
+    # TODOcz:
```
imron, line 115: What is left todo here?

scalyr_agent/monitor_utils/k8s.py
czerwingithub: Thanks, I didn't know about that nuance.  Implemented as you suggested.

scalyr_agent/monitor_utils/tests/k8s_test.py
czerwingithub: Done.

scalyr_agent/configuration.py
czerwingithub: Done.

scalyr_agent/monitor_utils/k8s.py
czerwingithub: Sorry, that was a left over marker for myself to modify some lines which I did.  That code was all dead there.

scalyr_agent/monitor_utils/k8s.py
imron: I had to look it up myself, but I had a vague memory that python2 had different behavior.

scalyr_agent/monitor_utils/k8s.py
imron: Yeah I figured that was likely the case.

czerwingithub:
@imron Thanks, yeah, you are right it would be good to have those tests.  I am about to sign off for the evening, so if you feel it's something you can put together in a few hours, yes, that would be great if you could do it.   I'm going to merge this change.

imron:
Sounds good, merge in the change and then I'll submit a separate PR for the tests.

PR #530, opened 2020-05-01
Add monitor utils tests to tox targets.

ArthurKamalov:
Tests for monitors utils were not added into tox targets, so it could not calculate coverage for them too.

Also, there is a small fix that does sorting of the namespaces in `K8sNamespaceFilter` before creating a string from it, to provide stable results.

### DIFF
--- a/scalyr_agent/monitor_utils/k8s.py
+++ b/scalyr_agent/monitor_utils/k8s.py
@@ -345,9 +345,9 @@ class K8sNamespaceFilter(object):
 
     def __str__(self):
         if self.__whitelist is not None:
-            return "include_only=%s" % ",".join(self.__whitelist)
+            return "include_only=%s" % ",".join(sorted(self.__whitelist))
         else:
-            return "exclude=%s" % ",".join(self.__blacklist)
+            return "exclude=%s" % ",".join(sorted(self.__blacklist))
 
     def __eq__(self, other):
         if self.__whitelist is not None:
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ commands =
     # isolated from other tests
     # NOTE 2: We write junit XML file for nicer visualization on Circle CI
     rm -rf test-results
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -35,7 +35,7 @@ commands =
     # isolated from other tests
     # NOTE 2: We write junit XML file for nicer visualization on Circle CI
     rm -rf test-results
-    py.test scalyr_agent/tests scalyr_agent/builtin_monitors/tests -vv --durations=5 -m "not memory_leak and not json_lib" --junitxml=test-results/junit-1.xml
+    py.test scalyr_agent/tests scalyr_agent/builtin_monitors/tests scalyr_agent/monitor_utils/tests -vv --durations=5 -m "not memory_leak and not json_lib" --junitxml=test-results/junit-1.xml
```
Kami, line 38: Good catch.

In the future we could perhaps also re-organize this a bit more to utilize py.test markers for everything and only specify top-level directory for tests.

Right now some of the tests intentionally run as part of a separate process to avoid cross test pollution.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -345,9 +345,9 @@ def __contains__(self, key):
 
     def __str__(self):
         if self.__whitelist is not None:
-            return "include_only=%s" % ",".join(self.__whitelist)
+            return "include_only=%s" % ",".join(sorted(self.__whitelist))
```
Kami, line 348: :+1:

tox.ini
ArthurKamalov: Yeah, good idea.

czerwingithub: We could also make a sane directory structure for tests.  I was going to suggest we check test inclusion using a pylint checker, but since this is an `ini` file, it makes importing some sort of shared "these are the test directories" variable a little more difficult (I believe).

Kami: I agree.

@ArthurKamalov and I already briefly talked about that in the past.

I personally think all the tests should either live in ``tests/`` or ``scalyr_agent/tests`` directory (and sub-directories, as necessary).

PR #531, opened 2020-05-01
Fix issue with dynamic module includes

czerwingithub:
Fix issue caused by passing in a `set` instead of a `list`.


yanscalyr:
> Fix issue caused by passing in a `set` instead of a `list`.

I assume you mean the other way around right?

czerwingithub:
Eh, yeah, I worded that wrong.  I'll fix that on merge.

PR #532, opened 2020-05-01
Fix incorrect namespace in service account

czerwingithub:


czerwingithub:
@yanscalyr I'm pretty sure this is the right fix to make.  I have to think we didn't get reports of this issue because either (1) users saw the problem on their own and fixed it or (2) if they had previous installed the Scalyr Agent on the cluster in the default namespace, they wouldn't have hit the issue (but they could be possibly using the old service account permissions).

PR #533, opened 2020-05-04
Move all the agent unit tests to the tests/unit/ directory

Kami:
This pull request moves all the existing agent unit tests to the ``tests/unit/`` directory.

This way we follow a consistent testing directory structure (all the various type of tests are contained inside ``tests/`` directory) and we can avoid issues like the ones fixed in #530 in the future.

Do keep in mind that we still (sadly) need multiple py.test invocations inside the tox targets. The reason for that is that some tests pollute global namespace and they need to be run in isolated inside a dedicated process.

---

~~NOTE 1: It might take a couple of commits to get coverage reports sorted out after this change.~~

NOTE 2: We still have some "legacy" smoke tests which rely on using ``run_tests.py`` script instead of py.test and live inside ``.circleci/`` directory.

We should eventually update those tests as well to utilize py.test like all the rest of the tests.


Kami:
While working on this, I also added some direct unit tests for ``_ignore_old_dead_container`` utility function - a65be571f19a115996b51df540f2dfd725b107d5.

---

I noticed that over coverage test runs, coverage jumps by +/- 2% - sometimes that function is covered and sometimes it's not.

The reason for that is that this function is indirectly exercised by legacy smoke tests - and depending on test timing, this function may or may not be executed.

With those direct unit tests, this function should now always be covered and coverage data should be more stable.

PR #534, opened 2020-05-04
[DONT MERGE] Use zstandard compression algorithm by default

Kami:
This pull request updates agent code to use zstandard compression algorithm by default on systems where ``zstandard`` Python library is available / installed.

In addition to that, I also updated requirements used by Docker containers to install that dependency so it will be used by default for Docker installations.

## Background, Context

We've done quite extensive testing and for most workloads, zstandard has shown to be a lot more efficient / faster than the current default (deflate level 9).

Because of that, it makes sense to use it by default in a future major agent release.

### Should we install zstandard Python library by default?

zstandard library contains a C extension which means we can't bundle it with the agent in the same manner as we bundle other pure-Python dependencies.

Good news though is that pre-compiled wheels are available for most common platforms and architectures so installing it using pip should work out of the box on most systems and doesn't require any C headers or compilation.

The question is if we should try to install ``zstandard`` library by default using pip inside our installer script or perhaps pre / post install script.

If we used a virtual environment or similar for agent, the answer would definitely be yes (or when we ship it as a single binary), but now the answer is not so clear / straight forward..

I personally think it should be fine to try to install it as part of the installer script using "best effort" approach - try to install it, but if the install fails (e.g. precompiled wheel is not available for the platform / architecture where the user is trying to install the agent on), ignore the failure (it simply means deflate instead of zstandard will be used).

## TODO

- [x] Decide if we should try to install zstandard library as part of the installer script or package pre / post install step


czerwingithub:
Just a record this note here...  I'm going to raise the issue with Mark O'Brien in the release meeting today.  I'd like to get Technical Services input on this.  After all, we could be making their life a harder (if this makes the installation more unstable or at least not as predictable).

Like I said in our meeting, we generally take the approach that we can do white glove service for customers who do push a large number of bytes into Scalyr (and therefore are more in need for something like this).

czerwingithub:
@Kami Ok, Mark O'Brien was fine with the idea of attempting to install zstd via the install script as long as there is a way to disable it.  We'll need to make it fairly robust so that it is not likely to cause problems.. and also do smart things like check to see if zstd is already installed.  Of course, the whole python2 / python3 thing might make this more difficult.

I'm also wondering if we should change how we specify the compression algorithm option to better support scenarios where the default is not available on the system.  Specifically, maybe we change the compression option to be a list of compression algorithms and compression levels.. we use the first one that is supported.  So, we could change the default to `zstd:6,deflate:6" (etc).

I believe we could make this backward compatible with existing option by just using the compression level option if none is specified for a particular algorithm.

Thoughts?

Kami:
Just to wrap up what we discussed  on the call.

With the current implementation, the end result is already the same (we fall back to deflate if zstandard is not available) so allowing a user to specify a list wouldn't really change that and it would just introduce a bit more complexity to the code.

In addition to that, most users who would override the value probably have a good reason for it and want to user single specific algorithm without a fallback.

Kami:
NOTE: smoke tests which rely on production API will be failing until the server side change is deployed there.

I could temporary update them to utilize deflate until the change is not deployed there, but it's probably just better to wait a day or two for that change to also be deployed there (this way we don't need to revert those changes again in a couple of days).

Kami:
As part of this PR, I also pushed some changes related to how we run unit tests.

We have some tests which rely on ``reload()`` and other functionality which can affect other tests running in the same process.

By default, py.test runs all the tests inside a single process which means we can't run those tests as part of the same target to avoid cross-test pollution.

Currently we are solving that by invoking py.test multiple times inside our tox unit test targets.

Downside of that approach is that it would not do anything if you manually run "conflicting" tests using py.test directly.

For example:

```bash
pytest -vv tests/unit/util/default_json_lib_test.py tests/unit/util/util_test.py
```

You would see errors similar to this:

```bash
ERROR: test_close_from_server (tests.unit.util.util_test.TestRedirectorClient)
  File "/home/kami/w/scalyr/scalyr-agent-2/tests/unit/util/util_test.py", line 788, in test_close_from_server
    self._client.join()
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 1205, in join
    if not self.isAlive() and self.__exception_info is not None:
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 1216, in isAlive
    return self.is_alive()
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 1232, in is_alive
    or not super(StoppableThread, self).is_alive()
TypeError: super(type, obj): obj must be an instance or subtype of type
```

That's because some of the tests imported StopptableThread which relies in old version of that class pre-reload which breaks things.

That + having multiple py.test invocations inside corresponding tox targets is not ideal so I decided to re-organize tests a bit (to make it very clear and explicit which tests need to run in isolation - those tests now have ``need_to_run_isolated`` pytest marker) and utilize pytest-forked plugin which allows us to run specific tests inside a subprocess.

@yanscalyr ^ The errors you were seeing a while back were likely related to that (or stale *.pyc file).

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ Packaged by Steven Czerwinski <czerwin@scalyr.com> on May 5, 2020 08:30 -0800
 
 Features
 * Add support for ``zstandard`` and ``lz4`` compression of outgoing requests to the Scalyr API. For backward compatibility reasons, the default value for ``compression_type`` configuration option still is ``deflate``.
+* Agent will now default to using ``zstandard`` instead of ``deflate`` for compressing outgoing requests on systems where ``zstandard`` Python library is available. For more information on this change and how to install zstandard library, please refer to the [release notes](https://github.com/scalyr/scalyr-agent-2/blob/master/RELEASE_NOTES.md).
```
czerwingithub, line 12: The entry above saying `compression_type` is still `deflate` seems incorrect now?

scalyr_agent/scalyr_client.py
```diff
@@ -565,12 +565,26 @@ def __receive_response(self, body_str, send_time):
...
+                    if status == "error/client/badRequest":
+                        log.error(
+                            (
+                                "Request to '%s' failed due to an error.  Returned error code was '%s'. "
+                                "This error likely indicates client is trying to use a compression "
+                                "algorithm which is not supported by the server."
+                            ),
+                            self.__full_address,
+                            status,
+                            error_code="error/client/badParam",
```
czerwingithub, line 579: I think you mean to use `error/client/badRequest` here.

tests/unit/configuration_test.py
```diff
@@ -1888,6 +1907,86 @@ def test_parse_compression_algorithm_invalid_compression_level(self):
...
+
+    @skipIf(
+        sys.version_info < (2, 7, 0),
+        "Skipping under Python < 2.7 (zstandard not available for Python 2.6)",
+    )
+    @mock.patch("scalyr_agent.util.zstandard", None)
+    @mock.patch("scalyr_agent.configuration.DEFAULT_COMPRESSION_ALGORITHM", "deflate")
+    @pytest.mark.forked
+    @pytest.mark.relies_on_reload
+    @pytest.mark.needs_to_run_isolated
```
czerwingithub, line 1937: If we have a mark for `relies_on_reload`, we should always run it in isolation, correct?  Is there a way to enforce that by updating the definition for  `relies_on_reload` to make sure they are run in isolation?  I don't think `forked` implies `needs_to_run_isolated` since you have test cases down below that fork but don't specify to run in isolation.

I'm also wondering if there's a good way to make sure folks know to use these annotations.  It seems adding documentation to `reload` should be one way.  But, for `forked`, I don't know if there's one clear place we can document this.

PR #535, opened 2020-05-05
Prepare version for the next release

Kami:
This pull request updates VERSION with information for the next release.

---

I wanted to build a new internal pre-release package (``2.1.3.pre4.1`` version), but I received the following error:

```bash
FATAL ERROR: Preproduction version: 2.1.3.pre4.1 must be less than VERSION: 2.1.3
```

Looking at the build script, we compare provided version against the one in ``VERSION`` file.

So unless I bump version in that file, the build will always fail.

@czerwingithub Please let me know if I misunderstood or missed something.'

Also, that's a more long term / best practice thing, but most projects bump version to a next development version as soon as the release is out.

For example, if ``2.1.3`` was just released, we would bump a version in master to ``2.1.4dev`` or similar.

This way versions don't get mixed up and "in development" packages also contain a correct version number.


czerwingithub:
@Kami Sorry, yes, I missed one step for you.  You need to update the `VERSION` file to be `2.1.4`.  If you do that, then you won't get the error you were seeing.

In terms of long term, I'm more than fine adopting better practices.

PR #536, opened 2020-05-05
Don't look for  field in Logstash smoketest

yanscalyr:


.circleci/docker_unified_smoke_unit/smoketest/smoketest.py
```diff
@@ -835,8 +835,6 @@ def _verify_queried_attributes(self, att, stream_name, process_name):
         if not all(
             [
                 # att.get('stream') in stream.name,  # we haven't setup server-side parser so $stream is not available
-                # Since the input streams are locally mounted, the event origins are all the same as the agent hostname
-                att.get("origin") == self._agent_hostname,
```
czerwingithub, line 839: I think we need something like this though to make sure your query is restricted to events that were created by this test instance.  Otherwise, you might be matching other events uploaded by other tests running on this same Scalyr account.

yanscalyr: From what I saw the query it does is limited by serverHost, the check we would replace it with would be checking against hostname which seems pointless in that case.

czerwingithub: Ok.. maybe we should be checking for `serverHost` here instead of `origin`.  Would that work?  I can understand that we want to make sure we are getting back the right events.

yanscalyr: My thought was since the query has "serverHost=whatever" in it the check would be a waste but it wouldn't hurt to add.

czerwingithub: Yeah, it wouldn't hurt.  It would also make sure that part of the code (including the server host in the query) doesn't get screwed up.  If we don't include the server host, it will lead to all sorts of nasty false positives that it would be nice to avoid.

PR #537, opened 2020-05-07
Include git build revision as part of the agent start up log line

Kami:
This pull request updates agent start up log line to also include the git revision on which the package is based on.

## Background, Context & Why

Recently I spent multiple hours troubleshooting an issue which, it turns out, was related to the same internal / dev package being re-built with the same version identifier, but a different git commit revision.

This means that the same version (``2.1.3.pre4.1``) was actually representing two different commits.

I assumed the root cause for the issue I was investigating lives somewhere else (initially I just verified the version agent was reporting in the logs is correct) so I spent tons of time digging into wrong places.

We already include ``build_info`` file with each package install which includes this information, but that file will only be accessible to people who have SSH (and likely root) access to the server where the agent is running on.

If this information was available in the log line (which is accessible in agent.log which we ingest), it would save me tons of time.

It's unlikely that something like this would happen with non dev / internal packages, but I still think it's a good idea to include this change.

## Implementation Details

The implementation simply reads the version from ``build_info`` file which is bundled with each package.

## TODO

- [x] Unit tests
- [x] Package tests


scalyr_agent/agent_main.py
```diff
@@ -907,19 +908,23 @@ def __run(self, controller):
...
-                    "Starting scalyr agent... (version=%s) %s (Python version: %s)"
-                    % (SCALYR_VERSION, scalyr_util.get_pid_tid(), python_version_str)
-                )
-                log.log(
-                    scalyr_logging.DEBUG_LEVEL_1,
-                    "Starting scalyr agent... (version=%s) %s (Python version: %s)"
-                    % (SCALYR_VERSION, scalyr_util.get_pid_tid(), python_version_str),
+                build_revision = get_build_revision()
+
+                # TODO: Why do we log the same line under info and debug? Intentional?
```
czerwingithub, line 914: Maybe it is because all debug output goes to `agent_debug.log` and we thought that normal non-debug logging wouldn't get there.. so to be sure that we had the version number, etc in `agent_debug.log`, we added this in?

Kami:
Looks like I forgot to add a file with unit tests for build info functionality.

Fixed this in 0b1d0022bd698faf47713cb126ff1fb08d770778.

PR #538, opened 2020-05-11
Windows packaging with pyinstaller.

ArthurKamalov:
This PR changes  `py2exe` packager to 'PyInstaller'

Also, there are changes to add ability to run libcloud deployments for Windows instances. 
There is one small hack that I had to make to bypass the problem with liblouid.ParamikoSSHClient. On short, after `ScriptDeployment` puts deployment script into remote Windows machine,  the paramiko SSH client gets the path where it was copied path as UNIX path ( like - `/c:/Users/Administrator`), so when it is time to execute the script, the script can not be found because of the invalid path. The quick solution is to monkey patch SSH client and make it to remove the first slash from the resulting path.


tests/ami/packages_sanity_tests.py
```diff
@@ -89,7 +89,30 @@
...
+
+
+class ParamikoSSHClient(libcloud.compute.ssh.ParamikoSSHClient):
+    def put(self, path, contents=None, chmod=None, mode="w"):
+        result = super(ParamikoSSHClient, self).put(
+            path, contents=contents, chmod=chmod, mode=mode
+        )
+        import re
+
+        if re.match(r"^\/\w\:.*$", result):
```
Kami, line 108: Can you please add a comment on why we need this?

And could this live upstream in Libcloud? :)

ArthurKamalov: I tried to explain this in the PR description. This is the only problem with libloud that I faced. After we `put` our deployment script into remote machine it returns remota path in unix format, and this path can not be accepted later when we execute this script.

tests/ami/packages_sanity_tests.py
```diff
@@ -213,7 +254,15 @@ def main(
         test_type,
         random.randint(0, 1000),
     )
-    step = ScriptDeployment(rendered_template, timeout=120)
+    test_package_step = ScriptDeployment(
+        rendered_template, name="deploy.ps1", timeout=120
+    )
+
+    if os.path.exists(to_version) and os.path.isfile(to_version):
+        put_file_step = FileDeployment(to_version, "./ScalyrAgentInstaller.msi")
```
Kami, line 262: I had a brief look at Libcloud code and you might not need to use Paramiko path workaround above if you use a full absolute path here (e.g. ``C:\\foo\\bar``), but would need to confirm.

And ``run()`` method onScriptDeployment might also need a small change to support windows style path separators (perhaps by adding path_separator argument to that class constructor or similar).

ArthurKamalov:
@Kami 
Sanity testing scripts now can accept package file and upload it to EC2 instances. It can be easier because we do not need to work with CircleCI artifacts API, which is very limited, in my opinion.

Also, there is a job that gathers new installation packages from previous windows and linux build jobs and then runs AMI tests with them.

.circleci/config.yml
```diff
@@ -1386,6 +1387,175 @@ jobs:
...
+            export PRIVATE_KEY_PATH="~/.ssh/id_rsa_7b26916ba09d31504c8d31a3a6b8f27a"
+            python -m pip install --user -r tests/ami/requirements.txt
+            export PYTHONPATH="."
+            mkdir -p outputs
+
+            # run sanity test for each image concurrently in backgroung.
+            python tests/ami/packages_sanity_tests.py --distro=WindowsServer2012 --type=install --to-version=/tmp/workspace/ScalyrAgentInstaller.msi &> outputs/WindowsServer2012.log & \
+            python tests/ami/packages_sanity_tests.py --distro=WindowsServer2016 --type=install --to-version=/tmp/workspace/ScalyrAgentInstaller.msi &> outputs/WindowsServer2016.log & \
+            python tests/ami/packages_sanity_tests.py --distro=WindowsServer2019 --type=install --to-version=/tmp/workspace/ScalyrAgentInstaller.msi &> outputs/WindowsServer2019.log & \
+            python tests/ami/packages_sanity_tests.py --distro=ubuntu1804 --type=upgrade --from-version=current --to-version=/tmp/workspace/scalyr-agent-2.deb &> outputs/ubuntu1804.log & \
```
Kami, line 1500: Great, thanks for also hooking Linux ones :)

I planned to do it at some point, but you beat me to it. Thanks again.

.circleci/config.yml
```diff
@@ -1386,6 +1387,175 @@ jobs:
...
+              mv .\ScalyrAgentInstaller-*.msi package\ScalyrAgentInstaller.msi
+
+      - persist_to_workspace:
+          root: package
+          paths: ScalyrAgentInstaller.msi
+
+      - store_artifacts:
+          path: package
+
+  build-linux-packages:
```
Kami, line 1430: Now that we have this job, we can probably get rid of ``tests/distribution/{rpm_deb}_package.py`` file, right?

.circleci/config.yml
```diff
@@ -1386,6 +1387,175 @@ jobs:
...
+            export ACCESS_KEY=${AWS_ACCESS_KEY}
+            export SECRET_KEY=${AWS_SECRET_KEY}
+            export KEY_NAME=circleci
+            export PRIVATE_KEY_PATH="~/.ssh/id_rsa_7b26916ba09d31504c8d31a3a6b8f27a"
+            python -m pip install --user -r tests/ami/requirements.txt
+            export PYTHONPATH="."
+            mkdir -p outputs
+
+            # run sanity test for each image concurrently in backgroung.
+            python tests/ami/packages_sanity_tests.py --distro=WindowsServer2012 --type=install --to-version=/tmp/workspace/ScalyrAgentInstaller.msi &> outputs/WindowsServer2012.log & \
```
Kami, line 1497: Do we have any timing information - aka how long this step takes?

I only see failed builds here so not sure - https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2?branch=windows-package-to-pyinstaller.

---

On that note, we could probably run tests for different distros in parallel since they are independent.

This would probably save quite some time (IIRC, Linux ones take around - 80 -120 seconds per distro and I imagine Windows one take quite a bit longer).

.circleci/config.yml
```diff
@@ -1386,6 +1387,175 @@ jobs:
...
+            python tests/ami/packages_sanity_tests.py --distro=WindowsServer2016 --type=install --to-version=/tmp/workspace/ScalyrAgentInstaller.msi &> outputs/WindowsServer2016.log & \
+            python tests/ami/packages_sanity_tests.py --distro=WindowsServer2019 --type=install --to-version=/tmp/workspace/ScalyrAgentInstaller.msi &> outputs/WindowsServer2019.log & \
+            python tests/ami/packages_sanity_tests.py --distro=ubuntu1804 --type=upgrade --from-version=current --to-version=/tmp/workspace/scalyr-agent-2.deb &> outputs/ubuntu1804.log & \
+            python tests/ami/packages_sanity_tests.py --distro=ubuntu1604 --type=upgrade --from-version=current --to-version=/tmp/workspace/scalyr-agent-2.deb &> outputs/ubuntu1604.log & \
+            python tests/ami/packages_sanity_tests.py --distro=ubuntu1404 --type=upgrade --from-version=current --to-version=/tmp/workspace/scalyr-agent-2.deb &> outputs/ubuntu1404.log & \
+            python tests/ami/packages_sanity_tests.py --distro=centos7 --type=upgrade --from-version=current --to-version=/tmp/workspace/scalyr-agent-2.rpm &> outputs/centos7.log &
+
+            # Wait for every job.
+            for job in `jobs -p`
```
Kami, line 1506: Would it be worthwhile to refactor this into a script file? E.g. ``wait-for-running-bg-js.sh`` or similar.

.circleci/config.yml
```diff
@@ -1386,6 +1387,175 @@ jobs:
...
+            if [ "$FAIL" -ne "0" ];
+            then
+                echo "FAIL"
+                exit 1
+            fi
+      - run:
+          name: ubuntu1804 result.
+          when: always
+          command: |
+            cat outputs/ubuntu1804.log
```
Kami, line 1521: We should probably also preserve those as build artifacts.

build_package.py
```diff
@@ -1606,7 +1633,7 @@ def get_build_info():
             "git config user.email", fail_quietly=True, command_name="git"
         )
         if rc != 0:
-            packager_email = u"unknown"
+            packager_email = "unknown"
```
Kami, line 1636: IIRC, This failed in the past if it didn't contain ``u`` prefix.

scalyr_agent/__scalyr__.py
```diff
@@ -117,7 +117,7 @@ def __determine_package_root():
             file_path = os.path.join(base, file_path)
         file_path = os.path.dirname(os.path.realpath(file_path))
     else:
-        return os.path.dirname(text_type(sys.executable, sys.getfilesystemencoding()))
```
Kami, line 120: Is there a specific reason why the second argument (encoding) has been removed?

scalyr_agent/builtin_monitors/windows_system_metrics.py
```diff
@@ -102,8 +104,10 @@ def _gather_metric(method, attribute=None, transform=None):
     def gather_metric():
         """Dynamically Generated """
         try:
+            print(attribute)
```
Kami, line 107: I assume that was a debugging change and is not needed anymore :)

tests/ami/packages_sanity_tests.py
```diff
@@ -229,21 +311,21 @@ def main(
             ex_security_groups=SECURITY_GROUPS,
             ssh_username=distro_details["ssh_username"],
             ssh_timeout=10,
-            timeout=140,
-            deploy=step,
+            timeout=240,
```
Kami, line 314: Perhaps we should use different (longer) timeout just for Windows? Or you think there is a need to use it for Linux as well?

I spun up many Linux tests and it never needed longer to finish (unless there was some upstream issue which would cause it to time out in any case).

tests/ami/packages_sanity_tests.py
```diff
@@ -292,13 +374,19 @@ def render_script_template(
...
-    )
     template_context["package_to_version"] = to_version
-    template_context["package_to_version_is_url"] = (
-        "http://" in to_version or "https://" in to_version
-    )
+
+    def get_source_type(version_string):
+        if "http://" in version_string or "https://" in version_string:
+            return "url"
+        elif os.path.exists(version_string) and os.path.isfile(version_string):
```
Kami, line 382: :+1:

tests/ami/scripts/fresh_install_deb.sh.j2
```diff
@@ -28,20 +28,20 @@ cat /etc/*-release
...
 
-{% if not package_to_version_is_url -%}
+# Install provided version
+{% if package_to_version_source_type == "install_script" -%}
 wget -O "install-scalyr-agent-2.sh" -q "{{installer_script_url}}"
 sudo bash ./install-scalyr-agent-2.sh --set-api-key "{{scalyr_api_key}}"
-{% endif -%}
+sudo apt-get install -y "scalyr-agent-2={{package_to_version}}"
+export PACKAGE_TO_VERSION="{{package_to_version}}"
+{% elif package_to_version_source_type in ("file", "url") -%}
```
Kami, line 37: Perhaps it would be a good idea to add a comment if the source is file that the file is uploaded by the script / Libcloud itself.

tests/ami/scripts/fresh_install_rpm.sh.j2
```diff
@@ -28,19 +28,20 @@ cat /etc/*-release
...
 
-{% if not package_to_version_is_url -%}
+# Install provided version
+{% if package_to_version_source_type == "install_script" -%}
 wget -O "install-scalyr-agent-2.sh" -q "{{installer_script_url}}"
 sudo bash ./install-scalyr-agent-2.sh --set-api-key "{{scalyr_api_key}}"
-{% endif -%}
+{% elif package_to_version_source_type in ("file", "url") -%}
+    {% if package_to_version_source_type == "url" -%}
+    wget -O to_package.rpm "{{ package_to_version }}"
```
Kami, line 37: Probably good to be consistent with fresh install deb script and just use ``package`` instead of ``to_package`` file name here.

tests/ami/scripts/fresh_install_rpm.sh.j2
```diff
@@ -28,19 +28,20 @@ cat /etc/*-release
...
+    {% endif -%}
 
-# Install latest version
-{% if package_to_version_is_url -%}
-wget -O package.rpm "{{ package_to_version }}"
 sudo rpm -iv package.rpm
```
Kami, line 40: Did you confirm this script works correctly? Since it kinda seems it's incorrectly mixing to_package.rpm and package.rpm file names.

tests/ami/scripts/fresh_install_rpm.sh.j2
```diff
@@ -28,19 +28,20 @@ cat /etc/*-release
...
 sudo sed -i 's/REPLACE_THIS/{{scalyr_api_key}}/' /etc/scalyr-agent-2/agent.json
-export PACKAGE_TO_VERSION=$(rpm -qp --queryformat '%{VERSION}\n' "package.rpm" | sed "s/_.*//g")
+export PACKAGE_TO_VERSION=$(rpm -qp --queryformat '%{VERSION}\n' "to_package.rpm" | sed "s/_.*//g")
 {% else -%}
-sudo yum install -y "scalyr-agent-2-{{package_to_version}}"
+sudo apt-get install -y "scalyr-agent-2={{package_to_version}}"
```
Kami, line 44: Hm, why the apt-get step here? This script file is used for rpm packages.

tests/ami/scripts/fresh_install_windows.ps1.j2
```diff
@@ -0,0 +1,92 @@
...
+
+Start-Sleep 5
+
+$status_output=scalyr-agent-2 status -v;
+
+if ($LASTEXITCODE -ne 0) { Throw "Status command failed" }
+
+Write-Output $status_output
+
+if ("$status_output" -notmatch "$msi_version")   { Throw "Can not find version." };
```
Kami, line 89: Looking at those checks, ami tests should indeed fail if the agent process fails to start, but we should still double check this scenario here - https://github.com/scalyr/scalyr-agent-2/pull/554#issuecomment-636448903.

tests/ami/scripts/upgrade_install_rpm.sh.j2
```diff
@@ -64,11 +68,19 @@ echo ""
...
-{% if package_to_version_is_url -%}
-wget -O package-to.rpm "{{ package_to_version }}"
-sudo rpm -Uv package-to.rpm
+{% if package_to_version_source_type in ("file", "url") -%}
+    {% if package_to_version_source_type == "url" -%}
+    wget -O to_package.rpm "{{ package_to_version }}"
+    {% endif -%}
+
+
+echo "1"
```
Kami, line 77: I assume you used those echo lines for debugging and they are not needed anymore?

tests/ami/windows/install_openssh.ps1
```diff
@@ -0,0 +1,74 @@
...
+# limitations under the License.
+#
+# This script downloads and installs OpenSSH server from https://github.com/PowerShell/Win32-OpenSSH.OpenSSH.
+# Script requires following environment variables being set.
+#   PUBLIC_KEY - string that contains public key for future authorization.
+#
+
+
+
+function Expand-ZIPFile($file, $destination) {
```
Kami, line 23: Is there some kind of PowerShell linter out there? Something similar to shellcheck which we use for bash / sh scripts.

win32/scalyr-agent.spec
```diff
@@ -0,0 +1,99 @@
+# -*- mode: python ; coding: utf-8 -*-
```
Kami, line 1: Should probably add license header just in case.

Kami:
I added some in-line comments.

Overall nice work - just needs a bit more minor polish.

tests/ami/windows/install_openssh.ps1
Kami: Looks like there is https://github.com/PowerShell/PSScriptAnalyzer, but no idea how useful it actually is.

build_package.py
ArthurKamalov: I guess this is the result of the `black` refactorer.
And it also looks like `packager_email` needed only for `print`

scalyr_agent/builtin_monitors/windows_system_metrics.py
ArthurKamalov: Woops.

tests/ami/scripts/upgrade_install_rpm.sh.j2
ArthurKamalov: Yes.. Thanks for pointing out.

tests/ami/scripts/fresh_install_rpm.sh.j2
ArthurKamalov: Just the same reason as above..

tests/ami/scripts/fresh_install_rpm.sh.j2
ArthurKamalov: Oops, yes, I forgot to check install scripts one more time.

tests/ami/scripts/fresh_install_rpm.sh.j2
ArthurKamalov: I changed naming from `to/from_package` to `install_package`  and `upgrade_package`.  I hope now it easier to understand.

tests/ami/scripts/fresh_install_deb.sh.j2
ArthurKamalov: Yeah, good idea.

tests/ami/packages_sanity_tests.py
ArthurKamalov: I guess there were some timeouts and it was with Linux instances IIRC.  Job time should increase because now we upload agent package to install, so it can be quite slow.

.circleci/config.yml
ArthurKamalov: Yes. I think new build job does the same thing as `tests/distribution/{rpm_deb}_package`

.circleci/config.yml
ArthurKamalov: Good suggestion. Done.

tests/ami/packages_sanity_tests.py
ArthurKamalov: Also, It still does not have to wait for 240 seconds if it finishes earlier.

scalyr_agent/__scalyr__.py
ArthurKamalov: on python3 `sys.executable` returns `str`(unicode), end `text_type` is `str` too, so `str(str(), encodeing="some_encoding")` will lead to error because we can not decode `str` from `str`.
But I agree that it might be wrong to remove it. I added a condition to handle this.

tests/ami/windows/install_openssh.ps1
ArthurKamalov: I don't think I so familiar with that too. But I'll have a look.

ArthurKamalov:
@Kami 
I tried to reduce code duplication in ami scripts, so there is only one script for each platform now. install and upgrade logic now located in the same file but upgrade com be optionally disabled through template parameters.

.circleci/config.yml
```diff
@@ -1395,91 +1396,276 @@ jobs:
...
+
+            # run AMI package sanity tests concurrently.
+            ./scripts/circleci/wait-for-running-bg-jobs.sh
+
+
+      - run:
+          name: ubuntu1804 result.
+          when: always
+          command: |
+            cat outputs/ubuntu1804.log
```
Kami, line 1522: You could run this as part of a single command to avoid a bunch of duplication :)

scripts/circleci/wait-for-running-bg-jobs.sh
```diff
@@ -0,0 +1,38 @@
...
+# Wait for every job.
+for job in `jobs -p`
+do
+    wait $job || let "FAIL+=1"
+done
+
+# Exit with error if some of the jobs failed.
+if [ "$FAIL" -ne "0" ];
+then
+    echo "FAIL"
```
Kami, line 36: Should probably print something along the lines of "One of the jobs failed, please inspect the logs".

And you can probably also just cat the logs here in case of failure.

.circleci/config.yml
ArthurKamalov: But wouldn't it be too hard to navigate in this one big log file?

ArthurKamalov: Sorry, not a log file, but only one output frame in CircleCI.

Kami:
@ArthurKamalov I've pushed various fixes and improvements to this branch:

- Output of the run ami tests script is now more user-friendly. Previously it didn't output anything so it was quite confusing and if you didn't dig in, you didn't know if tests even ran. On failure, we  now print which job has failed and we also try to cat the log file of the failed job.
- Removed "cat" jobs which are now not necessary anymore.
- Made sure we also cache Python dependencies for ami-tests job
- Made sure we also cache Python dependencies for "windows-package" job
- Various other Circle CI config fixes and improvements

Example output on failure: https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/3078/workflows/323e6b48-247c-4e74-a600-e519c519112f/jobs/41694

Example output on success: https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/3082/workflows/3ee7ae12-5ad4-447d-bf3e-5b6843c91780/jobs/41706

.circleci/config.yml
```diff
@@ -1395,91 +1396,256 @@ jobs:
...
+            mkdir artifacts
+            mkdir package
+
+            ls
+            pushd artifacts
+
+            # update agent version to avoid version conflict with current scalyr repository package.
+            #echo "$(cat VERSION).1" > VERSION
+
+            python ../build_package.py deb
```
Kami, line 1484: To make it easier to distinguish packages, I think it would be a good idea to include git revision in the filename (I believe we did the same thing with previous build package job).

tests/ami/packages_sanity_tests.py
```diff
@@ -156,13 +204,42 @@
...
 SECURITY_GROUPS_STR = get_env_throw_if_not_set(
-    "SECURITY_GROUPS", "allow-ssh"
+    "SECURITY_GROUPS", "allow-ssh-rdp"
 )  # sg-02efe05c115d41622
 SECURITY_GROUPS = SECURITY_GROUPS_STR.split(",")  # type: List[str]
 
 SCALYR_API_KEY = get_env_throw_if_not_set("SCALYR_API_KEY")
 
 
+def _get_source_type(version_string):
```
Kami, line 214: :+1:

tests/ami/scripts/test_deb.sh.j2
```diff
@@ -0,0 +1,127 @@
...
+sudo apt-get update -y
+sudo apt-get install -y "{{python_package}}"
+
+# Install Sclayr agent.
+{% if install_package["type"] in ("file", "url") -%}
+{# type of the package is 'file' or 'url', that means that file
+should be uploaded by script or already uploaded by Libloud. #}
+    {% if install_package["type"] == "url" -%}
+    wget -O install_package.deb "{{ install_package["source"] }}"
+    {% elif upgrade_package["type"] == "file" -%}
```
Kami, line 37: Should this line say ``install_package["type"]``?

tests/ami/scripts/test_rpm.sh.j2
```diff
@@ -0,0 +1,126 @@
...
+# cloud-init fully finished
+sleep 20
+
+uname -a
+cat /etc/*-release
+
+sudo yum install -y wget
+sudo yum install -y "{{python_package}}"
+
+# Install Sclayr agent.
```
Kami, line 31: Typo - Sclayr -> Scalyr, same in the install script.

tests/ami/scripts/test_rpm.sh.j2
```diff
@@ -0,0 +1,126 @@
...
+{% if install_package["type"] in ("file", "url") -%}
+{# type of the package is 'file' or 'url', that means that file
+should be uploaded by script or already uploaded by Libloud. #}
+    {% if install_package["type"] == "url" -%}
+    wget -O install_package.deb "{{ install_package["source"] }}"
+    {% elif upgrade_package["type"] == "file" -%}
```
Kami, line 37: Same here, I assume it should be install_package["type"].

Or we could remove this line all together since it's a no-op anyway.

Kami:
Please also re-enable all the tests which have been disabled during development before merging.

win32/scalyr-agent.spec
```diff
@@ -0,0 +1,112 @@
...
+                'source_root\\scalyr_agent\\third_party',
+				'source_root\\scalyr_agent\\third_party_python2',
+				],
+             hiddenimports = collect_submodules('scalyr_agent.builtin_monitors') + ['pkg_resources.py2_warn'],
+             datas=[('data_files\\VERSION', '.'), ('data_files\\licenses', 'third_party_licenses')],
+             )
+
+
+
+MERGE( (main_a, 'scalyr-agent-2', 'scalyr-agent-2.exe'), (config_a, 'scalyr-agent-2-config', 'scalyr-agent-2-config.exe'), (service_a, 'ScalyrAgentService', 'ScalyrAgentService.exe'))
```
Kami, line 55: Can you please also make sure we include ``build_info`` file and install it in the same directory where the agent is installed?

This file is created as part of ``build_package.py`` build pricess and it contains information about the git revision which was used to build the package.

(Might also need some changes to also create that file for Windows builds, but hopefully it shouldn't be too hard)

Kami:
While doing a quick manual test, I noticed that ``build_info`` file is not correctly included with the Windows package so I'm also working on fixing that (that was an oversight from me in the past while working on #537).

Basically making sure we include that file, making sure build_info.py functionality works on Windows and adding additional checks to the Linux and Windows AMI tests to assert that this file is indeed present.

.circleci/config.yml
Kami: Basically same thing as we do in one of the Dockerfiles - we append git commit sha to the version.

And we should also bump version in ``VERSION`` file so we can distinguish between different versions on upgrade tests.

Right now if you look at the script output, same version will be printed in both cases (pre and after the upgrade) which is confusing and incorrect.

Kami: And to give some context - right now we use the same version for current / from and to package version which means our upgrade tests are not 100% correct and robust - the version is the same in both cases so we can't correctly assert if the upgrade was successful or not.

Anyway, I pushed some changes which I believe should address that.

Before: https://41733-23852161-gh.circle-artifacts.com/0/outputs/ubuntu1804.log (as you can see the version pre and post upgrade is the same)
After: https://41741-23852161-gh.circle-artifacts.com/0/outputs/ubuntu1804.log (version after the upgrade is different)

Kami:
Another thing we should look into in the near future.

Right now if a build is canceled (either manually or via push), we will leave running EC2 instances laying around.

We need to ensure we periodically perform a cleanup.

I think one option would be to have a daily build job which looks for automated test instances (based on the instance name) which are older than X hours and deleted them.

We just need to be careful we correctly scope the query so we don't accidentally delete non-testing related instances.

tests/ami/scripts/test_deb.sh.j2
ArthurKamalov: Oh yeah, thank you for fixing that.

Kami:
@ArthurKamalov I removed testing changes, fixed build info tests, synced up package-tests jobs (removed the ones which are not needed anymore and made sure ``build-*-package`` jobs also run on non-master branch).

This should now be ready to be merged once you sort out the Windows build VM :)

PR #540, opened 2020-05-19
AGENT-398: Leaky bucket meter sleep API

yanscalyr:
Need to decide on an interface for this class, and write up tests for it.


yanscalyr:
We have a class for this, it just needs modifications

PR #541, opened 2020-05-19
Add additional config path as an option

imron:
Under kubernetes, it is not convenient to specify custom configuration snippets in the agent.d directory without overriding core configuration snippets.

This PR allows users to keep the default configuration snippets we provide, but also provide their own configuration snippets in an additional configuration directory.

This additional configuration directory is specified via an environment variable or a command line parameter.

If present, the agent will process all .json files in this directory and add them to its current configuration.

This only applies to the main agent, and not the agent config scripts.

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -149,6 +149,10 @@ class ScalyrAgent(object):
         self.__default_paths = platform_controller.default_paths
 
         self.__config_file_path = None
+
+        # An extra directory for config snippets
+        self.__extra_config_dir = None
+
         # If the current contents of the configuration file has errors in it, then this will be set to the config
         # object produced by reading it.
         self.__current_bad_config = None
@@ -244,6 +248,10 @@ class ScalyrAgent(object):
         status_format = command_options.status_format
         self.__no_fork = command_options.no_fork
         no_check_remote = False
+
+        self.__extra_config_dir = Configuration.get_extra_config_dir(
+            command_options.extra_config_dir
+        )
 
         # We process for the 'version' command early since we do not need the configuration file for it.
         if command == "version":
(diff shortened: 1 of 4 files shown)


scalyr_agent/agent_main.py
```diff
@@ -1750,6 +1763,12 @@ def stop(self):
         help="Read configuration from FILE",
         metavar="FILE",
     )
+    parser.add_option(
+        "--additional-config-dir",
```
czerwingithub, line 1767: Ok, something feels a bit awkward about calling this additional-config-dir and then treating it as something extra, on top, of the normal one.  What if we just allowed for this to be a list and default it to `agent.d`.  So, we would have a config option `--config-dir` or `--config-dirs` where it could be a comma separated list of configuration directories.  Same with the environment variable.

I know, it's bit of a change from what you've done.  But, these two items are really the same type of thing and it's more general to treat it as a list.

What do you think?  I'm going back and forth on it.  If don't change it to a list, then maybe a better name. 
 The `additional-config-dir` seems just too long / awkward.  Maybe `extra-config-dir`, or `custom-config-dir`, or `secondary-config-dir`?

Ok, I'll leave it up to.

imron: The problem with having it as a list is that someone could redefine the list - and accidentally leave of `agent.d`, which would then break the agent (at least in k8s world where we have a bunch of essential files in `agent.d`).

Happy to rename to `extra-config-dir`?

imron: I also imagine the most common usage of this (k8s) will be done with environment vars, and for non-k8s usage it's not really that necessary (this feature came about because k8s configmaps can't map sub-directories to files, but no such problem exists in standalone world).

PR #542, opened 2020-05-19
AGENT-373 K8 agent doesn't collect logs from a CronJob 

imron:
When the k8s monitor is using the docker socket for querying containers, it misses out on short-lived pods.

The reason this happens is that when we query the docker socket, we query all containers (running and stopped), but filter them by containers started after the previous time we queried the containers.

This allows us to pick up short-lived containers that started and stopped since our previous query.

However, since the k8s ratelimiter changes were added, if no matching pod is found in the k8s cache then the container is also skipped. This will always happen the first time a pod is queried on the cache, and so it means that short-lived pods are now ignored, because they won't show up when we next query the list of running containers, because it will only show containers created after our previous call.

There is no clean way to address this issue, however after investigating a number of different options and discussions with Steven, the way we will handle this is to extend the amount of time we allow for dead containers.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -1000,6 +1000,7 @@ class ControlledCacheWarmer(StoppableThread):
         # The list of containers whose pods are eligible to be warmed in the cache.  Essentially, the active_pods
         # minus those already warmed or blacklisted.
         self.__containers_to_warm = []
+
         self.__max_failure_count = max_failure_count
         self.__blacklist_time_secs = blacklist_time_secs
         self.__max_query_retries = max_query_retries
@@ -1053,6 +1054,49 @@ class ControlledCacheWarmer(StoppableThread):
             self.__condition_var.notify_all()
         finally:
             self.__condition_var.release()
+
+    def mark_has_been_used(self, container_id, unmark_to_warm=False):
+        """
+        Marks a WarmingEntry as having been used.  This typically means that
+        the container id has been returned as a container that should be
+        watched by the agent.
+        @param container_id: the container id of the warming entry to mark as used
+        @param unmark_to_warm: whether or not to unmark the WarmingEntry as something to warm
+        """
+        self.__lock.acquire()
+        try:
+            entry = self.__active_pods.get(container_id, None)
+            if entry:
+                entry.has_been_used = True
+
+                if unmark_to_warm:
+                    entry.is_recently_marked = False
+
+        finally:
+            self.__lock.release()
+
+    def pending_first_use(self, container_id):
+        """
+        Returns whether a WarmingEntry for a container_id is awaiting its first use
+
+        @param container_id: the container id to check
+
+        @return: Whether or not the entry has been marked as used. If this is
+        False, it typically means this container's log has not yet been
+        included in the list of logs to copy. If this container is not part of
+        the active set, always returns False.
+        """
+        self.__lock.acquire()
+        try:
+            entry = self.__active_pods.get(container_id, None)
+            if entry:
+                return not entry.has_been_used
+
+        finally:
+            self.__lock.release()
+
+        # if we are here, then the container id didn't have a warming entry.
+        return False
 
     def begin_marking(self):
         """Must be invoked before any container is marked as being active, by invoking `mark_to_warm`.
@@ -1569,9 +1613,11 @@ class ControlledCacheWarmer(StoppableThread):
             self.pod_namespace = pod_namespace
             # The associated pod's name.
             self.pod_name = pod_name
-            # The time when this pod was first requested to be warmed
+            # Whether or not this is a new warming entry that has never been used
+            self.has_been_used = False
+            # The most recent time when this pod was requested to be warmed - gets updated after cache expiry
             self.start_time = None
-            # Whether or not it has been successfully warmed.
+            # Whether or not it is currently warmed.
             self.is_warm = False
             # Used to indicate if it is been marked in the most recent marking run started by `begin_marking`.
             self.is_recently_marked = False
@@ -1673,9 +1719,18 @@ def _get_containers(
                 # Note we need to *include* results that were created after the 'running_or_created_after' time.
                 # that means we need to *ignore* any containers created before that
                 # hence the reason 'create_before' is assigned to a value named '...created_after'
-                if _ignore_old_dead_container(
+                is_old_dead_container = _ignore_old_dead_container(
                     container, created_before=running_or_created_after
-                ):
+                )
+
+                # AGENT-373/270 - we don't want to ignore containers that are still
+                # pending first use, otherwise we will miss out on short-lived logs
+                pending_first_use = False
+                if controlled_warmer is not None:
+                    pending_first_use = controlled_warmer.pending_first_use(cid)
+
+                # ignore containers that are old and dead and whose logs are already being copied
+                if is_old_dead_container and not pending_first_use:
                     continue
 
                 if len(container["Names"]) > 0:
@@ -1814,6 +1869,15 @@ def _get_containers(
                                             ignore_k8s_api_exception=True,
                                         )
                                         if pod:
+                                            # We've read the pod from the cache, so any WarmingEntry associated
+                                            # with this has now been used.  This ensures we pick up short lived
+                                            # logs that finished before they had a warm entry in the k8s_cache
+                                            if controlled_warmer is not None:
+                                                controlled_warmer.mark_has_been_used(
+                                                    cid,
+                                                    unmark_to_warm=is_old_dead_container,
+                                                )
+
                                             k8s_info["pod_info"] = pod
 
                                             k8s_container = k8s_info.get(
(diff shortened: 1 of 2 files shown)


imron:
I had spotted the same flaw and was trying to think of a good way out of it and I think what you've suggested is good.

The code in this commit is because I'm trying to get in to the habit of creating [WIP] draft PRs once I start working on a ticket, and I couldn't find a way to do that without having at least one commit on the branch, hence that initial commit.

imron:
Ok. This approach doesn't work either.

Imagine:

1. 0s - first iteration of loop
2. 2s - container 123 created and finished
3. 5s - first iteration of the loop where 123 is detected
   previous iteration was at 0s, and 123 was created after 0s so it is included
   but... 123 isn't in cache so it is marked as 'waiting to warm' and dropped from this iteration
   cachewarmer then sends query to api for container 123
4. 7s - cachewarmer receives response from api, and container 123 is no longer pending because it is now warm
5. 10s - second iteration through the loop where 123 is detected
   we see container 123 was created before 5s so plan to ignore it
   we check cachewarmer for pending containers, but 123 is no longer pending, so we do ignore it

Instead I've gone with a variation on this approach

* When a WarmingEntry is created, it is marked as `new`.
* We only ignore dead containers if the WarmingEntry for the container is not new
* We don't set the WarmingEntry to 'not new' until we have read it from the cache at least once.

There's no need to worry about building up dead containers because, because k8s will eventually remove them from docker, and then we'll stop trying to query them.

I've tested this and it works, and will be uploading it in a moment.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1054,6 +1055,38 @@ def _prepare_to_stop(self):
         finally:
             self.__condition_var.release()
 
+    def mark_not_new(self, container_id):
```
czerwingithub, line 1058: I think `new` is too overloaded in this context.  Can we call this `mark_as_used` where we define if it we've used this warmed cache entry or not.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1054,6 +1055,38 @@ def _prepare_to_stop(self):
...
+        self.__lock.acquire()
+        try:
+            entry = self.__active_pods.get(container_id, None)
+            if entry:
+                entry.is_new = False
+
+        finally:
+            self.__lock.release()
+
+    def is_new(self, container_id):
```
czerwingithub, line 1072: Likewise, this turns into `has_been_used`.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1569,9 +1602,11 @@ def __init__(self, pod_namespace, pod_name):
             self.pod_namespace = pod_namespace
             # The associated pod's name.
             self.pod_name = pod_name
-            # The time when this pod was first requested to be warmed
+            # Whether or not this is a new warming entry that has never been warmed
+            self.is_new = True
```
czerwingithub, line 1606: s/is_new/has_been_used

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1670,13 +1705,22 @@ def _get_containers(
...
-                # Note we need to *include* results that were created after the 'running_or_created_after' time.
-                # that means we need to *ignore* any containers created before that
-                # hence the reason 'create_before' is assigned to a value named '...created_after'
-                if _ignore_old_dead_container(
-                    container, created_before=running_or_created_after
-                ):
-                    continue
+                # AGENT-373/270 - we don't want to ignore containers that are still
+                # waiting to be warmed, otherwise we will miss out on short-lived
+                # logs
```
czerwingithub, line 1710: Ok, I would rewrite this a little bit differently in order to better capture what's going on.  I also want to use the `is_old_dead_container`:

```
       is_old_dead_container = _ignore_old_dead_container(container, created_before=running_or_created_after)
       has_been_used = controlled_warmer.has_been_used(container) if controlled_warmer is not None else False
       if is_old_dead_container and has_been_used:
           continue:
```

It kind of would be better to flip the sense of this, but that's hard to do given everything else in this loop just uses continues.  Basically, I want it to be pretty apparent that we allow in alive containers, recently dead containers, and old dead containers if we haven't used their cache entry yet.  Maybe we can list all of those cases just in a comment.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1814,6 +1858,12 @@ def _get_containers(
                                             ignore_k8s_api_exception=True,
                                         )
                                         if pod:
+                                            # We've read the pod from the cache, so any WarmingEntry associated
+                                            # with this is no longer new.  This ensures we pick up short lived
+                                            # logs that finished before they had a warm entry in the k8s_cache
+                                            if controlled_warmer is not None:
+                                                controlled_warmer.mark_not_new(cid)
```
czerwingithub, line 1865: I would also propose we add in:

```
   if is_old_dead_container:
       controlled_warmer.unmark_to_warm(cid, namespace, pod_name)
```

The main reason is that my head hurts that when we finish marking in this iteration, old dead containers will still in the active set.  I think that would be surprising ... even though with how the code currently is written, it doesn't make a difference.  However, who knows how we could change this in the future, so I think it's better to just make is so that the dead container doesn't get included.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: As per conversation earlier, this was changed to `pending_first_use`.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Done, but note that I didn't use the ternary syntax `pending_first_use = controlled_warmer.pending_first_use(container) if controlled_warmer is not None else False` because the Black formatter made it just as many lines as the original but made it more awkward to read.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Added a parameter to `mark_has_been_used` to say whether or not to also unmark from the warmer.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1054,6 +1055,45 @@ def _prepare_to_stop(self):
         finally:
             self.__condition_var.release()
 
+    def mark_has_been_used(self, container_id, unmark_to_warm=False):
+        """
+        Marks a WarmingEntry as having been used
```
czerwingithub, line 1060: I would just write this as
```
Marks a WarmEntry as having been used.  This typically means that the container id has been returned as a container that should be watched by the agent. 
```

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1054,6 +1055,45 @@ def _prepare_to_stop(self):
...
+        @param container_id: the container id of the warming entry to mark as used
+        @param unmark_to_warm: whether or not to unmark the WarmingEntry as somethign to warm
```
czerwingithub, line 1062: s/somethign/something/

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1054,6 +1055,45 @@ def _prepare_to_stop(self):
...
+        finally:
+            self.__lock.release()
+
+    def pending_first_use(self, container_id):
+        """
+        Returns whether a WarmingEntry for a container_id is awaiting its first use
+
+        @param container_id: the container id to check
+
+        @return: Whether or not the entry has been consumed by the agent, or False if the
```
czerwingithub, line 1082: I would rewrite this as

```Whether or not the entry has been marked as used.  If this is False, it typically means this container's log has not yet been included in the list of logs to copy.  If this container is not part of the active set, always returns False.```

File: scalyr_agent/builtin_monitors/kubernetes_monitor.py [hunk 9]
```diff
@@ -1569,9 +1609,11 @@ def __init__(self, pod_namespace, pod_name):
             self.pod_namespace = pod_namespace
             # The associated pod's name.
             self.pod_name = pod_name
-            # The time when this pod was first requested to be warmed
+            # Whether or not this is a new warming entry that has never been warmed
```
czerwingithub, line 1612: s/warmed/used/

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -1673,9 +1715,18 @@ def _get_containers(
...
-                ):
+                )
+
+                # AGENT-373/270 - we don't want to ignore containers that are still
+                # pending first use, otherwise we will miss out on short-lived logs
+                pending_first_use = False
+                if controlled_warmer is not None:
+                    pending_first_use = controlled_warmer.pending_first_use(cid)
+
+                # ignore containers that are old and dead, and that have been seen by the agent at least once
```
czerwingithub, line 1728: How about "ignore containers that are old and dead and whose logs are already being copied"

PR #543, opened 2020-05-20
AGENT-398: Add a method to RateLimiter that will block until there is capacity

yanscalyr:


scalyr_agent/util.py
```diff
@@ -1257,6 +1257,25 @@ def charge_if_available(self, num_bytes, current_time=None):
 
         @return: True if there are enough room in the rate limit to allow the operation.
         """
+        if self._get_time_to_sleep(num_bytes, current_time) > 0.0:
+            return False
+        else:
+            self.__bucket_contents -= num_bytes
+            return True
+
+    def sleep_and_charge(self, num_bytes, current_time=None):
```
czerwingithub, line 1266: I would call this `block_until_charge_succeeds`.  It's a bit wordy, but probably better to use `block` instead of `sleep` (it's a bit of an implementation detail we sleep) and I wanted to make it a little clearer as to what we are blocking on.

scalyr_agent/util.py
```diff
@@ -1257,6 +1257,25 @@ def charge_if_available(self, num_bytes, current_time=None):
...
+        """Sleeps until there are enough bytes available for an operation costing num_bytes, then charges that amount.
+
+        @param num_bytes: The number of bytes to consume from the rate limit.
+        @param current_time: If not none, the value to use as the current time, expressed in seconds past epoch. This
+            is used in testing.
+        """
+        time_to_sleep = self._get_time_to_sleep(num_bytes, current_time)
+        if time_to_sleep > 0.0:
+            time.sleep(time_to_sleep)
```
czerwingithub, line 1275: Technically, if someone sets a zero fill rate, we could try to sleep infinity.  I would instead check at the start of this method if fill rate is zero, and if so, throw an exception saying you cannot use this method if you have a zero fill rate.

scalyr_agent/util.py
```diff
@@ -1257,6 +1257,25 @@ def charge_if_available(self, num_bytes, current_time=None):
...
         """
+        if self._get_time_to_sleep(num_bytes, current_time) > 0.0:
+            return False
+        else:
+            self.__bucket_contents -= num_bytes
+            return True
+
+    def sleep_and_charge(self, num_bytes, current_time=None):
+        """Sleeps until there are enough bytes available for an operation costing num_bytes, then charges that amount.
+
```
czerwingithub, line 1268: I would also say "As a reminder, this abstraction is not thread-safe so no other method should be invoked on this instance while it is blocking."

scalyr_agent/util.py
```diff
@@ -1257,6 +1257,25 @@ def charge_if_available(self, num_bytes, current_time=None):
...
         @return: True if there are enough room in the rate limit to allow the operation.
         """
+        if self._get_time_to_sleep(num_bytes, current_time) > 0.0:
+            return False
+        else:
+            self.__bucket_contents -= num_bytes
+            return True
+
+    def sleep_and_charge(self, num_bytes, current_time=None):
+        """Sleeps until there are enough bytes available for an operation costing num_bytes, then charges that amount.
```
czerwingithub, line 1267: s/Sleeps/Blocks

scalyr_agent/util.py
czerwingithub, line 1268: Also, I see I didn't call this out originally (that the abstraction is not thread-safe.)  Can you go ahead and add a comment to the class comment about it not being thread safe please?

scalyr_agent/util.py
```diff
@@ -1257,6 +1257,25 @@ def charge_if_available(self, num_bytes, current_time=None):
...
+        @param num_bytes: The number of bytes to consume from the rate limit.
+        @param current_time: If not none, the value to use as the current time, expressed in seconds past epoch. This
+            is used in testing.
+        """
+        time_to_sleep = self._get_time_to_sleep(num_bytes, current_time)
+        if time_to_sleep > 0.0:
+            time.sleep(time_to_sleep)
+        self.__bucket_contents -= num_bytes
+
+    def _get_time_to_sleep(self, num_bytes, current_time=None):
```
czerwingithub, line 1278: Add a little documentation here.

tests/unit/util/util_test.py
```diff
@@ -415,6 +425,105 @@ def test_refill(self):
         self.advance_time(1)
         self.assertTrue(self.charge_if_available(60))
 
+    def test_basic_use_sleep(self):
+        def fake_sleep(self2, seconds):
```
czerwingithub, line 429: Using mock patch, there's an easier way to do this now.  Search for `@mock.patch("scalyr_agent.scalyr_client.time.time", mock.Mock(return_value=0))` in the unittests to see examples of how you can override methods with decorators and context objects.  Since you actually need to advance time, you probably should use a context manager.  It would look something like this:

```
  with mock.patch("scalyr_agent.util.time.sleep", self.fake_sleep):
      ## Do the rest of your test here.
``` 

In this example, I'm assuming you define a `fake_sleep` method at the `TestRateLimiter` level.  I'm not quite sure why you keep declaring `fake_sleep` within the method.. I think you just need to do it once and reuse that.

The context object will handle cleaning up the patch state.. so you don't have to do the `try .. finally` yourself.

We could also use this approach to patch `time.time` as well so that we don't have to pass in `current_time` everywhere.  I would mark that as a TODO for now... I'm not crazy about using different testing approaches in the same class (i.e., passing down current time to override time.time vs patching time.sleep) but it's going to be a little bit of work to change all of this.

yanscalyr: The redefinition was due to some weirdness with how many parameters the method ends up expecting when defined outside the method, probably won't be an issue with this approach.

scalyr_agent/util.py
```diff
@@ -1218,7 +1218,7 @@ def is_alive(self):
 
 class RateLimiter(object):
     """An abstraction that can be used to enforce some sort of rate limit, expressed as a maximum number
-    of bytes to be consumed over a period of time.
+    of bytes to be consumed over a period of time. This abstraction is not thread-safe.
```
czerwingithub, line 1221: Put the comment about it not being thread-safe on a separate line to draw more attention to it.

PR #545, opened 2020-05-21
Method to parse strings representing data rates

yanscalyr:
I'm conflicted on if we should check for bytes vs bits here, as someone might get tripped up on that without realizing.

### DIFF
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -20,6 +20,8 @@ from __future__ import division
 from __future__ import absolute_import
 from __future__ import print_function
 
+import re
+
 if False:
     from typing import Union
     from typing import Tuple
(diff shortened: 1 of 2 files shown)


scalyr_agent/util.py
```diff
@@ -677,6 +679,46 @@ def get_web_url_from_upload_url(server):
...
+            value = value / 60
+        elif denominator == "H":
+            value = value / 60 / 60
+        elif denominator == "D":
+            value = value / 60 / 60 / 24
+        elif denominator == "W":
+            value = value / 60 / 60 / 24 / 7
+
+        return value
+    raise ValueError("Could not parse data rate string")
```
czerwingithub, line 719: Include the input string here.  It's usually good to see what value tripped us up.

scalyr_agent/util.py
```diff
@@ -677,6 +679,46 @@ def get_web_url_from_upload_url(server):
...
+    """
+    m = re.search(r"(-?\d+\.?\d*)\s*([kKmMgGtT]?)([bB])/([sSmMhHdDwW])", value)
+    if m:
+        value = float(m.group(1))
+        numerator = m.group(2).upper()
+        bit_or_byte = m.group(3)
+        denominator = m.group(4).upper()
+
+        if numerator == "K":
+            value = value * 1024
```
czerwingithub, line 698: This (and below) should be multiplying by 1000.  There's a difference between KB and KiB.  KB is 1000 and KiB is 1024.  See this article

https://www.linkedin.com/pulse/20140814132922-176099595-mb-vs-mib-gb-vs-gib/

For clarity, maybe you should write these out in using powers.  For example, this should be 10^3 and M should be 10^6, etc.

czerwingithub, line 698: For bonus points, you could be in support for MiB vs MB... but I don't think it is really necessary.

czerwingithub:
@yanscalyr Yeah, I'm a little concerned too about people accidentally using bits when they mean bytes.  Let's disallow using `b` and if they do, throw a value error with an explicit error message saying we are disallowing `b` to ensure they only specify it in terms of bytes per second.

PR #548, opened 2020-05-22
AGENT-401: Implement new send rate defaults and fallback switches

yanscalyr:
Updating the defaults for several configuration parameters as part of AGENT-396:

| Config option                | Old default | New default |
|------------------------------|-------------|-------------|
| max_allowed_request_size     | 1,048,576   | 5,900,000   |
| pipeline_threshold           | 1.1         | 0           |
| min_request_spacing_interval | 1.0         | 0.1 or 0?   |
| max_request_spacing_interval | 5.0         | 1.0 or 5?   |
| max_log_offset_size          | 5,242,880   | 200,000,000 |
| max_existing_log_offset_size | 104,857,600 | 200,000,000 |

These can be reverted using one of two new config options, setting `max_send_rate_enforcement` to `"legacy"` or `disable_max_send_rate_enforcement_overrides` to `true`.

The main use of `max_send_rate_enforcement`, which is to put an explicit limit on the upload rate of logs from the Agent, is not implemented in this PR.


Kami:
Is this ready for a review?

If / when it is, please add some more context, etc. to the PR description.

yanscalyr:
Not quite, considering all the failing tests especially. I'll mark it as draft for now

Kami:
Once the test issue has been solved, let's please also document this default values and behavior change in the upgrade notes.

scalyr_agent/configuration.py
```diff
@@ -223,6 +223,53 @@ def parse(self):
                 if https_server != server:
                     self.__config["scalyr_server"] = https_server
 
+            # Set defaults based on `max_send_rate_enforcement` value
+            if (
+                self.__config["max_send_rate_enforcement"] == "legacy"
+                or self.__config["disable_max_send_rate_enforcement_overrides"]
+            ):
+                if "max_allowed_request_size" not in self.__config:
+                    self.__config["max_allowed_request_size"] = 1048576
```
Kami, line 232: To make it easier to read it would be nice to use the NB to bytes notation we used before - ``1 * 1024 * 1024``

Kami: Same goes for other values.

Kami:
It appears 085b706 fixed the weird test failure issue we've been seeing.

It appears the issue was the same one we originally noticed after merging #466 and before merging #495.

This kinda makes sense because the symptoms were the same - some data / files were not getting ingested.

This could be explained if log handlers and filters some how got mixed up and some data was incorrectly filtered out.

At that time, I suggested the same fix, but #495 seemed to have worked as well so I was fine with it.

Looks like #495 didn't fully fix it though.

Why does this issue only happen in some situations / scenarios? No idea, likely timing related.

yanscalyr:
I can't figure out a way to get this working while also having debug logs printed to stdout when running unforked. For now we decided to live with no debug logs in stdout and will return to this issue later.

Kami:
> I can't figure out a way to get this working while also having debug logs printed to stdout when running unforked. For now we decided to live with no debug logs in stdout and will return to this issue later.

Yeah, this seems like a reasonable compromise.

At least now we know where the problem / root cause lies (StdoutFilter) and focus on this code path in the future when trying to fix.

Kami:
While looking into this, I noticed ``stdout_severity`` configuration option doesn't work correctly.

I fixed it in 4a44081.

I still need to add a changelog entry and some tests would be good as well :)

Kami:
While looking into this I was able to track down and reproduce yet another issue - bb9035987994394fe9cd7104bbe39df26b5f8b53.

This was reported by Imron quite a while ago, but I never encountered the issue myself until today.

That commit should fix that issue.

Kami:
Noticed and fixed yet another issue - d8a9baf976b86465c2a7030f630acb2c5b022302.

For details, please refer to the commit message, but basically that caused some log messages which were produced after calling close_handlers() and before agent actually shutting down being lost.

Before (lost messages):

```bash
2020-05-28 15:55:33.464Z DEBUG [core] [platform_posix.py:640] Received SIGTERM signal
2020-05-28 15:55:33.464Z DEBUG [core] [platform_posix.py:644] Invoking termination handler...
2020-05-28 15:55:33.465Z INFO [core] [agent_main.py:610] Received signal to shutdown, attempt to shutdown cleanly.
No handlers could be found for logger "scalyr_agent"
```

After (no lost messages):

```bash
2020-05-28 15:59:29.406Z DEBUG [core] [platform_posix.py:640] Received SIGTERM signal
2020-05-28 15:59:29.406Z DEBUG [core] [platform_posix.py:644] Invoking termination handler...
2020-05-28 15:59:29.408Z INFO [core] [agent_main.py:610] Received signal to shutdown, attempt to shutdown cleanly.
2020-05-28 15:59:29.413Z INFO [core] [agent_main.py:1534] agent_status launch_time="Thu May 28 15:59:22 2020 UTC" version="2.1.4" watched_paths=9 copying_paths=4 total_bytes_copied=41551 total_bytes_skipped=0 total_bytes_subsampled=0 total_redactions=0 total_bytes_failed=0 total_copy_request_errors=0 total_monitor_reported_lines=392 running_monitors=2 dead_monitors=0 user_cpu_=0.255594 system_cpu=0.062967 ram_usage=36068
2020-05-28 15:59:29.413Z DEBUG [core] [agent_main.py:1755] Shutting down monitors
2020-05-28 15:59:29.413Z INFO [core] [monitors_manager.py:187] Stopping monitor linux_system_metrics()
2020-05-28 15:59:29.414Z INFO [core] [monitors_manager.py:187] Stopping monitor linux_process_metrics(agent)
2020-05-28 15:59:29.461Z INFO [monitor:linux_process_metrics(agent)] [scalyr_monitor.py:292] Monitor has finished
2020-05-28 15:59:29.462Z INFO [monitor:linux_system_metrics()] [tcollector.py:1173] shutting down children
2020-05-28 15:59:29.670Z Level 4 [core] [scalyr_client.py:522] Response was received with body "{"bytesCharged":0,"status":"success"}"
2020-05-28 15:59:29.671Z DEBUG [core] [copying_manager.py:802] Sent 184 bytes and received response with status="success".
2020-05-28 15:59:29.672Z DEBUG [core] [copying_manager.py:709] At top of copy log files loop.
2020-05-28 15:59:29.672Z DEBUG [core] [copying_manager.py:1145] Getting batch of events to send. (pipelining=False)
2020-05-28 15:59:29.673Z DEBUG [core] [copying_manager.py:1242] Information for batch of events. (pipelining=False): bytes=973.0 files=4.0 lines=6.0
2020-05-28 15:59:29.673Z DEBUG [core] [copying_manager.py:759] Sending an add event request
```

yanscalyr:
There is a plan to loadtest these changes once the overall PR is done, there is a couple of options we need to settle on a real value for. In case you aren't aware here is the notion doc for these changes: https://www.notion.so/scalyr/Fixing-Skipped-Logs-Improved-Rate-Limiting-574924085be24aaab42e98beb375c017

For release notes I was planning to do that as its own PR to avoid annoying conflicts since we have a release planned before this makes it in, but honestly it probably wouldn't be that bad.

Kami:
@yanscalyr Ideally we would add corresponding changelog + release note entries in the PR which adds those changes, but adding it later is fine as well.

As far as notion document goes - yeah, I just talked about that to Steven earlier and checked the documentation so it should be fine.

scalyr_agent/configuration.py
```diff
@@ -223,6 +223,53 @@ def parse(self):
...
+                if "min_request_spacing_interval" not in self.__config:
+                    self.__config["min_request_spacing_interval"] = 1.0
+                if "max_request_spacing_interval" not in self.__config:
+                    self.__config["max_request_spacing_interval"] = 5.0
+                if "max_log_offset_size" not in self.__config:
+                    self.__config["max_log_offset_size"] = 5 * 1024 * 1024
+                if "max_existing_log_offset_size" not in self.__config:
+                    self.__config["max_existing_log_offset_size"] = 100 * 1024 * 1024
+            else:
+                if "max_allowed_request_size" not in self.__config:
```
czerwingithub, line 244: Sorry if it wasn't clear from the spec, but these values (like `5900000`) should be used no matter what, even if the customer has some value defined in the config.

And yes, thinking about it, calling it "overriding defaults" is misleading in this way.  My apologies for that.

So, you shouldn't be doing any of the `not in self.__config` checks.

scalyr_agent/configuration.py
```diff
@@ -1500,10 +1568,27 @@ def __verify_main_config_and_apply_defaults(
...
+            description,
+            apply_defaults,
+            env_aware=True,
+        )
+
         self.__verify_or_set_optional_int(
             config,
             "max_allowed_request_size",
-            1 * 1024 * 1024,
+            None,
```
czerwingithub, line 1591: Still set the defaults here.  Do that here and below.

scalyr_agent/platform_posix.py
```diff
@@ -634,10 +634,14 @@ def handle_terminate(signal_num, frame):
 
             It takes care of invoking the agent termination handler function.
             """
-            logger.debug("Received SIGTERM signal")
+            # Termination handler will cause log handlers to be closed so we need to check if they
+            # are still opened before trying to log a message.
+            if not scalyr_logging.HANDLERS_CLOSED:
```
czerwingithub, line 639: @Kami Are we getting an error if the handlers are closed?  I would think we want to handle that error down in the logger classes then.  If we are getting an error, it sounds like we are trying to write to a closed handler.

tests/unit/configuration_test.py
```diff
@@ -2376,3 +2389,45 @@ def test___verify_or_set_optional_string_with_valid_values(self):
...
+        self.assertIsNone(config.parsed_max_send_rate_enforcement)
+
+        self.assertEquals(config.max_allowed_request_size, 1048576)
+        self.assertEquals(config.pipeline_threshold, 1.1)
+        self.assertEquals(config.min_request_spacing_interval, 1.0)
+        self.assertEquals(config.max_request_spacing_interval, 5.0)
+        self.assertEquals(config.max_log_offset_size, 5242880)
+        self.assertEquals(config.max_existing_log_offset_size, 104857600)
+
+    def test_disable_max_send_rate_enforcement_overrides(self):
```
czerwingithub, line 2414: Also add a test case that verifies that if `max_send_rate_enforcement` is set to `unlimited` and you have values set in the configuration file for all of these parameters, then you still get the unlimited settings.

scalyr_agent/platform_posix.py
Kami: Correct, we will see a bunch of messages like this on SIGTERM in some situations when agent is busy and doesn't shut down immediately in first SIGTERM (bb9035987994394fe9cd7104bbe39df26b5f8b53):

```bash
IOError: [Errno 0] Error
Logged from file platform_posix.py, line 637
  File "/usr/local/lib/python2.7/logging/__init__.py", line 889, in emit
    stream.write(fs % msg)
```

When shutting down the agent using agent_main.py stop we try to send SIGTERM for up to 5 seconds (with 0.01 second delay in between) before sending SIGKILL if the agent still didn't shut down after this period has passed.

And the first SIGTERM signal invocation will cause log handlers to be closed so subsequent logger messages will fail.

Another option would also be to have some global variable which indicates if SIGTERM handler function has already be called and do a no-op if it has.

czerwingithub: But, if a log handler has been closed, why hasn't it been removed?  In my mental model of how the Python logging modules work... you insert messages to be logged, there's some dispatching logic, and a set of handlers decide which messages they will emit.  If we close a handler, I would think it should have been removed from the list of handlers that can emit logs.  Is that not the case, or am I not understanding something?

scalyr_agent/configuration.py
```diff
@@ -223,6 +223,58 @@ def parse(self):
...
+                )
+                self._warn_of_override_due_to_rate_enforcement(
+                    "max_log_offset_size", 5 * 1024 * 1024
+                )
+                self._warn_of_override_due_to_rate_enforcement(
+                    "max_existing_log_offset_size", 100 * 1024 * 1024
+                )
+
+                if self.__config["max_send_rate_enforcement"] == "legacy":
+                    self.__config["max_allowed_request_size"] = 1024 * 1024
```
czerwingithub, line 248: Ok... what happens here if the customer has set a value for `max_allowed_request_size`?  If `legacy` is selected, then we should either use the old default value or the value from the configuration file if it has been set.

yanscalyr: Currently it forces the old default value (I should add a test for that too actually).

czerwingithub: Right, but we want it do is:  if `legacy` is selected, first use the value from the configuration file, or if there is not a value, the old default value.

PR #549, opened 2020-05-23
[WIP] Shard Logs Across Multiple Copying Managers

imron:
Problem
---

Due to limitations with the current ingest API, the Scalyr Agent can
only have a single request in flight for a given Scalyr Agent session,
and a single request can have a maximum of 6MB of data after
decompression.

This places a limit on the amount of data a single agent can upload to
the server, and a number of users are now running in to this limitation.

This branch is a draft to experiment with having multiple
CopyingManagers, each monitoring a subset of logs, and each with its own
connection and session.

This should allow us to have multiple requests in flight - each from a
different session.

Proposed Solution
---

Create a CopyingManagerFacade with an interface identical to the current
CopyingManager.  The rest of the application interacts with the facade
as normal, and the facade is then responsible for managing the
complexity of managing multiple CopyingManagers and sharding logs
between them.

This initial commit doesn't contain anything except a stub for the
CopyManagerFacade.  The main purpose of this commit is to create a draft
PR for design discussion.


scalyr_agent/copying_manager.py
```diff
@@ -200,6 +200,340 @@ def __init__(self, add_events_request, completion_callback):
         self.next_pipelined_task = None
 
 
+class CopyingManagerFacade(StoppableThread, LogWatcher):
+    """
+    The CopyingManagerFacade is used to hide the complexity of having sharded CopyingManagers.
+
+    It conforms to the same interface as the original CopyingManager, and is responsible for creating
+    and managing multiple copying managers and splitting logs between them.
```
czerwingithub, line 208: In general I think the facade approach is the right way to go with.  Have you given any thought to which methods are needed by the facade version of the CopyManager?  I don't think you need all of the ones below... for example, it won't need a `_send_events` method and a bunch of the private methods.

We'll need to figure out division of responsibility as well.  Does each CopyManager shard have its own checkpoint file that it writes to?  How do we handle that when the number of shards changes? etc

I'm also not a big fan of using `Facade` in the name for this class.. I might go with something like `ShardedCopyManager`.  But, I'm also not going to fight hard on that one.

imron: > for example, it won't need a _send_events method and a bunch of the private methods.

Initially it needed them for the tests because I had made CopyingManager private to the module and just exposed CopyingManagerFacade passing through to an internal copying manager and all the tests failed because the tests override CopyingManager.

I'm moving away from that and will just keep the tests for the CopyingManager as is (but using the private class).

I have no preference for using `Facade`, just thought it made the intent clearer in the design if other people read it.  Happy to switch to `ShardedCopyManager` as we both agree on the general approach.

I'm leaving the checkpoints decision until after I've proven whether it works or not.

PR #550, opened 2020-05-28
CT-52: Increase default max_line_size

yanscalyr:


czerwingithub:
Actually, @yanscalyr Can you do me a favor and add a note about the change of default for this in `CHANGELOG.md`.  It should include a note to customers that if they have set this value in their own configuration files, they may wish to not set it anymore to use the new default.

PR #551, opened 2020-05-28
AGENT-402: Use `max_send_rate_enforcement` to rate limit sent log bytes

yanscalyr:
Currently blocked waiting for #548


scalyr_agent/copying_manager.py
```diff
@@ -846,6 +856,13 @@ def run(self):
 
                             if result == "success":
                                 last_success = current_time
+
+                            # Rate limit based on amount of copied log bytes in a successful request
+                            if self.__rate_limiter:
+                                self.__rate_limiter.block_until_charge_succeeds(
```
Kami, line 862: I know there are tests for the rate limiter class itself in #543, but it would also be good to have some tests which cover that scenario for copying manager as well.

scalyr_agent/copying_manager.py
```diff
@@ -224,6 +224,15 @@ def __init__(self, configuration, monitors):
         """
         StoppableThread.__init__(self, name="log copier thread")
         self.__config = configuration
+
+        # Rate limiter
+        self.__rate_limiter = None
+        if self.__config.parsed_max_send_rate_enforcement:
+            self.__rate_limiter = RateLimiter(
+                self.__config.parsed_max_send_rate_enforcement * 4.0,
```
Kami, line 232: Can you please a comment on why we use 4 multiplier on bucket size?

scalyr_agent/copying_manager.py
Kami: And ideally, we would also have end to end test for it - e.g. we would set ``max_send_rate_enforcement`` to some low value and then we would try to consume some large-ish file and make sure it's eventually consumed (taking into account sleep, etc.).

Kami, line 862: I also wonder if we should log a message in the rate limiter itself when we sleep when the limit is reached (with some limit on a maximum amount of log messages in some time period).

This would likely come handy in various troubleshooting scenario and answering the question of how many times we would actually exceeded the configured send rate and we need to wait / sleep.

And we could of course then also grep for that string in the corresponding end to end tests :)

yanscalyr: For the logging its planned as part of AGENT-397, a 5 minute log output if we rate limit. I'll probably have that in a separate PR.

yanscalyr: Due to time constraints on this release and me personally not seeing any easy to add tests for this the plan is to leave these for later. If anyone has suggestions for easy to add tests I will gladly take them though.

scalyr_agent/copying_manager.py
```diff
@@ -1232,6 +1256,8 @@ def handle_completed_callback(result):
                 else:
                     processor.close()
 
+            return total_bytes_copied
```
imron, line 1259: Can you add a `@return:` value to the function comments.

scalyr_agent/copying_manager.py
Kami: Alright, let's then please get back to this once the release is out.

PR #552, opened 2020-05-30
Fix syslog strings incompatibility

ArthurKamalov:
Small fix in `syslog` monitor which converts received data to unicode before pass it to the logger.


ArthurKamalov:
@Kami 
Now `SyslogHandler.handle` method receives only unicode, so other abstractions, such as UDP or TCP handlers have to convert data before pass it. This allows us to add string type check by creating a small mock.

ArthurKamalov:
>How about our new end to end monitor tests, why didn't those catch that? Or we still don't have end to end syslog monitor test?

Syslog monitor end-to-end test does checking by reading the resulting log file, so it is always string for it, and it looks for substrings, so it does not matter if there are lines which begin with b''

>I assume our unit tests didn't catch this because we mock stdout?

The same thing with unit tests, they just looked for a substring after reading the log file, so it was ok for them.

Kami:
@ArthurKamalov Thanks for the clarification.

Could we also add additional assertion to the end to end tests to ensure there is no ``b''`` prefix in those strings?

PR #553, opened 2020-05-30
Add lint check which will fail if any of the bundled certs will expire in 6 months or sooner

Kami:
This pull request adds a script which will fail the build if any of the bundled certs expires in 6 months or sooner.

It's unlikely that this issue will bite us again or any time soon since all the new certs expire many years in the future, but it doesn't hurt either.


scripts/check-bundled-ca-certs-expirations.py
```diff
@@ -0,0 +1,63 @@
...
+
+        with open(file_path, "rb") as fp:
+            content = fp.read()
+
+        cert = x509.load_pem_x509_certificate(content, default_backend())
+
+        now_dt = datetime.datetime.utcnow()
+        expire_in_days = (cert.not_valid_after - now_dt).days
+
+        if now_dt + datetime.timedelta(days=30 * 6) >= cert.not_valid_after:
```
imron, line 47: I'd increase this to 2 years or something.  We have customers that don't upgrade for years and years.

imron: We should also hook it up to [build_package.py](https://github.com/scalyr/scalyr-agent-2/blob/master/build_package.py#L943) and fail the build if the certs are due to expire within the same timeframe.

Kami: Yeah, I agree about increasing it even further than 6 months - for typical updating the web server cert scenario, 1 month notification is plenty, but in this case we are bundling the certs and like you said many of the users still run very old agent versions so it's good we give ourselves and our users plenty of heads up.

Kami: As far as build_package.py goes - I will look into it separately since this script depends on ``cryptography`` library and just need to first make sure it installs on Jenkins environment we use to roll out the release.

(Probably the best time to update build_package.py + dev requiremetns is to just before we make a release - to ensure it's indeed working)

Kami: I updated it to warn if it expires in 2 years or sooner.

I also had a look into hooking it up to ``build_package.py`` script (which I agree is a good idea), but right now this script doesn't depend on any 3rd party library so getting it to work will be a bit more involved and should probably be done separately (we also use that script in various places - Jenkins, various Dockerfiles, tests, etc.).

Kami: @imron ^are you OK with that?

One thing also worth noting is that we plan to move to a Circle CI + packagecloud based package release pipeline in the near future and at that point build_package.py will become obsolete so probably not worth spending too much time on it at this point.

scripts/check-bundled-ca-certs-expirations.py
```diff
@@ -0,0 +1,91 @@
...
+
+"""
+Script which errors out if any of the bundled certs will expire in 24 months or sooner.
+"""
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+if False:
+    from typing import List
```
imron, line 24: If this is never True, do we need to include it?

Kami: It's only used for type checking purposes is only evaluated when running mypy.

We can drop "if False" part once we drop support for Python 2.

PR #554, opened 2020-05-30
Don't fail if extra_config_dir is not defined

Kami:
I believe this should fix the issue we are seeing on Windows.

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -219,6 +219,7 @@ class ScalyrAgent(object):
         my_options.no_fork = True
         my_options.no_change_user = True
         my_options.no_check_remote = False
+        my_options.extra_config_dir = None
 
         if perform_config_check:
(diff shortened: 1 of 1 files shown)


Kami:
To add some context, we've seen this error in the Windows event log:

```bash
Error, causing Windows Service to exit early 'Options' object has no attribute 'extra_config_dir'
```

It looks like this code path is only exercised on Windows when daemonizing / forking the agent, but not on Linux (that's why none of the existing tests caught this issue).

It also works fine when using ``--no-fork`` since then we exercise a different code path (aka then we actually parse command line arguments).

Kami:
And another thing - once #538 is merged, we will have basic installer sanity tests for Windows as well and those should catch issues like that.

In fact, I think it would be good idea to confirm that - run tests in that branch against this commit 26f8e0a2385c3ed3dede1d204b9086344cc6ee85 and make sure they fail.

/cc @ArthurKamalov

PR #556, opened 2020-06-02
CT-25 - add support for k8s_logs in configuration file

imron:
This commit adds support for k8s_logs in the agent config file (or
config directories).

Specifically config files support a new field `k8s_logs` which is
identical to a regular `logs` stanza except:

* There is no `path` field supported
* There are 3 new optional fields, `k8s_pod_glob`, `k8s_namespace_glob`
  and `k8s_container_glob`, which are globs that default to `*`.

If any of those fields are present for a `k8s_logs` config then any
container log files added by the kubernetes_monitor for a given pod are
filtered against all of those globs, and if all of them match, then the
config is applied to the logs for that container.

Only the first matching config will be applied to a given log.

This means that users should order `k8s_logs` entries from most specific
to least specific.

Example config usage:

```
{
  "k8s_logs": [
    {
      // most specific, filtering on pod name and namespace
      "k8s_pod_glob": "nginx*",
      "k8s_namespace_glob": "web",
      "attributes" : { "parser": "nginxLog" }
    },
    {
      // filtering on just the container name defined in the k8s yaml
      "k8s_container_glob": "*redis*",
      "attributes" : { "parser": "redisLog" }
    },
    {
      // filtering just on namespaces - won't match anything already
      // matched by the previous config stanzas
      "k8s_namespace_glob": "*monitoring*",
      "attributes" : { "parser": "monitoringLogs" }
    },
    {
      // least specific, `k8s_pod_glob`, `k8s_namespace_glob` and
      // `k8s_container_glob` all default to `*`, so this will match
      // everything not matched in the previous stanzas
      "attributes" : { "parser": "defaultParser" }
    },

  ]
}
```

Note: You can match against a deployment, by setting `k8s_pod_glob` to
`"deployment*"` because pods names for a given deployment will always
start with the deployment name.

With the exception of `path`, all log configuration settings supported
by `logs` are supported by `k8s_logs`, including parsers, attributes,
redaction_rules and so on.

Configuration options set by any k8s annotations will override the
values specified in the `k8s_logs`.


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2983,16 +2985,114 @@ def __get_last_request_for_log(self, path):
...
-            "parse_format": parse_format,
-            "attributes": attributes,
+            "parse_format": self.__parse_format,
+            "attributes": container_attributes,
+            "rename_logfile": rename_logfile,
         }
 
+        # This is for a hack to prevent the original log file name from being added to the attributes.
+        if self.__use_v2_attributes and not self.__use_v1_and_v2_attributes:
+            result["rename_no_original"] = True
```
czerwingithub, line 3013: I think we want to apply this hack implicitly to any log configuration we return.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2983,16 +2985,114 @@ def __get_last_request_for_log(self, path):
...
+
+        # If we don't have any k8s information then we don't match against k8s_log_configs
+        if k8s_info is None:
+            return result
+
+        # Now apply log configs
+        for i, config in enumerate(self.__k8s_log_configs):
```
czerwingithub, line 3020: Ok, I would suggest breaking essentially this method out as a small abstraction for testing purposes.  I think its initialization should take container_attributes, parser, rename_logfile, and rename_no_original (either true or false).

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2983,16 +2985,114 @@ def __get_last_request_for_log(self, path):
...
+            # We check for glob matches against `k8s_pod_glob`, `k8s_namespace_glob` and `k8s_container_glob`
+
+            # Check for the pod glob
+            pod_glob = config.get("k8s_pod_glob", None)
+            pod_name = k8s_info.get("pod_name", None)
+            if pod_glob is not None and pod_name is not None:
```
czerwingithub, line 3026: When you make the new abstraction, let's see if you can remove this duplicate code here (this check is pretty much repeated for each attribute).

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2983,16 +2985,114 @@ def __get_last_request_for_log(self, path):
...
+                    namespace_glob,
+                    k8s_container,
+                    container_glob,
+                ),
+            )
+            # We have the first matching config.  Apply the log config and break
+            # Note, we can't just .update() because otherwise the attributes dict
+            # may get overridden, plus we also need to exclude `path`
+            for key, value in six.iteritems(config):
+                # Ignore `path` so people can't override it
```
czerwingithub, line 3079: Hmm.. so you are essentially performing a merge with the default attributes on every match you get.  When you move this to an abstraction, I would advocate you merge in the default attributes during the instance creation so that you don't have to do this over and over again.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: This is now done when creating the config builder abstraction.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Unless we are instantiating this abstraction for every single container, then the only ones of these it can take are `rename_no_original`.

Everything other field will contain values specific to the container.

So, the abstraction I made takes the k8s_logs from the config, the logger, the log file parse format (json or cri) and the rename_no_original value (all of which are consistent across all containers).

There is then a `get_log_config` method that takes the remaining config specific items and returns the appropriate config.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: Done.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
imron: After further thought, we need to be doing this, because it's not the default attributes we are merging, but the container specific attributes (pod_name, pod_namespace etc).  We need to do that for every container anyway so I figure it's better to do it once rather than holding state.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2400,6 +2401,138 @@ def query_stats(self):
...
+            return result
+
+        # Now apply log configs
+        for i, config in enumerate(self.__k8s_log_configs):
+            # We check for glob matches against `k8s_pod_glob`, `k8s_namespace_glob` and `k8s_container_glob`
+
+            # Check for the pod glob
+            pod_glob = config.get("k8s_pod_glob", None)
+            pod_name = k8s_info.get("pod_name", None)
+            if not self._check_match(i, "pod_name", pod_name, pod_glob):
```
czerwingithub, line 2484: How about modifying `_check_match` to take in both the `k8s_info` and `config` objects, and then just pass in `k8s_pod_glob`.  That way you could do the gets for the `pod_glob` and `pod_name` in that method.  It would mean just passing in one more argument that you are now, but then you can eliminate 6 duplicate lines in this code block.  You can then also then write this block to be a single `if` statement:

```
  if self._check_match(i, "pod_name", ... ) or
    self._check_match(i, "k8s_container_name", ...) or
    self._check_match(i, "pod_namespace", ...):
    # match found.
```

Avoiding the sprinkle of `continue` statements throughout a body like this usually is better for readability.

imron: The main reason is because I also wanted those values for use in the following debug statement saying that we passed.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2400,6 +2401,138 @@ def query_stats(self):
...
+            return None
+
+        # Build up the default config that we will use
+        result = {
+            "parser": parser,
+            "path": path,
+            "parse_format": self.__parse_format,
+            "attributes": container_attributes,
+            "rename_logfile": rename_logfile,
+            "rename_no_original": self.__rename_no_original,
```
czerwingithub, line 2470: If I understand this correctly, `parser` could either be `docker` or a value from the pod label.  rename_logfile will always be a fixed format (a combination of CRI and the container name) just varies based on the actual container specifics.  And, attributes is always just the K8s attributes that we add in.

Is that correct?

I wonder if we should respect the parser from the k8s pods if there is a configuration that matches this pod.  I worry that it might be confusing to the customer if we try to merge in information from the various configuration mechanisms we have.  Part of me wants to say "pick one and stick with it" because otherwise it is confusing.

What do you think?

scalyr_agent/monitor_utils/k8s.py
czerwingithub: Ah, right... I missed that part.  Ok, never mind about my suggestion.

scalyr_agent/monitor_utils/k8s.py
czerwingithub: Part of the reason I'm pushing on these questions is that I don't think it's a good user experience if overriding some options for the log file forces you change something you may not want to change.

Take the `rename_logfile` config option.  By default, k8s logs are renamed to "/${CONTAINER_RT}/${CONTAINER_NAME}".  However, if you override the log config for particular log to set a parser, there's no way for you to also get the same "/${CONTAINER_RT}/${CONTAINER_NAME}" name.  That's because you have to specify some value for the `rename_logfile` config option.. but you can't specify one that will provide the exact same behavior.  Now, we can take the approach you took and just have that one be the default if you don't specify `rename_logfile`.  But, that seems like almost a hidden behavior.

Instead, we should just officially same the default for the K8s log file rename is "/${CONTAINER_RT}/${CONTAINER_NAME}".  We actually already support using template strings for filename substitution.  We do have container name... we just don't have container runtime.  So, if we add that, then we could make the default be a template string that then doesn't change on a per-container basis.  yes, the value of the string is different after the substitution, but the default value is the same.

What to do for the parser is a little tougher.  Part of me thinks we should just have a different user-visible way for the customer to specify the parser, since it has a different meaning anyway.  Call it `default_parser` instead of `parser`.  We document `default_parser` as the parser to use *if* there is no parser specified for the container via an annotation.  That's different than `parser` which means "always use this value as the parser".

What do you think about those ideas?  Use `default_parser` and have a templated-value for the default rename_logfile.  We just would have to add in the runtime name as a templated variable.

imron: I think that makes sense.  I'll write some longer comments once I have made some changes.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2400,6 +2401,137 @@ def query_stats(self):
...
+                    % (element, name, value, glob,),
+                )
+        return result
+
+    def get_log_config(self, info, k8s_info, parser):
+        """
+        Creates a log_config from various attributes and then applies any `k8s_logs` configs that
+        might apply to this log
+        @param info: A dict containing docker information about the container we are creating a config for
+        @param k8s_info: A dict containing k8s information about hte container we are creating a config for
```
czerwingithub, line 2460: s/hte/the/

czerwingithub:
Ok, everything looks good.  However, we should check out why the smoke-k8s test is failing.  It looks like the unittest-35-windows might be an executor problem -- it says all of the tests have passed but it fails to shutdown the tests properly.

PR #558, opened 2020-06-03
Fix codecov coverage data upload

Kami:
This pull request fixes codecov coverage data uploads.

## Background, Context

Codecov coverage upload started to fail with "400 bad request" error a couple of days ago (https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/2930/workflows/05136b23-cf27-41ed-8b69-74bce5c4cdec/jobs/40054/parallel-runs/0/steps/0-108).

It looks like it's an issue with the version of the client we are currently using and a new one should fix it - https://github.com/codecov/codecov-python/pull/268.

One of the reasons why we didn't spot this issue earlier was that we didn't consider "codecov data upload fatal" error as fatal.

I fixed that by adding ``--required`` flag to ``codecov`` binary which will make it return with non-zero if the upload fails. This way we will spot issues like that sooner.

### DIFF
--- a/scripts/submit-codecov-data.sh
+++ b/scripts/submit-codecov-data.sh
@@ -22,7 +22,8 @@ RETRY_DELAY=${RETRY_DELAY:-5}
 
 # Work around for temporary codecov API timing out
 for (( i=0; i<$MAX_ATTEMPTS; ++i)); do
-    # NOTE: We pass --required flag to the binary since we also want it to rAeturn non-zero if upload fails
+    # NOTE: We pass --required flag to the binary since we also want it to return non-zero (and fail
+    # the build) if upload fails
     codecov --root=/home/circleci/scalyr-agent-2/ --disable gcov --file coverage.xml --required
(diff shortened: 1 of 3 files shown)


scripts/submit-codecov-data.sh
```diff
@@ -22,7 +22,8 @@ RETRY_DELAY=${RETRY_DELAY:-5}
 
 # Work around for temporary codecov API timing out
 for (( i=0; i<$MAX_ATTEMPTS; ++i)); do
-    codecov --root=/home/circleci/scalyr-agent-2/ --disable gcov --file coverage.xml
+    # NOTE: We pass --required flag to the binary since we also want it to rAeturn non-zero if upload fails
```
ArthurKamalov, line 25: typo

Kami: Yep, thanks, fixed in 78401f5a22509541f639cd488a6071ccb0c54558.

PR #561, opened 2020-06-03
AGENT-397: Improved logging for investigating skipped bytes

yanscalyr:
Still needs tests, and there is some more manual testing I need to do.

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -884,12 +884,13 @@ class ScalyrAgent(object):
         # The stats we track for the lifetime of the agent.  This variable tracks the accumulated stats since the
         # last stat reset (the stats get reset every time we read a new configuration).
         base_overall_stats = OverallStats()
+
         # We only emit the overall stats once ever ten minutes.  Track when we last reported it.
-        last_overall_stats_report_time = time.time()
+        last_overall_stats_report_time = self.__start_time
         # We only emit the bandwidth stats once every minute.  Track when we last reported it.
-        last_bw_stats_report_time = last_overall_stats_report_time
+        last_bw_stats_report_time = self.__start_time
         # We only emit the copying_manager stats once every 5 minutes.  Track when we last reported it.
-        last_copy_manager_stats_report_time = last_overall_stats_report_time
+        last_copy_manager_stats_report_time = self.__start_time
 
         # The thread that runs the monitors and and the log copier.
         worker_thread = None
@@ -1598,7 +1599,7 @@ class ScalyrAgent(object):
         """Logs the copy_manager_status message that we periodically write to the agent log to give copying manager
         stats.
 
-        This includes such metrics as the amount of times through the main look and time spent in various sections.
+        This includes such metrics as the amount of times through the main loop and time spent in various sections.
 
         @param overall_stats: The overall stats for the agent.
         @type overall_stats: OverallStats
@@ -1756,17 +1757,19 @@ class ScalyrAgent(object):
             result.avg_bytes_copied_rate = (
                 result.total_bytes_copied - self.__last_total_bytes_copied
             ) / self.__config.copying_manager_stats_log_interval
+
+            total_bytes_produced = (
+                result.total_bytes_skipped
+                + result.total_bytes_copied
+                + result.total_bytes_pending
+            )
+            last_total_bytes_produced = (
+                self.__last_total_bytes_skipped
+                + self.__last_total_bytes_copied
+                + self.__last_total_bytes_pending
+            )
             result.avg_bytes_produced_rate = (
-                (
-                    result.total_bytes_skipped
-                    + result.total_bytes_copied
-                    + result.total_bytes_pending
-                )
-                - (
-                    self.__last_total_bytes_skipped
-                    + self.__last_total_bytes_copied
-                    + self.__last_total_bytes_pending
-                )
+                total_bytes_produced - last_total_bytes_produced
             ) / self.__config.copying_manager_stats_log_interval
             if result.total_bytes_skipped > self.__last_total_bytes_skipped:
                 if self.__config.parsed_max_send_rate_enforcement:
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -186,6 +186,10 @@ class ScalyrAgent(object):
         self.__last_verify_config = None
 
         self.__no_fork = False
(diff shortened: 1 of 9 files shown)


scalyr_agent/agent_main.py
```diff
@@ -1658,6 +1749,59 @@ def __calculate_overall_stats(self, base_overall_stats):
...
+                    self.__last_total_bytes_skipped
+                    + self.__last_total_bytes_copied
+                    + self.__last_total_bytes_pending
+                )
+            ) / self.__config.copying_manager_stats_log_interval
+            if result.total_bytes_skipped > self.__last_total_bytes_skipped:
+                rate_limited_time = (
+                    result.total_rate_limited_time - self.__last_total_rate_limited_time
+                ) / self.__config.copying_manager_stats_log_interval
+                if self.__config.parsed_max_send_rate_enforcement:
```
Kami, line 1772: Let's add tests for those two warnings.

Will likely need to mock a couple of things, but it should be mostly straight forward if we test this method directly.

scalyr_agent/agent_main.py
```diff
@@ -884,6 +888,8 @@ def __run(self, controller):
         last_overall_stats_report_time = time.time()
         # We only emit the bandwidth stats once every minute.  Track when we last reported it.
         last_bw_stats_report_time = last_overall_stats_report_time
+        # We only emit the copying_manager stats once every 5 minutes.  Track when we last reported it.
+        last_copy_manager_stats_report_time = last_overall_stats_report_time
```
czerwingithub, line 892: To improve readability here, let's declare a variable called `current_time` and set `time.time()` and use it in all of the above statements.  It looks a little odd when reading this line without that context.

scalyr_agent/agent_main.py
```diff
@@ -1563,7 +1594,36 @@ def __log_bandwidth_stats(self, overall_stats):
             )
         )
 
-    def __calculate_overall_stats(self, base_overall_stats):
+    def __log_copy_manager_stats(self, overall_stats):
+        """Logs the copy_manager_status message that we periodically write to the agent log to give copying manager
+        stats.
+
+        This includes such metrics as the amount of times through the main look and time spent in various sections.
```
czerwingithub, line 1601: /look/loop/

scalyr_agent/agent_main.py
```diff
@@ -1658,6 +1749,59 @@ def __calculate_overall_stats(self, base_overall_stats):
             result.rss_size,
         ) = self.__controller.get_usage_info()
 
+        if copy_manager_warnings:
+            result.avg_bytes_copied_rate = (
+                result.total_bytes_copied - self.__last_total_bytes_copied
+            ) / self.__config.copying_manager_stats_log_interval
+            result.avg_bytes_produced_rate = (
```
czerwingithub, line 1756: You could define two variables here called `total_bytes_produced` and `last_total_bytes_produced`.  It'd probably make the code a little more readable.  Alternatively, add a comment here that says you compute total produced by adding bytes_skipped, bytes_copied, and bytes_pending.

scalyr_agent/configuration.py
```diff
@@ -2330,6 +2338,13 @@ def __verify_main_config_and_apply_defaults(
         self.__verify_or_set_optional_bool(
             config, "disable_leak_bandwidth_stats", False, description, apply_defaults
         )
+        self.__verify_or_set_optional_bool(
+            config,
+            "disable_leak_copy_manager_stats",
```
czerwingithub, line 2343: I don't think you mean `leak` here.

yanscalyr: I went with how the other 2 status message disable flags were defined, I have no idea why they have "leak" there actually. I'm fine removing it.

czerwingithub: Those were status messages put in to help look specifically for memory leak issues.

scalyr_agent/agent_main.py
yanscalyr: We have a `self.__start_time` set a bit above this, I think that would make sense to use here. Naming wise it seems pretty clear.

scalyr_agent/copying_manager.py
```diff
@@ -800,7 +812,16 @@ def run(self):
                                 pipeline_time = 0.0
 
                             # Now block for the response.
+                            blocking_response_time_start = time.time()
                             (result, bytes_sent, full_response) = get_response()
+                            blocking_response_time_end = time.time()
```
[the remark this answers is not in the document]
yanscalyr: The `total_request_time` you mean? The notion doc does mention that this one will overlap with other timers. Should we exclude whatever time is spent in the send methods for this, or just avoid the compression specifically?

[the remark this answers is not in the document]
czerwingithub: Sorry, I meant removing it from `total_blocking_response_time`.  I'd like to try to have the `blocking_response_time` just be the time we are blocking on the server response.  How hard would that be to do?

yanscalyr:
I don't believe `get_response` has any compression happening in it. Is there something specific you say that makes you think it has the compression time included?

Kami:
@yanscalyr I believe it eventually calls this code - https://github.com/scalyr/scalyr-agent-2/blob/f96346854742f7e565d098549d5b30ba8cdd6810/scalyr_agent/scalyr_client.py#L653 which also prepares request, compresses it and sends it.

yanscalyr:
That part of the code is reached after calling `send_events` here: https://github.com/scalyr/scalyr-agent-2/blob/f96346854742f7e565d098549d5b30ba8cdd6810/scalyr_agent/copying_manager.py#L785
The method we call in the blocking_time section is the callback that is returned by that call that did the compression, it only reads the response.

czerwingithub:
> That part of the code is reached after calling `send_events` here:
> 
> https://github.com/scalyr/scalyr-agent-2/blob/f96346854742f7e565d098549d5b30ba8cdd6810/scalyr_agent/copying_manager.py#L785
> 
> 
> The method we call in the blocking_time section is the callback that is returned by that call that did the compression, it only reads the response.

Ok, thanks for verifying that.  Sgtm.

scalyr_agent/agent_main.py
Kami: And eventually (post release is fine) we should also have some end tests which would cover that in combination with the new ``max_send_rate_enforcement`` configuration option.

We now have end to end tests in place so it shouldn't be too painful to add a test case for that scenario.

PR #568, opened 2020-06-04
Add back reverted tempfile.mkstemp changes, fix Windows compatibility

Kami:
This pull request adds back ``tempfile.mkstemp`` and other changes which were reverted in #566.

It also fixes the code which utilizes ``tempfile.mkstemp`` so it works correctly on Windows in places where we delete temporary created file at the end.

## Background, Context

Those changes broke the build, because the default``os.remove`` / ``os.unlink`` behavior is not the same on Linux and Windows.

On Linux, if you will try to delete a file which is still open it will work, but on Windows it will fail.

By default, ``tempfile.mkstemp()`` returns open file handle (in addition to the temporary file path).

If we don't explicitly close that file handle, deletion of that file later on will fail on Windows.

## Proposed Fix / Change

In this PR, I use one of two approaches (depending on the context where the code it's used - if it's security sensitive where open a race could be an issue or if it's not).

In places where we can and where we want to avoid potential race with re-opening the file, we directly use open FD which is returned by tempfile.mkstemp.

In other places (mostly tests) where re-opening a file is not an issue and that file path is later on re-opened by some other code, we simply close the FD returned by tempfile.mkstemp.

### DIFF
--- a/.circleci/modernize/modernize.py
+++ b/.circleci/modernize/modernize.py
@@ -23,7 +23,7 @@
 #   -j --processes <jobs count> run concurrently.
 # Note: This is only python3 script.
 
-if False:
+if False:  # NOSONAR
     # NOTE: This is a workaround for old Python versions where typing module is not available
     # We should eventually improve that once we start producing distributions with Python
     # interpreter and dependencies bundled in.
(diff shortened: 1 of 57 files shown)


Kami:
I'm not too happy that this issue was caught by our build system and not tests so I'm also working on trying to get at least some unit tests to run on Windows (this will require more work though since a lot of unit tests rely on Linux specific behavior).

Kami:
I've confirmed those changes work by merge them into #569, but we should still keep an eye on them once they are merged (and perhaps trigger an internal build to ensure it indeed works).

PR #569, opened 2020-06-04
Also run unit tests on Windows

Kami:
This pull request updates our Circle CI config so we also run our unitests on Windows Server 2019 under Python 3.5.

In addition it also updates a bunch of the existing tests so we either skip them on Windows (if they are Linux specific) or so they work correctly on Windows.

## Background, Context

#563 introduced an issue with Windows build which was only caught by the build system which is unacceptable.

#538 will make some of that better since it introduces basic end to end sanity tests, but we also need a way to run unit tests on Windows and catch various Windows related issues early (end to end tests and unit tests cover different code paths - unit tests tend to cover many more edge cases whereas end to end tests just mostly ensure agent starts up and is consuming and sending logs correctly).

Keep in mind that due to Circle CI Windows limitations we only run tests under single Python version (3.5) on Windows Server 2019, but that's better than nothing.

## Notes

Running tests on Windows uncover a couple of small real life issues, but nothing major.

For example - AutoFlushingRotatingFile class didn't take into account line separator differences between Linux and Windows so the actual file would end up being larger on Windows (1 additional bye per line in file - that's because of carriage return aka \r character).

# TODO

- [x] Update affected tests and code and make sure they also pass on Windows (this will be quite involved, a bunch of tests rely on Linux like behavior and paths)


scalyr_agent/test_base.py
```diff
@@ -360,7 +360,11 @@ def tearDown(self):
             )
 
         if not self.__assertion_failed:
-            shutil.rmtree(self.logs_directory)
+            try:
+                shutil.rmtree(self.logs_directory)
+            except PermissionError:
+                # Workaround for Windows when sometimes deleting fails because file is still in use
```
ArthurKamalov, line 366: Could it be better to keep it failing and to try to close all opened descriptors?

Kami: Yeah, that's why it's failing.

Problem is that it's not easy to keep track of all the open FDs to random files by various tests and ensure they are closed.

I will look again if I can identify tests which leave open FDs, but I may not be able to track down the root cause.

On Linux we could also use something like lsof or similar to find all the open FDs for this process and trying to close it, but that won't work on Windows...

Kami: Looks like this did the trick - c3bc98a26f5e6eb54a710a5ecada8b20647b6ab0.

Main offender was ``scalyr_logging.set_log_destination`` which opens handles to the log files, but we (tests) never closed them.

I will also quickly audit other base test classes for this issue and fix them, where / if necessary.

Kami:
@ArthurKamalov Thanks for the review.

PR #571, opened 2020-06-05
SonarCloud integration though our CI system (Circle CI)

Kami:
This pull request updates Circle CI config so we integrate with SonarCloud through our CI/CD System instead of using Automatic Analysis.

## Background, Context

Currently we integrate with SonarClouad using Automatic Analysis (https://sonarcloud.io/documentation/analysis/automatic-analysis/).

It works by automatically running a scan on SonarCloud servers once a new commit is added to a branch (sonar cloud is notified about changes via Github webhook).

This works fine for simple projects and common scenarios, but it breaks with larger and more complex projects where we want to control exactly when analysis runs and also integrate with other tools such as pylint, bandit and code coverage.

To be able to integrate via CI, it means we need to trigger sonar cloud scan as part of our CI build.

## Proposed Implementation

I added a new ``sonnarcloud-scan`` job which runs after all the long running smoke tests jobs finish.

This is not 100% ideal, but Circle CI doesn't allow you to specify cross workflow dependencies which means we can't also wait on ``unittest-27`` job to finish, but this shouldn't be a big issue since smoke tests job we wait on always take much longer to complete than the unittest-27 one.

In addition to that, to get coverage reporting to work correctly, I updated each job which generates coverage data to also upload coverage data to S3 bucket and then download all those reports later and combine them before passing combined file to sonar scan binary.

In ideal world, we would be able to utilize Circle CI artifacts directly for that, but Circle CI API is quite limited and there is no easy and simple way for us to retrieve build artifacts for all the jobs which are part of a specific build (in theory, it's possible, but it requires polling, etc. and it's much more error prone and messy than S3 approach).

Actual coverage data is small and I set up auto purge after 14 days for that S3 bucket so costs should be minimum.


Kami:
Here we go - we now also submit code coverage data to sonarcloud.io.

Code coverage number on codecov.io dropped a bit, because now we also correctly report branch code coverage data for package and monitor tests as well.

Previously we only reported simple line code coverage, but to be able to combine coverage for all the tests and submit it to sonarcloud.io, we need all the coverage data files to contain branch coverage data or none.

Kami:
@ArthurKamalov Yeah, retrieving artifacts from a corresponding Circle CI jobs would be possible, but quite more complex due to how their API is structured.

PR #572, opened 2020-06-11
AGENT-413 K8S Scalyr agent uploads duplicate logs on restart/config change

oliverhsu77:
There is a race condition in the `copying_manager` that causes duplicate logs to be uploaded with the Kubernetes monitor.

The `copying_manager` main loop performs the following:

1. Check for new logs (this doesn't apply to Kubernetes, since the new logs are provided dynamically through the Kubernetes monitor)
2. Read new log lines
3. Send logs to Scalyr
4. Scan for new dynamically added log files (from the Kubernetes monitor) / close log processors that no longer match a matcher
5. Remove log files scheduled for deletion
6. Remove finished log matchers
7. Write out checkpoint file

On a separate thread, the `kubernetes_monitor.stop` calls
`copying_manager.remove_log_path`,  which removes the log matcher for the Kubernetes log file.

The `Kubernetes_monitor` and `copying_manager` are stopped and restarted on 1) Scalyr agent config file change and 2) Agent restart.

**Scenario 1**
`kubernetes_monitor.stop` calls `copying_manager.remove_log_path` in the loop before step 4.  In this scenario, the K8S log processors will be removed and not included in the checkpoint file since the K8S log matchers are removed.

**Scenario 2**
`kubernetes_monitor.stop` calls `copying_manager.remove_log_path` in the loop after step 4.  In this scenario, the K8S log processors are still active and will will be included in the checkpoint file.

**Fix**
Since both the `kubernetes_montor` and `copying_manager` will be shutdown on config file update and agent restart, there is not a need for `kubernetes_monitor.stop` to call `copying_manager.remove_log_path`.  Shutting down `copying_manager` will cause all the log processors to be shutdown anyways.  The fix is to not call `copying_manager.remove_log_path` on `kubernetes_monitor.stop`.

I confirmed that this fixed the issue through manual testing.  Ideally, we should also have a unit test for this.  Since this issue is an interaction between modules, it is not straight forward to unit test.

Primary reviewer: @imron


Kami:
Good catch.

And like you said, tests would be good, but yeah, probably not easy to introduce this race scenario inside a test case.

Perhaps we could mock that + some other methods and verify ``remove_log_path`` is only called ``n`` times (where n is number of watched logs) instead of ``n * 2`` times.

PR #573, opened 2020-06-16
Also run install only tests inside the AMI tests job

Kami:
This pull request updates AMI-tests script and Circle CI job so we also run install only tests which utilize latest stable version of the installer script to verify that installation of the latest stable agent package using the installer script works correctly.

---

We already have upgrade tests which also tests the install scenario / path, but it also doesn't hurt to have additional separate tests for the install only path (at least right now when we run those jobs in parallel and costs are very low).

### DIFF
--- a/scripts/circleci/run-ami-tests-in-bg.sh
+++ b/scripts/circleci/run-ami-tests-in-bg.sh
@@ -23,7 +23,7 @@ mkdir -p outputs
 
 # Run sanity test for each image concurrently in background
 # Tests below utilize installer script to test installing latest stable version of the package
-python tests/ami/packages_sanity_tests.py --distro=ubuntu1804 --type=install--to-version=current --installer-script-url="https://www.scalyr.com/scalyr-repo/stable/latest/install-scalyr-agent-2.sh" &> outputs/ubuntu1804-install.log &
+python tests/ami/packages_sanity_tests.py --distro=ubuntu1804 --type=install --to-version=current --installer-script-url="https://www.scalyr.com/scalyr-repo/stable/latest/install-scalyr-agent-2.sh" &> outputs/ubuntu1804-install.log &
 python tests/ami/packages_sanity_tests.py --distro=ubuntu1604 --type=install --to-version=current --installer-script-url="https://www.scalyr.com/scalyr-repo/stable/latest/install-scalyr-agent-2.sh" &> outputs/ubuntu1604-install.log &
 python tests/ami/packages_sanity_tests.py --distro=ubuntu1404 --type=install --to-version=current --installer-script-url="https://www.scalyr.com/scalyr-repo/stable/latest/install-scalyr-agent-2.sh" &> outputs/ubuntu1404-install.log &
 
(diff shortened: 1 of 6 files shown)


scripts/circleci/run-ami-tests-in-bg.sh
```diff
@@ -21,14 +21,23 @@ echo "Runing AMI sanity tests concurrently in the background (this may take up t
...
-python tests/ami/packages_sanity_tests.py --distro=ubuntu1804 --type=upgrade --from-version=current --to-version=/tmp/workspace/scalyr-agent-2.deb &> outputs/ubuntu1804.log &
-python tests/ami/packages_sanity_tests.py --distro=ubuntu1604 --type=upgrade --from-version=current --to-version=/tmp/workspace/scalyr-agent-2.deb &> outputs/ubuntu1604.log &
-python tests/ami/packages_sanity_tests.py --distro=ubuntu1404 --type=upgrade --from-version=current --to-version=/tmp/workspace/scalyr-agent-2.deb &> outputs/ubuntu1404.log &
-python tests/ami/packages_sanity_tests.py --distro=centos7 --type=upgrade --from-version=current --to-version=/tmp/workspace/scalyr-agent-2.rpm &> outputs/centos7.log &
+# Run sanity test for each image concurrently in background
+# Tests below utilize installer script to test installing latest stable version of the package
+python tests/ami/packages_sanity_tests.py --distro=ubuntu1804 --type=install --to-version=current --installer-script-url="https://www.scalyr.com/scalyr-repo/stable/latest/install-scalyr-agent-2.sh" &> outputs/ubuntu1804-install.log &
```
ArthurKamalov, line 26: @Kami Do we really need to check the stable release from the repo here? I thought the idea is that the stable release gets its own portion of tests when we merge changes to master and release branches?

Kami: Yeah, one option is to split it out in a separate job, since ideally we would also run those stable install tests periodically (e.g. every 24 hours).

Kami: And to expand on that - doing that (adding a new job for running install tests using installer script and stable packages) would give us flexibility to run that job periodically as well (e.g. every 24 hours which is a good idea), but it does increase complexity quite a bit (new Circle CI job, additional script to run those tests).

Kami:
@ArthurKamalov As discussed on Slack, I merged your changes from #574 into this branch.

I also updated Circle CI config so we run ami-tests-stable job on daily basis - this may help us catch any installer script or upstream related issues.

scripts/circleci/run-ami-tests-in-bg.sh
Kami: This has now been resolved - thanks for your assistance with #574.

Kami:
I also enabled install and upgrade tests on CentOS 8.

Latest package doesn't declare dependency on ``python`` package anymore so we can now also run upgrade tests on CentOS 8.

PR #574, opened 2020-06-16
Arthur/ami tests installer script

ArthurKamalov:
sanity test bash script now gets a positional argument `stable` or `development` and starts needed sanity tests according to this.


Kami:
Mostly LGTM, thanks.

It does look like there was a bit of confusion though since we are both working on a similar set of changes.

Are you OK with me merging those changes into #573 and finishing them there?

PR #575, opened 2020-06-16
Also run AMI end to end package tests in Debian 10 & Amazon Linux 2

Kami:
This pull request updates install and upgrade AMI package end to end tests so they also run under Debian 10.

As discussed before, we run tests for various distros in parallel so adding another distro doesn't increase overall build time as long as those distro tests don't take longer than the slowest distro tests (which is Windows at this point - it can take up to 4 minutes whereas most other distros finish in 2 minutes or less).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -634,7 +634,16 @@ commands:
       test_type:
         description: "Type of tests to run."
         type: enum
-        enum: ["development", "stable"]
+        enum:
+          - "development"
+          - "stable"
+        # URL to the installer script to use with the tests. We default to stable version, but
+        # you can change that value if you made changes to the installer script and you want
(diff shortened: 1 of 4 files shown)


Kami:
I also pushed a change which allows user to specify ``installer_script_url`` parameter for the ``ami-test`` Circle CI pipeline - 192b07383a9532f9f30ab8b8466fdd15efecebda.

This change makes it easy to test any installer script changes with all the distros we support in our AMI tests script.

---

Right now it relies on installer script from stable / beta / internal repo, but in the future, we should also work on producing installer script with bundled package artifact as part of every Circle CI build.

This will remove dependency on running "push installer scripts" Jenkins job and make changes to the installer script much easier, faster and less risky.

PR #579, opened 2020-06-18
Fix Pyinstaller specification.

ArthurKamalov:
This pull request fixes dynamic monitor loading error. This error appears only on particular machines, and it very hard to track it, because the agent works well on other machines.


scalyr_agent/__scalyr__.py
```diff
@@ -184,18 +184,27 @@ def __add_scalyr_package_to_path():
...
-
-    if os.path.isfile(in_package):
-        version_path = in_package
-    elif os.path.isfile(in_install):
-        version_path = in_install
-    else:
+
+    file_names = ["VERSION"]
+
+    if __is_frozen__:
```
Kami, line 190: It wouldn't hurt to add unit test for this, but I believe we should also be covered by new AMI end to end tests.

ArthurKamalov: @Kami That's a good idea. I think it would be better to start a new PR for that, to stop blocking you for now.

PR #581, opened 2020-06-23
Log redis cluster replication offsets

huchengming:
This commit enables current redis monitor to log replication information of redis cluster.

Specifically, if the agent is running on a redis cluster master, redis
monitor will log replication offsets of both master and replica, as well
as offset differences and how many seconds the replica is falling behind

### DIFF
--- a/scalyr_agent/builtin_monitors/redis_monitor.py
+++ b/scalyr_agent/builtin_monitors/redis_monitor.py
@@ -25,9 +25,11 @@ import binascii
 import re
 import time
 
+from six.moves import range
+
 from scalyr_agent import ScalyrMonitor
 
-from redis.client import Redis  # pylint: disable=import-error
+from redis.client import Redis, parse_info  # pylint: disable=import-error
 from redis.exceptions import (  # pylint: disable=import-error
     ConnectionError,
     TimeoutError,
@@ -56,6 +58,12 @@ class RedisHost(object):
 
         # Timestamp of the last slowlog entry
         self.last_timestamp = 0
+
+        # master offset of the last replication entry
+        self.last_master_offset = 0
+
+        # Timestamp of the last replication entry
+        self.last_replication_sample_timestamp = 0
 
         # Redis object
         self.__redis = None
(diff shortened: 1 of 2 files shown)


scalyr_agent/builtin_monitors/redis_monitor.py
```diff
@@ -187,6 +195,65 @@ def log_slowlog_entries(self, logger, lines_to_fetch):
...
+            return
+
+        master_replid = replication_info["master_replid"]
+        connected_replicas = replication_info["connected_slaves"]
+
+        # If there are more than one replicas, log the most up-to-date
+        max_replica_offset = 0
+        for n in range(connected_replicas):
+            max_replica_offset = max(
+                max_replica_offset, replication_info["slave{}".format(n)]["offset"]
```
imron, line 218: For consistency reasons with the rest of the agent, can you use `%` formatting instead of `.format`?

scalyr_agent/builtin_monitors/redis_monitor.py
```diff
@@ -402,11 +469,11 @@ def _initialize(self):
             self.__redis_hosts.append(redis_host)
 
     def gather_sample(self):
-
         for host in self.__redis_hosts:
             new_connection = not host.valid()
             try:
                 host.log_slowlog_entries(self._logger, self.__lines_to_fetch)
+                host.log_cluster_replication_info(self._logger)
```
imron, line 476: Could you put this behind a configuration option defaulting to `False`?  This way existing customers already using the Redis monitor won't start getting extra logs.

See [here](https://github.com/scalyr/scalyr-agent-2/blob/master/scalyr_agent/builtin_monitors/shell_monitor.py#L51) for an example of how to define configuration options.  You can then enable it in the config option in the monitor snippet of `agent.json` e.g. 

```
{
  monitors:[
    {
      "module": "scalyr_agent.builtin_monitors.redis_monitor",
      "log_cluster_replication_info": true

    }
  ]
}
```

huchengming:
@imron ready for another look

imron:
Oh, also, can you rebase off master before you merge.

PR #586, opened 2020-06-25
Correctly exclude some tests directories from coverage data on Windows.

Kami:
This pull request updates ``.coveragerc`` file used by Circle CI jobs to correctly exclude some test directories from coverage reports.

This is needed because paths on Windows and Linux are different and after re-enabling code coverage for Windows unit tests, some directories were now being incorrectly included in the reports.


Kami:
I also pushed a change to exclude the same set of directories from coverage reports on sonarcloud.

So hopefully now sonarcloud coverage data should be more in line with codecov.

PR #587, opened 2020-06-26
End-to-end test for rate limiting functionality

yanscalyr:


tests/smoke_tests/verifier.py
```diff
@@ -359,6 +359,215 @@ def _verify(self):
...
+            "$logfile=='{0}'".format(
+                self._runner.get_file_path_text(self._data_json_log_path)
+            )
+        )
+        self._request.add_filter("$stream_id=='{0}'".format(self._timestamp))
+
+        self._message = {
+            "count": 0,
+            "stream_id": self._timestamp,
+            "filler": "aaajhghjgfijhgfhhjvcfgujhxfgdtyubn vcgfgyuhbnvcgfytuhvbcftyuhjgftyugftyuyygty7u7y8f8ufgfg8fgf8f"
```
Kami, line 389: To make it slightly easier on the eyes, you could also do ``"aaajhghjgfijhgfhhjvcfgujhxfgdtyubn vcgfgyuhbnvcgfytuhvbcftyuhjgftyugftyuyygty7u7y8f8ufgfg8fgf8f gfyfgufyu" * 60`` :)

tests/smoke_tests/verifier.py
```diff
@@ -359,6 +359,215 @@ def _verify(self):
...
+        return self._verify()
+
+    def _verify_ingest_began(self):
+        """
+        Check that any lines have been uploaded, this helps keep the test consistent by working around ingest time.
+        """
+        try:
+            response = self._request.send()
+        except Exception:
+            print("Query failed.")
```
Kami, line 517: Let's please also log the original exception error message.

tests/smoke_tests/verifier.py
```diff
@@ -359,6 +359,215 @@ def _verify(self):
...
+            print("Response data: %s" % (str(response)))
+            return False
+
+        return len(response["matches"]) > 0
+
+    def _verify(self):
+        try:
+            response = self._request.send()
+        except Exception:
+            print("Query failed.")
```
Kami, line 531: Same here.

tests/smoke_tests/verifier.py
```diff
@@ -359,6 +359,215 @@ def _verify(self):
...
+
+    def prepare(self):
+        print(("Write test data to log file '{0}'".format(self._data_json_log_path)))
+        for i in range(self._lines_count):
+            self._message["count"] = i
+            json_data = json.dumps(self._message)
+            self._runner.write_line(self._data_json_log_path, json_data)
+        return
+
+    def verify(self, timeout=2 * 60):
```
Kami, line 477: Do you happen to know what amount of time it takes to ingest all the data we send?

The reason I'm asking is if it's substantially less than 2 minutes, we could decrease the timeout so we time out earlier on actual issues / failures.

Kami: And perhaps even have two timeouts - I imagine ingestion should being much earlier than in 2 minutes (unless there is an issue on the server side or a bug / similar in the client).

tests/smoke_tests/verifier.py
```diff
@@ -359,6 +359,215 @@ def _verify(self):
...
+            "aaajhghjgfijhgfhhjvcfgujhxfgdtyubn vcgfgyuhbnvcgfytuhvbcftyuhjgftyugftyuyygty7u7y8f8ufgfg8fgf8f gfyfgufyu"
+            "aaajhghjgfijhgfhhjvcfgujhxfgdtyubn vcgfgyuhbnvcgfytuhvbcftyuhjgftyugftyuyygty7u7y8f8ufgfg8fgf8f gfyfgufyu"
+            "aaajhghjgfijhgfhhjvcfgujhxfgdtyubn vcgfgyuhbnvcgfytuhvbcftyuhjgftyugftyuyygty7u7y8f8ufgfg8fgf8f gfyfgufyu",
+        }
+
+        self._lines_count = 5000
+        self._upload_wait_time = 30
+        # Estimate of line size
+        self._line_size = len(json.dumps(self._message))
+        self._rate_limit_bytes_per_second = 500000
```
Kami, line 464: It would probably be good to add a comment that this value needs to match rate value defined in environment variable / similar.

.circleci/config.yml
```diff
@@ -1729,6 +1729,22 @@ jobs:
             export PYTHONPATH=.
             ./scripts/send-benchmark-completion-notification-to-slack.py /tmp/codespeed-screenshots/
 
+  smoke-standalone-35-rate-limit:
+    working_directory: ~/scalyr-agent-2
+    environment:
+      SCALYR_MAX_SEND_RATE_ENFORCEMENT: "500KB/s"
```
Kami, line 1735: Since we define those values on the Circle CI environment variable level, test won't work correctly if you run it locally directly using tox, right?

If so, it may be better to put those values inside the tox target (``setenv``) or inside the code itself - this way tests should be reproducible no matter how you run them (via Circle CLI, using tox or directly using pytest binary).

.circleci/config.yml
```diff
@@ -2076,6 +2092,7 @@ workflows:
       - smoke-standalone-26
       - smoke-standalone-27
       - smoke-standalone-35
+      - smoke-standalone-35-rate-limit
```
Kami, line 2095: IIRC, each new smoke tests job adds quite some overhead due to image caching, etc.

It may make sense to combine it with some other existing standalone target. (e.g. just run two tox targets as part of the existing ``smoke-standalone-35`` job).

@ArthurKamalov what do you think?

Kami:
I pushed a couple of debugging / print changes which should make it slightly easier to troubleshoot various smoke test related failures.
 
 It turns out we get such error with different smoke tests every now and then:
 
 ```bash
 Response data: {'message': 'query timed out or could not be fully evaluated', 'status': 'error/client/badParam'}
 ```
 
 Previously we didn't always log the response data so we couldn't know what was the underlying cause when we've seen errors similar to this one:
 
 ```bash
 >       if len(response["matches"]) < 220:
 
E       KeyError: 'matches'
 ```
 
In addition to that, I pushed a change so we also try to retry multiple times actual verification step which verifies data has been ingested.

Part of the problem is that the debugging changes haven't been reflected because ``.circleci/docker_unified_smoke_unit/smoketest/smoketest.py`` file is bundled in pre-built Docker image which we pull down during those test run.

In the short term, we should figure out how to get that file to that testing container - one option is to re-build the container and the other one is to simply copy it during the tests.

In the long term - we should move those tests to a new end to end smoke testing framework which is easier to maintain since we re-build images as part of every run, etc.

Kami:
@yanscalyr Here is a workaround I used to test smoketest.py changes - 1015100e3c37964455e421baaf0362584cad72af.

It's not great, but we may as well do that since it's probably quite a bigger hassle to push a new docker base image (we may just want to update all the docker test runs in that script to use that script + to use branch from CIRCLE_BRANCH env variable or fall back to ``master`` if that variable is not defined).

The output we now see with those changes shows that ``parser`` attribute is missing:

```bash
Sample response for matches[0]
 
{'severity': 3, 'session': 'log_session_8423015096460696226', 'attributes': {'containerName': 'ci-agent-docker-syslog-46203-uploader', 'monitor': 'agentSyslog', 'containerId': '71f3f2ef5e6a', 'serverHost': '51571d535beb'}, 'thread': 'log_4', 'message': 'count=1000,line_stream=stderr,verifier_type=Docker Syslog\n', 'timestamp': '1593773762663043072'}
 
"monitor" or "parser" attribute value doesn't match
 
Expected(monitor): 'agentSyslog', got 'agentSyslog'
Expected(parser): 'agentSyslogDocker', got 'None'
 ```

I'm not sure why that is happening though. 

Could be related to recent $logfile changes on the API side.

Kami:
I also pushed a change which decreases the max wait / timeout for those tests - fa2fae7ad1e83bc4d41430c951e6a8bde83e5f77.

I checked previous runs and on average those tests finish much sooner than in 150 seconds (actual Circle CI step takes longer because of the container set up time, but actual run time in the verifier code is much shorter than 150 seconds).

And if there is a failure, we will need to wait until this timeout is reached and it makes no sense to wait that long (and waste Circle CI credits and cycles on it).

yanscalyr:
I think the only thing left is the question asked to Arthur about possibly bundling this into another test. Looks like you already lowered the timeout on this test and removed those missing attribute checks.

.circleci/config.yml
ArthurKamalov: > IIRC, each new smoke tests job adds quite some overhead due to image caching, etc.
I think standalone smoke tests do not use image caching, but I also think it is possible to combine them.

PS: Sorry, I had a tons of notifications from other repos and I missed this.

PR #590, opened 2020-07-07
Add a release note about possibly not having expected configuration in custom docker images.

yanscalyr:
The expected configuration for `stdout_severity` in k8s is set in our own Dockerfile for it, which customers may not be using. Adding a note about this to the release it was added in.

Note: this PR is against the `release` branch, let me know if this is wrong.


Kami:
smoke-docker-syslog will fail until my fix from your other PR is merged.

So we can either finish and merge that PR, or if we think that may take a while, I can cherry pick that fix into another PR.

yanscalyr:
Huh, my UI was saying I was making a PR against `release`, thats annoying. Cherrypicking may be more annoying than just creating another PR, if you aren't against that.

yanscalyr:
Closing to open a new one against the release branch.

PR #592, opened 2020-07-08
Add a release note about possibly not having expected configuration in custom docker images. 

yanscalyr:
The expected configuration for stdout_severity in k8s is set in our own Dockerfile for it, which customers may not be using. Adding a note about this to the release it was added in.


yanscalyr:
This brought in a bunch of changes I didn't intend from somewhere. Closing and trying again...

PR #594, opened 2020-07-08
Merge updated release notes in `release` into master.

yanscalyr:


yanscalyr:
Ah, right, I can't make a PR for this because it will require merging master into release... Im closing this quick so no one does it accidentally.

PR #595, opened 2020-07-09
Artifact upload from circleci

ArthurKamalov:
This pull request adds new CircleCi jobs that build Windows install package and agent install script.

There are also PR in the [devops](https://github.com/scalyr/devops/pull/322) repository that allows accepting artifacts from circleci and sign them.
And PR in the [agent-release](https://github.com/scalyr/agent-release/pull/3) repo with changes to combine install script push and regular release pipeline.


scripts/circleci/release/create-agent-installer.sh
```diff
@@ -0,0 +1,349 @@
+#!/bin/bash
+#
```
Kami, line 2: Since it's a public file, please add license header to it.

scripts/circleci/upload_circleci_artifacts.py
```diff
@@ -0,0 +1,404 @@
...
+    artifact_output_path = os.path.join(
+        output_path, os.path.basename(artifact_info["path"]),
+    )
+
+    with requests.Session() as session:
+        resp = session.get(url=artifact_info["url"], allow_redirects=True, stream=True)
+
+    with open(artifact_output_path, "wb") as file:
+        for chunk in resp.iter_content(chunk_size=8192):
+            file.write(chunk)
```
Kami, line 152: I wanted to say that it would be good to verify the checksum of the downloaded file to make sure we are working with a correct and non-corrupted file, but it looks like Circle CI API (https://circleci.com/docs/api/#artifacts) doesn't expose the checksum.

---

EDIT: I had a look and it looks the the server also doesn't return E-Tag or a similar header so there isn't much we can do (option would be to to publish ``<file name>.sha256`` file as part of the build and use that, but this may be an overkill and not needed for now).

Kami:
Please also update ``LINT_FILES_TO_CHECK`` variable defined in ``tox.ini`` and make sure we also run lint checks on all the new Python files (``scripts/circleci/*.py``).

Same story for shell check, need to update ``scripts/shell-scripts-lint.sh`` script (if it gives too many false positives on some of the existing scripts we added to this repos, we can perhaps ignore those specific scripts to begin with).

.circleci/config.yml
```diff
@@ -2010,14 +2066,32 @@ jobs:
...
+
+      - when:
+          condition: << pipeline.parameters.agent_version >>
+          steps:
+            - run:
+                name: Change package version to << pipeline.parameters.agent_version >>.
+                shell: bash
+                command: |
+                  echo << pipeline.parameters.agent_version >> > VERSION
+            - unless:
```
Kami, line 2086: Hm, will the second step ever get executed?

Since in the first condition we check if ``agent_version`` is truthy / defined and only if it is, we try to run the first and the second step.

But the second step only runs if ``agent_version`` is falsy / not defined (``unless``) and this will never be true because of the conditional in the parent ``condition`` clause?

Or am I missing / misunderstanding something?

.circleci/config.yml
```diff
@@ -2010,14 +2066,32 @@ jobs:
...
-          command: |
-            # Update agent version to avoid version conflict with current scalyr repository package.
-            # This way we can also correctly assert that the upgrade has worked - if we didn't do
-            # that, from and to version would be the same
-            echo "$(cat VERSION).dev1-$(git rev-parse --short HEAD)" > VERSION
-            cat VERSION
+#      - run:
+#          name: Bump Agent Version
+#          command: |
+#            # Update agent version to avoid version conflict with current scalyr repository package.
```
Kami, line 2072: Probably a good idea to leave this comment in place so people who look at the pipeline know why that step is needed when ``agent_version`` parameter is not explicitly defined.

.circleci/config.yml
ArthurKamalov: Wow, yes, thank you for noticing that. It was a copy-paste from the other job, and I missed that the editor made a wrong indent. Actually, I did not expect that circleci would accept such config as valid.

ArthurKamalov:
@Kami I forgot to say that there is a job called `build-linux-install-script` and it is not used for now. I  moved some of its logic to Linux packages build job which is called now `build-linux-packages-and-installer-script`.

.circleci/config.yml
Kami: No problem and I believe this would have been caught sooner or later by package tests (we verify that the upgrade works by grepping for a specific git revision in the package version).

Kami:
> @Kami I forgot to say that there is a job called build-linux-install-script and it is not used for now. I moved some of its logic to Linux packages build job which is called now build-linux-packages-and-installer-script.

It's probably not a bad idea to run it every now and then as part of a build cron job - just to make sure it's working and doesn't get out of date (otherwise it's probably better to remove it).

Kami:
Probably also not a bad idea to add README.md to ``scripts/circleci/release/`` directory.

PR #596, opened 2020-07-10
AGENT-386: Health check functionality for Kubernetes.

yanscalyr:
Current design doesn't really match up to the Youtrack ticket since that was mostly about an HTTP health endpoint that we decided against.

Works by defining a new CLI option for use with the `status` command, with a 0 return code for healthy and 1 for unhealthy, and a message about that. K8s can use this to automatically restart the Agent pod when the health check fails.

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -739,10 +739,17 @@ class ScalyrAgent(object):
             )
             return 1
 
+        health_check_error = None
         fp = open(status_file)
         for line in fp:
             print(line.rstrip())
+            if "Health check:" in line and line.rstrip() != "Health check: Good":
+                health_check_error = line.rstrip()
         fp.close()
+
+        if health_check_error:
+            print(health_check_error, file=sys.stderr)
+            return 1
 
         return 0
 
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -90,7 +90,10 @@ from scalyr_agent.agent_status import AgentStatus
 from scalyr_agent.agent_status import ConfigStatus
 from scalyr_agent.agent_status import OverallStats
 from scalyr_agent.agent_status import GCStatus
-from scalyr_agent.agent_status import report_status
+from scalyr_agent.agent_status import (
+    report_status,
+    report_health,
+)
 from scalyr_agent.platform_controller import (
     PlatformController,
     AgentAlreadyRunning,
@@ -223,6 +226,7 @@ class ScalyrAgent(object):
         my_options = Options()
         my_options.quiet = True
         my_options.verbose = False
+        my_options.health_check = False
         my_options.status_format = "text"
         my_options.no_fork = True
         my_options.no_change_user = True
@@ -254,6 +258,7 @@ class ScalyrAgent(object):
         """
         quiet = command_options.quiet
         verbose = command_options.verbose
+        health_check = command_options.health_check
         status_format = command_options.status_format
         extra_config_dir = command_options.extra_config_dir
         self.__no_fork = command_options.no_fork
@@ -321,9 +326,9 @@ class ScalyrAgent(object):
                 return self.__run(self.__controller)
             elif command == "stop":
                 return self.__stop(quiet)
-            elif command == "status" and not verbose:
+            elif command == "status" and not (verbose or health_check):
                 return self.__status()
-            elif command == "status" and verbose:
+            elif command == "status" and (verbose or health_check):
                 if self.__config is not None:
                     agent_data_path = self.__config.agent_data_path
                 else:
@@ -333,7 +338,9 @@ class ScalyrAgent(object):
                         file=sys.stderr,
                     )
                 return self.__detailed_status(
-                    agent_data_path, status_format=status_format
+                    agent_data_path,
+                    status_format=status_format,
+                    health_check=health_check,
                 )
             elif command == "restart":
                 return self.__restart(quiet, no_check_remote)
@@ -618,8 +625,10 @@ class ScalyrAgent(object):
             log.info("Received signal to shutdown, attempt to shutdown cleanly.")
             self.__run_state.stop()
 
-    def __detailed_status(self, data_directory, status_format="text"):
-        """Execute the status -v command.
+    def __detailed_status(
+        self, data_directory, status_format="text", health_check=False
+    ):
+        """Execute the status -v or -H command.
 
         This will request the current agent to dump its detailed status to a file in the data directory, which
         this process will then read.
@@ -690,7 +699,10 @@ class ScalyrAgent(object):
             fp.write(status_format)
 
         # Signal to the running process.  This should cause that process to write to the status file
-        result = self.__controller.request_agent_status()
+        if health_check:
+            result = self.__controller.request_agent_health_check()
+        else:
+            result = self.__controller.request_agent_status()
         if result is not None:
             if result == errno.ESRCH:
                 print(AGENT_NOT_RUNNING_MESSAGE, file=sys.stderr)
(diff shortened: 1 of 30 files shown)


yanscalyr:
Looks like I need to reduce a bit of code duplication which shouldn't be an issue

scalyr_agent/agent_main.py
```diff
@@ -690,7 +699,10 @@ def __detailed_status(self, data_directory, status_format="text"):
             fp.write(status_format)
 
         # Signal to the running process.  This should cause that process to write to the status file
-        result = self.__controller.request_agent_status()
+        if health_check:
+            result = self.__controller.request_agent_health_check()
```
czerwingithub, line 703: I don't think we should have a separate signal for the health check.  Rather, the health check should just use the normal `-v` signal but be selective about what it emits to stdout. 

Most likely, the "right" way to do this is to have the health check request the verbose status in JSON and then it just parses that JSON looking for the health check field.  It returns success or not based on that.

czerwingithub, line 703: This also should allow you to remove a bunch of the changes you made related to registering and handling the new health check handler.

scalyr_agent/platform_controller.py
```diff
@@ -323,7 +323,7 @@ def stop_agent_service(self, quiet):
         pass
 
     def request_agent_status(self):
-        """Invoked by a process that is not the agent to request the current agent dump the current detail
+        """Invoked by a process that is not the agent to request the current agent dump the current health
```
czerwingithub, line 326: I think you meant to leave this as `detail`

scalyr_agent/copying_manager.py
```diff
@@ -1026,6 +1026,18 @@ def generate_status(self, warn_on_rate_limit=False):
             for entry in self.__log_matchers:
                 result.log_matchers.append(entry.generate_status())
 
+            if self.__config.enable_health_check:
+                result.health_check_result = "Good"
+                if (
+                    time.time()
+                    > self.__last_attempt_time
```
czerwingithub, line 1033: What is `__last_attempt_time` initialized to before it actually has made an attempt?  `None`?  We will need to handle that.

scalyr_agent/copying_manager.py
```diff
@@ -1026,6 +1026,18 @@ def generate_status(self, warn_on_rate_limit=False):
...
+                    + self.__config.healthy_max_time_since_last_copy_attempt
+                ):
+                    result.health_check_result = (
```
czerwingithub, line 1036: Can you add in a unit test that tests this method?  In particular, verify it returns a bad message if the last copy attempt has been exceeded.

scalyr_agent/configuration.py
```diff
@@ -1221,6 +1221,23 @@ def strip_domain_from_default_server_host(self):
         """
         return self.__get_config().get_bool("strip_domain_from_default_server_host")
 
+    @property
```
czerwingithub, line 1224: Add test cases to `configuration_test.py` for your two new options here.

scalyr_agent/agent_main.py
```diff
@@ -1949,6 +1984,14 @@ def stop(self):
         default=False,
         help="For status command, prints detailed information about running agent.",
     )
+    parser.add_option(
+        "-H",
+        "--health_check",
+        action="store_true",
+        dest="health_check",
+        default=False,
+        help="For status command, prints health check status. Return code will be 0 for a passing check, and 1 for failing",
```
czerwingithub, line 1993: What happens if the health check option is not enabled for the agent process?  We should print a helpful message when they invoke `scalyr-agent-2 status -h`.

scalyr_agent/agent_main.py
```diff
@@ -739,12 +751,14 @@ def __detailed_status(self, data_directory, status_format="text"):
             )
             return 1
 
+        return_code = 0
         fp = open(status_file)
         for line in fp:
             print(line.rstrip())
+            if health_check and line.rstrip() != "Health check: Good":
```
czerwingithub, line 758: I think we should, even if we are doing `status -v` (so not an explicit health check), we should return a non-zero status code if the health check has failed.  We should also return different return code, like 2.  That way, scripts can tell the difference between the agent not running and a failed health code.

Of course, if we do `status -v` and the agent does not have health checks enabled, then we shouldn't return a status code based on the health check.    We could do that by only emitting the health check information in the `status -v` output if the health check is enabled.  Then, in this code, we just look for a line that starts with `Health check` and see if it is good or bad... setting some variable to True or False depending on the status.  If we don't find any line that starts with `Health check` then we set that variable to None.

We should also be sure this handles the JSON format for the status correctly as well.

scalyr_agent/agent_main.py
```diff
@@ -1814,8 +1860,8 @@ def __calculate_overall_stats(
 
         return result
 
-    def __report_status_to_file(self):
-        # type: () -> str
+    def __report_status_to_file(self, health_check=False):
```
czerwingithub, line 1863: I don't think you really need to pass in `health_check` anymore, right?  The format should already be set to JSON and so we should just return back the normal JSON status object.

scalyr_agent/agent_main.py
```diff
@@ -887,6 +929,10 @@ def __run(self, controller):
         # a file because a user has run the 'detailed_status' command.
         self.__controller.register_for_status_requests(self.__report_status_to_file)
 
+        # Register handler for when we get an interrupt signal.  That indicates we should dump the status to
+        # a file because a user has run the 'detailed_status' command.
+        self.__controller.register_for_health_check(self.__report_health_to_file)
```
czerwingithub, line 934: Do you need this anymore?

scalyr_agent/agent_main.py
```diff
@@ -1878,6 +1924,16 @@ def __report_status_to_file(self):
 
         return final_file_path
 
+    def __report_health_to_file(self):
```
czerwingithub, line 1927: I don't believe you should need this anymore.

scalyr_agent/platform_controller.py
```diff
@@ -355,6 +355,18 @@ def register_for_status_requests(self, handler):
         """
         pass
 
+    def register_for_health_check(self, handler):
```
czerwingithub, line 358: I don't think you should need this anymore.

scalyr_agent/platform_windows.py
```diff
@@ -125,6 +125,8 @@
 _SERVICE_DESCRIPTION_ = "Collects logs and metrics and forwards them to Scalyr.com"
 # A custom control message that is used to signal the agent should generate a detailed status report.
 _SERVICE_CONTROL_DETAILED_REPORT_ = win32service.SERVICE_USER_DEFINED_CONTROL - 1
+# A custom control message that is used to signal the agent should generate a health check report.
+_SERVICE_CONTROL_HEALTH_CHECK_ = win32service.SERVICE_USER_DEFINED_CONTROL - 2
```
czerwingithub, line 129: I don't think you should need this anymore.  We should just use the normal status -v signal.

scalyr_agent/platform_windows.py
```diff
@@ -250,6 +252,8 @@ def __init__(self):
         self.__termination_handler = None
         # The method to invoke when status is requested by another process.
         self.__status_handler = None
+        # The method to invoke when health check is requested by another process.
+        self.__health_check_handler = None
```
czerwingithub, line 256: I don't think you should need this anymore.

scalyr_agent/platform_windows.py
```diff
@@ -647,11 +651,23 @@ def register_for_status_requests(self, handler):
         """
         self.__status_handler = handler
 
+    def register_for_health_check(self, handler):
```
czerwingithub, line 654: I don't think you should need this anymore.

Kami:
I assume we also plan to add some documentation  on how to utilize this new functionality in the context of Kubernetes livenessProbe?

I think we should also add release notes entry and point users to the usage example in our docs.

scalyr_agent/copying_manager.py
Kami: Hm, I'm actually wondering why even have this ``enable_health_check`` config option.

I assume it doesn't add that much overhead and we could simply remove the flag and have it enabled by default? Or other people feel strongly about having a flag?

yanscalyr: Only thing I can think of is having `status -v` return an error code might cause some issues if the user isnt prepared for it, but I can't actually think of a good example of that. I'll let @czerwingithub comment but I would be fine with removing that option.

czerwingithub: Yeah, I agree w/ Tomaz.  It doesn't make sense to not always have this on.  My apologies for not also thinking of that earlier.

yanscalyr:
We have it on by default for k8s actually, although we should still have documentation for it. A release note and documentation on the main site makes sense.

PR #599, opened 2020-07-16
DO NOT MERGE AGENT-425 Add debug log stmts for uploading duplicate logs

oliverhsu77:
Add debug statements where files are closed and where new log processors are created.


scalyr_agent/log_processing.py
```diff
@@ -926,6 +926,10 @@ def prepare_for_inactivity(self, current_time=None):
         if close_file:
             for pending in self.__pending_files:
                 self.__close_file(pending)
+                log.info(
```
imron, line 929: Can you add information about what the values are - both of the field, and also the results of the calculation used to trigger this branch? i.e. the value of `delta`, `current_time`, `self.__modification_time_raw` and `self.__max_modification_duration`.

This way we can not only tell what is happening, but also why.

scalyr_agent/log_processing.py
```diff
@@ -3287,12 +3291,19 @@ def find_matches(
                     checkpoint_state = None
                     # Get the last checkpoint state if it exists.
                     if matched_file in previous_state:
+                        log.info(
+                            "%s has previous state %s" % (matched_file, previous_state)
```
imron, line 3295: This was the log message I was using in my test version, but when I was looking through the logs, I found it difficult to pick it out because it starts with the filename.  I'd change this to say something like:

`Copying %s from checkpoint %s`, then we can just search the logs for "Copying" and will match either of those lines.

scalyr_agent/log_processing.py
oliverhsu77: The data type in `self.__pending_files` is `FileState`.  Newbie Python question, will it print correctly?

imron: It should, if using the `%s` placeholder and you use `%` for formatting, then python will invoke `str()` on the object, which will print a string representation.

imron: Worst case scenario is that it will be something like `<FileState object at 0x105e764a8>`

imron:
Hi Oliver,

I've added new debug statements and tested them for coverage, and they all appear good.

Are you able to review the changes, and then prepare a debug build?

oliverhsu77:
@imron Changes LGTM.

oliverhsu77:
@ArthurKamalov LGTM

imron:
Added extra debug statements to see _why_ we are closing all the files.  Can someone please create a new debug file for release?  (Note: the failing test is a problem with uploading codecov results to s3 and is unrelated to these changes).

PR #600, opened 2020-07-22
CT-109: Unmatched group error

yanscalyr:
Turns out if `re` tries to use an optional group that it never matched it will throw an exception instead of just using an empty string like most other implementations, which causes us to not upload a lot of logs if a customer trips over it. This PR makes us handle that case a bit more gracefully, only not uploading the one log that actually had the issue and giving a hopefully helpful error.

### DIFF
--- a/scalyr_agent/log_processing.py
+++ b/scalyr_agent/log_processing.py
@@ -2876,9 +2876,11 @@ class LogLineRedacter(object):
             return input_line, False
 
         modified_it = False
+        redaction_rule_applying = None
 
         try:
             for redaction_rule in self.__redaction_rules:
+                redaction_rule_applying = redaction_rule
                 (input_line, redaction) = self.__apply_redaction_rule(
                     input_line, redaction_rule
                 )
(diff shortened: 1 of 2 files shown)


yanscalyr:
This issue only happens under python 2 apparently so the new test I added is failing under python 3.

scalyr_agent/log_processing.py
```diff
@@ -2877,11 +2877,29 @@ def process_line(self, input_line):
...
+        try:
+            for redaction_rule in self.__redaction_rules:
+                (input_line, redaction) = self.__apply_redaction_rule(
+                    input_line, redaction_rule
+                )
+                modified_it = modified_it or redaction
+        except re.error as e:
+            if e.message == "unmatched group":  # pylint: disable=no-member
+                log.error(
+                    "Error while getting the default gateway: %s. Please make sure any redaction rules only reference groups that are guaranteed to match.",
```
czerwingithub, line 2889: Default gateway?  I think you mean something like "Error while executing redaction rules".

Also, let's include the text of the redaction rule as well here.

scalyr_agent/log_processing.py
```diff
@@ -2877,11 +2877,29 @@ def process_line(self, input_line):
...
+                    six.text_type(e.message),  # pylint: disable=no-member
+                    limit_once_per_x_secs=300,
+                    limit_key="redaction_unmatched_group",
+                )
+            else:
+                log.error(
+                    "Error while applying redaction rule: %s",
+                    six.text_type(e.message),  # pylint: disable=no-member
```
czerwingithub, line 2897: Let's also add in the text of the redaction rule here.

scalyr_agent/log_processing.py
yanscalyr: I have no idea how that got there...

PR #603, opened 2020-07-26
Fix various issues

Kami:
This pull request fixes a couple of issues I noticed when auditing some code:

- Make sure we correctly open a temporary file when writing cert content. Previously it would fail in some scenarios.
- Make sure we also close all the files we open.
- Use the correct certificate name (looks like we missed that spot a while back when we changed the certs)

Keep in mind that all those code changes refer to "pure python" TLS code path which has other issues as well (most notably, us bundling cffi which will only work under scenarios where a customer already has the same version of cffi installed locally).

## TODO

- [x] Unit tests for ``_validate_chain_openssl`` function
- [x] Unit tests for ``_validate_chain_certvalidator`` function


Kami:
While working on tests, I noticed ``_validate_chain_openssl`` doesn't work correctly in all the scenarios (or perhaps it does, but other functions don't?).

It appears that right now we return a different set of certs if client sends SNI or not (``openssl s_client -showcerts -connect agent.scalyr.com:443 </dev/null`` vs ``openssl s_client -showcerts -connect agent.scalyr.com:443 -noservername </dev/null``).

That function will succeed on the cert returned when we don't send the SNI, but fail when we send it.

Rest of the code for cert validation seems to work correctly in both scenarios.

I didn't have time to dig in yet why that is the case, but just throwing this out.

PR #606, opened 2020-07-31
Fix pyinstaller spec monitors

ArthurKamalov:
Fix pyinstaller specifications to be able to import the rest of the agent monitors.


win32/scalyr-agent.spec
```diff
@@ -20,7 +20,19 @@ from PyInstaller.utils.hooks import collect_submodules
 
 block_cipher = None
 
-windows_monitors = ['windows_process_metrics', 'windows_system_metrics', 'windows_event_log_monitor']
+windows_monitors = [
```
czerwingithub, line 23: This is a duplicate of what we do here: https://github.com/scalyr/scalyr-agent-2/blob/master/setup.py#L78

I suspect there's not an easy way to have both setup.py and this file point at a common definition of this.  If that's true, can you at least add comments in both places reference to the other location so we know we have to update both?

You should know I also created a static pylint checker that made sure the one in setup.py is up to date. So, if we ever get rid of the setup.py, we should change that checker to look at this location.  You could add that to the comment.  See this: https://github.com/scalyr/scalyr-agent-2/blob/master/pylint_plugins/py2exe_checker.py#L44

ArthurKamalov:
@czerwingithub 
I adapted `pylint` checker to work with Pyinstaller. I removed everything connected with `six.moves` and only kept monitor modules checking because `PyInstaller` is able to handle `six.moves` "out-of-the-box".  I also created a python module `dynamic_modules.py` where we can specify what we want to import dynamically. I moved the `WINDOWS_MONITOR_MODULES_TO_INCLUDE` list from the `setup.py` file into this module too and now the `pylint` checker and the Pyinstaller itself get the needed monitors information from it.

PyInstaller does not rely on `setup.py`, so with its introduction(PyInstaller), we can remove the `py2exe` related code in 'setup.py'.

ArthurKamalov:
> LGTM. It would be nice if we could add a lint check for it at some point.

Thanks for the good advice. Done.

scalyr_agent/third_party/licenses/LICENSE_activestate.txt
```diff
@@ -19,7 +19,7 @@ ACTIVESTATE SOFTWARE INC. ("ACTIVESTATE") IS WILLING TO LICENSE THE SOFTWARE ONL
...
+   b. You may: (i) copy the Software for archival purposes, (ii) copy the Software for personal use purposes, (iii) use, copy, and distribute the Software solely for Your organization's internal non-production use and or internal business operation purposes including copying the Software to other workstations inside Your organization, (iv) redistribute parts of the Software outside of Your organization only as part of a Wrapped Application utilizing executable generators such as PerlApp, Perl2Exe, PAR, TclApp, py2app, py2exe or PyInstaller. Any copy must contain the original Software's proprietary notices in unaltered form. "Wrapped Application" means a single-file executable wherein all binary components are encapsulated in a single binary however You may not expose the base programming language as a scripting language within your own application program to end users.
```
czerwingithub, line 22: I don't think we should be changing the text for another library's LICENSE.  We should rever this change.  After all, they do say you cannot alter the notices in any form.

In practice, we don't need to have PyInstaller listed explicitly here anyway because they say "executable generators such as " -- it is not meant to be an exhaustive list.

PR #607, opened 2020-08-03
Bug/duplicate upload due to race

czerwingithub:
# Background

In Scalyr Agent release 2.1.6, we essentially turned on pipelining upload requests by default.  Pipelining is when we start to prepare next `/addEvents` upload request while still waiting on the response from the previous request (i.e., reading log content, serializing the log bytes into a network request, etc).  This allows us to overlap the round-trip-time of the first request with the construction of the second.  As long as the first request is accepted by the server, we can immediately turn around and send the second request.

The pipelining feature has been in our code base for several years now and used by customers who needed this performance optimization to maximum their upload throughput.  By making it default, we now have a larger number of customers using it.  Unfortunately, it turned out there was a somewhat subtle bug in how we constructed pipeline requests that caused issues for customers with particular log patterns.

# Bug

The essence of the bug is that we erroneously rely on the list of log processors being the same from when we create an `/addEvents` request and when we receive the response for that request.  That's because we keep a map for the callbacks that must be invoked for all of the log files with content in that request keyed by the index of the log processor associated with that log file in the processor list.  If the processor list changes between when we create the request and when we receive its response, then we will mistakenly associate the wrong callbacks with the wrong log processors (log files).

In non-pipelined requests, this is not an issue because that state can't really change between when the request is created and we get the response.  No other work is done while that request is in-flight.  However, with pipelined requests, the second request is created before the first request's response is processed.  If processing the first request results in changes to the log processor list (such as removing a log processor because its log file has been deleted or become stale), then the second request's callbacks will be invoked on the wrong processors.

# Symptoms

This bug manifested in a very puzzling manner.  The end result was that a given log file would be reuploaded to Scalyr starting from its start (i.e., byte offset zero).  This was caused by a fairly complex chain of events.  At the core, because the wrong callbacks were being invoked for the wrong log file, a given log file would be marked as "done" even though it wasn't really done.  Typically, a log file is only done when it has been deleted from the file system (otherwise, we are always expecting new bytes to show up in the log file).  So, when the Scalyr Agent recorded a log file as done, but then saw it again when it next scanned for new log files -- it would assume it was a completely new log file.  After all, the old one has been "deleted" from the file system (or so it though).  This resulted in the Scalyr Agent reuploading the log file from the start.

# Solution

The solution is really pretty straight-forward.  Instead of mapping which callbacks are associated with which LogProcessors based on its position in the LogProcessor list, we use a more stable index.  In particular, LogProcessors already have a unique id.  We use that as the key to callback dictionary.  Since this key is stable even with changes to the LogProcessor list, there is no more mixing up the callbacks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Packaged by Steven Czerwinski <czerwin@scalyr.com> on Aug 3, 2020 12:30 -0800
 --->
 
 Bug fixes:
-* Fixed race condition in pipelined requests which could lead to duplicate log upload, especially for systems with a large number of inactive log files.  Log files would be reuploaded from their start over short period of time (seconds to minutes).
+* Fixed race condition in pipelined requests which could lead to duplicate log upload, especially for systems with a large number of inactive log files.  Log files would be reuploaded from their start over short period of time (seconds to minutes).  This bug is triggered when pipelining is enabled, either by explicitly setting the `pipeline_threshold` config option or by using a Scalyr Agent release >= 2.1.6 (pipelining was turned on by default in 2.1.6).
 
 Misc:
 * ``compression_level`` configuration option now defaults to ``6`` when using ``deflate`` ``compression_type`` (``deflate`` is the default value for the ``compression_type`` configuration option). 6 offers the best trade off between compression ratio and CPU usage. For more information, please refer to the release notes document.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,14 @@
 Scalyr Agent 2 Changes By Release
 =================================
 
-## 2.1.8 "TBD" - July 10, 2020
+## 2.1.8 "Titan" - August 3, 2020
 
 <!---
-Packaged by Steven Czerwinski <czerwin@scalyr.com> on Jul 20, 2020 08:30 -0800
+Packaged by Steven Czerwinski <czerwin@scalyr.com> on Aug 3, 2020 12:30 -0800
 --->
+
+Bug fixes:
+* Fixed race condition in pipelined requests which could lead to duplicate log upload, especially for systems with a large number of inactive log files.  Log files would be reuploaded from their start over short period of time (seconds to minutes).
 
 Misc:
 * ``compression_level`` configuration option now defaults to ``6`` when using ``deflate`` ``compression_type`` (``deflate`` is the default value for the ``compression_type`` configuration option). 6 offers the best trade off between compression ratio and CPU usage. For more information, please refer to the release notes document.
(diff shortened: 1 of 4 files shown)


czerwingithub:
@ArthurKamalov @Kami I'd appreciate if one of you could take a look at this.  This is the fix for the duplicate log upload issue and we are looking to cut a release on Monday.  Thanks.

CHANGELOG.md
```diff
@@ -1,12 +1,15 @@
...
-## 2.1.8 "TBD" - July 10, 2020
+## 2.1.8 "Titan" - August 3, 2020
 
 <!---
-Packaged by Steven Czerwinski <czerwin@scalyr.com> on Jul 20, 2020 08:30 -0800
+Packaged by Steven Czerwinski <czerwin@scalyr.com> on Aug 3, 2020 12:30 -0800
 --->
 
+Bug fixes:
+* Fixed race condition in pipelined requests which could lead to duplicate log upload, especially for systems with a large number of inactive log files.  Log files would be reuploaded from their start over short period of time (seconds to minutes).
```
Kami, line 11: Is it also worth adding which versions and configurations would it affect (aka > 2.6.1 or pipeling enabled)?

scalyr_agent/log_processing.py
```diff
@@ -2034,6 +2034,10 @@ def __init__(
         # happens).  However, this is how the old agent worked and the UI relies on it, so we just keep the old system
         # going for now.
         self.__thread_name = "Lines for file %s" % file_path
+
+        # Note: thread id is also used as a unique identifier for LogFileProcessors (see #CT-107)
+        # So even if "thread_id" is no longer used by the Scalyr API, we still need it as a unique
+        # integer identifier for the LogFileProcessor
         self.__thread_id = LogFileProcessor.generate_unique_id()
```
Kami, line 2041: I had a look and it looks like this returns a monotonically increasing integer which should be fine.

If we ever needed to avoid the need to use a lock we could probably also use uuid v4, but since we already have that locking code in place, this is fine as well.

CHANGELOG.md
czerwingithub: Good point.  I'd added that note to this comment.

scalyr_agent/log_processing.py
czerwingithub: Yeah, Imron and I had that exact same mini-debate.  We agreed this is good enough for now.

CHANGELOG.md
```diff
@@ -1,12 +1,15 @@
...
-## 2.1.8 "TBD" - July 10, 2020
+## 2.1.8 "Titan" - August 3, 2020
 
 <!---
-Packaged by Steven Czerwinski <czerwin@scalyr.com> on Jul 20, 2020 08:30 -0800
+Packaged by Steven Czerwinski <czerwin@scalyr.com> on Aug 3, 2020 12:30 -0800
 --->
 
+Bug fixes:
+* Fixed race condition in pipelined requests which could lead to duplicate log upload, especially for systems with a large number of inactive log files.  Log files would be reuploaded from their start over short period of time (seconds to minutes).  This bug is triggered when pipelining is enabled, either by explicitly setting the `pipeline_threshold` config option or by using a Scalyr Agent release >= 2.1.6 (pipelining was turned on by default in 2.1.6).
```
Kami, line 11: LGTM :+1:

PR #608, opened 2020-08-03
Various fixes and improvements to the AMI tests scripts

Kami:
This pull request includes the following fixes and improvements to the AMI tests scripts:

* Remove out of date comments

* Fix Windows script and make sure we close the DB connection after querying to version info

* Fix Windows script and make sure we correctly assert on the agent version

* Enable Strict Mode for the Windows script and make sure it fails with non zero exit code on failure

* Add support for debug / verbose mode to the Windows script

* Move duplicated code into shared partial files which are included

* User longer wait interval while polling and waiting for instance to come online to make it less likely to hit AWS API rate limits

* Make sure we remove and clear log files before each set of checks / assertions to make sure we are not accidentally asserting on old log messages.


tests/ami/scripts/test_windows.ps1.j2
```diff
@@ -15,44 +14,55 @@
...
+        $q = "SELECT Value FROM Property WHERE Property = 'ProductVersion'"
+        $View = $database.GetType().InvokeMember(
+            "OpenView", "InvokeMethod", $Null, $database, ($q)
+        )
+
+        $View.GetType().InvokeMember("Execute", "InvokeMethod", $Null, $View, $Null)
+        $record = $View.GetType().InvokeMember( "Fetch", "InvokeMethod", $Null, $View, $Null )
+        $version = $record.GetType().InvokeMember( "StringData", "GetProperty", $Null, $record, 1 )
+
+        $View.GetType().InvokeMember("Close", "InvokeMethod", $Null, $View, $Null)
```
ArthurKamalov, line 54: Very nice catch, thanks.

tests/ami/packages_sanity_tests.py
```diff
@@ -525,9 +534,9 @@ def render_script_template(
 
     template_context["verbose"] = verbose
 
-    template = Template(script_template)
+    env = Environment(loader=FileSystemLoader(SCRIPTS_DIR),)
+    template = env.from_string(script_template)
```
ArthurKamalov, line 538: Does it also allow us to substitute text in the additional templates (such as `partial/reencode_config_without_bom.ps1.j2`) or is only possible for the main template?

Kami: I believe nested includes should also be supported, but not 100% :)

PR #609, opened 2020-08-03
Fix formatting of the health check line in the agent status output

Kami:
This pull request updates formatting of the "Health Check" line in the agent status output - it makes sure the line value is left padded and consistent with other lines.

The changed code handles both format - old and new when parsing the line.


scalyr_agent/agent_main.py
```diff
@@ -766,8 +766,8 @@ def __detailed_status(
                     print("Cannot get health check result.")
             elif (
                 status_format == "text"
-                and "Health check" in line
-                and line.rstrip() != "Health check: Good"
+                and "Health check:" in line
+                and "Good" not in line.rstrip()
```
yanscalyr, line 770: This line would alternatively contain the error message, which could in theory contain "Good", seems like a long shot but something that we could trip over without noticing. I think some kind of more strict check is needed here.

Kami: Yeah, I wanted to keep it simple, but I can add regex which handler both scenarios: ``^Health check\:\s+Good$"``

Kami: Here we go - 2012790598effa1917007e768f1929618d1641c1.

PR #616, opened 2020-08-05
Log OpenSSL version against which Python is linked to in the agent start up line

Kami:
This pull request updates agent start up line to also log OpenSSL version against which Python binary which the agent is using to run is linked to.

This change should make it easier for us to troubleshoot some type of issues.

### DIFF
--- a/tests/ami/packages_sanity_tests.py
+++ b/tests/ami/packages_sanity_tests.py
@@ -256,8 +256,14 @@ def _get_source_type(version_string):
         return "url"
     elif os.path.exists(version_string) and os.path.isfile(version_string):
         return "file"
-    else:
+    elif re.match(r"\d+\.\d+\.\d+", version_string) or version_string in ["current"]:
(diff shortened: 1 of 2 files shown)


tests/ami/packages_sanity_tests.py
```diff
@@ -256,8 +256,14 @@ def _get_source_type(version_string):
         return "url"
     elif os.path.exists(version_string) and os.path.isfile(version_string):
         return "file"
-    else:
+    elif re.match(r"\d+\.\d+\.\d+", version_string) or version_string in ["current"]:
```
ArthurKamalov, line 259: I may be missing something, but can we use`version_string == "current"` here instead of the list lookup?)

Kami: Yeah, I thought we also support stable (as an alias for current) so I used list first, but you are right - it's not needed.

PR #617, opened 2020-08-05
Also run Windows AMI tests when running AMI tests for stable package versions

Kami:
This pull request updates our AMI end to end tests so we also run Windows tests when we run tests for stable package versions.

I believe this was an oversight when originally adding tests for Windows.

To run those tests, we first download latest Scalyr Agent installer from the repo since those tests rely on uploading the binary to the server.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2234,8 +2234,7 @@ workflows:
            - build-windows-package
            - build-linux-packages-and-installer-script
          <<: *master_and_release_only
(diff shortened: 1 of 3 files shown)


Kami:
Confirmed it's working - https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/3898/workflows/51d0cc35-80bc-47f7-a30a-940af865fab4/jobs/51381.

PR #619, opened 2020-08-17
CT-127: Change Dockerfile to install journald dependencies

yanscalyr:
Seems like at least one customer wants to run the journald monitor from docker but newer versions of the `slim` python base image we use lacks the ability to download the journald dependencies. This PR changes the base image of the first stage of our build and installs the journald monitor dependencies.

Questions to consider:
  1. Do we need these changes in the k8s images? The change in size is under 1MB so it might be easier to maintain if we change everything
  2. What specific version of `systemd-python` to use in the `requirements.txt` file


docker/Dockerfile
```diff
@@ -1,10 +1,12 @@
...
 # the scalyr-agent
 # NOTE: multi-stage builds require Docker 17.05 or greater
-FROM python:2.7-slim as scalyr-dependencies
+FROM python:2.7-buster as scalyr-dependencies
 MAINTAINER Scalyr Inc <support@scalyr.com>
 
-RUN apt-get update && apt-get install -y build-essential
+RUN apt-get update && apt-get install -y build-essential libsystemd-dev
+# necessary for journald monitor
+RUN apt-get install libsystemd-dev -y
```
dlobue, line 9: Why does `libsystemd-dev` need to be installed twice?

yanscalyr: Oops, good catch

docker/requirements.txt
```diff
@@ -7,3 +7,4 @@ yappi==1.2.3
 # to prevent versions conflict, install the appropriate version of 'requests', to force 'docker' to reuse it.
 requests==2.15.1
 docker==4.1.0
+systemd-python
```
dlobue, line 10: probably should pin a version here.

Kami (approved):
Please add a changelog entry, besides that, LGTM.

PR #620, opened 2020-08-27
CT-129: Support redaction rules in the Scalyr Agent's syslog monitor

yanscalyr:
Still needed:
- Same changes in Docker mode, that mode already splits logs into different files so we will need to handle that.
- Figure out what to do about the `message_log` monitor configuration parameter, since we may have multiple files now.
- Update the tests for the new functionality.


yanscalyr:
I'm not familiar with our AMI tests so I'm not sure why they are failing, could use a tip on that.

Kami:
AMI tests failure looks like a temporary issue related to some upstream yum repo or similar. Restart should likely fix it.

scalyr_agent/builtin_monitors/syslog_utils.py
```diff
@@ -0,0 +1,211 @@
...
+            self,
+            global_config,
+            formatter,
+            max_log_size=max_log_size,
+            max_log_rotations=max_log_rotations,
+            extra_config=extra_config,
+        )
+        self.current_log_files = []
+        self.log_deleter = LogDeleter(
+            extra_config.get("docker_check_for_unused_logs_mins", 60),
```
Kami, line 41: Do we intentionally use keys starting with ``docker_`` here?

I assume we want to use the same additional options for docker and syslog monitor, or?

scalyr_agent/builtin_monitors/syslog_utils.py
```diff
@@ -0,0 +1,211 @@
...
+                            mtime = os.path.getmtime(rotated_file)
+                            if current_time - mtime > self._delete_interval:
+                                result.append(rotated_file)
+                        else:
+                            if added:
+                                result.append(rotated_file)
+
+                    except OSError as e:
+                        global_log.warn(
+                            "Unable to read modification time for file '%s', %s"
```
Kami, line 168: If we are only catching this specific exception, it would make sense to only wrap that line where we can except that exception (aka ``os.path.getmtime``).

scalyr_agent/builtin_monitors/syslog_utils.py
```diff
@@ -0,0 +1,211 @@
...
+        )
+
+        self._last_check = time.time()
+        self.existing_logs = []
+
+    def add_current_log(self, path):
+        self.existing_logs.append(path)
+
+    def _get_old_logs_for_glob(
+        self, current_time, glob_pattern, existing_logs, check_rotated, max_rotations
```
Kami, line 136: Docstrings + type annotations for all those arguments would be nice.

scalyr_agent/builtin_monitors/syslog_utils.py
```diff
@@ -0,0 +1,211 @@
...
+                )
+            except OSError as e:
+                global_log.warn(
+                    "Error deleting old log file '%s', %s"
+                    % (filename, six.text_type(e)),
+                    limit_once_per_x_secs=300,
+                    limit_key="delete-%s" % filename,
+                )
+
+        self.existing_logs = []
```
Kami, line 211: Not sure if I understand why we need to reset this variable here.

scalyr_agent/builtin_monitors/syslog_utils.py
yanscalyr: Yes this is intentional, these are fields that are only useful in cases where the syslog monitor is running on docker mode, and the log deleter only does anything in docker mode as well.

scalyr_agent/builtin_monitors/syslog_utils.py
Kami: EDIT: Now I see you moved this code from a different module - in any case, now is our opportunity to improve it :)

scalyr_agent/builtin_monitors/syslog_utils.py
yanscalyr: This is something I attempted to bring over from the old implementation of LogDeleter, my understanding of the old implementation is that this list gets populated by logs that have been written to recently (since the last time we checked for old logs) to avoid checking them for deletion. If we didn't clear it nothing would ever get deleted, and we clear it here because to set all logs to once again be able to be deleted.

Having said that, I have no idea why we really need this. If a log was written recently we would find a fresh write time and they wouldn't get deleted anyway. If you see any reason to keep this please let me know, but if you can't find any reasoning behind it either then I would be happy to remove it.

scalyr_agent/builtin_monitors/syslog_utils.py
yanscalyr: Type annotations aren't supported for python 2 right? Or did you just mean document them in the docstring?

Kami: They are supported using comments notation.

Kami:
Mostly LGTM, but let's please also add some end to end tests which test this new functionality.

IIRC, Arthur already added some basic one in the past (``tests/smoke_tests/monitors_test/syslog_test.py``) so you should be able to utilize this as a starting point.

yanscalyr:
I did add some tests to that file which confirm we still write expected data to the expected file when `syslog_logs` are configured, but any of the actual processing is not part of the monitor itself so won't be tested there.

PR #623, opened 2020-09-08
Use new client library

yanscalyr:


yanscalyr:
So far I haven't removed any existing functionality, and for new client uploads to work the old method needs to be configured correctly too. There is also no buffering at the moment.

yanscalyr:
Now there is a config option `use_new_ingestion` that will use the new client and not fail if a real Scalyr connection can't be made. There is a cert validation error at startup because it still creates a Scalyr client, this can be avoided with `verify_server_certificate` false, or just ignored since it runs fine after that.

Also to clarify about buffering, we don't send one log line at a time, each `LogFileProcessor` loop will build up a collection of whatever lines it picks up and send them all in one request. I need to review the logic but this might actually recreate our minimum wait time for buffering correctly, but there is no upper limit on how many events it sends at once yet so we could run into issues there.

scalyr_agent/copying_manager.py
```diff
@@ -224,6 +226,32 @@ def __init__(self, configuration, monitors):
...
+                DataPlaneAPIClient,
+            )
+
+            self._session = Session(uuid=str(uuid.uuid4()))
+
+            self._control_plane_client = ControlPlaneAPIClient(
+                api_token=str(configuration.api_key),
+                cert_path=str(configuration.ca_cert_path),
+            )
+            manager_address = self._control_plane_client.send_client_hello()
```
Kami, line 247: For now doing this here is fine, but in the future we will likely need to handle this elsewhere and also handle retries, etc.

scalyr_agent/copying_manager.py
```diff
@@ -821,7 +857,12 @@ def run(self):
 
                             # Now block for the response.
                             blocking_response_time_start = time.time()
-                            (result, bytes_sent, full_response) = get_response()
+                            if self.__config.use_new_ingestion:
+                                result = "success"
+                                bytes_sent = 0
```
Kami, line 862: In theory you should be able to use this value from the IngestionHeadersPb, but this may not be accessible here yet (likely need to make small change to the client so it's available here).

scalyr_agent/log_processing.py
```diff
@@ -2395,6 +2409,22 @@ def perform_processing(self, add_events_request, current_time=None):
...
                     event = self.__create_events_object(line_object, sample_result)
+
+                    if self._data_plane_client:
+                        import scalyr_ingestion_client.log_line as ingestion_client_line  # pylint: disable=import-error
+
+                        new_event_timestamp = line_object.timestamp
+                        if not new_event_timestamp:
+                            new_event_timestamp = int(current_time)
+                        new_event = ingestion_client_line.LogLine(
+                            str(self._event_id),
```
Kami, line 2420: id should be start offset in the log file where that line was read from.

scalyr_agent/configuration.py
```diff
@@ -1163,6 +1163,11 @@ def ca_cert_path(self):
         """Returns the configuration value for 'ca_cert_path'."""
         return self.__get_config().get_string("ca_cert_path")
 
+    @property
+    def use_new_ingestion(self):
```
Kami, line 1167: I wonder if a better name would be use_new_ingestion_api / use_new_api

scalyr_agent/copying_manager.py
Kami: Basically, we should do similar as we with the existing client - we use an exponential back off and try to retry connecting to the API.

scalyr_agent/copying_manager.py
```diff
@@ -224,6 +226,32 @@ def __init__(self, configuration, monitors):
...
+                Session,
+            )
+            from scalyr_ingestion_client.client import (  # pylint: disable=import-error
+                ControlPlaneAPIClient,
+                DataPlaneAPIClient,
+            )
+
+            self._session = Session(uuid=str(uuid.uuid4()))
+
+            self._control_plane_client = ControlPlaneAPIClient(
```
Kami, line 243: While testing those changes, I noticed ``send_hello`` gets called every time we send events.

> 3:21:00.722 110 I requestDeserializationNanos=60022 tag=ingestionManager.clientHello 
3:21:00.722 111 I requestDeserializationNanos=188475 tag=ingestionManager.sendEvents 
3:21:00.727 115 I requestDeserializationNanos=189423 tag=ingestionManager.sendEvents 
3:21:01.280 46 I path=/home/kami/w/scalyr/ingestion-pipeline/.scalyrDbState/env/eventPersistence/directdb/meta debugName=DirectDB freeSpace=388148609024 tag=getFreeSpace usableSpace=343632363520 value=343632363520 
3:21:01.321 52 I source=counters tag=directDbAllMemtableMemory_hires value=0 
3:21:01.336 51 I path=/home/kami/w/scalyr/ingestion-pipeline/.scalyrDbState/env/eventPersistence/directdb debugName=DirectDB freeSpace=388148609024 tag=getFreeSpace usableSpace=343632363520 value=343632363520 
3:21:05.734 118 I requestDeserializationNanos=166601 tag=ingestionManager.sendEvents 
3:21:05.747 122 I requestDeserializationNanos=156924 tag=ingestionManager.sendEvents 
3:21:05.753 123 I requestDeserializationNanos=2172218 tag=ingestionManager.clientHello 
3:21:06.153 58 I node=localhost service=INGESTION_MANAGER action=heartbeat tag=serviceDiscovery isDrained=false isHealthy=true 
3:21:06.321 52 I source=counters tag=directDbAllMemtableMemory_hires value=0 
3:21:10.836 126 I requestDeserializationNanos=176803 tag=ingestionManager.clientHello 
3:21:11.284 46 I path=/home/kami/w/scalyr/ingestion-pipeline/.scalyrDbState/env/eventPersistence/directdb/meta debugName=DirectDB freeSpace=388148506624 tag=getFreeSpace usableSpace=343632261120 value=343632261120 
3:21:11.322 52 I source=counters tag=directDbAllMemtableMemory_hires value=0 
3:21:11.338 51 I path=/home/kami/w/scalyr/ingestion-pipeline/.scalyrDbState/env/eventPersistence/directdb debugName=DirectDB freeSpace=388148506624 tag=getFreeSpace usableSpace=343632261120 value=343632261120 
3:21:11.354 53 I tag=cpuPing 
3:21:11.838 49 I tag=compactionScannerStatus position=[[empty]] recompressScanPosition=[[empty]] level=2 
3:21:15.721 58 I node=localhost service=INGESTION_MANAGER action=heartbeat tag=serviceDiscovery isDrained=false isHealthy=true 
3:21:15.756 129 I requestDeserializationNanos=604205 tag=ingestionManager.sendEvents 
3:21:15.864 132 I requestDeserializationNanos=105244 tag=ingestionManager.clientHello 
3:21:16.323 52 I source=counters tag=directDbAllMemtableMemory_hires value=0 
3:21:20.773 135 I requestDeserializationNanos=686218 tag=ingestionManager.sendEvents 
3:21:20.885 138 I requestDeserializationNanos=91936 tag=ingestionManager.clientHello 
3:21:21.286 46 I path=/home/kami/w/scalyr/ingestion-pipeline/.scalyrDbState/env/eventPersistence/directdb/meta debugName=DirectDB freeSpace=388148428800 tag=getFreeSpace usableSpace=343632183296 value=343632183296 
3:21:21.322 52 I source=counters tag=directDbAllMemtableMemory_hires value=0 

This needs to be refactored so we only call it once on start up - similar to when and how we do remote server check with the current API.

Kami (changes requested):
To recap the feedback:

1. We should send client hello when initially establishing a connection on agent startup (similar to how we currently handle connectivity check). Right now it happens each time when CopyingManager is instantiating and when we send the events to the server. This way we reduce unnecessary object churn.

It's true that the server may at some point tell the client to move / start sending data to different Ingestion Manager on response, but this should be handled elsewhere, likely in the client library.

2. Connection establish logic should follow the same "exponential retry" logic as we currently follow in the client.

3. For event ids and start and end sequences, we need to use start offsets for particular lines in the log file.

For all those changes, we should ideally introduce a new higher level client abstraction - right now this could be contained in the agent repo, but eventually, we may want to move it to the client library repo (although this depends since some of that code will be agent specific).

yanscalyr:
4 tests are failing due to me moving when we create `LogMatcher`s, they used to be created immedietly when the `CopyingManager` was and I moved them to when the thread for it is started. I believe this should have no negative effects on how the agent runs normally but since we have these tests I'm not actually sure. So the question is do I fix the tests or do I change the code and figure out a different way to get the new clients where I need them?

Kami:
If end to end tests still pass those are likely related to unit tests relying on a specific time of creation.

Having said that it's probably still safer to leave it as it was and figure out a different way to get the client to where you need it.

scalyr_agent/copying_manager.py
```diff
@@ -821,7 +833,12 @@ def run(self):
 
                             # Now block for the response.
                             blocking_response_time_start = time.time()
-                            (result, bytes_sent, full_response) = get_response()
+                            if self.__config.use_new_ingestion:
+                                result = "success"
+                                bytes_sent = 0
```
Kami, line 838: I had a look again and the ``SendEventsAPIResponse`` message already contains ``bytes_ingested`` field which you could use.

This indicates actual bytes ingested though so if we are interested in raw number of bytes sent (pre compression), I will modify a client to expose this information on ``send_events``.

Kami: This may help - https://github.com/scalyr/python-ingestion-library/pull/12

Keep in mind that ``headers_pb.events_bytes`` field only includes total bytes which represent the size of ``attributes`` field for all the events in the payload - aka it doesn't include gRPC metadata + headers overhead (which is probably fine for here).

yanscalyr: Yeah the raw number of bytes is what we care about, its needed for the rate limiting mostly.

yanscalyr:
Moved LogMatcher creation back to where it was and instead I pass in a reference to the copying manager that created it, I'm not sure if this is actually totally okay, I feel like I'm forgetting something...

Kami:
> Moved LogMatcher creation back to where it was and instead I pass in a reference to the copying manager that created it, I'm not sure if this is actually totally okay, I feel like I'm forgetting something...

This could be problematic because of a potential cyclic reference (and a chance for a memory leak) so we should try to find a different approach.

Kami:
LGTM, only thing which is still missing is correct sequence ids for the events (those should contain beginning offsets from a file where the line was read from).

For example, if we read the following lines from the file 1.txt in this order:

- line 1\n (start offset = 0, end offset = 7)
- line 2\n (start offset = 8, end offset = 15)
- line 3\n (start offset = 16, end offset = 23)

start offset would be the event id and ``sequence_range_start`` would be ``0`` and ``sequence_range_end`` would be ``16``.

Kami (approved):
I've done quick check locally and it appears sequence ids are being set correctly.

This should be fine for now, but in the very near future we will also need some end to end tests which verify that.

PR #624, opened 2020-09-15
Also define pytest_benchmark_scale_unit function even if the nvironment variable is not set

Kami:
Small change to ``pytest_util`` module to also define and export ``pytest_benchmark_scale_unit`` function even if the corresponding environment variable is not defined.

This way, the function can more easily be re-used by other projects which rely on ``pytest_util`` module.


ArthurKamalov (approved):
LGTM

PR #632, opened 2020-10-01
Postgres and mysql monitors fix

ArthurKamalov:
* Fix postgres_monitor error caused by wrong type conversion during the parsing of the config file.
* Fall back to native json library if orjson can not handle unsigned 64bit integers. 

Monitor tests and log analyzing tools were also improved.

### DIFF
--- a/tests/smoke_tests/monitors_test/mysql_test.py
+++ b/tests/smoke_tests/monitors_test/mysql_test.py
@@ -32,6 +32,7 @@ from tests.utils.agent_runner import AgentRunner
 from tests.utils.dockerized import dockerized_case
 from tests.image_builder.monitors.common import CommonMonitorBuilder
 from tests.utils.log_reader import LogMetricReader
+from tests.utils.log_reader import AgentLogReader
 
 import six
 
@@ -107,7 +108,9 @@ def _test(request, python_version):
     )
 
     reader = MySqlLogReader(runner.mysql_log_path)
-    reader.start()
+    agent_log_reader = AgentLogReader(runner.agent_log_file_path)
+
+    agent_log_reader.wait(5)
 
     metrics_to_check = [
         "mysql.global.com_insert",
@@ -116,7 +119,7 @@ def _test(request, python_version):
         "mysql.global.com_update",
     ]
 
-    previous_metric_values = reader.get_metrics(metrics_to_check)
+    previous_metric_values = reader.wait_for_metrics_exist(metrics_to_check)
 
     rows = mysql_cursor.execute("SELECT * FROM test_table")
 
@@ -155,7 +158,7 @@ def _test(request, python_version):
 
     reader.wait_for_metrics_increase(metrics_to_check, previous_metric_values)
 
-    current_metrics = reader.get_metrics(metrics_to_check)
+    current_metrics = reader.wait_for_metrics_exist(metrics_to_check)
 
     diff = dict(
         (name, current_metrics[name] - previous_metric_values[name])
@@ -169,7 +172,7 @@ def _test(request, python_version):
         "mysql.global.com_insert": 2,
     }
 
-    runner.stop()
+    agent_log_reader.go_to_end()
 
 
 @pytest.mark.usefixtures("agent_environment")
--- a/tests/smoke_tests/monitors_test/mysql_test.py
+++ b/tests/smoke_tests/monitors_test/mysql_test.py
@@ -119,7 +119,17 @@ def _test(request, python_version):
         "mysql.global.com_update",
     ]
 
-    previous_metric_values = reader.wait_for_metrics_exist(metrics_to_check)
+    reader.wait_for_metrics_exist(metrics_to_check, timeout=60)
+
+    reader.wait_for_metrics_equal(
+        expected={
+            "mysql.global.com_insert": 0,
+            "mysql.global.com_select": 2,
+            "mysql.global.com_delete": 0,
+            "mysql.global.com_update": 0,
+        },
+        timeout=60,
+    )
 
     rows = mysql_cursor.execute("SELECT * FROM test_table")
 
@@ -156,21 +166,15 @@ def _test(request, python_version):
     (row1_text,) = mysql_cursor.fetchone()
     assert row1_text == "updated_row1"
 
-    reader.wait_for_metrics_increase(metrics_to_check, previous_metric_values)
-
-    current_metrics = reader.wait_for_metrics_exist(metrics_to_check)
-
-    diff = dict(
-        (name, current_metrics[name] - previous_metric_values[name])
-        for name in metrics_to_check
+    reader.wait_for_metrics_equal(
+        expected={
+            "mysql.global.com_insert": 2,
+            "mysql.global.com_select": 9,
+            "mysql.global.com_delete": 1,
+            "mysql.global.com_update": 1,
+        },
+        timeout=60,
     )
-
-    assert diff == {
-        "mysql.global.com_select": 7,
-        "mysql.global.com_delete": 1,
-        "mysql.global.com_update": 1,
-        "mysql.global.com_insert": 2,
-    }
 
     agent_log_reader.go_to_end()
 
--- a/tests/smoke_tests/monitors_test/postgres_test.py
+++ b/tests/smoke_tests/monitors_test/postgres_test.py
@@ -45,7 +45,9 @@ DATABASE = "scalyr_test_db"
 def postgresql_client():
     os.system("service postgresql start")
 
-    os.system('psql -c "CREATE USER {} WITH PASSWORD \'{}\'";'.format(USERNAME, PASSWORD))
+    os.system(
+        "psql -c \"CREATE USER {} WITH PASSWORD '{}'\";".format(USERNAME, PASSWORD)
+    )
     os.system('psql -c "CREATE DATABASE {};"'.format(USERNAME))
     os.system('psql -c "CREATE DATABASE {};"'.format(DATABASE))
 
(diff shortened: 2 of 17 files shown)


tests/smoke_tests/monitors_test/postgres_test.py
```diff
@@ -0,0 +1,188 @@
...
+@pytest.fixture()
+def postgres_cursor(postgresql_client):
+    cursor = postgresql_client.cursor()
+
+    yield cursor
+
+    cursor.close()
+
+
+class PostgresAgentRunner(AgentRunner):
```
Kami, line 71: Great, thanks for working on those tests :+1:

tests/smoke_tests/monitors_test/postgres_test.py
```diff
@@ -0,0 +1,188 @@
...
+USERNAME = "scalyr_test_user"
+PASSWORD = "scalyr_test_password"
+DATABASE = "scalyr_test_db"
+
+
+@pytest.fixture()
+def postgresql_client():
+    os.system("service postgresql start")
+
+    os.system(
```
Kami, line 48: I know this code is only used in the test context and we use hard-coded values, but still, to be on the safe side and to avoid various static analysis tools to complaining, we should quote all the arguments which get passed to those commands - https://docs.python.org/3/library/shlex.html#shlex.quote (aka username, password).

tests/unit/util/json_util_test.py
```diff
@@ -79,6 +79,9 @@ def test_long(self):
     def test_negative_long(self):
         self.__test_encode_decode(r"-1234567890123456789", -1234567890123456789)
 
+    def test_long_long_int(self):
+        self.__test_encode_decode("18446744073709551615", 18446744073709551615)
```
Kami, line 83: This check should be fine and cover most of the scenarios, but perhaps to be on the safe we could also add explicit test case for orjson - ensure that it throws overflow error on large int and verify our wrapper works correctly (although, yeah, the second part should already be covered by this test case).

tests/utils/log_reader.py
```diff
@@ -27,77 +28,109 @@
...
+    def wait_for_next_line(self, timeout=10):
+        timeout_time = time.time() + timeout
+        while True:
+            line = self._read_line()
+            if line:
+                return line
+
+            if time.time() >= timeout_time:
+                raise LogReaderTimeoutError(
+                    "Timeout of %s seconds reached while waiting for metrics" % timeout
```
Kami, line 81: :+1:

tests/utils/log_reader.py
```diff
@@ -27,77 +28,109 @@
...
+                )
+
+            time.sleep(0.01)
+
+    def go_to_end(self):
+        while True:
+            line = self._read_line()
```
Kami, line 88: IIRC, in some scenarios, underlying ``readline()`` will block until there is data available / EOF is reached, right?

tests/utils/log_reader.py
```diff
@@ -27,77 +28,109 @@
...
+            if not line:
+                break
+
+        pass
```
Kami, line 92: This pass is redundant/

tests/utils/log_reader.py
```diff
@@ -27,77 +28,109 @@
...
+
+    def wait(self, seconds):
+        """
+        Keep checking for new lines for some period of time in 'seconds'.
+        """
+        stop_time = time.time() + seconds
+        while True:
+            try:
+                self.wait_for_next_line()
+            except LogReaderTimeoutError:
```
Kami, line 118: Should we log a message here, perhaps under debug?

tests/utils/log_reader.py
```diff
@@ -27,77 +28,109 @@
...
+                break
 
-        with self._wait_for_line_cv:
-            while not self._has_new_line:
-                if time.time() >= timeout_time:
-                    raise ValueError(
-                        "timeout of %s seconds reached while waiting for metrics"
-                        % (timeout)
-                    )
+    def add_error_check(self, pattern, message=None):
```
Kami, line 124: Please add doc strings (and type annotations) to all the new methods you added or changed.

tests/utils/log_reader.py
```diff
@@ -130,56 +157,67 @@ def _parse_line(self, line):
...
+                return needed_metrics
 
-            if time.time() >= timeout_time:
-                raise ValueError(
-                    "timeout of %s seconds reached while waiting for metrics"
-                    % (timeout)
-                )
+            try:
+                self.wait_for_next_line()
+            except LogReaderTimeoutError():
```
Kami, line 179: Perhaps also log a message under debug or similar here - something along the line of "Timeout of x seconds reached when waiting for a new line, will retry"

tests/utils/log_reader.py
```diff
@@ -130,56 +157,67 @@ def _parse_line(self, line):
...
+    ):
+        timeout_time = time.time() + timeout
+        while True:
+            try:
+                self.wait_for_next_line()
+            except LogReaderTimeoutError:
+                if time.time() >= timeout_time:
+                    raise
+                else:
+                    continue
```
Kami, line 199: Same here - let's log a message to make troubleshooting easier.

tests/utils/log_reader.py
```diff
@@ -130,56 +157,67 @@ def _parse_line(self, line):
...
-        return metrics[name]
+    def wait_for_metrics_change(
+        self, metric_names, previous_values, predicate, timeout=10
+    ):
+        timeout_time = time.time() + timeout
+        while True:
+            try:
+                self.wait_for_next_line()
+            except LogReaderTimeoutError:
+                if time.time() >= timeout_time:
```
Kami, line 196: Perhaps we should throw a different exception here (same in other places where we use the same pattern) so we can differentiate between timeout being hit on wait_for_next_line and global timeout for that method being reached.

Kami:
Once this PR is merged, can you please verify that code coverage data for postgresql monitor is correctly updated?

Kami:
It does appear it's correctly updated and reflected in sonar cloud step - https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/4135/workflows/16fac9fa-e4a0-4219-a63c-1e3b5f621ca9/jobs/54561, but for some reason codecov.io data doesn't appear to be getting updated.

I pushed a couple of commits which will hopefully shed some light on why it's not being updated.

Kami:
OK, I fixed it - issue was related to the testing change which was added to get that step to pass in the private repo :)

tests/unit/util/json_util_test.py
ArthurKamalov: Added orjson.dumps error catch assertions.

tests/utils/log_reader.py
ArthurKamalov: readline() does not block if I am not mistaken. It just returns an empty string.

tests/utils/log_reader.py
ArthurKamalov: This code part was refactored :)

tests/utils/log_reader.py
ArthurKamalov: +

tests/utils/log_reader.py
ArthurKamalov: I've added line reading based on generators, so I could remove all those timeout checks.

tests/smoke_tests/monitors_test/postgres_test.py
ArthurKamalov: Thanks for the good advice. Done.

ArthurKamalov:
@Kami part of the code was refactored and something became outdated.

tests/smoke_tests/monitors_test/mysql_test.py
```diff
@@ -153,23 +166,17 @@ def _test(request, python_version):
...
     assert row1_text == "updated_row1"
 
-    reader.wait_for_metrics_increase(metrics_to_check, previous_metric_values)
-
-    current_metrics = reader.get_metrics(metrics_to_check)
-
-    diff = dict(
-        (name, current_metrics[name] - previous_metric_values[name])
-        for name in metrics_to_check
+    reader.wait_for_metrics_equal(
```
Kami, line 169: Cool, I like this approach more :+1:

tests/smoke_tests/monitors_test/syslog_test.py
```diff
@@ -66,28 +66,27 @@ def _test(python_version):
...
 
     os.system("logger message3 --port {0} --udp --server 127.0.0.1".format(UDP_PORT))
 
-    last_line = reader.wait_for_new_line()
+    last_line = reader.wait_for_next_line()
 
     assert "message3" in last_line
 
-    runner.stop()
+    agent_log_reader.go_to_end()
```
Kami, line 89: This works for now, but since we always need to call this ``go_to_end()`` method it may make sense in the future to implement ContextManager which always calls that method on ``__exit__``.

tests/utils/log_reader.py
```diff
@@ -27,101 +32,172 @@
...
+
+    def _line_generator(self):  # type: () -> Generator
+        """Generator that reads all available lines.
+        """
+        for line in self._file:
+            line = line.strip("\n")
+            self._new_line_callback(line)
+            yield line
+
+    def _line_generator_blocking(self, timeout=10):  # type: (float) -> Generator
```
Kami, line 100: Some tests for this ``LogReader`` class functionality itself would be nice as well :)

Kami (approved):
LGTM.

When you get a chance, some tests for LogReader functionality would be nice as well.

tests/utils/log_reader.py
ArthurKamalov: There is already a file `tests/utils/tests/test_log_reader.py ` with some basic tests in it:)
Those tests are invoked by the tox target before the main tests.

tests/smoke_tests/monitors_test/syslog_test.py
ArthurKamalov: Good advice. Added reminder for that.

Kami:
@ArthurKamalov I noticed we are missing corresponding changelog entries - can you please add them.

Issue #631, opened 2020-09-22
Kubernetes monitor requires nodes/proxy authorization


mgreensmith:
Greetings, I've discovered that the kubernetes monitor can't connect to the local kubelet API `/pods` endpoint for me unless I add an additional rule of `verb=get, resource=nodes, subresource=proxy` to the `scalyr-clusterrole` ClusterRole resource. I'm running k8s 1.17 on EKS.

Without `nodes/proxy` alongside the existing `node/stats` [permission](https://github.com/scalyr/scalyr-agent-2/blob/master/k8s/scalyr-service-account.yaml#L29), the kubelet `/pods` endpoint returns a `403`. The [official docs](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/#kubelet-authorization) concur that this permission is required.

Additionally, this rabbit hole was a deep one because the scalyr agent log output about this failure is suppressed (debug only), and it doesn't log anything to `agent.log` until it attempts (and fails) to connect to the hardcoded `_fallback_kubelet_url`  on port `10255`. It was frustrating to debug this, because no matter which kubelet-related config vars I changed (cert paths, template URIs, etc), it kept telling me that it couldn't connect to the fallback URI, and gave me no information about the failures to the primary kubelet URI. Maybe you could log [this failure](https://github.com/scalyr/scalyr-agent-2/blob/c89ab9dcba6d772313c09dbd897cdb102bc0236f/scalyr_agent/monitor_utils/k8s.py#L2376) at INFO instead?

Thanks!

weilliu:
@mgreensmith I couldn't reproduce the issue on my end. I just sent you an email with additional information on how I deploy the agent and verify it. Let me know if you have any questions.

Kami:
@mgreensmith I'm looking into it this, but I'm also having troubles reproducing it.

Can you please provide any additional information or custom cluster configuration which may help us to reproduce this issue?

Having said that - I agree that the current user-experience when hitting an error is not the best so I made the change you suggested above - #633. Now we log those errors under warning log level and also include a response body.

PR #633, opened 2020-10-02
Kubernetes monitor improvements

Kami:

1. Make sure we log any errors we receive from Kubernetes API under warn log level and also include actual response body. This should make it easier to troubleshoot various configuration and permission related issue - as suggested in #631.

2. While trying to reproduce #631, I encountered weird behavior when trying to stop the agent before the Kubernetes monitor has been fully initialized (e.g. if we receive an error from the API or we fail to connect to it).

In such scenario, CTRL+C would have no effect - the reason for that is that the "stop" signal was not correctly propagated to all the underlying objects and threads.

My change is probably not 100% ideal, but based on my testing, it works. The problem is that the current code is quite convoluted so doing any other major refactoring would be more risky.

I'm also looking into adding a test case for it, but it's not totally straight forward.

## TODO

- [x] Tests for stop process edge case fix


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2734,12 +2748,16 @@ def start(self):
...
     def stop(self, wait_on_join=True, join_timeout=5):
+        self.__stopped = True
+
         if self.__controlled_warmer is not None:
             self.__controlled_warmer.stop(
                 wait_on_join=wait_on_join, join_timeout=join_timeout
             )
 
-        self.__thread.stop(wait_on_join=wait_on_join, join_timeout=join_timeout)
+        if self.__thread.isAlive():
```
Kami, line 2758: This is needed because threads might not have been started yet when stop is called.

```bash
2020-10-02 14:00:26.258Z ERROR [core] [scalyr_logging.py:618] Failed to stop the metric log due to an exception :stack_trace:
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/kubernetes_monitor.py", line 4195, in stop
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/kubernetes_monitor.py", line 2746, in stop
      self.__thread.stop(wait_on_join=wait_on_join, join_timeout=join_timeout)
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 1226, in stop
      self.join(join_timeout)
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 1248, in join
      threading.Thread.join(self, timeout)
    File "/home/kami/.pythonz/pythons/CPython-3.6.9/lib/python3.6/threading.py", line 1051, in join
      raise RuntimeError("cannot join thread before it is started")
  RuntimeError: cannot join thread before it is started
```

yanscalyr (approved):
LGTM

PR #635, opened 2020-10-06
Add a workaround / mitigation for a possible race condition which can occur during concurrent "scalyr-agent status" command executions

Kami:
This pull request includes a fix / workaround for an issue reported by a user.

## Problem Description, Background

Every now and then the user would see this error:

> Caught exception when attempt to execute command status.  Exception was [Errno 2] No such file or directory: '/var/lib/scalyr-agent-2/last_status'

After digging into the code, I noticed there is a possible race condition if ``scalyr-agent status`` command is executed multiple times concurrently or around the same time.

I was also able to reproduce the issue locally by trying to force a race condition using (aka multiple executions around the same time):

>   FileNotFoundError: [Errno 2] No such file or directory: '/home/kami/scalyr-agent-dev/data/last_status.tmp' -> '/home/kami/scalyr-agent-dev/data/last_status'
2020-10-06 14:18:55.952Z ERROR [core] [scalyr_logging.py:618] [error="failedStatus"] Exception caught will try to report status :stack_trace:
    File "scalyr_agent/agent_main.py", line 1927, in __report_status_to_file
      os.rename(tmp_file_path, final_file_path)
  FileNotFoundError: [Errno 2] No such file or directory: '/home/kami/scalyr-agent-dev/data/last_status.tmp' -> '/home/kami/scalyr-agent-dev/data/last_status'

I believe the issue was already there in the past, but adding health check option to the status command which likely gets executed more frequently (and perhaps around the same time some other script calls scalyr-agent status or similar) makes it more likely to hit this issue.

Especially when mixing text and json based status formats.

## Proposed Fix / Workaround

As per the code comment, only 100% correct and robust fix would be to use a global agent status file based read write lock and only allow a single agent status invocation at the same time.

Problem is that this approach is a bit more involved - we support multiple platforms, etc. There are libraries there for handling file based locks and I wrote that code in the past before, but I would like to avoid this, at least for now.

Instead, I implemented a simple mitigation / workaround - in case we receive "no such file or directory" error when trying to move a ``.tmp`` file, and actual final ``last_status`` file already exists, we simply ignore that error since it means file was already moved by other invocation of agent status.

In addition to that, even though ``-H`` should imply JSON based status format, we also accept and handle non JSON format - this could happen when mixing / running concurrently status -v and status -H command (or status -v and status -v --format=json command).

Another more robust approach here would be to use different files (e.g. ``last_status.text`` and ``last_status.json``), but this again complicates things and we would still need to introduce locking for 100% corectness.

## TODO

- [x] Tests

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -279,7 +279,10 @@ class ScalyrAgent(object):
         self.__config_file_path = config_file_path
 
         try:
-            self.__config = self.__read_and_verify_config(config_file_path)
+            log_warnings = command not in ["status"]
+            self.__config = self.__read_and_verify_config(
+                config_file_path, log_warnings=log_warnings
+            )
 
             # check if not a tty and override the no check remote variable
             if not sys.stdout.isatty():
@@ -365,27 +368,35 @@ class ScalyrAgent(object):
                     % (command, six.text_type(e), traceback.format_exc())
                 )
 
-    def __read_and_verify_config(self, config_file_path):
+    def __read_and_verify_config(self, config_file_path, log_warnings=True):
         """Reads the configuration and verifies it can be successfully parsed including the monitors existing and
         having valid configurations.
 
         @param config_file_path: The path to read the configuration from.
         @type config_file_path: six.text_type
 
+        @param log_warnings: True if config.parse() should log any warnings which may come up during
+                             parsing.
+        @type log_warnings: bool
+
         @return: The configuration object.
         @rtype: scalyr_agent.Configuration
         """
-        config = self.__make_config(config_file_path)
+        config = self.__make_config(config_file_path, log_warnings=log_warnings)
         self.__verify_config(config)
         return config
 
-    def __make_config(self, config_file_path):
+    def __make_config(self, config_file_path, log_warnings=True):
         """Make Configuration object. Does not read nor verify.
 
         You must call ``__verify_config`` to read and fully verify the configuration.
 
         @param config_file_path: The path to read the configuration from.
         @type config_file_path: six.text_type
+
+        @param log_warnings: True if config.parse() should log any warnings which may come up during
+                             parsing.
+        @type log_warnings: bool
 
         @return: The configuration object.
         @rtype: scalyr_agent.Configuration
@@ -395,6 +406,7 @@ class ScalyrAgent(object):
             self.__default_paths,
             log,
             extra_config_dir=self.__extra_config_dir,
+            log_warnings=log_warnings,
         )
 
     def __verify_config(
@@ -409,6 +421,7 @@ class ScalyrAgent(object):
 
         @param config: The configuration object.
         @type config: scalyr_agent.Configuration
+
         @return: A boolean value indicating whether or not the configuration was fully verified
         """
         try:
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -630,7 +630,11 @@ class ScalyrAgent(object):
             self.__run_state.stop()
 
     def __detailed_status(
-        self, data_directory, status_format="text", health_check=False
+        self,
+        data_directory,
+        status_format="text",
+        health_check=False,
+        zero_content=True,
     ):
         """Execute the status -v or -H command.
 
@@ -639,6 +643,11 @@ class ScalyrAgent(object):
 
         @param data_directory: The path to the data directory.
         @type data_directory: str
+
+        :param zero_content: True to zero the content so we can detect when agent writes a new
+                             content. This is primary meant to be used in testing where we can
+                             set it to False which means we can avoid a lot of nasty mocking if
+                             open() and related functiond.
 
         @return:  An exit status code for the status command indicating success or failure.
         @rtype: int
@@ -696,7 +705,7 @@ class ScalyrAgent(object):
             return 1
 
         # Zero out the current file so that we can detect once the agent process has updated it.
-        if os.path.isfile(status_file):
+        if os.path.isfile(status_file) and zero_content:
             f = open(status_file, "w")
             f.truncate(0)
             f.close()
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -634,7 +634,7 @@ class ScalyrAgent(object):
         data_directory,
         status_format="text",
         health_check=False,
-        zero_content=True,
+        zero_status_file=True,
     ):
         """Execute the status -v or -H command.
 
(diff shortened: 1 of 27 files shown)


Kami:
I confirmed that with those changes, the error doesn't appear in the agent log anymore and the health check result is always correctly reported.

But as mentioned above, this is still a workaround / mitigating and not a 100% robust and correct solution which would involve locking and making sure that the only single signal handler can be running at once / status can only be invoked by process at a time.

scalyr_agent/agent_main.py
```diff
@@ -757,40 +757,52 @@ def __detailed_status(
...
+
+        if not health_check:
+            # Regular non-health check invocation, just print the stats
+            print(content)
+            return return_code
+
+        # Health check invocation, try to parse status from the report and print and handle it here
+        health_result = self.__find_health_result_in_status_json(content)
+
+        if health_result != "Good":
```
yanscalyr, line 772: Do we need this check here considering we have `if health_result in ["good", "bad"]:` right below it? Also since `__find_health_result_in_status_json` returns `.lower()`ed strings can we ever hit this?

yanscalyr (changes requested):

scalyr_agent/agent_main.py
Kami: I pushed more refactoring so that code is now out of update, please have a look again.

yanscalyr (approved):
LGTM

scalyr_agent/agent_main.py
```diff
@@ -757,40 +757,54 @@ def __detailed_status(
...
-                return_code = 2
-        fp.close()
+
+        with open(status_file, "r") as fp:
+            content = fp.read()
+
+        if not health_check:
+            # Regular non-health check invocation, just print the stats
+            print(content)
+            return return_code
```
czerwingithub, line 767: I believe you aren't covering a case here.  We will return `0` for `return_code` if the status format is `json` and `health_check` is not True.  That's a change in behavior.  We are always trying to return a status code based on whether or not the agent is healthy.

scalyr_agent/agent_main.py
```diff
@@ -1925,7 +1944,19 @@ def __report_status_to_file(self):
...
+            # If agent status or health check is requested multiple times or concurrently around the
+            # same time, it's likely the race will occur and rename will fail because the file  was
+            # already renamed by the other process.
+            # This workaround is not 100%, only 100% solution is to use a global read write lock
+            # and only allow single invocation of status command at the same time.
+            if tmp_file is not None:
+                tmp_file.close()
+
+            if os.path.isfile(final_file_path):
+                return final_file_path
```
czerwingithub, line 1958: Is returning something here right?  I don't think this method has a return value.

czerwingithub (changes requested):

scalyr_agent/agent_main.py
Kami: Yep, that method does have a return value - https://github.com/scalyr/scalyr-agent-2/blob/f952e8fbeae7d6937245feb210ca207f8df91233/scalyr_agent/agent_main.py#L1971.

On that note, that's where type annotations and MyPy would also come handy - if method didn't declare a return, but we returned something, mypy would catch that.

scalyr_agent/agent_main.py
Kami: Ah, that's a good catch - I forgot we want to do that (and it doesn't appear we have any tests for it either).

I will make that change and also look into adding some tests to avoid such accidental regressions in the future.

I will probably start with a simple mock based unit tests, but once I have some more time, I will work on end to end tests (those will take more work since some of those things will be a bit hard to test in the end to end context).

Kami: Here we go - d66c36a6f7cb57d67a2498c74a1eb02379da21fb.

Now there are also test cases which cover all the possible scenarios (which are quite many) - b7c01042b511492dea3c2ffe0b8f5eedb9793e00.

As mentioned above, those are not full blown end to end tests (those would be better since we would assert on the actual exit code, but that would require more work and it's on TODO for future improvements). It's still better than what we had before though (no tests for those scenarios).

A couple of things worth noting:

1. Added new argument to the method which is only used for testing purposes. I almost always try to avoid that (especially in dyamic languages like Python where it's usually easy to patch thing), but this is a bit special case. We could patch ``open`` method, but it would get much more messy and require much more glue code since we perform many different operations on the open call and open is also used as context manager...

2. We now return 3 if case we can't get health check data and -H flag is used - imo that's desired and better than previous behavior when we would return 0 in such case - https://github.com/scalyr/scalyr-agent-2/blob/master/scalyr_agent/agent_main.py#L772. This scenario should only really happen if status file for some reason is missing status data (aka running old version of the agent with new version of status command or similar - unlikely). We exit with 0 if we can't get health check data is -H flag is not used (same as before)

3. Those tests are intentionally a bit on the "copy paste" side - I could reduce some code by changing it a bit and using a loop. But in the tests, I tend to be careful - many times I've seen broken / invalid tests, especially "dynamic" ones which utilizes loops and similar.

czerwingithub (approved):

Kami:
While testing this some more, I noticed another potential issue:

```bash
(scalyr-agent-py3) kami ~/w/scalyr/scalyr-agent-2 (git:agent_status_possible_race_mitigation_workaround)$  python scalyr_agent/agent_main.py status -H
2020-10-07 08:49:07.130Z WARNING [core] [configuration.py:343] Configured option max_log_offset_size is being overridden due to max_send_rate_enforcement setting.
Health check: Good
```

So depending on the configuration, this override warning would also get printed when running agent status command.

Imo, that's not desired nor a good idea (it should only be logged in the main agent process) since it may break any script or similar which relies on parsing the output of the status command (status output and log messages are both output to stdout).

/cc @yanscalyr

scalyr_agent/configuration.py
```diff
@@ -238,7 +245,8 @@ def parse(self):
 
             # Set defaults based on `max_send_rate_enforcement` value
             if (
-                not self.__config["disable_max_send_rate_enforcement_overrides"]
+                self.__log_warnings
```
yanscalyr, line 248: This path includes the setting of the default values which we probably don't want to turn off along with the warnings, line 271 for example: `self.__config["max_allowed_request_size"] = 5900000`. If I remember correctly `_warn_of_override_due_to_rate_enforcement` only conditionally logs a warning so putting this check in there might be a better place. Setting these values probably won't matter in the case of requesting a status but better that the option did what it said on the tin (in that vein maybe `_warn_of_override_due_to_rate_enforcement` could do with a rename too).

yanscalyr (changes requested):

scalyr_agent/configuration.py
Kami: Yeah, like like you said - it shouldn't really matter since those settings really only apply to main agent process and have no affect on the status command itself.

Having said that, I can change it and put that logic in the __warn functions.

Tests would also be nice, but as mentioned above, for that we would need to end to end tests which will take more work.

Kami: b559eefab6d3f83f10264c0cff9e66dc9f63a85a

yanscalyr (approved):
LGTM

PR #637, opened 2020-10-07
CT-193: Sequence range start must always be less (not equal) to Sequence range end

yanscalyr:
https://scalyr.myjetbrains.com/youtrack/issue/CT-193
I don't believe the server side changes are done or fully decided on so this may still need to be changed.

### DIFF
--- a/scalyr_agent/log_processing.py
+++ b/scalyr_agent/log_processing.py
@@ -2021,6 +2021,7 @@ class LogFileProcessor(object):
         """
         self._new_scalyr_client = new_scalyr_client
         self._log_stream = None
+        self._last_sequence_end_number = -1
         if config.use_new_ingestion:
             from scalyr_ingestion_client.log_stream import (  # pylint: disable=import-error
                 LogStream,
@@ -2350,7 +2351,6 @@ class LogFileProcessor(object):
             added_thread_id = False
 
             new_events_buffer = []
-            sequence_start_number = 0
             sequence_end_number = 0
 
             # Keep looping, add more events until there are no more or there is no more room.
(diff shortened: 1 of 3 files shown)


scalyr_agent/log_processing.py
```diff
@@ -2417,12 +2417,10 @@ def perform_processing(self, add_events_request, current_time=None):
                     if self._log_stream:
                         import scalyr_ingestion_client.log_line as ingestion_client_line  # pylint: disable=import-error
 
-                        if not new_events_buffer:
-                            sequence_start_number = sequence_number
                         sequence_end_number = sequence_number
                         new_event_timestamp = line_object.timestamp
                         if not new_event_timestamp:
-                            new_event_timestamp = int(current_time)
+                            new_event_timestamp = int(current_time * 1000000000)
```
Kami, line 2423: You could also use new ``scalyr_ingestion_client.time_utils.get_nanoseconds_from_epoch`` method.

Kami, line 2423: Per Steven's comment on the other PR - we only want to set ``timestamp`` attribute on the ``LogLine`` if it's different from a timestamp on the previous log line.

Aka since we periodically poll for new data / lines, all the lines we read as part of a single poll will contain the same timestamp so we will only need to set it on the first ``LogLine`` object in that group.

yanscalyr: That method would get us a slightly different timestamp for each event as we processed through them, if we have no parsed timestamp to work with the expected behavior is that we would have the same timestamp for every event right? Using that method elsewhere in the function to get a nanosecond current time would also require another conditional import so I think using the existing `current_time` is our best bet, though I did move this calculation to the beginning fo the method.

scalyr_agent/log_processing.py
```diff
@@ -2272,6 +2273,7 @@ def perform_processing(self, add_events_request, current_time=None):
         """
         if current_time is None:
             current_time = time.time()
+        current_time_nano = int(current_time * 1000000000)
```
Kami, line 2276: Can we use ``get_nanoseconds_since_epoch`` from the new client library here? (will need to use a conditional import)

yanscalyr: If you're okay with another conditional import here then yes, I wanted to avoid it since for something this simple its very ugly but I'm not married to that idea.

PR #638, opened 2020-10-12
Fix `read_checkpoint_state`

anton-ryzhov:
Current code updates `full_checkpoints` rather than `full_checkpoints["checkpoints"]` — active checkpoints data is ignored in fact

### DIFF
--- a/scalyr_agent/copying_manager.py
+++ b/scalyr_agent/copying_manager.py
@@ -1176,10 +1176,9 @@ class CopyingManager(StoppableThread, LogWatcher):
 
             if active_checkpoints["time"] > full_checkpoints["time"]:
                 full_checkpoints["time"] = active_checkpoints["time"]
-                for path, checkpoint in six.iteritems(
-                    active_checkpoints["checkpoints"]
-                ):
-                    full_checkpoints[path] = checkpoint
+                full_checkpoints["checkpoints"].update(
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the contribution / bug fix.

The fix indeed seems correct to me, but since the code is quite critical it needs a deeper dive and tests (I plan to do a deeper dive and work on a tests soon).

Kami:
Thanks for the contribution / bug fix.

The fix indeed seems correct to me, but since the code is quite critical it needs a deeper dive and tests (I plan to do a deeper dive and work on a tests soon).

Kami:
It appears that this change has been there for a while - a02e534c8a1efd797450863599e9034cc1d40465.

If I understand this code correctly, because of this bug, it's possible that the agent would re-send some of the data which has already sent if it crashed (or similar) after the active check point file was written, but before a full checkpoint file has been written.

Since default value for ``full_checkpoint_interval_in_seconds`` config option is 60 second, that window is quite small, but still something which indeed needs to be fixed.

anton-ryzhov:
>  it's possible that the agent would re-send some of the data which has already sent if it crashed (or similar) after the active check point file was written, but before a full checkpoint file has been written

Yes, that's what I saw

anton-ryzhov:
Fixed in https://github.com/scalyr/scalyr-agent-2/pull/639

PR #639, opened 2020-10-12
Fix merging of active checkpoint data into full checkpoint data on full checkpoint write

Kami:
This pull request builds on top of #638 and adds mock based unit tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1159,7 +1159,7 @@ jobs:
           command: |
             docker container create --name dummy -v shared_vol:/app alpine && \
             docker cp $(pwd)/.circleci/smoketest_standalone.sh dummy:/app/ && \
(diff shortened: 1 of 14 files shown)


Kami:
I've added mock based unit tests to begin with, but perhaps it would also be good to add end to end tests (that may be a bit harder and in the end we will still need to mock some data).

Kami:
Re test failures - per discussion on Slack those appear to be related to a server-side / API change.

Our verifier script doesn't escape / url quote ``"`` and ``$`` character and so the check / API request fails.

For now, I will go ahead and make the changes to that script so we correctly url escape / quote all the special characters.

Kami:
I updated our verifier script so it correctly url encodes double quotes, but it turned to be a bit of a rabbit hole.

I also added test case for that one method, but then encounter a bunch of other related issues since that script wasn't working under Python 2, etc.

Of course after fixing Python 2 compatibility, other "random" unrelated issues showed up which I also needed to fix...

Good news is that url quoting of ``"`` does seem to fix the smoke test failures.

czerwingithub (approved):
No comments from me.  It does look to me like this change is pretty safe to merge in.  Thanks for fixing all of the issues that cropped up due to the change in how we handle the API calls.

PR #643, opened 2020-10-20
More debugging statements to help with CT-207

yanscalyr:
The intent of this PR is to help others review the contents of a Debug build we will give to the customer, though if the changes end up being just more DEBUG level logs this might make sense to merge in for future issues.


czerwingithub (commented):
Do you have a sense for how much debug output we will be generating?  It would be good to understand it in terms of MB / minute.

PR #644, opened 2020-10-23
Kubernetes monitor INFO level logging improvements

Kami:
This pull request includes the following logging changes to the Kubernetes monitor:

1. "Cluster name detected" message is now logged under INFO log level
2. Monitor start up info log line now also containers the value for the following monitor config options: ``log_mode``, ``report_k8s_metrics``
3. On monitor start up we now also print the values of the following environment variables: ``SCALYR_REPORT_CONTAINER_METRICS``, ``SCALYR_REPORT_K8S_METRICS``

Those changes may make it a bit easier and faster for us to troubleshoot some user issues in the future since it may allow us to skip asking the user for their config file in some scenarios.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -3458,6 +3458,7 @@ cluster.
         self.__glob_list = self._config.get("container_globs")
         self.__include_all = self._config.get("k8s_include_all_containers")
 
+        self.__log_mode = self._config.get("log_mode")
         self.__report_container_metrics = self._config.get("report_container_metrics")
         self.__report_k8s_metrics = (
             self._config.get("report_k8s_metrics") and self.__report_container_metrics
@@ -3501,7 +3502,7 @@ cluster.
         )
 
         self.__container_checker = None
-        if self._config.get("log_mode") != "syslog":
+        if self.__log_mode != "syslog":
             self.__container_checker = ContainerChecker(
                 self._config,
                 self._global_config,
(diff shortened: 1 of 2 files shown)


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -4156,11 +4159,16 @@ def run(self):
             self.__report_k8s_metrics = False
 
         global_log.info(
-            "kubernetes_monitor parameters: ignoring namespaces: %s, report_controllers: %s, report_metrics: %s"
+            (
+                "kubernetes_monitor parameters: ignoring namespaces: %s, report_controllers: %s, "
+                "report_metrics: %s, report_k8s_metrics: %s, log_mode: %s"
```
Kami, line 4164: I added new values to the end, just in case someone has custom parser defined for that log line (highly unlikely).

czerwingithub: If we are going to centralize this stuff, it might be a good idea to also record here if we are using docker or CRI to get a list of containers.

The line I was referring to is at: https://github.com/scalyr/scalyr-agent-2/blob/master/scalyr_agent/builtin_monitors/kubernetes_monitor.py#L2649

Essentially, if we aren't in syslog mode, then there are still two different ways we get containers.  Either using Docker or using a filesystem based CRI method.  If we aren't using Docker, than I don't think we'll get metrics.

Kami: Thanks, will update it to also log that.

Kami: Pushed a change so we now also log that value.

czerwingithub (approved):

PR #645, opened 2020-10-26
Update syslog monitor so we allow messages which span multiple TCP packets and exceed "tcp_max_buffer" in size when "tcp_unlimited_buffer_size" monitor config option is true

Kami:
This pull request updates syslog monitor so it supports messages / frames which potentially span multiple TCP packets and exceed ``tcp_max_buffer`` in size.

## Background, Context

Currently when running syslog monitor in TCP mode, all the message / frame data needs to fit in a single TCP packet which can be a maximum of ``tcp_max_buffer`` bytes in size.

In general, for efficiency reasons, it's good to set this option to a high value, but there could still be a scenario where we have a frame or a messages which is larger than`tcp_max_buffer``.

In the current version, this would cause agent to log a warning and dump any partial frame / message bytes which have been accumulated to far.

## TODO

- [x] Unit tests
- [x] End to end tests
- [x] Tests with actual syslog log stream
- [ ] ~~Implement safe guards (issue a warning and dump contents if we failed to receive full frame in X seconds or if currently accumulated frame / message size exceeds some reasonable pre-defined value)~~
- [x] Should we also change a default value to something more reasonable - e.g. at least 8K?


Kami:
As discussed, I decided to change the approach a bit and made the following changes:

1. Changes the default value for ``tcp_buffer_size`` from 2k to 8k (reasonable and good default value change)

2. Added new ``tcp_unlimited_buffer_size`` config option. When this config option is true (defaults to false for backward compatibility reasons), the monitor supports syslog messages / frames split across arbitrary number of TCP packets. In this scenario, ``tcp_buffer_size`` setting only tells the monitor how many bytes we try to read from a socket at once and doesn't serve as a hard limit (which is the case when unlimited buffer size is false).

3. Improved warning message which is logged when ``tcp_buffer_size`` limit is reached

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1140,6 +1159,7 @@ def handle(self, data):  # type: (six.text_type) -> None
         else:
             self.__logger.info(data)
         # We add plus one because the calling code strips off the trailing new lines.
+        print("innn")
```
ArthurKamalov, line 1162: Missed debug print, I guess.

ArthurKamalov (approved):
LGTM

scalyr_agent/builtin_monitors/syslog_monitor.py
Kami: whoops 

removed

czerwingithub (changes requested):

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -128,11 +128,20 @@
...
     "support messages up to 8192 bytes long.",
     default=8192,
-    min_value=2048,
     max_value=65536 * 1024,
     convert_to=int,
 )
 
+define_config_option(
+    __monitor__,
+    "tcp_unlimited_buffer_size",
```
czerwingithub, line 137: Yeah, this name might not be the best because TCP/IP buffer size probably means something very specific to folks.

How about these as alternatives:
1. `limit_line_size_to_tcp_buffer`
2. `allow_message_to_exceed_tcp_buffer`
3. `message_size_can_exceed_tcp_buffer`
4. `lines_can_span_tcp_buffers`

Ok, I'm just rattling off combinations at this point.  See if any of those suggestions work for you.

PR #646, opened 2020-10-27
CT-207: Configurable time range to accept old and dead k8s containers on monitor startup

yanscalyr:
In autoscaling k8s environments it's not impossible that a pod will start and fail before the k8s monitor is able to start up and catch it, meaning those logs will not be uploaded. This change will make the monitor accept dead pods of a configurable age to have their logs uploaded on the first iteration of the container checking loop, aka when it starts up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Improvements:
 * Improve logging in the Kubernetes monitor.
 
 Features:
-* Add new ``initial_stopped_container_collection_window`` configuration option to the kubernetes monitor, which can be configured by setting the ``SCALY_INITIAL_STOPPED_CONTAINER_COLLECTION_WINDOW`` environment variable. By default, the Scalyr Agent does not collect the logs from any pods stopped before the agent was started. To override this, set this parameter to the number of seconds the agent will look in the past (before it was started). It will collect logs for any pods that was started and stopped during this window.
+* Add new ``initial_stopped_container_collection_window`` configuration option to the Kubernetes monitor, which can be configured by setting the ``SCALY_INITIAL_STOPPED_CONTAINER_COLLECTION_WINDOW`` environment variable. By default, the Scalyr Agent does not collect the logs from any pods stopped before the agent was started. To override this, set this parameter to the number of seconds the agent will look in the past (before it was started). It will collect logs for any pods that was started and stopped during this window. This can be useful in autoscaling environments to ensure all pod logs are captured since node creation, even if the Scalyr Agent daemonset starts just after other pods.
 
 ## 2.1.13 "Celaeno" - October 15, 2020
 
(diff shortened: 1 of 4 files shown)


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -102,6 +102,16 @@
     env_aware=True,
 )
 
+define_config_option(
+    __monitor__,
+    "first_time_old_container_window",
```
czerwingithub, line 107: I don't think "old" container is really descriptive here.  I would suggest calling this option "initial_stopped_container_collection_window".  I know it's wordy, but I'd definitely like to use the word "stopped" instead of "old".

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -102,6 +102,16 @@
...
+    "Optional (defaults to 0). For the first container check, how old a dead container can be and still have its "
```
czerwingithub, line 108: Let's rephrase this documentation to: "By default, the Scalyr Agent does not collect the logs from any pods stopped before the agent was started.  To override this, set this parameter to the number of seconds the agent will look in the past (before it was started).  It will collect logs for any pods that was started and stopped during this window."

Note, I wrote the comment assuming the window compares the pod start time, which I believe is the case.  If a pod is started before this window but stopped during the window, I believe we still won't collect the logs.  Which is fine, we just have to be clear about it.

czerwingithub (changes requested):
Also, be sure to add an entry for this in the change log for the next release.  Just FYI, I just approve a change for Tomaz where he added in the initial entry for the next release.

czerwingithub (commented):
Ok, everything essentially looks good, but please add an entry to this in the `CHANGELOG.md` file.  Be specific about what it does and reference the environment variable name they need to set (for K8s, we want to encourage customers to use the environment variable).

yanscalyr:
CHANGELOG update is more or less just the configuration option description you posted earlier with a mention of the environment variable. I couldn't think of a better or more clear way to express it.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Packaged by Oliver Hsu <oliver@scalyr.com> on Nov 15, 2020 19:00 -0800
 Improvements:
 * Improve logging in the Kubernetes monitor.
 
+Features:
+* Add new ``initial_stopped_container_collection_window`` configuration option to the kubernetes monitor, which can be configured by setting the ``SCALY_INITIAL_STOPPED_CONTAINER_COLLECTION_WINDOW`` environment variable. By default, the Scalyr Agent does not collect the logs from any pods stopped before the agent was started. To override this, set this parameter to the number of seconds the agent will look in the past (before it was started). It will collect logs for any pods that was started and stopped during this window.
```
czerwingithub, line 14: Capitalize Kubernetes.

czerwingithub, line 14: I would end comment by saying. "This can be useful in autoscaling environments to ensure all pod logs are captured since node creation, even if the Scalyr Agent daemonset starts just after other pods."

czerwingithub (changes requested):
Ok, just two small nits.

czerwingithub (approved):

PR #647, opened 2020-10-27
Docker monitor "docker_raw_logs: false" and related bug fixes

Kami:
This pull request includes various fixes to the ``docker_raw_logs: false`` mode functionality in the Docker monitor.

## Background, Context

While troubleshooting another issue, I decided to use / test ``docker_raw_logs: false`` functionality of the Docker monitor (using direct mounts for docker containers is a bit of a pain in Fedora due to selinux, etc.).

It turns out that functionality was quite badly broken in many different way.

I assume it has been broken since the time we did Python 2 -> 3 switch and moving from docker-py to docker Python library (aka upgrading the library version).

Some of the issues I found:

1. Our wrapper classes didn't expose iterator functionality so that code was broken and raised an exception out of the box

```bash
2020-10-26 14:51:04.882Z WARNING [monitor:docker_monitor] [docker_monitor.py:1596] Unhandled exception in DockerLogger.process_request for web:
	'WrappedRawResponse' object is not an iterator
```

In addition to that, that piece of code was broken in many more ways - we passed ``response`` and ``decode`` as regular and not keyword arguments to the parent method. 

That parent method actually takes three arguments (response, chunk_size, decode) and decode argument is actually 3rd and not second one which means the code would incorrectly interpret it as a chunk size instead of decode argument, etc.

2. ``sock.shutdown()`` function is not available

```bash
2020-10-26 15:34:55.368Z INFO [monitor:linux_process_metrics(agent)] [scalyr_monitor.py:292] Monitor has finished
2020-10-26 15:34:55.368Z ERROR [core] [scalyr_logging.py:618] Failed to stop the metric log due to an exception :stack_trace:
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/docker_monitor.py", line 2477, in stop
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/docker_monitor.py", line 1044, in stop
      logger.stop(wait_on_join, join_timeout)
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/docker_monitor.py", line 1531, in stop
      sock.shutdown(socket.SHUT_RDWR)
  AttributeError: 'SocketIO' object has no attribute 'shutdown'
```

3. Code will never ingest new container log lines in case the timezone where the agent is running is not UTC (well, to be exact, in case timezone is > UTC so likely a person in the US wouldn't spot this issue)

At first I was a bit puzzled on why logs don't make it in even after fixing other things.

Thing got apparent after printing out the timestamp - most of the code correctly and intentionally operates with UTC, but the code which retrieves last request for log (``__get_last_request_for_log()``) incorrectly calls ``datetime.datetime.fromtimestamp`` which returns local and not UTC time.

I also included some logging improvements and other fixes and improvements (potentially leaking open file handles on some errors, etc.).

## TODO

- [x] Tests (I need to think of a good way to approach those tests which won't take a week since there are a lot of scenarios we want to test which are relatively hard to replicate in automated tests)

## Open Issues / To decide

- [x] I grepped the code and I see more pieces of code which utilize ``datetime.fromtimestamp`` instead of ``datetime.utcfromtimestamp``. Likely the code which utilizes that function exhibits similar issues / bugs when the agent is running on a server where local time / timezone is not UTC (likely not in that many scenarios since most people run servers in UTC, which is really the correct thing to do, but still).

- [x] There are some edge cases where don't gracefuly handle 404 erors which could get throw by the client library streaming methods in case container is deleted.

```bash
2020-10-27 10:08:58.911Z WARNING [monitor:docker_monitor] [docker_monitor.py:1631] Unhandled exception in DockerLogger.process_request for web:
	404 Client Error: Not Found ("No such container: a8462a150e8cdbd2c246989421110eaa7fbeccf423cdbe67c137a252232c8eb7").

  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/api/client.py", line 261, in _raise_for_status
    response.raise_for_status()
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/third_party/requests/models.py", line 934, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http+docker://localhost/v1.40/containers/a8462a150e8cdbd2c246989421110eaa7fbeccf423cdbe67c137a252232c8eb7/json

During handling of the above exception, another exception occurred:

  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/docker_monitor.py", line 1554, in process_request
    follow=True,
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/api/container.py", line 855, in logs
    output = self._get_result(container, stream, res)
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/api/client.py", line 451, in _get_result
    return self._get_result_tty(stream, res, self._check_is_tty(container))
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/api/client.py", line 447, in _check_is_tty
    cont = self.inspect_container(container)
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/api/container.py", line 758, in inspect_container
    self._get(self._url("/containers/{0}/json", container)), True
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/api/client.py", line 267, in _result
    self._raise_for_status(response)
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/api/client.py", line 263, in _raise_for_status
    raise create_api_error_from_http_exception(e)
  File "/home/kami/.virtualenvs/scalyr-agent-py3/lib/python3.6/site-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation)
docker.errors.NotFound: 404 Client Error: Not Found ("No such container: a8462a150e8cdbd2c246989421110eaa7fbeccf423cdbe67c137a252232c8eb7")
```

We have additional safe guard later on which will still cause that log file to be removed (aka on retry), but we should still handle this edge case better.

- [x] Is this still relevant - https://github.com/scalyr/scalyr-agent-2/blob/master/scalyr_agent/builtin_monitors/docker_monitor.py#L268?


Kami:
I added new smoke tests framework based end to end tests called ``docker-smoke-api`` which test the ``docker_raw_logs: false`` code path end to end.

It appears we had no (either unit or end to end) tests for that functionality before hand, so no wonder it was broken for so long (please correct me if I'm wrong).

Kami:
Good to see code coverage stuff for all the gazillion jobs we have is indeed working correctly (previously a lot of that non raw log mode code was uncovered):

Screenshot from 2020-10-27 18-10-43

Kami:
Just to be on the safe side, I reverted the fix and verified the new tests I added indeed correctly fail with the fix not in place - https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/4292/workflows/1456df36-f0a5-4002-99dd-019072fc367e/jobs/57112.

ArthurKamalov (approved):
Everything looks good to me as far as I can see.

Many things were missed during the transitioning to python3... Thanks for finding them.

PR #648, opened 2020-10-27
Fix a bug / race-condition with scheduled log removal in Docker monitor

Kami:
This pull request fixes a bug / race-condition in Docker monitor when monitoring containers on the same host which could cause, in some scenarios (more on that below), for container logs to stop being ingested after the container restart.

## Background, Context

There was a race condition in the Docker monitor which could cause container logs to stop being ingested, under some scenarios, after container restart.

If our code queried Docker API for running containers just after the container was stopped and before it was started back it, up it would determine that container to be dead and mark / schedule that container log file for removal.

Since we didn't delete the log file immediately, but scheduled it for removal (so we could still upload any data which was not uploaded yet), it could happen that the container would come back up before the scheduled removal was actually performed which means that ``add_log_config`` method call for the started container would not add the log file, but instead return the message that the log file is already being monitored.

Do keep in mind that the time window / frame when this could happen was relatively small - container checker code would need to run just after stopping the container, but before starting it up. With the default interval of ``5`` seconds, that was not all that likely to happen.

---

Special thanks to @ArthurKamalov for the initial investigation.

It took me quite a while to reproduce the issue, but mostly due to encountering other bugs on the way, which I was working on fixing (#647).

In addition to that, to make the race condition more likely to appear, I needed to set ``container_check_interval`` monitor config option to ``0.2`` seconds (there is a relatively short time window when this bug can manifest itself - if we query for running containers after the container has been stopped, but before it has started back up again).

## Proposed Fix

There are multiple possible approaches to fix the issue.

I decided to with the one which I think it's the safest and simplest, but it would be good if someone more familiar with that code (most likely @czerwingithub) would confirm that this approach is indeed safe and that I'm not missing anything.

In my proposed fix, I add new ``force_add`` argument to ``add_log_config`` method (open to suggestions for a better argument name - e..g ``cancel_scheduled_removal`` or similar) and if this argument value is ``True``, we will simply remove that log path from log paths which are scheduled to be removal and continue monitoring it.

We only pass ``force_add=True`` in the Docker monitor which means the change is very scope and doesn't affect anything but Docker monitor when running in raw logs mode.

## Open Questions / To Decide

- [ ] An unrelated issue, but we need to decide how to handle it - https://github.com/scalyr/scalyr-agent-2/pull/648/files#r512737906


scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -76,7 +76,7 @@
     __monitor__,
     "container_check_interval",
     "Optional (defaults to 5). How often (in seconds) to check if containers have been started or stopped.",
-    convert_to=int,
+    convert_to=float,
```
Kami, line 79: This makes it easier / faster to try to reproduce the scenario locally since we can use interval smaller than 1 second and the change is also fully backward compatible.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -763,6 +763,20 @@ def _get_containers(
                                     info["LogPath"] if "LogPath" in info else None
                                 )
 
+                                if not log_path:
+                                    # NOTE: If docker_raw_logs is True and we hit this code path it
+                                    # really means we won't be ingesting any logs so this should
+                                    # really be treated as a fatal error.
+                                    logger.warning(
```
Kami, line 770: An unrelated issue I encountered when trying to reproduce the original issue.

``LogPath`` being None will in 95% of cases represent a misconfiguration on the Docker daemon side - e.g. the daemon is using some other log-driver such as ``jourdnald`` and not ``json-file`` one.

It could likely also happen under some other scenarios, but it basically represents a fatal error if ``docker_raw_logs`` is ``True`` (aka default value), since it means we won't be monitoring / ingesting logs for that container.

Right now we simply silently ignore this issue which is obviously bad and a nightmare to troubleshoot.

We can't really throw here (even if it most likely represents a configuration error), but I think we should at the very least log a message under error level and be very explicit - e.g. something along the lines of "LogPath for container foo is None. This means we won't be monitoring and ingesting logs for this container if docker_raw_logs mode is used".

What do other people think?

scalyr_agent/copying_manager.py
```diff
@@ -370,10 +370,14 @@ def __path_in_globs(self, path, path_globs):
 
         return None
 
-    def add_log_config(self, monitor_name, log_config):
+    def add_log_config(self, monitor_name, log_config, force_add=False):
```
czerwingithub, line 373: Part of me wonders if we shouldn't just always have `force_add=True` (and therefore remove it as a parameter).  I know you are trying to limit the scope of this change, but I do wonder if other similar bugs will crop up on paths where `force_add=False`.  Did you take a look to get a sense?

I have to admit, it was not on my radar to handle the case where a docker container is stopped and started again.

I'm a little worried what happens in the case where the container is stopped for a longer period of time and then restarted, while the original log file still remains on disk.  I'm a little worried that we will start copying the log again from the start of the log file, rather than where we left off.  Here's what I am thinking:

1. Docker container stops.
2. Scalyr Agent the log file for removal.
3. (I believe) Once the Scalyr Agent catches up to the end of the current log file, the LogProcessor is removed.
4. (I believe) At some later point, the checkpoint for that log file is removed.
5. Docker container starts again.  The original log file content is still in the docker log file.
6. (I believe) The Scalyr Agent will notice the container running, and add back in the log file to be copied, configured with copying from start (because I think that's how we handle Docker containers).

Can you take a look to see if that is what would happen?  Step 4 probably doesn't matter (if we have a checkpoint state for it or not) since we are going to copy from the start anyway.  But, I wanted to highlight that a checkpoint for a file is removed at some point just in case it does matter.

czerwingithub (commented):
Ok, I asked more questions than gave review comments, so you can curse me in the night if you'd like.

Kami:
> Part of me wonders if we shouldn't just always have force_add=True (and therefore remove it as a parameter). I know you are trying to limit the scope of this change, but I do wonder if other similar bugs will crop up on paths where force_add=False.

Yeah, I was indeed thinking about doing that. It should likely be safe, but then again I also wanted to keep the scope down and avoid any potential inadvertent regressions.

So maybe for this first iteration, we release this fix as is and then we make ``force_add=True`` a default behavior, make sure all the tests pass and have the agents running on staging, etc. for a while with that value and ensure there are no regressions.

As far as your second comment goes, I believe we have additional safe-guard which should prevent this from happening (and that's actually the place where there was a timezone related bug) - https://github.com/scalyr/scalyr-agent-2/blob/36e810cec93829629cba955a1f2b30292fe346a0/scalyr_agent/builtin_monitors/docker_monitor.py#L1261

Basically, if the log file is still on disk, we will try to read timestamp from the last line (well, last 5k bytes in the file) and use that as a starting point.

I don't think we have any (direct and explicit) tests for that functionality, but I will look into at least adding some unit test ones (end to end would be a big pain).

Kami:
@czerwingithub I've pushed log message change (4c498408c967797b92d7ff6e71da170d247235c1) and unit tests for the existing ContainerChecker functionality (3826838761f10528b9640988b1eed56d453cf035).

czerwingithub (approved):

PR #649, opened 2020-10-28
Correctly work with UTC timestamps / dates everywhere in the code base

Kami:
This pull request updates all the code to correctly work and utilize UTC timestamps everywhere.

Previously some of the code incorrectly used ``datetime.datetime.now()`` and ``datetime.datetime,fromtimestamp()`` functions which return time in local server timezone and not UTC.

Since rest of the code assumes and works with UTC everywhere, that's not correct and could result in broken / incorrect behavior when running the agent on servers which use some other time zone and not UTC (depending on the code in question and also the local time zone set).

This bug was originally discovered when fixing other issues in Docker monitor in #647.


czerwingithub (approved):
Thanks for doing this.

PR #650, opened 2020-10-29
New diskperf error check

ArthurKamalov:
This small PR adds another check for an error caused by disabled "diskperf". This error may happen on older versions of windows and the existing exception catch does not work.


Kami:
I pushed a small logging change (40d7c8aed2c3ff6ee507bb146de41c6905357a4e) and added unit tests (a4305854fb4570166fad1b2833b5efe2bbdd7d45).

Ideally we would have full blown end to end AMI based tests, but that would be much harder to orchestrate.

Kami (approved):

PR #651, opened 2020-10-30
Warn user on non-UTF-8 coding / locale and other locale related improvements

Kami:
This pull request updates the agent start up code to log a warning if non ``UTF-8`` coding / locale is detected.

## Background, Context

I was troubleshooting a customer issue and noticed multiple issues which could arise when locale / coding under which the agent is running is not set to UTF-8 and user is trying to monitor files with non-ascii characters in the file names / content.

Ultimately, if we encounter this scenario this represents a misconfiguration on the user side so there is not much we can do beyond warning the user and advising them to change the locale to UTF-8 (I mean in theory we could try setting locale using ``locale.setlocale``, but I'm worried this may open a can of worms and is something we should avoid - just the other day I was reading this "rant" wrt to locale mess on POSIX systems - https://github.com/mpv-player/mpv/commit/1e70e82baa9193f6f027338b0fab0f5078971fbe).

Keep in mind that some of those issues will only arise under Python 2 and not Python 3 (e.g. ``os.path.join`` only seems to throw under Python 2 if coding is not UTF-8, but not under Python 3).

In addition to the warning change, I also added a small idna encoding issue workaround (https://bugs.python.org/issue29288) and updated orjson code to fall back to native Python json implementation on unicode related errors - it appears orjson will throw whereas native Python json library won't if we try to  load / dump data with non-ascii characters when locale / coding is not set to UTF-8.

## TODO

- [x] Unit tests
- [x] Verify change on Windows

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,12 +114,6 @@ commands:
       - checkout
       - restore_cache:
           key: deps2-tox-{{ .Branch }}-<< parameters.python_version >>-venv-{{ checksum "dev-requirements.txt" }}
-      - run:
-          name: Configure Locales
-          command: |
-            # Needed for locale tests
-            locale-gen en_US.UTF-8
-            locale-gen en en_US en_US.UTF-8
       - when:
           condition: << parameters.apt_dependencies >>
           steps:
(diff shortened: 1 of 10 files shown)


czerwingithub (changes requested):
@Kami Do you know how this change will behave under Windows?  Will we be emitting this warning?

Beyond that, looks good to me.  Can you add an entry for this in `CHANGLELOG.md`?

Kami:
@czerwingithub I added a changelog entry.

As far as behavior on the Windows goes - it should be the same - we warn if coding is not set to utf8.

ArthurKamalov (approved):
all looks good to me 👍

czerwingithub (approved):

PR #654, opened 2020-11-12
Sharded copying manager

ArthurKamalov:
@czerwingithub, as it was said previously, the first commit is just a mechanical refactoring.


scalyr_agent/agent_main.py
```diff
@@ -192,10 +193,6 @@ def __init__(self, platform_controller):
         self.__copying_manager = None
         # The current monitors manager.
         self.__monitors_manager = None
-        # The current ScalyrClientSession to use for sending requests.
```
ArthurKamalov, line 195: Agent main does not have client sessions anymore. All sessions now are created in the workers.

scalyr_agent/agent_main.py
```diff
@@ -1457,65 +1446,6 @@ def __create_new_client(self):
             result = NewScalyrClientSession(self.__config)
         return result
 
-    def __create_client(self, quiet=False):
```
ArthurKamalov, line 1460: client creation function moved to `scalyr_client.py`

scalyr_agent/agent_main.py
```diff
@@ -1849,24 +1779,28 @@ def __calculate_overall_stats(
...
-        )
-        delta_stats.total_request_latency_secs = (
-            self.__scalyr_client.total_request_latency_secs
-        )
-        delta_stats.total_connections_created = (
-            self.__scalyr_client.total_connections_created
-        )
-        delta_stats.total_compression_time = self.__scalyr_client.total_compression_time
+        # get client session states from all workers of the copying manager and sup up all their stats.
+        session_states = self.__copying_manager.get_worker_session_statuses()
```
ArthurKamalov, line 1783: Naw we sum stats from clients from all workers.

scalyr_agent/configuration.py
```diff
@@ -3640,6 +3714,11 @@ def __perform_substitutions(self, source_config):
                 object_value=source_config, substitutions=substitutions
             )
 
+    def __getstate__(self):
```
ArthurKamalov, line 3717: the configuration object can not be serialized due to the logger object, so it is impossible to pickle them and send to the multiprocess workers.
Another solution is to read the configuration one more time on each worker,

scalyr_agent/log_processing.py
```diff
@@ -2234,13 +2238,7 @@ def is_active(self):
...
         self.__is_active = False
 
-    @property
-    def log_path(self):
-        """
-        @return:  The log file path
-        @rtype: str
-        """
-        # TODO:  Change this to just a regular property?
+    def get_log_path(self):
```
ArthurKamalov, line 2241: We have to add more code to make Multiprocessing proxy objects work with properties, so it is easier just to use member functions.

scalyr_agent/scalyr_client.py
```diff
@@ -70,8 +73,90 @@ def _set_last_timestamp(val):
     __last_time_stamp__ = val
 
 
+def create_new_client(config, api_key=None):
```
ArthurKamalov, line 76: this is moved from agent_main.py

tests/unit/agent_main_test.py
```diff
@@ -182,68 +186,81 @@ def test_skipped_bytes_warnings(self):
...
-        self.assertIsNotNone(test)
-        with open(self.agent_log_path, "r") as f:
-            print((f.read()))
-        self.assertLogFileContainsLineRegex(
-            ".*"
-            + re.escape(
-                "Warning, skipping copying log lines.  Only copied 0.0 MB/s log bytes when 0.0 MB/s were generated over the last 1.0 minutes.  This may be desired (due to excessive bytes from a problematic log file).  Please contact support@scalyr.com for additional help."
-            )
-        )
+        with mock.patch("scalyr_agent.__scalyr__.INSTALL_TYPE", __scalyr__.DEV_INSTALL):
```
ArthurKamalov, line 189: Test case if refactored to work without the client session object in the ScalyrAgent instance.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -41,23 +182,9 @@ def __init__(self, configuration, monitors):
...
-        self.__rate_limiter = None
-        if self.__config.parsed_max_send_rate_enforcement:
-            self.__rate_limiter = RateLimiter(
-                # Making the bucket size 4 times the bytes per second, it shouldn't affect the rate over a long time
-                # meaningfully but we don't have a rationale for this value.
-                # TODO: Make this configurable as part of the `max_send_rate_enforcement` configuration option
-                self.__config.parsed_max_send_rate_enforcement * 4.0,
-                self.__config.parsed_max_send_rate_enforcement,
-            )
-
```
ArthurKamalov: Moved to the worker.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -159,12 +265,17 @@ def __init__(self, configuration, monitors):
         self.__disable_copying_thread = configuration.disable_copying_thread
 
         # Statistics for copying_manager_status messages
-        self.total_copy_iterations = 0
-        self.total_read_time = 0
-        self.total_waiting_time = 0
-        self.total_blocking_response_time = 0
-        self.total_request_time = 0
-        self.total_pipelined_requests = 0
```
ArthurKamalov: Moved to worker.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -107,46 +234,25 @@ def __init__(self, configuration, monitors):
...
         self.__lock = threading.Lock()
 
-        # The current pending AddEventsTask.  We will retry the contained AddEventsRequest serveral times.
-        self.__pending_add_events_task = None
-
-        # The next LogFileProcessor that should have log lines read from it for transmission.
-        self.__current_processor = 0
-
-        # The client to use for sending the data.  Set in the start_manager call.
-        self.__scalyr_client = None
```
ArthurKamalov: Moved to the worker.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -107,46 +234,25 @@ def __init__(self, configuration, monitors):
...
-        self.__last_attempt_time = None
-        self.__last_success_time = None
-        self.__last_attempt_size = None
-        self.__last_response = None
-        self.__last_response_status = None
-        self.__total_bytes_uploaded = 0
+        self.__last_scan_attempt_time = None
         self.__total_errors = 0
-        self.__total_rate_limited_time = 0
-        self.__rate_limited_time_since_last_status = 0
```
ArthurKamalov: Moved to the worker.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -518,282 +610,78 @@ def run(self):
...
 
-                if checkpoints_state is None:
-                    checkpoints = None
-                else:
-                    checkpoints = checkpoints_state["checkpoints"]
+                log.info("Starting copying manager workers.")
+
+                # gather and merge all checkpoints from all active workers( or workers from previous launch)
+                # into single checkpoints object.
+                checkpoints = self.__get_checkpoints(current_time)
```
ArthurKamalov: Moved checkpoint reading into __get_checkpoints() function

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -518,282 +610,78 @@ def run(self):
...
-                last_success = current_time
-                last_success_status = None
-
-                # Force the agent to write a new full checkpoint as soon as it can
-                last_full_checkpoint_write = 0.0
-
-                pipeline_byte_threshold = self.__config.pipeline_threshold * float(
-                    self.__config.max_allowed_request_size
-                )
-
```
ArthurKamalov: moved to the worker.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -518,282 +610,78 @@ def run(self):
...
-                                self.__pending_add_events_task = None
-                            # Tell all of the processors to go to the end of the current log file.  We will start
-                            # copying
-                            # from there.
-                            for processor in self.__log_processors:
-                                processor.skip_to_end(
-                                    "Too long since last successful request to server.",
-                                    "skipNoServerSuccess",
-                                    current_time=current_time,
-                                )
```
ArthurKamalov: Moved to the worker.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -518,282 +610,78 @@ def run(self):
...
-                        self.__lock.acquire()
-                        copying_params.update_params(result, bytes_sent)
-                        self.__last_attempt_time = current_time
-                        self.__last_success_time = last_success_status
-                        self.__last_attempt_size = bytes_sent
-                        self.__last_response = six.ensure_text(full_response)
-                        self.__last_response_status = result
-                        if result == "success":
-                            self.__total_bytes_uploaded += bytes_sent
-                        self.__lock.release()
```
ArthurKamalov: Moved into the worker.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -803,38 +691,31 @@ def run(self):
...
-                        self.__write_full_checkpoint_state(current_time)
-                        last_full_checkpoint_write = current_time
-
-                    if pipeline_time < copying_params.current_sleep_interval:
-                        self._sleep_but_awaken_if_stopped(
-                            copying_params.current_sleep_interval - pipeline_time
-                        )
-                        self.total_waiting_time += (
-                            copying_params.current_sleep_interval - pipeline_time
-                        )
```
ArthurKamalov: Moved into the worker.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -803,38 +691,31 @@ def run(self):
...
-            except Exception:
+                    self.total_scan_iterations += 1
+            except Exception as e:
                 # If we got an exception here, it is caused by a bug in the program, so let's just terminate.
                 log.exception("Log copying failed due to exception")
                 sys.exit(1)
         finally:
-            self.__write_full_checkpoint_state(current_time)
-            for processor in self.__log_processors:
-                processor.close()
```
ArthurKamalov: Moved into the worker.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1148,19 +1027,22 @@ def __create_log_processors_for_log_matchers(
...
         # iterate over the log_matchers while we create the LogFileProcessors
         for matcher in log_matchers:
             for new_processor in matcher.find_matches(
                 self.__log_paths_being_processed,
                 checkpoints,
                 copy_at_index_zero=copy_at_index_zero,
+                get_log_processor_factory=self.__get_log_processor_factory,
             ):
-                self.__log_processors.append(new_processor)
-                self.__log_paths_being_processed[new_processor.log_path] = new_processor
```
ArthurKamalov: Now log processor is added to the worker inside the `matcher.find_matches` function.

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -501,7 +490,6 @@ def run(self):
                                 >= pipeline_byte_threshold
                                 and self.__pending_add_events_task.next_pipelined_task
                                 is None
-                                and not disable_pipelining
```
ArthurKamalov: I am not sure how to be withpipelining. In original  copying manager the function `self.__has_pending_log_changes()` checks for padding new log matches. I guess here we can do the same for new pending log processors and disable pipeline if there are pending logprocessors.

czerwingithub: I'll have to think through this a little more.  Add a TODO here to say that we need to handle the `disable_pipelining` case.  That was added to solve a very specific bug with respect to how we invoked the callback handlers on the processors.  It may be that with the new design that bug goes away.  Let's save it as one of the last things we need to address for this PR.

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -851,9 +846,12 @@ def __get_next_add_events_task(self, bytes_allowed_to_send, for_pipelining=False
 
         # Initialize the looping variable to the processor we last left off at on a previous run through this method.
         # This is an index into the __log_processors list.
+        log_processors = list(self.__log_processors.values())
```
czerwingithub, line 849: You should move the above comments to `current_processor`.

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -915,16 +913,14 @@ def handle_completed_callback(result):
             # We could be susceptible to exceptions thrown in the middle of this method, though now should.
 
             # Copy the processor list because we may need to remove some processors if they are done.
-            processor_list = self.__log_processors[:]
-            self.__log_processors = []
-            # A dict that maps file paths to the processors assigned to them.  This is used to ensure multiple
-            # processors do not try to process the same file.
-            self.__log_paths_being_processed = {}
+            processor_list = self.__log_processors
```
czerwingithub, line 916: I actually think you want to use the `log_processors` variable here that you created.  Of course, that's a list instead of a dictionary, but you can make those changes as well.

If you make a temporary copy of self.__log_processors, that would handle the bug that we added `disable_pipelining` to cover.  Essentially, the contents `self.__log_processors` could have changed between the sending the request and the callback invocation.  So, we should keep a copy of `self.__log_processors` as it was when we issued the request.

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -1028,19 +1025,271 @@ def __write_full_checkpoint_state(self, current_time):
...
+                copying_manager_class.COPYING_MANAGER_CLASS,
+                exposed=copying_manager_class.EXPOSED,
+                method_to_typeid={six.ensure_str("get_log_processors"): six.ensure_str("list")}
+            )
+
+            # register proxy type for the LogFilePRocessor
+            CopyingManagerMemoryManager.register(
+                six.ensure_str("LogFileProcessorProxy"),
+                LogFileProcessor,
+                exposed=[
```
czerwingithub, line 1206: Are these supposed to be all of the method that can be invoked across the process boundary?  This is what the copy manager would invoke, is that right?

If so, I'm a little surprised.  I would think that the only ones that need to be invoked across process boundaries are `close_at_eof` and `close`.  Maybe `add_missing_attributes`.  I have to dig into that one a little more.  But, the worker (which is in the same process as the processor) should be handling the checkpointing and the status generation, I would think.

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -1028,19 +1025,271 @@ def __write_full_checkpoint_state(self, current_time):
...
+            # create shared memory manager
+            self.__memory_manager = CopyingManagerMemoryManager()
+
+            # start memory manager.
+            # Important for the understanding. When the __memory_manager is started, it creates a new process.
+            # We use this process as a process for the worker.
+            self.__memory_manager.start()
+
+            # create proxy object of the worker. The real worker instance is created in the new process
+            # which was created when __mameory_manager started.
```
czerwingithub, line 1226: Misspell of `memory_manager`.

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -1028,19 +1025,271 @@ def __write_full_checkpoint_state(self, current_time):
...
+            self._real_worker = self.__memory_manager.CopyingManagerWorker(
```
czerwingithub, line 1227: This seems like a bad name for this... because in the original process, isn't this variable really just a proxy object for the `CopyingManagerWorker`?

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -1028,19 +1025,271 @@ def __write_full_checkpoint_state(self, current_time):
...
+                ]
+            )
+
+            # create shared memory manager
+            self.__memory_manager = CopyingManagerMemoryManager()
+
+            # start memory manager.
+            # Important for the understanding. When the __memory_manager is started, it creates a new process.
+            # We use this process as a process for the worker.
+            self.__memory_manager.start()
```
czerwingithub, line 1223: When the process is started, what gets executed first?  Is it `CopyingManagerWorkerContainer.__init__`?  I think we should document some of the magic that goes on here.  Also, we still need to have some code somewhere to handle closing down the agent log logfile.  We need to know where we should do that.

czerwingithub (commented):
Here are some initial comments.  I haven't looked at the tests in detail yet.

The biggest technical design issue I have is about which methods you are exposing in the LogProcessorProxy object and why.  I don't think that's a major issue, so should be quick to work through.  We also need to discuss how you are handling writing out to the `agent.log` in the worker processes.  I think, once we work through those two issues, you could start doing performance benchmarking without worry that there would be significant changes that would invalidate them.

scalyr_agent/configuration.py
```diff
@@ -3092,6 +3128,44 @@ def __verify_monitor_entry_and_set_defaults(
...
+        # if the id for the api_key entry is not specified, just set the entry index as id.
+        if "id" not in api_key_entry:
+            api_key_entry["id"] = six.text_type(entry_index)
+
+        self.__verify_or_set_optional_string(
+            api_key_entry, "type", "thread", description
+        )
+
+        # windows does not support copying manager backed with multiprocessing workers. Change worker types to thread.
+        if api_key_entry.get("type") == "process" and platform.system() == "Windows":
```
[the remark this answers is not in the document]
ArthurKamalov: Oh, that makes things even easier. The `CopyingManagerWorkerContainer` class was created just to deal with different worker types and there is no need for it if all workers have the same type. Now, the `ApiKeyWorkerPool` deals with workers and proxies themselves, without any container class. 
https://github.com/scalyr/scalyr-agent-2/blob/d0dc47790615f14fc7ca209b0f985b02399a668c/scalyr_agent/sharded_copying_manager/copying_manager.py#L46-L85

scalyr_agent/sharded_copying_manager/worker.py
ArthurKamalov: This class is removed now. See here for more info.
https://github.com/scalyr/scalyr-agent-2/pull/654#discussion_r526289575

> because in the original process, isn't this variable really just a proxy object for the CopyingManagerWorker?

The   `CopyingManagerWorkerContainer` was not supposed to be used in the worker's original process. It is just a wrapper for the `self.real_worker` which **is** a proxy of the `CopyingManagerThreadedWorker` (or CopyingManagerThreadedWorker itself in case of "no multiprocess" mode)

scalyr_agent/sharded_copying_manager/worker.py
ArthurKamalov: Yes, I removed the unneeded methods. Now there are only:
*  close_at_eof
* close
* add_missing_attributes
* get_log_path # this one does not change, so I made it cacheable on the proxy side.
* generate_status # used by LogMatcher's

scalyr_agent/sharded_copying_manager/worker.py
ArthurKamalov: I made the `self.__log_processors`as a list again, I make it as a `dict` in my previous approaches and there is no need for it now.

scalyr_agent/log_processing.py
```diff
@@ -3318,6 +3320,8 @@ def find_matches(
             then if copy_at_index_zero is True, the file will be processed from the first byte in the file.  Otherwise,
             the processing will skip over all bytes currently in the file and only process bytes added after this
             point.
+        @param get_log_processor_factory:  Callable which returns another callable which creates
```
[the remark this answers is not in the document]
ArthurKamalov: Yes, I made it more complicated than it has to be.

The new approach is that the LogMatchers do not create log processors directly by instantiating the 'LogFileProcessor' class. Instead of that, we find an appropriate worker pool for this matcher. (by the `api_key_id`) and then pass a new method of the worker pool named `create_and_add_new_log_processor` as a `create_log_processor` parameter.

Inside the `find_matches` method, the LogMatcher will call the `create_log_processor` parameter to create the log processor.

https://github.com/scalyr/scalyr-agent-2/blob/d0dc47790615f14fc7ca209b0f985b02399a668c/scalyr_agent/sharded_copying_manager/copying_manager.py#L1091-L1102


**How `worker_pool.create_and_add_new_log_processor` creates  a new log processor?**

This method finds a worker and calls its method - `create_and_schedule_new_log_processor` 

https://github.com/scalyr/scalyr-agent-2/blob/d0dc47790615f14fc7ca209b0f985b02399a668c/scalyr_agent/sharded_copying_manager/copying_manager.py#L112-L123

This method finally creates the new log processor and also immediately adds it to the worker itself. 

https://github.com/scalyr/scalyr-agent-2/blob/d0dc47790615f14fc7ca209b0f985b02399a668c/scalyr_agent/sharded_copying_manager/worker.py#L1080-L1097

In other words, the log processor creation and adding it to the worker is now combined into one operation.

scalyr_agent/log_watcher.py
```diff
@@ -30,22 +30,50 @@ def add_log_config(self, monitor_name, log_config):
         param: log_config - a log_config object containing at least a path
         returns: the log_config variable with updated path and default information
         """
-        pass
+        self._add_log_config(log_config, monitor_name=monitor_name)
```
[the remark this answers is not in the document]
ArthurKamalov: oops, forgot to remove it long time ago.

scalyr_agent/single_threaded_copying_manager.py
```diff
@@ -370,14 +370,10 @@ def __path_in_globs(self, path, path_globs):
 
         return None
 
-    def add_log_config(self, monitor_name, log_config, force_add=False):
+    def add_log_config(self, monitor_name, log_config):
         """Add the log_config item to the list of paths being watched
         param: monitor_name - the name of the monitor adding the log config
         param: log_config - a log_config object containing the path to be added
-        param force_add: True or force add this file and cancel any removal which
```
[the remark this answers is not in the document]
ArthurKamalov: This should be a file with an old copying manager implementation, everything should be fine on the `sharded_copying_manager.copying_manager.py` But I'll still fix that too.

scalyr_agent/single_threaded_copying_manager.py
```diff
@@ -975,7 +956,7 @@ def run(self):
                                     "%s%s"
                                     % (
                                         self.__config.copying_thread_profile_output_path,
-                                        datetime.datetime.utcnow().strftime(
+                                        datetime.datetime.now().strftime(
```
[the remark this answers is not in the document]
ArthurKamalov: same here https://github.com/scalyr/scalyr-agent-2/pull/654#discussion_r526311296

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -518,276 +614,72 @@ def run(self):
...
                             if seconds_past_epoch % profile_dump_interval == 0:
                                 profiler.disable()
                                 profiler.dump_stats(
-                                    "%s%s"
+                                    "%s%s%s"
                                     % (
                                         self.__config.copying_thread_profile_output_path,
-                                        datetime.datetime.utcnow().strftime(
+                                        "copying_manager_",
+                                        datetime.datetime.now().strftime(
```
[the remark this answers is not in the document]
ArthurKamalov: Yes, I missed that, thanks.

scalyr_agent/sharded_copying_manager/worker.py
ArthurKamalov: This class is removed now. See here for more info.
#654 (comment)

> When the process is started, what gets executed first? Is it CopyingManagerWorkerContainer.__init__?
`CopyingManagerWorkerContainer.__init__` was not executed in the worker process at all. It was just a container for the proxy object or the real worker(in case of "non-process").

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1,24 +1,213 @@
...
+        self.__shared_memory_managers = dict()  # type: Dict[six.text_type, CopyingManagerSharedMemory]
+
+        for i in range(api_key_config["workers"]):
+            # combine workers positional number in pool and pool id.
+            worker_id = "%s_%s" % (self.__api_key_id, i)
+
+            if not self.__config.use_multiprocess_copying_workers:
+                worker = CopyingManagerThreadedWorker(self.__config, api_key_config, worker_id)
+            else:
+                memory_manager = CopyingManagerSharedMemory()
```
czerwingithub, line 71: Same here.  We should have a better name for this class.  Maybe `MultiprocessWorkerManager`.

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -1001,22 +991,41 @@ def __write_checkpoint_state(
...
 
             if full_checkpoint:
                 processor.set_inactive()
 
+        checkpoints_dir_path = os.path.join(
+            self.__config.agent_data_path, "checkpoints"
+        )
+
+        try:
+            os.mkdir(checkpoints_dir_path)
```
czerwingithub, line 1007: Why was this code added (I believe it wasn't there originally, or was it)?

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1256,4 +1202,168 @@ def __scan_for_new_logs_if_necessary(
...
+                log.warn(
+                    "The worker checkpoint file '%s' is too stale (written at '%s').  Ignoring it.  All log files will "
+                    "be copied starting at their current end.",
+                    full_checkpoint_file_path,
+                    scalyr_util.format_time(full_checkpoints["time"]),
+                    error_code="staleCheckpointFile",
+                )
+
+                # checkpoint files are stale, so remove them.
+                os.unlink(full_checkpoint_file_path)
```
czerwingithub, line 1319: This seems unsafe because it looks like we are removing the checkpoint state from disk.  What happens if the agent crashes before it is able to write it to disk?  We'd lose all of the checkpoint state.

I think we need to come up with a better strategy on how to handle start up and consolidating the checkpoint state.

My suggestion is that we have an initialization step that reads all checkpoints (including legacy file paths) and writes out a new, single consolidated file to disk before it goes and deletes all of the other checkpoint files.  Once we have cleaned up the state like that, then we can have the individual workers start working, writing out their own checkpoint files, etc.

czerwingithub (commented):
We should have a conversation about what you are working on and what your plan is to address all of the comments.  I don't believe you've addressed all of my comments from the previous round, which is fine if you have a plan for when you are going to get to them.

Depending on what you are working on, I think the next round of revisions from you should address my concerns in terms of the checkpoint management.  Currently, it is not clear to me if we properly handle the upgrade case (i.e., the customer updates to this new version of the code which writes checkpoints into a new location... we still have to read the checkpoints from the old location at start up).  Also, I believe there is possibility of losing checkpoint state if the agent crashes at a particular time.  Plus, I think we can simplify the checkpointing code.  Take a look at my comments.

Other major things we need to resolve:
* How the stats are rolled up
* The copying rate limiting

By the way, yup, your last round of refactoring was nice and did simplify things.  Good job on that. I'm glad you could get rid of that container abstraction you had.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1256,4 +1136,188 @@ def __scan_for_new_logs_if_necessary(
...
+            result.update(wc["checkpoints"])
+
+        return result
+
+    def __find_and_read_workers_checkpoints(self, current_time):
+        """
+        Read for all checkpoint files that
+        :return:
+        """
+        checkpoints_dir_path = os.path.join(
```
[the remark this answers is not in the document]
ArthurKamalov: Done.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1256,4 +1136,188 @@ def __scan_for_new_logs_if_necessary(
...
+            self.__config.agent_data_path, "checkpoints"
+        )
+
+        if not os.path.isdir(checkpoints_dir_path):
+            return []
+
+        glob_path = os.path.join(checkpoints_dir_path, "checkpoints-*.json")
```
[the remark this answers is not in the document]
ArthurKamalov: Yes, now it does. Now it should treat the checkpoint files from the older version as a regular checkpoint file. (Also test was added)

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1256,4 +1136,188 @@ def __scan_for_new_logs_if_necessary(
...
+
+        workers_checkpoints = list()
+
+        for full_checkpoints_path in glob.glob(glob_path):
+            m = re.search(
```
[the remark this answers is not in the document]
ArthurKamalov: Thanks for pointing on that. Yes, this was absolutely unnecessary.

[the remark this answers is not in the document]
ArthurKamalov: Yes, not it does. 
https://github.com/scalyr/scalyr-agent-2/pull/654#discussion_r527069144

scalyr_agent/configuration.py
```diff
@@ -3640,6 +3726,16 @@ def __perform_substitutions(self, source_config):
                 object_value=source_config, substitutions=substitutions
             )
 
+    def __getstate__(self):
+        """
+        Remove unpicklable fields from the configuration instance.
```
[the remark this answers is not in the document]
ArthurKamalov: This method is used only by SyncManager. It is a feature provided by the `pickle` library that allows changing the pickling behavior.
https://docs.python.org/3/library/pickle.html#object.__getstate__

scalyr_agent/single_threaded_copying_manager.py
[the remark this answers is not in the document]
ArthurKamalov: I deleted this file.

scalyr_agent/single_threaded_copying_manager.py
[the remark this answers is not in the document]
ArthurKamalov: This is actually the same file as in the previous comment. https://github.com/scalyr/scalyr-agent-2/pull/654#discussion_r526897274  Deleted.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1,24 +1,165 @@
...
+        current_worker = self.__workers[self.__current_worker]
+        if self.__current_worker == len(self.__workers) - 1:
+            self.__current_worker = 0
+        else:
+            self.__current_worker += 1
+
+        return current_worker
+
+
+    def start_workers(self):
```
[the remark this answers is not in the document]
ArthurKamalov: That's a good point. Thanks.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1059,24 +955,7 @@ def __scan_for_pending_log_files(self):
...
-                scalyr_util.format_time(checkpoints_state["time"]),
-                error_code="staleCheckpointFile",
-            )
-            checkpoints_state = None
-
-        if checkpoints_state is None:
-            checkpoints = {}
-        else:
-            checkpoints = checkpoints_state["checkpoints"]
+        checkpoints = self.__get_checkpoints(time.time())
```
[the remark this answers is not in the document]
ArthurKamalov: I am not sure. Looks like Tomaz hasn't done this as utc.
As you can see from the master branch, it still uses the `time.time()`
https://github.com/scalyr/scalyr-agent-2/blob/8cb4968accacfcb00c95ba228223c66410551cce/scalyr_agent/copying_manager.py#L703-L723

scalyr_agent/sharded_copying_manager/worker.py
ArthurKamalov: Update: also `is_closed` method is invoked externally by the LogMatcher.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1,24 +1,213 @@
...
+    """max function which ignores None values"""
+    values = [v for v in [v1, v2] if v is not None]
+    return max(values)
+
+
+class ApiKeyWorkerPool(object):
+    """
+    This abstraction is responsible for maintaining the workers
+    for a particular entry in the 'api_keys' list in the configuration.
+    It also creates worker instances according to the number in the "workers" field in each entry.
```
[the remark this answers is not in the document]
ArthurKamalov: Done.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1,24 +1,213 @@
...
+    """
+    def __init__(self, config, api_key_config):
+        # type: (Configuration, Dict) -> None
+
+        self.__api_key_id = api_key_config["id"]
+        self.__config = config
+
+        self.__workers = []  # type: List[CopyingManagerThreadedWorker]
+
+        # collection of shared memory managers. (Subclasses of the multiprocessing.BaseManager)
```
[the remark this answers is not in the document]
ArthurKamalov: IMO, the `shared object manager` is the best option.

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: Done.

scalyr_agent/sharded_copying_manager/worker.py
ArthurKamalov: Yes, it wasn't there originally. I've just added this to check if the `checkpoints` folder does not exist. 
But now I moved this check into the copying manager, now it does the check when it combines all worker checkpoint files into one master file.

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: As discussed, now the copying manager combines all checkpoint files into a single - master file. After that, it also removes the worker checkpoints.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1,24 +1,165 @@
...
+            worker = CopyingManagerWorkerContainer(
+                self.__config, api_key_config, worker_id
+            )
+            self.__workers.append(worker)
+            log.log(
+                scalyr_logging.DEBUG_LEVEL_1,
+                "CopyingManagerWorker #%s is created." % worker.get_id(),
+            )
+
+        self.__current_worker = 0
```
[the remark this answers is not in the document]
ArthurKamalov: Done.

scalyr_agent/single_threaded_copying_manager.py
```diff
@@ -1056,7 +1037,7 @@ def generate_status(self, warn_on_rate_limit=False):
             result.last_attempt_time = self.__last_attempt_time
             result.last_attempt_size = self.__last_attempt_size
             result.last_response = self.__last_response
-            result.last_response_status = self.__last_response_status
+            result.last_response_statuses = self.__last_response_status
```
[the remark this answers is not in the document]
ArthurKamalov: This file is unneeded and deleted now.

scalyr_agent/agent_status.py
```diff
@@ -751,8 +751,8 @@ def __report_copying_manager(output, manager_status, agent_log_file_path, read_t
                 )
 
         if (
-            worker_pool.has_good_health_checks is not None
-            and not worker_pool.has_good_health_checks
+            worker_pool.all_health_checks_good is not None
+            and not worker_pool.all_health_checks_good
```
czerwingithub, line 755: `None` is not truthy, so can't you just write this as `not worker_pool.all_health_checks_good`?  (i.e., eliminate the first predicate that checks for `is not None`)

scalyr_agent/util.py
```diff
@@ -2357,3 +2360,19 @@ def summarize(self):
             self.max(),
             self.estimate_median(),
         )
+
+
+def max_ignore_none(*args, **kwargs):
+    """max function which ignores None values"""
+    key =kwargs.get("key")
```
czerwingithub, line 2367: fix spacing on the equals sign.

scalyr_agent/util.py
```diff
@@ -2357,3 +2360,19 @@ def summarize(self):
             self.max(),
             self.estimate_median(),
         )
+
+
+def max_ignore_none(*args, **kwargs):
+    """max function which ignores None values"""
```
czerwingithub, line 2366: Add documentation for the arguments, especially what the first one (key) is for.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -149,62 +161,38 @@ def generate_status(self):
...
                 )
 
             if status.last_attempt_size:
-                last_attempt_requests_overall_size += status.last_attempt_size
+                result.last_attempt_requests_overall_size += status.last_attempt_size
 
             if status.last_response_status is not None:
-                last_response_statuses.add(status.last_response_status)
+                # do not overwrite the result if we already have a bad response.
+                if result.all_responses_successful is not False:
```
czerwingithub, line 183: Just write this as `not result.all_response_successful`.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -149,62 +161,38 @@ def generate_status(self):
...
+                    result.all_responses_successful = status.last_response_status == "success"
 
-            if (
-                status.health_check_result is not None
-            ):
-                health_checks.add(status.health_check_result)
+            if status.health_check_result is not None:
+                # do not overwrite the result if there is a bad health check.
+                if result.all_health_checks_good is not False:
```
czerwingithub, line 188: Just write this as `not result.all_health_checks_good`.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -799,69 +787,66 @@ def generate_status(self, warn_on_rate_limit=False):
 
             result.total_scan_iterations = self.total_scan_iterations
 
-            last_response_statuses = set()
-            health_checks = set()
+            all_health_checks_good = None
```
czerwingithub, line 790: Why not set this and the next one to `True`?

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -799,69 +787,66 @@ def generate_status(self, warn_on_rate_limit=False):
...
                 result.api_key_worker_pools.append(worker_pool_status)
 
                 result.total_errors += worker_pool_status.total_errors
+                result.total_bytes_uploaded += worker_pool_status.total_bytes_uploaded
 
                 if worker_pool_status.all_responses_successful is not None:
-                    # collect response status information from the worker pools.
-                    last_response_statuses.add(worker_pool_status.all_responses_successful)
+                    # do not overwrite the result if we already have a bad response.
+                    if all_responses_good is not False:
```
czerwingithub, line 806: Replace with `all_responses_good` (remove `is not False`) here and below.  Just do a search for that as well.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1204,99 +1137,94 @@ def _scan_for_new_logs_if_necessary(
...
+        result = {}
         # merge checkpoints from  all workers to one checkpoint.
 
         # checkpoints from different workers may contain checkpoint for the same file,
         # so we sort checkpoints by time and update resulting collection with the same order.
-        checkpoints.sort(key=operator.itemgetter("time"))
+        found_checkpoints.sort(key=operator.itemgetter("time"))
 
-        for wc in checkpoints:
+        for wc in found_checkpoints:
```
czerwingithub, line 1188: Right now, it doesn't seem to me like there's anything that will delete the old style checkpoint files.  (i.e., after upgrade, what would delete the `checkpoints.json` and `active-checkpoints.json` files?  For that matter, if the number of workers decrease, the checkpoints for the previous workers will never be deleted.

I think to do some garbage collection, we should first write out the full checkpoint state to `checkpoints.json` and then delete all the other checkpoint files in this directory (both active and regular).  That way we at least try to clean up old state.

czerwingithub (changes requested):
Arthur, do you have another commit you haven't pushed up?  some of your responses to my comments say you have applied the fixes (such as changing the name away from shared memory) but I don't see the changes in the code.

scalyr_agent/agent_status.py
ArthurKamalov: The main idea here is to not print anything if there is a None. 
Some fields like the health check and response status are not determined immediately from the start, so we just do not show them while they are unknown.  
In the code above there is a case when not all responses are successful and we want to show unsuccessful ones.

But I have to agree that I could make things smaller by "worker_pool.all_health_checks_good is False"

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: The status code was changed a little bit. Please take a look. As I tried to say above. The idea is that `None` and `False` are not the same. 
In the case of `None`, we do not show anything until it becomes determined.

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: https://github.com/scalyr/scalyr-agent-2/pull/654#discussion_r528000579

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: Because I thought that the idea - is to not show the health check until it is not determined.

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: Same as here https://github.com/scalyr/scalyr-agent-2/pull/654#discussion_r528000579

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: There is a new member function called `__prepare_checkpoints`. If creates a consolidated master checkpoint file(including reading old-style checkpoints) and, after the creation, it removes all checkpoints(except master), including old-style ones. 
https://github.com/scalyr/scalyr-agent-2/blob/137cde17d5a53d14102a58b1baaec22a9d93c0b1/scalyr_agent/sharded_copying_manager/copying_manager.py#L1216

tests/ami/scripts/test_rpm.sh.j2
```diff
@@ -90,6 +90,8 @@ export INSTALL_PACKAGE_VERSION="{{install_package["source"]}}"
 
 {% include "partial/restart_agent_and_remove_logs.sh.j2" %}
 
+echo "111"
```
czerwingithub, line 93: Delete this and similar debug comments.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1137,20 +1192,66 @@ def _scan_for_new_logs_if_necessary(
...
+        checkpoints_dir_path = os.path.join(
+            self.__config.agent_data_path, "checkpoints"
+        )
+
+        # if there is a file called as the checkpoints folder, remove it
+        if os.path.isfile(checkpoints_dir_path):
+            os.unlink(checkpoints_dir_path)
+
+        # create new checkpoint folder if does not exist.
+        if not os.path.exists(checkpoints_dir_path):
```
czerwingithub, line 1232: Are there cases where this would happen?  In general, our install scripts (I believe) create the checkpoint directories.  So, we shouldn't have files with their names.  Or, am I missing something and you are trying to address something here?

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -1137,20 +1192,66 @@ def _scan_for_new_logs_if_necessary(
...
+        write_checkpoint_state_to_file(
+            checkpoints, master_checkpoints_path, time.time()
+        )
+
+        # clear the checkpoints folder by removing all worker checkpoint files.
+        checkpoints_glob = os.path.join(checkpoints_dir_path, "*checkpoints*.json")
+
+        # also remove old version checkpoints
+        for path in glob.glob(
+            os.path.join(self.__config.agent_data_path, "*checkpoints.json")
```
czerwingithub, line 1250: Do you need this separate glob check?  I believe `*checkpoints*.json` will also match `*checkpoints.json`.  I think we could simplify the code and just do the glob for `*checkpoints*.json`.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -198,30 +204,48 @@ def generate_status(self, warn_on_rate_limit=False):
...
-            # the next fields are a sum of the same fields from all workers.
             result.total_bytes_uploaded += status.total_bytes_uploaded
 
             result.total_errors += status.total_errors
 
+        # get overall information about last responses of all workers in the worker pool.
+        # making this as a set should help in further decision making process.
+        responses = {
+            # compare each response status for "success" and store the boolean result.
+            # NOTE: we do not compare None values, they are stored as are.
```
czerwingithub, line 219: I would add to this comment by saying why `None` matters.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -198,30 +204,48 @@ def generate_status(self, warn_on_rate_limit=False):
...
+            None
+            if s.last_response_status is None
+            else s.last_response_status == "success"
+            for s in worker_statuses
```
czerwingithub, line 223: I think this loop is a little confusing with the line breaks, etc.  Maybe rewrite it using the more verbose form where the `for s in worker_statuses` comes first.

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -837,37 +865,45 @@ def generate_status(self, warn_on_rate_limit=False):
...
+            elif None in responses:
+                # if there is at least one unknown response - a whole result is unknown too.
+                # NOTE: As the above condition implies, the result will be unknown unly if there is no failed responses.
+                result.last_responses_status_info = None
+            else:
+                result.last_responses_status_info = "All successful"
+
+            # Get health checks the same way as with responses.
+            health_checks = {s.all_health_checks_good for s in api_key_statuses}
+            if False in health_checks:
```
czerwingithub, line 899: Why isn't `result.health_check_result` set here?

scalyr_agent/sharded_copying_manager/copying_manager.py
```diff
@@ -198,30 +204,48 @@ def generate_status(self, warn_on_rate_limit=False):
...
+        responses = {
+            # compare each response status for "success" and store the boolean result.
+            # NOTE: we do not compare None values, they are stored as are.
+            None
+            if s.last_response_status is None
+            else s.last_response_status == "success"
+            for s in worker_statuses
+        }
+
+        if False in responses:
```
czerwingithub, line 226: I see at least four places where you have this pattern of accumulating the responses together, handling the failure case and None cases in this manner.  Maybe we should just have a helper method that does this.  Something like:

```
def accumulate_responses(responses, result_on_at_least_one_fail=False, result_on_all_succeed=True):
    if False in response:
        return result_on_at_least_one_fail
    elif None in responses:
        return None
    else:
         return result_on_all_succeeded
```

I think that would work here and below but give it a try.

czerwingithub (changes requested):
Here are my latest round of comments.  Once you've addressed these, I will go back and look at it as a whole and then look over the tests as well.

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: Because we also do a health check of the copying manager thread itself, and we collect all possible error messages in the `all_health_check_error_messages` list, and only then we set an overall message to the `result.health_check_result`

tests/ami/scripts/partial/ssl_checks.sh.j2
```diff
@@ -18,7 +18,11 @@ echo_with_date ""
 echo_with_date "Verifying agent status output"
 echo_with_date ""
 
+echo "888"
```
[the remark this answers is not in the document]
ArthurKamalov: Whoops, yes, I missed that, thanks.

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: Yes, I guess it should be better to create it on installation. I've modified the `build_packages.py` file for that. Thanks for the good point.

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: Thanks, that helped to reduce the amount of code.

scalyr_agent/sharded_copying_manager/copying_manager.py
ArthurKamalov: Yes, that makes things easier, thanks.

build_package.py
```diff
@@ -567,6 +568,7 @@ def build_common_docker_and_package_files(create_initd_link, base_configs=None):
     make_directory("root/etc/init.d")
     make_directory("root/var/log/scalyr-agent-2")
     make_directory("root/var/lib/scalyr-agent-2")
+    make_directory("root/var/lib/scalyr-agent-2/checkpoints")
```
czerwingithub, line 571: Ok, I missed this during the review.  You are changing the directory where the checkpoints are stored?  Any particular reason?  I would vote against this unless there is a really good reason.

czerwingithub (commented):
Everything essentially looks good.  I just have one comment about whether or not we really should be changing the checkpoint directory.

build_package.py
ArthurKamalov: Sorry, I forgot to say you that there was no checkpoints folder before this PR. There were just 2 files - `checkpoints.json` and `active-checkpoints.json` directly in the agent `data` folder. 
I introduced the `checkpoints` directory to not flood the data folder with the worker checkpoint files.

czerwingithub: Sorry, I didn't see this response until just now.  Still, this is a data folder just for the Scalyr Agent.  I think it is fine to put all of the checkpoint files in it.  Introducing another directory just adds a little more complication.  For example, K8s needs to mount the checkpoint directory to a persistent place on the system.  This probably wouldn't break this, but still, I'd rather not take the chance.

tests/unit/agent_status_test.py
```diff
@@ -355,11 +418,26 @@ def test_basic(self):
 (these statistics cover the period from Fri Sep  5 11:14:13 2014 UTC)
 
 Bytes uploaded successfully:               10000
-Last successful communication with Scalyr: Fri Sep  5 23:13:13 2014 UTC
-Last attempt:                              Fri Sep  5 23:13:13 2014 UTC
-Last copy request size:                    10000
+Last requests:                             All successful
 Health check:                              Good
 
+Api keys statistics
```
czerwingithub, line 424: I would change this to "Uploads statistics by API key:"

czerwingithub, line 424: Also, what happens in the common case where you only have one api key?  Do we still print these headers?  I would actually special case it so that it looks like how it does now.

tests/unit/sharded_copying_manager_tests/copying_manager_test.py
```diff
@@ -0,0 +1,353 @@
...
+        config_data=None,
+        disable_flow_control=False,
+    ):  # type: (int, bool, bool, Dict, bool) -> Tuple[Tuple[TestableLogFile, ...], TestableCopyingManager]
+
+        pipeline_threshold = 1.1
+        if use_pipelining:
+            pipeline_threshold = 0.0
+
+        initial_config_data = {
+            "debug_level": 5,
```
czerwingithub, line 98: Debug level 5 is fairly verbose.  Is that what you really want here (if it is, that's fine).  I just want to make sure it is not left over from some testing.

tests/unit/sharded_copying_manager_tests/common.py
```diff
@@ -0,0 +1,919 @@
...
+        Wraps the next 'self._manager.controller.wait_for_rpc(), returns lines instead' of requests alongside with
+        response  callback.
+        :param response: response for request.
+        """
+        (request_lines, responder_callback) = self._wait_for_rpc()
+        responder_callback(response)
+
+        return request_lines
+
+    def _append_lines_with_callback(self, *lines, **kwargs):
```
czerwingithub, line 162: Is there any particular reason you aren't explicitly listing the kwargs here?  I'd rather we list them if possible because it improves readability.

Also, can you provide documentation here.  It is not immediately clear to me what the callback is used for.

tests/unit/sharded_copying_manager_tests/common.py
```diff
@@ -0,0 +1,919 @@
...
+        # type: (*str, **Dict[str, Union[str, TestableLogFile]]) -> Tuple[List[str], Callable]
+
+        self._append_lines(*lines, **kwargs)
+        (request, responder_callback) = self._instance.controller.wait_for_rpc()
+        request_lines = self._extract_lines(request)
+        return request_lines, responder_callback
+
+    def _append_lines_and_wait_for_rpc(self, *lines, **kwargs):
+        # type: (*str, **Union[str, TestableLogFile]) -> List[str]
```
czerwingithub, line 171: Same here.  Can we explicitly list kwargs and add a little documentation.

tests/unit/sharded_copying_manager_tests/common.py
```diff
@@ -0,0 +1,919 @@
...
+        response = kwargs.pop("response", "success")
+
+        # set worker into sleeping state, so it can process new lines and send them.
+        self._instance.run_and_stop_at(TestableCopyingManagerThreadedWorker.SLEEPING)
+
+        self._append_lines(*lines, **kwargs)
+        request_lines = self._wait_for_rpc_and_respond(response)
+        return request_lines
+
+    def _append_lines_and_check(self, *lines, **kwargs):
```
czerwingithub, line 187: Same here.

tests/unit/sharded_copying_manager_tests/common.py
```diff
@@ -0,0 +1,919 @@
...
+class TestableApiKeyWorkerPool(ApiKeyWorkerPool):
+    def _stop_shared_object_managers(self):
+        """
+        We may need to interact with workers even after manager is stopped,
+         so do not stop the shared object managers.
+        """
+        pass
+
+    def really_stop_shared_object_managers(self):
+        return super(TestableApiKeyWorkerPool, self)._stop_shared_object_managers()
```
czerwingithub, line 668: Add documentation, why is this "really"

tests/unit/sharded_copying_manager_tests/common.py
```diff
@@ -0,0 +1,919 @@
...
+        # when it invokes `_send_events` to send a new request, and when it blocks to receive the response.
+        # These three states are referred to as 'sleeping', 'sending', 'responding'.
+        #
+        # The CopyingManagerWorker will have state to record where it should next block (i.e., if it should block at
+        # 'sleeping' when it attempts to sleep).
+        #
+        # We can manipulate this state, notifying changes on the conditional variable
+        # by using following methods: 'wait_for_rpc', 'perform_scan', 'perform_pipeline_scan'
+        # The CopyingManagerWorker will block in this state (and indicate it is blocked) until the
+        # we set a new state to block at.
```
czerwingithub, line 457: This looks like the same comment as above in `TestableCopyingManagerFlowController`.  I would just say see those comments here.

tests/unit/sharded_copying_manager_tests/common.py
```diff
@@ -0,0 +1,919 @@
...
+):
+    """An instrumented version of the CopyingManagerWorker which allows intercepting of requests sent, control when
+    the worker processes new logs, etc.
+
+    This allows for end-to-end testing of the core of the CopyingManagerWorker.
+
+    Doing this right is a bit complicated because the CopyingManagerWorker runs in its own thread.
+
+    Many members of this class are common with the TestableCopyingManager class
+    and mare oved to the base class - TestableCopyingManagerFlowController
```
czerwingithub, line 430: Some misspellings here.

tests/unit/sharded_copying_manager_tests/common.py
```diff
@@ -0,0 +1,919 @@
...
+    """
+
+    __test__ = False
+
+    # The different points at which the CopyingManager can be stopped.  See below.
+    SLEEPING = "SLEEPING"
+    SENDING = "SENDING"
+    RESPONDING = "RESPONDING"
```
czerwingithub, line 438: Can't these be inherited from `TestableCopyingManagerFlowController`?  It seems like you have a moderate amount of duplication even though you are inheriting from that class.

tests/unit/copying_manager_test.py
```diff
@@ -1497,181 +1430,25 @@ def test_whole_response_is_logged_on_non_success(self):
...
-        if "events" in parsed_request:
-            for event in parsed_request["events"]:
-                if "attrs" in event:
-                    attrs = event["attrs"]
-                    if "message" in attrs:
-                        lines.append(attrs["message"].strip())
-
-        return lines
+        return extract_lines_from_request(request)
+        # parsed_request = test_util.parse_scalyr_request(request.get_payload())
```
czerwingithub, line 1437: Remove this dead code?

tests/unit/copying_manager_test.py
```diff
@@ -256,41 +289,6 @@ def test_update_config_not_monitored(self):
         matchers = self._manager.log_matchers
         self.assertEquals(0, len(matchers))
 
-    def test_schedule_log_path_for_removal_and_re_add_before_actual_removal(self):
```
czerwingithub, line 259: Did this test case get moved somewhere else?

tests/unit/copying_manager_test.py
```diff
@@ -1049,6 +1040,8 @@ def test_pipelined_requests(self):
 
         responder_callback("success")
 
+    # TODO: confirm that this test only makes sense in the context of the single worker, and then remove.
+    @skipIf(True, "This test is moved to worker tests.")
```
czerwingithub, line 1044: I think it makes sense that you just have to test it in the single worker case.  We still need the test though.

tests/unit/copying_manager_test.py
```diff
@@ -1370,48 +1364,6 @@ def test_start_with_non_utf8_checkpoint(self):
         # start reading the logfiles from the end. In this case, that means lines three and four will be skipped.
         self.assertEquals(0, len(lines))
 
-    @skipIf(platform.system() == "Windows", "Skipping failing test on Windows")
```
czerwingithub, line 1373: Did this get moved somewhere else?

tests/unit/copying_manager_test.py
```diff
@@ -1497,181 +1430,25 @@ def test_whole_response_is_logged_on_non_success(self):
                 )
             finally:
                 if controller:
-                    controller.stop()
-
-    def test__read_checkpoint_state(self):
```
czerwingithub, line 1502: Can you just double check this case is covered somewhere. I suspect it is but just want to double check.

tests/unit/sharded_copying_manager_tests/copying_manager_worker_test.py
```diff
@@ -0,0 +1,695 @@
...
+
+            worker.start_worker(stop_at=None)
+
+            # wait for copying start in separate thread
+
+            wait_future = executor.submit(worker.wait_for_copying_to_begin)
+
+            # check for waiting is finished, it must not be finished until event is not set.
+            assert not wait_future.done()
+            time.sleep(0.1)
```
czerwingithub, line 205: Why sleep here like this?  It always worries me to see lines like this.  Normally the right thing to do is to sleep until some condition becomes true.

tests/unit/sharded_copying_manager_tests/copying_manager_worker_test.py
```diff
@@ -0,0 +1,695 @@
...
+            _, worker = self._init_worker_instance(
+                2, auto_start=False, disable_flow_control=True
+            )
+
+            worker.start_worker()
+            worker.wait_for_copying_to_begin()
+
+            # wait until worker is crashed.
+            while worker.is_alive():
+                time.sleep(0.1)
```
czerwingithub, line 249: Probably we need sometime out, otherwise the test is going to just hang forever if the worker never dies.

czerwingithub (changes requested):
@ArthurKamalov Here are my comments for the testing code.  I asked you to double check in a few places where it looks like test cases were removed.  I suspect most of those are covered (I did see some when I was reviewing later files) but it would be good to double check.

The biggest change I'm suggesting is to rename `ConfigBuilder`.  Hopefully you agree with that and it is not too much of a hassle.

scalyr_agent/sharded_copying_manager/worker.py
czerwingithub: Let's add this TODO and then follow up in a separate PR to fix any issues.  This is a little bit of a complicated issue and I want to look at the code a little more.

czerwingithub (changes requested):
Ok, here are my last batch of comments.  I've re-reviewed the meat of the code.  We now just need to work through my remaining comments.

scalyr_agent/agent_main.py
```diff
@@ -1783,30 +1716,30 @@ def __calculate_overall_stats(
...
-            delta_stats.total_request_time = (
-                current_status.copying_manager_status.total_request_time
-            )
-            delta_stats.total_pipelined_requests = (
-                current_status.copying_manager_status.total_pipelined_requests
-            )
-            delta_stats.rate_limited_time_since_last_status = (
-                current_status.copying_manager_status.rate_limited_time_since_last_status
-            )
+            # delta_stats.total_rate_limited_time = (
```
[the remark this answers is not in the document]
ArthurKamalov: TODO has been added.

tests/unit/copying_manager_test.py
ArthurKamalov: Yes, there is a `test_checkpoints` test case.
https://github.com/scalyr/scalyr-agent-2/blob/255f25119ece2093afb8a0eb97aaee09355e3ffe/tests/unit/sharded_copying_manager_tests/copying_manager_test.py#L331

tests/unit/copying_manager_test.py
ArthurKamalov: I've got it back. I've missed it somehow, thanks.

tests/unit/sharded_copying_manager_tests/common.py
```diff
@@ -427,18 +413,13 @@ class TestableCopyingManagerThreadedWorker(
...
     __test__ = False
 
-    # The different points at which the CopyingManager can be stopped.  See below.
-    SLEEPING = "SLEEPING"
-    SENDING = "SENDING"
-    RESPONDING = "RESPONDING"
-
     # To prevent tests from hanging indefinitely, wait a maximum amount of time before giving up on some test condition.
-    WAIT_TIMEOUT = 5.0
+    WAIT_TIMEOUT = 500.0
```
czerwingithub: Ok, this seems long.  Can we set it to 60?

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -1189,6 +1178,15 @@ def get_log_path(self):
         return self.__cached_log_path
 
 
+def change_agent_log(new_path):
```
czerwingithub, line 1181: I'm a little worried about relying on the original process to invoke this remote procedure.  Can we instead use `initializer` from the manager's `start` method?  See this documentation https://docs.python.org/2.7/library/multiprocessing.html#multiprocessing.managers.BaseManager.start

It does seem like we can to tell the subprocess to run a specific function upon start up.  It would make more sense to do it then.

Does that work?

czerwingithub (changes requested):
Ok, here are my comments to your latest round.  I just have one more sticky point in terms of how have the subprocess set the new logging destination.

scalyr_agent/agent_main.py
```diff
@@ -1717,12 +1717,13 @@ def __calculate_overall_stats(
             )
 
             # TODO: refactor this to be able to work with multi-worker copying manager.
+            # Those stats are now part of the CopyingManagerThreadedWorker's
+            # status and they do not make too much sense outside if it.
+            # Caw we just sum them? Maybe avarege? Or just remove them?
```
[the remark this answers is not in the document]
ArthurKamalov: There only those which I am not sure about. Are those sums informative and make any sense? But I think it would not be worse if we keep them as sums.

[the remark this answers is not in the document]
czerwingithub: Yes, the sums make sense for these stats.  I think we just go ahead and do it.

tests/unit/sharded_copying_manager_tests/common.py
ArthurKamalov: 5.0 should be fine. I guess this one was outdated because it's 5 now.

scalyr_agent/sharded_copying_manager/worker.py
ArthurKamalov: Sure, thanks for reminding me of that. I completely forgot about this feature. That should be exactly what is needed.

czerwingithub (commented):
@ArthurKamalov Ok, the latest changes look good to me.

scalyr_agent/sharded_copying_manager/worker.py
```diff
@@ -493,7 +488,10 @@ def run(self):
                                 )
 
                             # Check to see if pipelining should be disabled
-                            disable_pipelining = self.__has_pending_log_changes()
+
+                            # disable_pipelining = (
```
czerwingithub, line 492: Add a TODO to undo this commenting out.

czerwingithub (changes requested):

czerwingithub (commented):
Ok, the latest changes look good to me as well.

czerwingithub (approved):
Ok, thanks for doing these final renaming.  All looks good.  Let's get this sucker merged in.

PR #655, opened 2020-11-13
Fix invalid handling of bad / partial unicode sequences in the LineGrouper

Kami:
This pull request fixes ``LineGrouper`` class and makes sure we ignore any decoding errors we get when we try to decode bytes to utf-8 string.

Previously, this code would throw on partial / bad / corrupted unicode sequences.

## Background, Context

A user reported an issue with agent throwing ``UnicodeDecodeError`` inside the line matcher code.

After digging in, I saw our code doesn't correctly handle and ignore bad unicode sequences, but it fails on decode.

Keep in mind that this issue would only affect users which had line grouping rules defined.

## Open Questions

There are some more places in our code base where we rely on a default "strict" behavior of ``bytes.decode()``.

To be on the safe side, I think we should update all of those places to gracefully handle unicode decoded errors.

## TODO

- [ ] Changelog entry
- [ ] Update rest of the code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 Scalyr Agent 2 Changes By Release
 =================================
 
+## 2.1.15 "TBD" - November 20, 2020
+
+<!---
+Packaged by Tomaz Muraus <tomaz@scalyr.com> on Nov 20, 2020 14:00 -0800
+--->
+
+Bug fixes:
+* Fix line grouping code and make sure we don't throw if line data contains bad or partial unicode escape sequence.
+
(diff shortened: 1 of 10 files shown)


ArthurKamalov (approved):
LGTM. And yes, I am sure there should be more such code parts with the `strict` error policy.

Kami:
On that note - you could also argue that we should try to catch those exceptions in some places and log them (we do this in Redis monitor, for example).

The issue with that is that it may add overhead (would need to use rate limited log, but it could still potentially add a lot of overhead in hot / log matcher code path) and a lot of existing code doesn't do that.

PR #656, opened 2020-11-19
Fix run_monitor.py so it also works with source installation

Kami:
This pull request includes a small fix for``run_monitor.py`` so it also works when agent is not installed as Python package and when running it directly from source.

It uses the same pattern we use in ``scalyr_agent/agent_main.py``.

Original error:

```bash
  File "/home/kami/.pythonz/pythons/CPython-3.6.9/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/kami/.pythonz/pythons/CPython-3.6.9/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/run_monitor.py", line 57, in <module>
    from __scalyr__ import scalyr_init
ModuleNotFoundError: No module named '__scalyr__'
```


Kami:
It looks like invalid path is referenced in inline repo docs and docs on the scalyr website - fade70ee.

I will also open a PR for website fix at some point.

czerwingithub (approved):
I should be the one most knowledgeable about this, but I have to admit, it's been several years since I dug into the relevant code here.

I looked around a little bit to refresh my memory and it does seem to me like what you are doing is right.  Unfortunately, it's not that unbelievable that we've had a mistake like that in our documentation.

Kami:
While troubleshooting test failures (which looks like some race), I noticed we need to improve disk perf detection logic a bit more so we also report user-friendly error for write_bytes metric - 89e62433599e7fa5cf201e4a44c7f7fe330332ef.

PR #657, opened 2020-11-23
Better handling of default agent config file permissions

Kami:
This pull request fixes deb and rpm packages to make sure that the ``agent.json`` config file which is installed and ``agent.d/`` directory itself is not readable by others by default (aka permissions in octal format are ``0`` for "other" users).

## Background, Context, Impact

Currently, ``agent.json`` file is readable by others by default which means other users which have access to the server where the agent is running could potentially read that config file.

Keep in mind that this only impacts regular server installations and not container and other "single user" deployments (Kubernetes, Docker, etc.). And even in those cases, user would need to have access to that server to potentially be able to read / view the config file (and likely the user who has access to the server already had a way to achieve that one way or another).

In addition to that, all the serious and large deployments usually use some kind of configuration management system (puppet, ansible, chef, etc.) to manage configs such as agent.json. In those cases, permissions bit are usually also explicitly managed and set to the least privileged approach (well, at least that's the best practice and most serious users follow it) so those installations are not affected - and that's still the best / recommended practice going forward.

## Proposed Fix

In my PR I fix it for the new installations inside ``build_package.py``. That change is relatively straight forward.

Other part of the change is "rectifying" that permissions bit for the existing installations. This part is more complex because there are more possible scenarios and edge cases.

One option would simplify be to chmod agent.d with 741 and *.json config files with 640.

This should work fine, but it may overwrite any changes to the permissions the user has made to the "owner" and "group" part. 

Because of that, I went with a slightly more complex approach where we preserve the existing owner, group permission bit and only change the "other" one to "0" for files and "1" for directories (agent.d/).

In theory, straight chmod should be fine as well, especially because as mentioned above, most serious shops explicitly manage config files content and permissions using some kind of configuration management tool (and that always happens post installation aka after postinst run).

It's also worth explicitly noting that this only fixes it for new installations and existing one with agent.json file. If user performs a fresh installation, deletes the config and then manually creates it - well, that's domain of ``umask`` and out of our scope and there is really not much we can or should do in those cases.

## TODO 

- [x] Fix permissions for new / clean installations
- [x] Rectify permissions for existing installations on postinst
- [x] Add release notes entries (document the behavior / change, info on impact, best practices going forward, etc.).
- [x] Decide on how to handle this on Windows and implement the changes needed
- [x] Update agent to warn on startup if ``agent.json`` is readable by "others" aka non-owner / group
- [x] Update agent to warn if config is not owned by the user under which the agent is running (this is already done when not using ``--no-change-user``, but not when using that flag)


build_package.py
```diff
@@ -294,11 +294,11 @@ def build_win32_installer_package(variant, version):
...
 
     # Copy the config file.
+    agent_json_path = make_path(agent_source_root, "config/agent.json")
     cat_files(
-        [make_path(agent_source_root, "config/agent.json")],
-        "agent_config.tmpl",
-        convert_newlines=True,
+        [agent_json_path], "agent_config.tmpl", convert_newlines=True,
     )
+    os.chmod(agent_json_path, int("640", 8))
```
czerwingithub, line 301: I would add a comment here that we are explicitly setting the "other" permission to zero to protect sensitive informatin in the config, or something like that.  I'd hate for someone who can't read chmod to change that without realizing how important it is.

build_package.py
```diff
@@ -338,6 +338,8 @@ def build_win32_installer_package(variant, version):
     make_directory("Scalyr/logs")
     make_directory("Scalyr/data")
     make_directory("Scalyr/config/agent.d")
+    os.chmod("Scalyr/config/agent.d", int("740", 8))
```
czerwingithub, line 341: Same here, let's make a comment so that folks know they shouldn't be mucking with it.

build_package.py
```diff
@@ -585,6 +587,7 @@ def build_common_docker_and_package_files(create_initd_link, base_configs=None):
 
     # Make sure there is an agent.d directory regardless of the config directory we used.
     make_directory("root/etc/scalyr-agent-2/agent.d")
+    os.chmod("root/etc/scalyr-agent-2/agent.d", int("740", 8))
```
czerwingithub, line 590: Comment here as well.

installer/scripts/postinstall.sh
```diff
@@ -80,19 +78,68 @@ check_python_version() {
...
+# Ensure agent.d/*.json files are note readable by others
+# NOTE: Most software gives +x bit on *.d directories so we do the same
+ensure_path_other_permissions "/etc/scalyr-agent-2/agent.d" "1"
+
+if [ -d "/etc/scalyr-agent-2/agent.d" ]; then
+  # NOTE: find + -print0 correctly handles whitespaces in  filenames and it's more robust than for,
+  # but it may have cross platform issues. In that case we may need to revert to for.
+  #for config_fragment_path in /etc/scalyr-agent-2/agent.d/*.json; do
+  find /etc/scalyr-agent-2/agent.d/ -name "*.json" -print0 | while read -r -d $'\0' config_fragment_path; do
+    ensure_path_not_readable_by_others "${config_fragment_path}" " 0"
```
czerwingithub, line 139: Should there be a space between the quote and the `0`?

czerwingithub (changes requested):
The general approach looks good to me.  I have a few small comments, maybe one that catches a bug.

build_package.py
Kami: Agreed - there are also tests which catch this issue.

installer/scripts/postinstall.sh
Kami: Yep, this was fixed in a later commit. And I will also add additional guard to the ensure function to validate the argument which is passed in.

Kami:
Just to add a quick update on the progress.

There turned out to be more edge cases.

One of the is that fpm is using ``771`` as a default permission for ``agent.d/`` directory which is not great.

I tried to solve it on ``build_package.py`` / fpm level with ``--rpm-defattrdir``, but sadly this breaks permissions for other directories such as ``/etc/init.d`` so in the end I went with ``postinst`` approach where we fix permission for that directory to ``751`` - aka we also change the permission bit for group and not just the owner.

Kami:
I got all the fixes to work, but after some more testing I decided to not run that config fix on Windows automatically post installation / upgrade - 6a777a37ef526321158bbb991c3cedd97b7d702d.

This fix could break installations where the agent is running as a user which is not Administrator or not part of the Administrators group.

I think this is likely sufficiently rare, but until we hear back on this from some of our Windows users, it's safer to just document the fix to avoid potentially breaking some installations.

Kami:
@czerwingithub As discussed, I added a change to include information if the agent is running as privileged user to the ``User-Agent`` string.

In addition to that, I replaced ``ssllib`` string with openssl version used by the agent.

After removing ``tlslite`` support, ``ssllib`` string provides no value any more and including OpenSSL version used by the agent / Python interpreter makes more sense (it makes it easier to us to determine if it makes sense to also support TLS v1.3 and similar - technically, we could also get some of that info from nginx logs, but that's more complicated and noisy).

I believe this may require us to update some Power Queries / Dashboard on the server side though.

Currently the fragment is called ``admin-1|0``, but if we wanted to save some bytes, we could just include ``admin`` string or similar if running as admin user and nothing if not. I don't think it would make much difference on the server / query side.

@czerwingithub Do you have any preference on this?

CHANGELOG.md
```diff
@@ -1,17 +1,23 @@
...
+Misc:
+* On startup and when parsing a config file, agent now emits a warning if file is readable by others.
+
 Bug fixes:
 * Fix line grouping code and make sure we don't throw if line data contains bad or partial unicode escape sequence.
 * Fix ``scalyr_agent/run_monitor.py`` script so it also works correctly out of the box when using source code installation.
 * Update Windows System Metrics monitor to better handle a situation when disk io counters are not available.
 
+Security fixes and improvments:
+* Agent installation artifacts have been updated so the default ``agent.json`` file which is bundled with the agent is not readable by "other" system users by default anymore. For more more context, details and impact, please see [RELEASE_NOTES](https://github.com/scalyr/scalyr-agent-2/blob/master/RELEASE_NOTES.md).
```
czerwingithub, line 19: remove repeat of `more`

scalyr_agent/config_main.py
```diff
@@ -1427,7 +1428,22 @@ def make_symlink(source, target):
...
+        # permissions inside the .wxs wix spec file.
+        # Right now it can only be used with "--init-config" flag on Windows.
+        parser.add_option(
+            "",
+            "--fix-config-permissions",
+            dest="fix_config_permissions",
+            action="store_true",
+            default=False,
+            help=(
+                "Fix permissions for agent.json file and agent.d/ directory and make sure it's. "
```
czerwingithub, line 1443: I would mention in the help string that this only applies for windows.

czerwingithub (commented):
I just had two minor documentation nits.  I know you've closed this one off so feel free to just do another commit for these and push it directly in.

scalyr_agent/config_main.py
Kami: Technically, we could only add this option is platform is windows, WDYT?

I think it should be fairly safe.

Kami: Made that change in - a0ee41bd.

When we check for that option we have a platform starts with win guard / check so it should be safe.

PR #659, opened 2020-11-28
Add support for ignoring / excluding mount points to the Linux system metrics monitor

Kami:
This pull request adds new ``ignore_mounts`` configuration option to the Linux system metrics monitor which allows user to specify which mount point to ignore and not report metrics for.

## Background, Context

On modern Linux systems (think recent versions of RHEL, Fedora, Ubuntu, Debian, etc.), system contains many mount points for special file systems such as cgroup namespaces, systemd mounts, etc.

Those mount points are special so standard inode and block size related metrics we collect in the agent are really of no use.

This results in a big churn of metrics which will likely never be used.

Here is an example of mount points on such system:

```bash
/dev/mmcblk0p2 on / type ext4 (rw,noatime,nodiratime)
devtmpfs on /dev type devtmpfs (rw,relatime,size=3900372k,nr_inodes=85483,mode=755)
sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
proc on /proc type proc (rw,relatime)
securityfs on /sys/kernel/security type securityfs (rw,nosuid,nodev,noexec,relatime)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev)
devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000)
tmpfs on /run type tmpfs (rw,nosuid,nodev,mode=755)
tmpfs on /run/lock type tmpfs (rw,nosuid,nodev,noexec,relatime,size=5120k)
tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,mode=755)
cgroup2 on /sys/fs/cgroup/unified type cgroup2 (rw,nosuid,nodev,noexec,relatime,nsdelegate)
cgroup on /sys/fs/cgroup/systemd type cgroup (rw,nosuid,nodev,noexec,relatime,xattr,name=systemd)
none on /sys/fs/bpf type bpf (rw,nosuid,nodev,noexec,relatime,mode=700)
cgroup on /sys/fs/cgroup/perf_event type cgroup (rw,nosuid,nodev,noexec,relatime,perf_event)
cgroup on /sys/fs/cgroup/net_cls,net_prio type cgroup (rw,nosuid,nodev,noexec,relatime,net_cls,net_prio)
cgroup on /sys/fs/cgroup/cpu,cpuacct type cgroup (rw,nosuid,nodev,noexec,relatime,cpu,cpuacct)
cgroup on /sys/fs/cgroup/blkio type cgroup (rw,nosuid,nodev,noexec,relatime,blkio)
cgroup on /sys/fs/cgroup/freezer type cgroup (rw,nosuid,nodev,noexec,relatime,freezer)
cgroup on /sys/fs/cgroup/pids type cgroup (rw,nosuid,nodev,noexec,relatime,pids)
cgroup on /sys/fs/cgroup/cpuset type cgroup (rw,nosuid,nodev,noexec,relatime,cpuset)
cgroup on /sys/fs/cgroup/devices type cgroup (rw,nosuid,nodev,noexec,relatime,devices)
systemd-1 on /proc/sys/fs/binfmt_misc type autofs (rw,relatime,fd=32,pgrp=1,timeout=0,minproto=5,maxproto=5,direct)
debugfs on /sys/kernel/debug type debugfs (rw,relatime)
mqueue on /dev/mqueue type mqueue (rw,relatime)
sunrpc on /run/rpc_pipefs type rpc_pipefs (rw,relatime)
configfs on /sys/kernel/config type configfs (rw,relatime)
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,noatime,size=256000k)
tmpfs on /var/log type tmpfs (rw,nosuid,nodev,noexec,noatime,size=307200k,mode=755)
/dev/mmcblk0p1 on /boot type vfat (rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,errors=remount-ro)
tmpfs on /run/user/1000 type tmpfs (rw,nosuid,nodev,relatime,size=806492k,mode=700,uid=1000,gid=1000)
```

## Proposed Change

To address this issue, I added new ``ignore_mounts`` config option to the Linux system metrics monitor which can contain a list of glob-style patterns of mount points to exclude / ignore.

Here is an example configuration:

```javascript
    monitors: [
        {
            module: "scalyr_agent.builtin_monitors.linux_system_metrics",
            ignore_mounts: ["/sys/*", "/dev/*", "/run/*", "/boot/*"],
        }
    ],
```

Keep in mind that tcollector module already contains some ignore rules based on the file system type (``tmpfs``, ``devfs``), but those don't cover all the cases.

Config option value defaults to an empty list so it's fully backward compatible.

Going forward though, I do think it would be a good idea to ignore some of those mount points by default though (at the very least, cgroup ones - we can either do it on fstype or based on the actual mount point path).

Here is an example of number of lines / metrics without this change and with filters defined per the sample config above.

Before

```bash
kami ~$ cat ~/scalyr-agent-dev/log/linux_system_metrics.log | grep "df." |wc -l
66
```

After

```bash
kami ~$ cat ~/scalyr-agent-dev/log/linux_system_metrics.log | grep "df." |wc -l
33
```


Kami:
I also done quick search on one of infra environments and excluding /sys/*, /dev/* and /run/* mount points, we would be able to reduce number of Linux system metrics log lines for around 130k per hour (basically reduce number of log lines for df. metrics by more than half).

Kami:
And if indeed feel comfortable that we should ignore this by default already, we can set default ``ignore_mounts`` config options to this value and instruct users on how to change / revert that behavior in RELEASE_NOTES by changing that monitor config option.

Kami:
Based on discussion with @czerwingithub, I will update the defaults to exclude those special mount points and document in the release notes on how to change or re-enable monitoring for those mounts if necessary.

czerwingithub (approved):

PR #660, opened 2020-11-28
Add support for adding random jitter to the monitor sample gather interval

Kami:
This pull request is a WIP proposal to add a global monitor wide configuration options which adds random jitter to the monitor sample gather time interval.

## Background, Context

Right now on startup agent starts configured monitors at about the same time which means ``gather_sample()`` for all of them will also run at about the same time as well.

In most cases this is not an issue, but when agent is running on more lower powered devices (think Raspberry Pi) and is configured with many monitors, this could cause relatively large agent CPU consumption / load spikes when ``gather_sample()`` is called for all the configured monitors.

## Proposed Change

In this change I add new ``global_monitor_sample_interval_enable_jitter`` option. When this option is enabled, random jitter is added to each monitor sample gather interval (right now that's between 1/10 and 2/10 of the configured monitor sample interval, but we could make this adjustable via config as well).

This random jitter means there is a less of a chance sample gathering for all the monitors will run at the same time (it will be spread out more) which should cause smaller / less CPU usage spikes for the agent process when configured with more monitors.

There are also other more complex possible approach, but in the past, in similar scenarios, I used a similar random jitter approach which worked good enough.

## TODO

- [x] Decide if we want a feature like that
- [x] Figure out if and how it the roll ups / aggregation on the server side
- [x] We may also want a black list for monitors where we don't want to use this jitter (e.g. kubernetes)


scalyr_agent/configuration.py
```diff
@@ -1709,6 +1716,14 @@ def __verify_main_config_and_apply_defaults(
             apply_defaults,
             env_aware=True,
         )
+        self.__verify_or_set_optional_bool(
+            config,
+            "global_monitor_sample_interval_enable_jitter",
+            False,
```
czerwingithub, line 1722: Maybe we set this to `True` by default?  It seems to me that this falls into that category of "turning it on shouldn't negatively impact anyone" and "if you need this, it is going to be hard to find this option".

scalyr_agent/scalyr_monitor.py
```diff
@@ -270,24 +271,32 @@ def run(self):
         # noinspection PyBroadException
         try:
             while not self._is_thread_stopped():
+                sample_interval = self._sample_interval_secs
+                if self._adjust_sleep_by_gather_time:
+                    random_jitter = 0
+                elif self._global_config.global_monitor_sample_interval_enable_jitter:
+                    # min jitter is 1/10 of the sample interval and max is 2/10
```
czerwingithub, line 278: Eh, I might do 8/10 for the max.  Delaying by one sample period should not be a big deal.

czerwingithub (changes requested):

scalyr_agent/scalyr_monitor.py
Kami: Yeah, I think this should be reasonable, especially for "normal" sample intervals (think 30 or 60 seconds).

If someone intentionally sets it to something large (e.g. 15 or 30 minutes), it may be a bigger issue. We could add some kind of upper bound for it (e.g. 1 minute), but I also don't think many people actually use that long intervals.

scalyr_agent/configuration.py
Kami: Yeah, I agree.

I may just wait till the next release so we have a chance to run it on staging for a while just to be sure.

scalyr_agent/scalyr_monitor.py
czerwingithub: Good idea about capping to a maximum size.

Kami:
@czerwingithub This is ready for a review now.

I decided to leave it off by default for now. In the future once it's running on staging for a while we can turn it on by default if there are no issues.

scalyr_agent/scalyr_monitor.py
```diff
@@ -250,6 +256,33 @@ def config_from_monitors(self, manager):
         """
         pass
 
+    def _get_sleep_delay(self):
```
czerwingithub, line 259: Ok, I thought this was just being added to the first sleep time, but now I see that it is being added on every sleep.  Yeah, that makes it more invasive and could hurt some of the dashboards.  (Some of the dashboards based on metrics emitted by the monitors do things like calculate the average value over a minute interval... if we keep do this, then in some intervals it will get two samples vs none in others.)

I think it would be sufficient to just delay the sleep for the first sample.  I.e, right after the agent starts, instead of having all of the monitors start collecting right then, we jitter each one so their start time is spread out over a 60 sec interval.  After that, they each sleep the normal sample interval.  So, the frequency is still fixed at 60 seconds but the phase is offset for all of the monitors.

What do you think of that?

Kami: Yeah, I think this should also work as long as we do it on on monitor init aka when monitors are started / restarted.

Kami: Alright, I pushed a change so we only sleep before first sample gather interval and after that we used fixed interval - 09d5e4a34797f8c5576deeb3ede18f0d810f94f7. 

I thought that adding random jitter to each sleep should work as well (as long as it's smaller than minimum roll up / aggregation window we use on the server side), since server should already take such things (time drifts, delayed metric, missed samples, etc.) into account, but yeah, this approach is probably indeed safer.

And since we only use this sleep before first sample gather interval, I'm also fine with enabling this option by default - it should have no negative impact or consequences.

@czerwingithub  - WDYT about enabling it by default? I would still leave config option / flag there to disable it, if needed.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Improvements:
 
 Misc:
 * On startup and when parsing a config file, agent now emits a warning if the config file is readable by others.
+* New ``global_monitor_sample_interval_enable_jitter`` config option has been added which is enabled by default. When this option is enabled, random sleep between 2/10 and 8/10 of the configured monitor sample gather interval is used before gathering the sample for the first time. This ensures that sample gathering for all the monitors doesn't run at the same time. This comes handy when running agent configured with many monitors on lower powered devices to spread the monitor sample gathering related load spike across a longer time frame.
```
czerwingithub, line 16: s/comes handy/comes in handy/

czerwingithub (changes requested):
One minor documentation nit, but beyond that, everything looks good.

As for enabling by default, yes, I think with the change as it is now, we should just enable it by default.  Like we discussed earlier, this is one of those things that would be hard for someone to find if they needed it.

CHANGELOG.md
Kami: Fixed in a42bf3d90548a1c166f24c5860f1dfd9c51d5f85.

PR #661, opened 2020-12-01
Fix failing AMI tests

Kami:
This pull request fixes failing AMI tests.

It updates one of the checks to use a specific CA bundle to validate the cert which only contains Scalyr CA info. This means if we try to use that CA and validate any other cert, it should fail.

Previously we relied on example.com having SSL cert issued by ca cert which we don't trust and that's not really the best and most robust approach :)

Ideally we would just spin a up mock http server inside AMI tests with self signed cert for that test scenario, but that would be more involved.


Kami:
There also seems to be some upstream networking related issue - AMI tests started to take twice as long as usually (usually then finish in sub 5 minutes, now they take almost 10 minutes)

For now, I bumped the timeout and when the things stabilize, I will decrease it.

IIRC, we had a similar issue in the past. If I recall correctly, it was indeed related to some upstream apt / yum repo timeout or similar.

And I do see some cloudfront yum related errors in the logs (it does line up since amazonlinux 2 and centos 8 tests are timing out).

PR #664, opened 2020-12-07
Fix copying manager config issues

ArthurKamalov:
* Expose missing methods for the LogFileProcessor proxy.
* Fix load config from config fragment files (e.g. /agent.d/*.json)

TODO: 
* Anable process metrics monitor for workers' child processes.


scalyr_agent/configuration.py
```diff
@@ -367,12 +372,118 @@ def parse(self):
...
+                        "The default api key is redefined. Please use only unique api keys."
+                        "The entries without 'api_key' field are also considered as default",
+                        "api_keys",
+                        "apiKeyDuplication",
+                    )
+
+        unique_keys = set()
+        other_api_key_configs = list()
+        # If there are multiple api key entries with the same api_key, then the duplication error is raised too.
+        for key_config in api_keys:
```
czerwingithub, line 411: As discussed, change this to exclude entries with the same `id` and not `api_key` value.

scalyr_agent/configuration.py
```diff
@@ -367,12 +372,118 @@ def parse(self):
...
+        Verify all api key config entries from the "api_keys" list in config.
+        """
+        api_keys = self.__config.get_json_array("api_keys")
+        default_key_config = None
+        for key_config in api_keys:
+            api_key = key_config.get("api_key", none_if_missing=True)
+            # if the 'api_key' from this entry equals to the 'main' api key,
+            # than this api key entry is considered as default.
+            if api_key is None or api_key == self.api_key:
+                if default_key_config is None:
```
czerwingithub, line 396: As discussed, we don't want to penalize the user if they are re-using the same api key.  HOWEVER, this code here does raise an interesting case.  If the user is trying to assign the "default" by just having it match the by api key, then there is ambiguity if there are multiple api keys with the same value.

1.  Loop through, finding first entry that doesn't have an `api_key`.  Assign that to be `default.`  If there is more than one entry missing `api_key`, throw an error.
2. If we didn't find a default key that way, then loop through, finding first entry whose `api_key` matches the main `api_key`.  Assign that to `default`.  If there is more than one, then throw an error.

What I'm trying to accomplish here is that we do not throw an error if the same `api_key` is used multiple times as long as only one entry does not contain the `api_key` field.

Let me know if that does not make sense.

scalyr_agent/configuration.py
```diff
@@ -3071,17 +3139,18 @@ def __verify_log_entry_and_set_defaults(
...
-            log_entry,
-            "api_key_id",
-            default_api_key_id,
-            description,
-            valid_values=api_key_ids,
-        )
+        # Because the log file are verified earlier than api keys we can not verify the default api_key_id field there,
+        # and this is done in the '__verify_and_match_api_keys_in_logs' function,
+        # when the default api key is already known.
+        # But we still need to verify api_key_id in case if log file was added after the config was verifyed.
```
czerwingithub, line 3145: s/verifyed/verified/

scalyr_agent/copying_manager/worker.py
```diff
@@ -1156,6 +1156,10 @@ def FixedRebuildProxy(func, token, serializer, kwds):
         six.ensure_str("add_missing_attributes"),
         six.ensure_str("get_log_path"),
         six.ensure_str("generate_status"),
+        six.ensure_str("add_sampler"),
+        six.ensure_str("add_redacter"),
+        six.ensure_str("skip_to_end"),
+        six.ensure_str("set_new_scalyr_client"),
```
czerwingithub, line 1162: Is this one really invoked across the mutliprocess boundary?  I thought the workers create their own scalyr client instance.

czerwingithub (changes requested):

scalyr_agent/configuration.py
ArthurKamalov: Yes, it totally does. But I think I have one option to suggest. What if we do allow users to re-use the api key but, instead of that, we do not allow leave the 'api_key' field empty? (In other words, make the  'api_key' field mandatory/required) I think it should make things easier.

Here are the cases:

1) If there are no api key entries  with the default api key, then the default api key entry is created automatically, based on another default options(such as `default_workers_per_api_key`)
2) If the user adds one(first) entry with a default api key, then it is considered as he wants to "override" the default values from the first case.
3) The second(and subsequent) api key entry with the default api key, as we discussed, is to "fine-tune" the load on workers.

czerwingithub: What if we just make this simpler and require both `api_key` and `id`.  We only look at `id == "default"` to say this is the default entry.  We throw an error if that entry's `api_key` does not match the main api key's value.

tests/unit/configuration_test.py
```diff
@@ -3095,3 +3098,141 @@ def test_default_workers_per_api_key(self):
...
+            ]
+          }
+        """
+        )
+
+        config = self._create_test_configuration_instance()
+        with pytest.raises(BadConfiguration) as err_info:
+            config.parse()
+
+        assert "Value must be greater than 1" in err_info.value.message
```
Kami, line 3238: Oops, I think configuration.py needs to be updated to say "Value must be greater than or equal to ...". Same goes for "max_value" (it's just a bug / typo in that error message).

Kami: 64db8892

Kami:
I spent some more time testing the changes using this PR.

As part of that, I pushed a couple of small changes / improvements:

1. I noticed we don't mask the API key in the ``api_keys`` config value when logging raw config values on startup if debug_level is >= 5 (minor thing)
2. I updated copy manager logging code to also log PID of all the spawned worker processes when multi processing mode is used. I believe we will need similar code anyway for supporting process monitor per-worker in the future.
3. Added small change so we log ``api_keys`` configuration value as part of "useful" config setting (this should make it easier for us to troubleshoot things when looking at the agent logs). Actual API key values in that object are masked / removed (in the future, we may want to look last couple of characters to make troubleshooting easier, but that's TBD).

Tomorrow I also plan to spend some more time trying to "automate" some of the tests I used for testing the checkpoints + workers stuff. I don't think we have any specific broad end to end tests like that so having some would be a good idea.

Basically those checks do something along the lines of:

1. Start the agent with multi process and multiple workers enabled for a file
2. Another process is writing data to a file which is consumed by multiple workers
3. Another process is randomly changing number of workers every X seconds to ensure scaling up and down + checkpoints work fine (log config reload interval is used so we can do that more often)
4. Every X seconds, also restart the agent
5. Verify all data makes it to Scalyr and there are no duplicated lines.

scalyr_agent/configuration.py
```diff
@@ -386,60 +386,117 @@ def __verify_api_keys(self):
...
+        unique_api_ids = {}
+        # Apply other defaults to all api key entries
+        for i, api_key_entry in enumerate(api_keys):
+            self.__verify_api_keys_entry_and_set_defaults(api_key_entry)
+            api_key_id = api_key_entry["id"]
+            if api_key_id in unique_api_ids:
                 raise BadConfiguration(
-                    "The api key with api_key_id '%s' is redefined in next entries. Please use only unique api keys."
-                    % key_id,
+                    "The entry field 'api_key_id' of the list 'api_keys' has to be unique.",
```
czerwingithub, line 401: We should include the specific value of `api_key_id` that caused a duplicate in the error message.

Also, because you set `"default"` as the id for any entry that doesn't have an api key set, this error message could be confusing if there are multiple entries that do not have the `id` field specified.   Maybe you should special case the error message for `"default"` and say either there are multiple entries that have `"default"` or mulitple entries that do not have `id` defined.

tests/unit/copying_manager_tests/common.py
```diff
@@ -938,6 +938,8 @@ def workers_log_processors_count(self):
         six.ensure_str("get_checkpoints_path"),
         six.ensure_str("_init_scalyr_client"),
         six.ensure_str("change_agent_log"),
+        six.ensure_str("get_checkpoints"),
+        six.ensure_str("write_checkpoints"),
```
czerwingithub, line 942: I'm still curious as to why `_init_scalyr_client` has to be proxied, and now also curious about `get_checkpoints` and `write_checkpoints`.  Do those all have to be invoked across the multiprocess boundary?

czerwingithub (changes requested):

tests/unit/copying_manager_tests/copying_manager_test.py
```diff
@@ -48,898 +43,174 @@
...
-            [0.36, ALMOST_SIX_MB],
-            [0.216, ALMOST_SIX_MB],
-            [0.1296, ALMOST_SIX_MB],
-        )
+    def __extract_lines(self, request):
+        return self._extract_lines(request)
 
-    def test_request_too_big(self):
-        self.test_params.current_sleep_interval = 1
+    def assertEquals(self, expected, actual):
```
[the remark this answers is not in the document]
ArthurKamalov: This was made to be able to run those tests on `pytest`  but without spending time on rewriting the tests cases code.
The reason for that is that the `pytest` is able to run `unittest` test cases but it can not parameterize them.
I've also left the TODO to rewrite it and to remove those methods.
https://github.com/scalyr/scalyr-agent-2/blob/91aa95fb47e1120aef1fb4f4dac487599285aff9/tests/unit/copying_manager_tests/copying_manager_test.py#L104-L114

scalyr_agent/configuration.py
ArthurKamalov: Sorry, I forgot to answer this comment https://github.com/scalyr/scalyr-agent-2/pull/664#discussion_r537845613
but I've made exactly as said there. 

> Also, because you set "default" as the id for any entry that doesn't have an api key set, ...
Now it just throws an error if the `api_key` file is missing. The `api_key` and `id` fields are required now

It is also impossible to have a duplication error with two `default` ids unless the user does not add them explicitly, because nothing is created automatically in this case and nothing can confuse the user.
If it adds only one default key, then everything is fine. In this case, we just do not create the default key automatically.

Does that make sense to you?

PS: I've also changed the error message to
 
_"The api key #0 already has id 'some id' but it has also been re-used by the api key #1. Api key id's must remain unique."_

`#0`, `#1` - positions in the `api_keys` list

tests/unit/copying_manager_tests/common.py
ArthurKamalov: Oh, right, it shouldn't be there. But we have to expose other functions in order to be able to test multi-process workers. Those methods are only members of the "testable" subclass of the worker to make testing easier.

scalyr_agent/configuration.py
```diff
@@ -386,60 +386,117 @@ def __verify_api_keys(self):
...
+
+        default_api_key_entry = unique_api_ids.get("default")
+
+        if default_api_key_entry is None:
+            default_key_entry = JsonObject(api_key=self.api_key, id="default")
+            self.__verify_api_keys_entry_and_set_defaults(default_key_entry)
+            api_keys.insert(0, default_key_entry)
+            self.__config.put("api_keys", JsonArray(*api_keys))
+
+        # for i, key_config in enumerate(api_keys):
```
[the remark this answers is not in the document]
ArthurKamalov: Yes. I forgot about it, sorry.

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -146,6 +156,27 @@ def api_key_id(self):
...
+        if not self.__shared_object_managers:
+            return []
+
+        result = []
+
+        for shared_object_manager in self.__shared_object_managers.values():
+            try:
+                result.append(shared_object_manager._process.ident)
+            except Exception:
+                pass
```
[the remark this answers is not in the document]
Kami: Technically it's an internal non-public API attribute so I added it as a safe guard for just in case (and IIRC, it's not available in Python 2.6).

I can add the comment to that line.

scalyr_agent/configuration.py
czerwingithub: I don't think using the indexes will be that meaningful in all cases.  Remember, those key entries could be created by joining together the information in across multiple `.json` files.  So, I think we should just say there are entries with duplicate ids and give the id name.

Ok, I saw that there is code down below in the `verify` method that adds in `default` if the `id` field is not specified.  Maybe that was dead code and you meant to remove it.  But yes, that makes sense.

scalyr_agent/copying_manager/copying_manager.py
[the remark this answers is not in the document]
ArthurKamalov: The only reason that I see is if `shared_object_manager` does not have a `_process` attribute.  I guess Tomaz had meant this.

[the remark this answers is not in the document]
Kami: Yep and as a safe-guard for "future proofing" since we are accessing "internal" attribute (sadly there is no public interface to get this info, afaik).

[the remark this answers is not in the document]
ArthurKamalov: @Kami sorry, I haven't seen your first comment. It hasn't been shown without page refresh for some reason.

czerwingithub (approved):
I think there might be a few comments from Tomas and we still think we should remove the index position from the dup key id .  But in the interest of unblocking I'll approve and we can follow up in a separate PR

PR #665, opened 2020-12-09
Add workaround for occassional AMI test failure on CI during chmod postinst phase

Kami:
This pull request is an attempt to get to the bottom of the ``getpwuid(): uid not found: 1001`` failure we obsessionally see on CI on package post install phase where we chmod the config files.

We already have ``chmod || true`` pattern in the code so that step should not fail even if ``set -e`` is used. I also can't reproduce the issue locally in any way.

I still believe the root cause is related to partial / unfinished AMI provisioning, hopefully new log statements will help us pinpoint the issue.


Kami:
OK, I re ran ami-tests-development many times now and it did never fail with that error.

I will merge this into master and keep an eye if that error appears again at some point.

ArthurKamalov (approved):
LGTM

PR #666, opened 2020-12-10
Make sure Docker monitor also ingests container stderr logs when running in non-raw log file mode

Kami:
A while back in #647, I fixed "docker_raw_logs: false" functionality of Docker monitor which has been broken for a while now.

Looking at the commit history, it appears that bug has been inadvertently introduced in cb4f9030455875a8b4619b794736a55a988436b7.

When fixing that functionality for #647, I originally thought we intentionally only retrieve stdout so I didn't add tests for stderr code path, but do think we want to ingest stderr as well since we do that when running in docker_raw_logs mode.

As part of this PR, I also added tess which verify that both container stdout and stderr are ingested and also verified it myself.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -730,7 +730,34 @@ commands:
               export INSTALLER_SCRIPT_URL="/tmp/workspace/installScalyrAgentV2.sh"
             fi
 
-            ./scripts/circleci/run-ami-tests-in-bg.sh << parameters.test_type >>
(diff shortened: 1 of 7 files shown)


ArthurKamalov (approved):
LGTM

Kami:
I'll merge this into master since the fix is verified and the AMI test failure is not related to this changes + Arthur is looking into that.

PR #667, opened 2020-12-10
Stop copying manager workers on agent kill

ArthurKamalov:
This PR fixes an issue when the agent process is killed but the orphan child processes of the workers still alive.

Since we use `multiprocessing.managers.SyncManager`, the problem has become a bit more complicated, because the `multiprocessing.managers.SyncManager` itself does not have facilities to detect the parent process kill.

The solution was to extend the current `SyncManager` class and to spawn an additional `watchdog` thread, which is responsible for checking if the parent process exits. This approach should be more correct rather than only checking the status of the parent process from the worker itself because we also have to take care of other threads of the `multiprocessing.managers.SyncManager`.

We also exclude the potential negative impact of those changes on the "default config" scenario because the worker class remains untouched and there is no additional "multiprocess" logic in it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2291,169 +2291,163 @@ jobs:
 
 workflows:
  version: 2
-
+ static-analysis:
+   <<: *skip_release_build_branch_push
+   jobs:
+     - lint-checks:
+         context: scalyr-agent
+ unit-tests:
+   <<: *skip_release_build_branch_push
+   # NOTE: To avoid massive and inflated build matrix, we currently don't run tests under multiple
+   # Python 3 versions
+   jobs:
+     - unittest-26:
+         context: scalyr-agent
+     - unittest-27:
+         context: scalyr-agent
+     - unittest-35:
+         context: scalyr-agent
+     - unittest-35-osx:
+         context: scalyr-agent
+     - unittest-35-windows:
+         context: scalyr-agent
+ smoke-tests:
+   <<: *skip_release_build_branch_push
+   jobs:
+     - smoke-standalone-26:
+         context: scalyr-agent
+     - smoke-standalone-27:
+         context: scalyr-agent
+     - smoke-standalone-35:
+         context: scalyr-agent
+     - smoke-standalone-35-rate-limit:
+         context: scalyr-agent
+     - smoke-standalone-26-tls12:
+         context: scalyr-agent
+     - smoke-standalone-27-tls12:
+         context: scalyr-agent
+      # NOTE: smoke test jobs below still use old smoke tests framework
+     - smoke-docker-json:
+         context: scalyr-agent
+     - smoke-docker-api-raw-logs-disabled:
+         context: scalyr-agent
+     - smoke-docker-syslog:
+         context: scalyr-agent
+     - smoke-k8s:
+         context: scalyr-agent
+     - smoke-monitors-tests:
+         context: scalyr-agent
+     - sonarcloud-scan:
+         requires:
+           # NOTE: This is not ideal, but Circle CI doesn't support cross
+           # workflow dependencies (we can't wait on unittest-27 here) so we just wait on the
+           # slowest jobs
+           - smoke-docker-json
+           - smoke-docker-api-raw-logs-disabled
+           - smoke-docker-syslog
+           - smoke-k8s
+           - smoke-monitors-tests
+ package-tests:
+   <<: *skip_release_build_branch_push
+   jobs:
+     - build-linux-packages-and-installer-script
+     - build-windows-package
+     - package-test-rpm:
+         context: scalyr-agent
+         <<: *master_and_release_only
+     - package-test-deb:
+         context: scalyr-agent
+         <<: *master_and_release_only
+     - smoke-rpm-package-py2:
+         context: scalyr-agent
+         requires:
+           - package-test-rpm
+         <<: *master_and_release_only
+     - smoke-rpm-package-py3:
+         context: scalyr-agent
+         requires:
+           - package-test-rpm
+         <<: *master_and_release_only
+     - smoke-deb-package-py2:
+         context: scalyr-agent
+         requires:
+           - package-test-deb
+         <<: *master_and_release_only
+     - smoke-deb-package-py3:
+         context: scalyr-agent
+         requires:
+           - package-test-deb
+         <<: *master_and_release_only
+#     - ami-tests-development:
+#          context: scalyr-agent
+#          requires:
+#            - build-windows-package
+#            - build-linux-packages-and-installer-script
+     - ami-tests-stable:
+          context: scalyr-agent
+          <<: *master_and_release_only
  benchmarks:
    <<: *skip_release_build_branch_push
    jobs:
-     - benchmarks-loaded-agent_2-growing_logs-monitors-multiprocess-2-worker-20mb-py-35
+     - benchmarks-micro-py-27
+     - benchmarks-micro-py-35
+     - benchmarks-idle-agent-py-27:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-idle-agent-py-35:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-idle-agent-no-monitors-py-27:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-idle-agent-no-monitors-py-35:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-loaded-agent-single-50mb-log-file-with-parser-py-27:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-loaded-agent-single-50mb-log-file-with-parser-py-35:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-loaded-agent-single-growing-log-file-20mb-py-27:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-loaded-agent-single-growing-log-file-20mb-py-35:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-loaded-agent-single-growing-log-file-180mb-py-27:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-loaded-agent_single_growing_with_monitors_2_process_workers-20mb-py-35:
+         <<: *benchmarks_master_and_release_only
+     - benchmarks-loaded-agent_2-growing_logs-monitors-multiprocess-2-worker-20mb-py-35:
+         <<: *benchmarks_master_and_release_only
+     - send-benchmark-results-and-graphs-to-slack:
+         requires:
+           - benchmarks-idle-agent-py-27
+           - benchmarks-idle-agent-py-35
+           - benchmarks-idle-agent-no-monitors-py-27
+           - benchmarks-idle-agent-no-monitors-py-35
+           - benchmarks-loaded-agent-single-50mb-log-file-with-parser-py-27
+           - benchmarks-loaded-agent-single-50mb-log-file-with-parser-py-35
+           - benchmarks-loaded-agent-single-growing-log-file-20mb-py-27
+           - benchmarks-loaded-agent-single-growing-log-file-20mb-py-35
+           - benchmarks-loaded-agent-single-growing-log-file-180mb-py-27
+           - benchmarks-loaded-agent_2-growing_logs-monitors-multiprocess-2-worker-20mb-py-35
+         <<: *benchmarks_master_and_release_only
+ weekly-circle-ci-usage-report:
+   triggers:
+     - schedule:
+         cron: "0 0 * * 0"
+         filters:
+           branches:
+             only:
+               - master
+   jobs:
+     - send-circle-ci-usage-report
 
-#workflows:
-# version: 2
-# static-analysis:
-#   <<: *skip_release_build_branch_push
-#   jobs:
-#     - lint-checks:
-#         context: scalyr-agent
-# unit-tests:
-#   <<: *skip_release_build_branch_push
-#   # NOTE: To avoid massive and inflated build matrix, we currently don't run tests under multiple
-#   # Python 3 versions
-#   jobs:
-#     - unittest-26:
-#         context: scalyr-agent
-#     - unittest-27:
-#         context: scalyr-agent
-#     - unittest-35:
-#         context: scalyr-agent
-#     - unittest-35-osx:
-#         context: scalyr-agent
-#     - unittest-35-windows:
-#         context: scalyr-agent
-# smoke-tests:
-#   <<: *skip_release_build_branch_push
-#   jobs:
-#     - smoke-standalone-26:
-#         context: scalyr-agent
-#     - smoke-standalone-27:
-#         context: scalyr-agent
-#     - smoke-standalone-35:
-#         context: scalyr-agent
-#     - smoke-standalone-35-rate-limit:
-#         context: scalyr-agent
-#     - smoke-standalone-26-tls12:
-#         context: scalyr-agent
-#     - smoke-standalone-27-tls12:
-#         context: scalyr-agent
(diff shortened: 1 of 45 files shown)


Kami:
If I understand it correctly, we run one watchdog thread per each multiprocessing process, right?

scalyr_agent/copying_manager/worker.py
```diff
@@ -1215,7 +1215,83 @@ def create_shared_object_manager(worker_class, worker_proxy_class):
     """
 
     class _SharedObjectManager(multiprocessing.managers.SyncManager):
-        pass
+        # the PID of the parent process.
+        pid = None
```
Kami, line 1219: Should be better to call it ``ppid`` or similar since you may think it refers to the actual worker / child process pid.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1215,7 +1215,83 @@ def create_shared_object_manager(worker_class, worker_proxy_class):
...
+                    try:
+                        # probing the parent process.
+                        if cls.pid is not None:
+                            os.kill(cls.pid, 0)
+                        time.sleep(1)
+                    except OSError:
+                        # parent is not found.
+                        # stop the worker if it is still alive.
+                        if cls.worker and cls.worker.is_alive():
+                            try:
```
Kami, line 1266: Should we also log something in this scenario? Does it even make sense? I assume it wouldn't make it up to scalyr until the next time the agent is started again, or?

ArthurKamalov:
> If I understand it correctly, we run one watchdog thread per each multiprocessing process, right?

@Kami right, in other cases there are no many options to check the parent's pid without rewriting come part of the manager's code. I guess this approach should be less invasive, considering that the implementations of the `multiprocess.managers` may differ from version to version.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: Sorry in advance if i haven't understood you correctly. The worker lives in the same process as this thread lives. When the parent probing is unsuccessful(parent is killed), then it's only needed to terminate/kill the current process and the worker thread will stop too. (But before that we try to stop it gracefully of course)

Kami: Sorry, I just wanted to say it would be a good idea to log a message when watchdog kills a worker process.

So we have it for "audit" purposes and to potentially make troubleshooting easier (to know why a worker process has exited. etc.).

My only question is, if we log a message here, would it even make it to the Scalyr API?

ArthurKamalov: The logger of the current worker should be still ok even if we stop the worker. But the agent*.log file for this worker may be sent to Scalyr by a completely different worker, which may be also stopped at this moment.

Kami: OK, in this case we should try to log it, I think.

Kami:
Some tests would be nice. I know full blow end to end may take some work, but some unit test based ones where we verify the behavior would be nice.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: But even considering what I've said above, it still should be a good idea, because at least, those log messages can be read locally from the file. Thank for pointing on that.

ArthurKamalov: PS: Btw, in the normal case, the agent should be restarted right after the crash, so those agent*.log messages should reach the servers anyway.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1215,7 +1215,83 @@ def create_shared_object_manager(worker_class, worker_proxy_class):
...
+                """
+                Periodically check the parent process for existence and terminate everything if process is not found.
+                :return:
+                """
+                while True:
+                    try:
+                        # probing the parent process.
+                        if cls.parent_pid is not None:
+                            os.kill(cls.parent_pid, 0)
+                        time.sleep(1)
```
Kami, line 1261: I think 1s is fine since this shouldn't have much impact on the load, but if we needed, we can also increase it a bit :)

ArthurKamalov: Yeah, it definitely can be increased, maybe 10 sec?)

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -99,12 +99,16 @@ def __init__(self, config, api_key_config):
                 # Important for the understanding. When the shared_object_manager is started, it creates a new process.
                 # We use this process as a process for the worker.
 
+                # set the parent process pid for the SharedObjectManager's watchdog
+                # which should detect if parent was killed.
+                SharedObjectManager.parent_pid = os.getpid()
```
czerwingithub, line 104: It seems odd to me from a code perspective to set this as a class instance variable.  I'm going to make some suggestions for refactoring this watch dog as a separate abstraction.  As part of that, I'm going to suggest removing this and instead invoking another method during the worker initialization.

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -99,12 +99,16 @@ def __init__(self, config, api_key_config):
...
+                # change agent log path for the new worker.
+                # this initializer function will be invoked in the worker's process.
                 worker_agent_log_path = os.path.join(
                     self.__config.agent_log_path, "agent-%s.log" % worker_id
                 )
 
-                # change agent log path for the new worker.
-                # this initializer function will be invoked in the worker's process.
                 def initializer():
                     self._change_worker_process_agent_log_path(worker_agent_log_path)
```
czerwingithub, line 113: Add a new call here `self._start_watchdog(ppid)`  I think that should be invoked after the `_change_worker_process_agent_log_path` call.

I'm suggesting this because it is more consistent with how the main agent process configures the worker processes.  We already are using the `initializer` to allow changes to be made.  We should just use that same approach for this as well.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1215,7 +1215,83 @@ def create_shared_object_manager(worker_class, worker_proxy_class):
...
+        worker = None
+
+        @classmethod
+        def _create_worker(cls, *args, **kwargs):
+            """
+            Create a new worker and save it as a 'cls.worker' class attribute
+            to be able to access the worker's instance by the watchdog.
+            :return: the proxy object for the 'cls.worker'
+            """
+            cls.worker = worker_class(*args, **kwargs)
```
czerwingithub, line 1232: I don't think it's right to have this be a class variable.  But, I think all of this can be simplified by using a separate watchdog abstraction as discussed above.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1215,7 +1215,83 @@ def create_shared_object_manager(worker_class, worker_proxy_class):
     """
 
     class _SharedObjectManager(multiprocessing.managers.SyncManager):
-        pass
+        # the PID of the parent process.
+        parent_pid = None
+        # When the shared object manager's process is going to be terminated,
+        # we need to be sure that the worker is stopped first.
+        # This class attribute should provide access to the worker for the watchdog that is implemented below.
+        worker = None
```
czerwingithub, line 1223: I think we should pull must of this logic out into a separate WatchDog abstraction that we define in one of the common utlity files.

It's interface should be something like:

```
     class WatchDog():
         def __init__(self, parent_pid, poll_interval, delay_before_kill):
              Here poll_interval is the number of seconds to sleep between checking if
              the parent is still alive.  delay_before_kill is the time to sleep between sending the
              TERM signal and sending the KILL signal.

         def start(self, has_been_stopped_callback):
             Here has_been_stopped_callback is a function that will be invoked by this
             abstraction once the parent is no longer running.  It can be used by the worker to
             make sure it has been stopped.  We might also want to return a bool from this callback
             that indicates if the worker has been stopped because that might change what we log
            out.

File: scalyr_agent/copying_manager/worker.py (hunk 3 above)
Comment (czerwingithub) [inline scalyr_agent/copying_manager/worker.py:1261]: Yeah, let's have this be replaced by a variable.  I suggested `poll_interval` above.

File: scalyr_agent/copying_manager/worker.py (hunk 2 above)
Comment (czerwingithub) [reply scalyr_agent/copying_manager/worker.py:1266]: I agree about us wanting to log something.  BUT, I think we should only do that if the worker has already stopped or if when we ask it to stop, it gracefully does so.  I know you could argue that the second case should be logged, but I worry that due to race conditions, it might be normal for the worker to not have been told to stopped before we see the parent go away.  Hmm... maybe not? 

 @ArthurKamalov in the normal case, will the main agent process wait until all of the workers have terminated.  If so, then yes, let's always log.

File: scalyr_agent/copying_manager/worker.py [hunk 8]
```diff
@@ -1215,7 +1215,83 @@ def create_shared_object_manager(worker_class, worker_proxy_class):
...
+                                cls.worker.stop_worker()
+                            except:
+                                # ignore if worker can not be stopped everything will be killed later anyway.
+                                pass
+                            finally:
+                                break
+
+                # try to terminate everything gracefully first.
+                os.kill(os.getpid(), signal.SIGTERM)
+                time.sleep(1)
```
czerwingithub, line 1279: This should be greater than 1.  I would suggest 5 seconds.  Also, replace with the `delay_before_kill` variable.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1215,7 +1215,83 @@ def create_shared_object_manager(worker_class, worker_proxy_class):
...
+
+                # kill everything to be sure that nothing is survived.
+                os.kill(os.getpid(), signal.SIGKILL)
```
czerwingithub, line 1282: Won't we always get here?  If we send the python process our a TERM, I think Python will attempt to wait for all threads to finish before it actually stops.  So, us just having this thread will mean that there is one thread that is still around, so kill always gets called.

czerwingithub (changes requested):

ArthurKamalov:
@czerwingithub 

As the source of the "killed parent" problem is a `multiprocessing.managers.SyncManager` and its inability to handle this situation, I decided, first of all, to create a subclass that solves that problem. I named it `ParentProcessAwareSyncManager` and put it as a general-purpose utility in the `utils` module.  It has a callback which is invoked when the parent process is killed.
In order to stop workers in such a situation, we just subclass this tool once more and override this callback.

.circleci/config.yml
```diff
@@ -2362,11 +2378,11 @@ workflows:
          requires:
            - package-test-deb
          <<: *master_and_release_only
-     - ami-tests-development:
-          context: scalyr-agent
-          requires:
-            - build-windows-package
-            - build-linux-packages-and-installer-script
+#     - ami-tests-development:
```
czerwingithub, line 2381: tomaz fixed the windows issue we believe so remember to uncomment this.

scalyr_agent/agent_status.py
```diff
@@ -326,6 +326,7 @@ class CopyingManagerWorkerStatus(BaseAgentStatus):
 
     def __init__(self):
         self.worker_id = None
+        self.pid = (None,)
```
czerwingithub, line 329: Add documentation for the var.

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -233,13 +246,26 @@ def _stop_shared_object_managers(self):
         # also stop all shared object managers.
         for worker_id, memory_manager in self.__shared_object_managers.items():
             try:
-                memory_manager.shutdown()
+                memory_manager.shutdown_and_wait()
```
czerwingithub, line 249: Can we pass a parameter here to specify the wait time?  If not, at least add a comment to say how long the wait time is.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1206,27 +1211,84 @@ class CopyingManagerWorkerProxy(_CopyingManagerWorkerProxy):  # type: ignore
...
+                        "Can not stop the worker. Wait for killing the process.."
+                    )
+                    # can not stop worker gracefully, just wait for the main thread of the process exits and
+                    # the the worker's thread(since it is a daemon)  will be terminated too.
+
+        @classmethod
+        def _on_exit(cls, error=None):
+            if error:
+                log.error(
+                    "The shared object manager thread has ended up with an error."
```
czerwingithub, line 1276: Remove "up"

scalyr_agent/util.py
```diff
@@ -2434,3 +2436,226 @@ def match_glob(pathname):
...
     return result
+
+
+class ProcessWatchDog(threading.Thread):
+    """
+    A watchdog that creates a new thread and pools a particular process.
+    If the process does not exist anymore, the watchdog invokes a callback.
+    """
+
+    def __init__(self, pid, poll_interval):
```
czerwingithub, line 2447: `pid` could mean the current process's id.  I would call this `pid_to_watch`.  Use that for the parameter and the class instance variable.  I don't love `pid_to_watch` so if you come up with something better, than that's fine as well.

scalyr_agent/copying_manager/copying_manager.py
ArthurKamalov: I agree that an initializer is the right place for that. But in our case, it reveals too many internals of the `multiprocesing.SyncManager` library and other stuff making the `copying_manager.py`  even more bloated. 
As I tried to explain here https://github.com/scalyr/scalyr-agent-2/pull/667#issuecomment-743477423, now `multiprocesing.SyncManager` solves its own problems in its own place.

scalyr_agent/util.py
```diff
@@ -2434,3 +2436,226 @@ def match_glob(pathname):
...
+        :return:
+        """
+        while True:
+            try:
+                # probing the parent process.
+                if self.pid is not None:
+                    os.kill(self.pid, 0)
+                time.sleep(self._poll_interval)
+            except OSError:
+                self._on_stop_callback()
```
czerwingithub, line 2474: For double protection, I think we should also wait X seconds here and then send a KILL signal to our own process.  I know the main agent process is also set to kill this process, but it is safest to do that here as well.

So, be sure to catch all exceptions when you invoke `self._on_stop_callback` and then sleep for 10 seconds, and then KILL this process.

czerwingithub: Let's have the kill time also be an instance variable set like the poll interval.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: I've managed to make it an instance variable, but anyway, we still have to bind the `shared object manager's` instance with the worker in order to be able to stop the worker in case of troubles.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: Good idea. Done.

czerwingithub (changes requested):

scalyr_agent/copying_manager/copying_manager.py
ArthurKamalov: It has a default keyword argument `timeout` which is 5 seconds for now.

scalyr_agent/util.py
ArthurKamalov: What about `target_pid`?

czerwingithub: Sure, that's fine.

czerwingithub (approved):

scalyr_agent/util.py
ArthurKamalov: In the case of our manager class, its member function `ParentProcessAwareSyncManager._terminate` is set as `_on_stop_callback` for a watchdog, so I guess it would be better to do a kill there and keep the watchdog class "less specialized".
https://github.com/scalyr/scalyr-agent-2/blob/3586c46d34bddf17aec1191ea173115e87213417/scalyr_agent/util.py#L2587-L2607

PR #668, opened 2020-12-10
CT-264: Support both Docker and containerd without configuration

yanscalyr:
We seem to detect docker vs containerd correctly, so the only thing that needed to change to get this working was to move the docker API socket check to after we are sure we are running in docker. Also this PR changes the default CRI behavior to check the filesystem, as this works better in AKS and is likely to work better in a lot of environments.

One concern left: Incorrect docker sockets would crash the agent causing pod restart with a nice message before, but after this change the pod stays running but not working. the `terminate_agent` method other exception catching uses in the same code seems to not actually terminate the agent and cause a pod restart as anticipated. This probably needs to be fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 2, 2020 14:00 -0800
 --->
 
 Bug fixes:
-* Fix to make sure we don't expect valid Docker socket when running in CRI mode.
+* Fix to make sure we don't expect valid Docker socket when running the kubernetes monitor in CRI mode.
 
 Misc:
 * The default value for the `SCALYR_K8S_CRI_QUERY_FILESYSTEM` kubernetes monitor configuration option has changed to `True`. This means that by default when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than first querying the Kubelet API. If you wish to revert to the original default to prefer using the Kubelet API, set `SCALYR_K8S_CRI_QUERY_FILESYSTEM` the environment variable to "false" for the Scalyr Agent daemonset.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 Scalyr Agent 2 Changes By Release
 =================================
 
+## 2.1.16 "TBD" - TBD
+
+<!---
+Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 2, 2020 14:00 -0800
+--->
+
+Misc:
+* The default value for the `k8s_cri_query_filesystem` kubernetes monitor configuration option has changed to `True`. This means that by default when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than first querying the Kubelet API. If you wish to query the Kubelet API for active containers configure `k8s_cri_query_filesystem` as false.
+
 ## 2.1.15 "Endora" - December 2, 2020
 
 <!---
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,11 @@ Scalyr Agent 2 Changes By Release
 Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 2, 2020 14:00 -0800
 --->
 
+Bug fixes:
+* Fix to make sure we don't expect valid Docker socket when running in CRI mode.
+
 Misc:
-* The default value for the `k8s_cri_query_filesystem` kubernetes monitor configuration option has changed to `True`. This means that by default when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than first querying the Kubelet API. If you wish to query the Kubelet API for active containers configure `k8s_cri_query_filesystem` as false.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
...
 =================================
 
+## 2.1.16 "TBD" - TBD
+
+<!---
+Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 2, 2020 14:00 -0800
+--->
+
+Misc:
+* The default value for the `k8s_cri_query_filesystem` kubernetes monitor configuration option has changed to `True`. This means that by default when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than first querying the Kubelet API. If you wish to query the Kubelet API for active containers configure `k8s_cri_query_filesystem` as false.
```
czerwingithub, line 11: I would also say you fixed a bug that prevented automatically failing over toe CRI mode when a Docker socket was not detected.  That should be a separate entry under `Bugs`.

czerwingithub, line 11: Since this is for K8s, I would document how to revert by using the environment vars.  I believe the environment var for this would be `SCALYR_K8S_CRI_QUERY_FILESYSTEM` but double check that.  

I would also say, "If you wish to revert to the original default to prefer using the Kubelet API, set `SCALYR_K8S_CRI_QUERY_FILESYSTEM` the environment variable to "false" for the Scalyr Agent daemonset.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -357,9 +357,9 @@
 define_config_option(
     __monitor__,
     "k8s_cri_query_filesystem",
-    "Optional (defaults to False). If True, then when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than first querying the Kubelet API. This is a useful optimization when the Kubelet API is known to be disabled.",
+    "Optional (defaults to True). If True, then when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than first querying the Kubelet API. This is a useful optimization when the Kubelet API is known to be disabled.",
```
czerwingithub, line 360: I would remove the last sentence of this documentation comment since it is really saying why the default was False.

czerwingithub (changes requested):

czerwingithub:
@yanscalyr Also, can you update the related bugs: https://scalyr.myjetbrains.com/youtrack/issue/CT-264 and CT-258.

CHANGELOG.md
```diff
@@ -7,8 +7,11 @@ Scalyr Agent 2 Changes By Release
 Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 2, 2020 14:00 -0800
 --->
 
+Bug fixes:
+* Fix to make sure we don't expect valid Docker socket when running the kubernetes monitor in CRI mode.
+
 Misc:
-* The default value for the `k8s_cri_query_filesystem` kubernetes monitor configuration option has changed to `True`. This means that by default when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than first querying the Kubelet API. If you wish to query the Kubelet API for active containers configure `k8s_cri_query_filesystem` as false.
+* The default value for the `SCALYR_K8S_CRI_QUERY_FILESYSTEM` kubernetes monitor configuration option has changed to `True`. This means that by default when in CRI mode, the monitor will only query the filesystem for the list of active containers, rather than first querying the Kubelet API. If you wish to revert to the original default to prefer using the Kubelet API, set `SCALYR_K8S_CRI_QUERY_FILESYSTEM` the environment variable to "false" for the Scalyr Agent daemonset.
```
czerwingithub, line 14: Oops, I'm sorry, I didn't mean for you to change the first reference.  Let's change the first part of the sentence to: "The default value for the `k8s_cri_query_filesystem` Kubernetes monitor config option (set via the `SCALYR_K8S_CRI_QUERY_FILESYSTEM` environment var) has changed to `True`."

CHANGELOG.md
```diff
@@ -7,8 +7,11 @@ Scalyr Agent 2 Changes By Release
 Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 2, 2020 14:00 -0800
 --->
 
+Bug fixes:
+* Fix to make sure we don't expect valid Docker socket when running the kubernetes monitor in CRI mode.
```
czerwingithub, line 11: We should always describe bugs in terms of the symptoms that customers would think in.  So, let's change this to:

"Fix to make sure we don't expect a valid Docker socket when running Kubernetes monitor in CRI mode.  This fixes an issue preventing the K8s monitor from automatically failing over to CRI mode if Docker is not available."

czerwingithub (approved):
@yanscalyr Sorry Yan, I want to just tweak your documentation once more.  Go ahead and submit once this is done.

CHANGELOG.md
yanscalyr: I dont think thats actually accurate though, because we dont fail over to CRI and never did. I guess saying "preventing the K8s monitor from running in CRI mode if Docker is not available." would work?

czerwingithub: Hmm, ok, let's make sure we are on the same page on this one.  We should be auto-detecting whether or not we should run using docker or CRI.  The requirement from the customer is they want one configuration that works in both environments.  We have these options `k8s_always_use_docker` and `k8s_always_use_cri` that both default to `False`.  From my understanding, if both of those are false and docker is not available, then we should automatically use CRI.  So, one configuration would work in both environments.

Am I wrong in my understanding?

yanscalyr: We are auto-detecting yes, I just think failover is the wrong term for it, since we don't default to Docker, we just pick the correct one from the start automatically. It's a minor detail now that I'm typing it out.

czerwingithub: Ok, no worries, it is better to double check these things.  Go ahead and make the documentation change as you suggested and submit.  Thanks for doing this.

czerwingithub (approved):

PR #670, opened 2020-12-13
Log a message when sending SIGTERM and SIGKILL to the child worker process

Kami:
I've tested #667 and everything seems to be working correctly :+1: 

I just added two additional log statements so we also log a message when sending SIGTERM and SIGKILL to the child worker process.

Keep in mind that those represent edge case which is triggered when parent is kill -9'ed so using info log level is fine.

### DIFF
--- a/scalyr_agent/copying_manager/worker.py
+++ b/scalyr_agent/copying_manager/worker.py
@@ -1238,7 +1238,8 @@ def create_shared_object_manager(worker_class, worker_proxy_class):
             # type: (Configuration, Dict, six.text_type) -> CopyingManagerWorker
             """
             Create a new worker and save it as an attribute.
-            to be able to access the worker's instance within the local process.
+
+            To be able to access the worker's instance within the local process.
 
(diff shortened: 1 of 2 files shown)


scalyr_agent/copying_manager/worker.py
```diff
@@ -1268,15 +1269,15 @@ def _on_parent_process_kill(self):
                     self._worker.stop_worker()
                 except:
                     log.exception(
-                        "Can not stop the worker. Wait for killing the process.."
+                        "Can not stop the worker. Wait before killing the process.."
```
czerwingithub, line 1272: I think that should be "Waiting"

czerwingithub (approved):

scalyr_agent/copying_manager/worker.py
Kami: Good catch, will fix it.

ArthurKamalov (approved):

PR #673, opened 2020-12-21
Add back config permissionns fixes changes

Kami:
This pull request adds back changes with config file and directory permissions fixes.

I'll dig into our ancient builder VM to try to see what is going on there and why the generated packages are corrupted when including those changes.


Kami:
I didn't confirm my theory end to end yet, but after digging through the builder VM and fpm code, I think my theory which I posted the other day is indeed correct - the root case for issues with the packages build on builder VM is ``*--use-file-permissions`` fpm flag.

On CI, job step runs as a user with uid 0 so it works fine, but on builder VM, it runs as a ``rpmbuilder`` user with uid ``1001`` and that uid will be preserved inside the deb package tarball which will cause issues on postinst.

In ideal world we would of course want to use this flag since it means created package artifacts contain correct permissions for agent.json and other files. Sadly that's not easily possible with current builder VM setup. It means we also need to rely in postint step for fresh / new installations - not great.

If that builder user had sudo / root access, we could include some other workarounds in build_package.py if the uid is not 0 (aka use chown, etc, but atm we can't do that).

Here is an example package layout with and without those flags.

Screenshot from 2020-12-21 11-16-26

Screenshot from 2020-12-21 11-16-44

As you can see, when the flags are not used, permissions for agent.json, agent.d, etc. are not preserved which is not great, but we don't have much choice atm.

As a long term solution we should just utilize packages built on Circle CI and ensure uid 0 is used there. Another possible workaround would also be to add rpmbuilder sudo access and add some chown steps to the build_package.py if os.getuid() is not 0...

Kami:
I'll merge this into master so I can build internal packages for testing.

Technically I could also build packages for this branch, but since we use rebase merge strategy, on merge commit sha would change which would make one used to build the package invalid and could cause issues.

PR #677, opened 2020-12-22
Include information on configured number of workers and unique API keys in the User-Agent header

Kami:
This pull request updates ``User-Agent`` string sent by the agent so it also includes a value which indicates if sharded copy manager functionality is used and total number of workers and unique API keys configured.

Possible values for this header fragment are documented here - https://github.com/scalyr/scalyr-agent-2/compare/user_agent_mpw?expand=1#diff-4cb9befeecb4697d270768d8487b47aa20810c540fa7a9b6c86800735e7927a2R951.

This will provide us with a better insight on usage of this functionality.

### DIFF
--- a/scalyr_agent/scalyr_client.py
+++ b/scalyr_agent/scalyr_client.py
@@ -921,7 +921,7 @@ class ScalyrClientSession(object):
                     + "-"
                     + str(openssl_version[3])
                 )
-                openssl_version_string = "openssl-%s" % (openssl_version_string)
+                openssl_version_string = "o-%s" % (openssl_version_string)
             except Exception:
                 openssl_version_string = None
         else:
@@ -941,10 +941,10 @@ class ScalyrClientSession(object):
             "\\Administrators"
         ):
             # Indicates agent running as a privileged / admin user
-            user_string = "admin-1"
+            user_string = "a-1"
         else:
             # Indicates agent running as a regular user
-            user_string = "admin-0"
+            user_string = "a-0"
 
         sharded_copy_manager_string = ""
 
--- a/scalyr_agent/scalyr_client.py
+++ b/scalyr_agent/scalyr_client.py
@@ -151,6 +151,7 @@ def create_client(config, quiet=False, api_key=None):
         disable_send_requests=config.disable_send_requests,
         disable_logfile_addevents_format=config.disable_logfile_addevents_format,
         enforce_monotonic_timestamps=config.enforce_monotonic_timestamps,
+        workers_api_keys_tuple=config.get_number_of_configured_workers_and_api_keys(),
     )
 
 
@@ -228,6 +229,7 @@ class ScalyrClientSession(object):
         disable_send_requests=False,
         disable_logfile_addevents_format=False,
         enforce_monotonic_timestamps=False,
+        workers_api_keys_tuple=None,
     ):
         """Initializes the connection.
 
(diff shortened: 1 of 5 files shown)


scalyr_agent/scalyr_client.py
```diff
@@ -151,6 +151,7 @@ def create_client(config, quiet=False, api_key=None):
         disable_send_requests=config.disable_send_requests,
         disable_logfile_addevents_format=config.disable_logfile_addevents_format,
         enforce_monotonic_timestamps=config.enforce_monotonic_timestamps,
+        workers_api_keys_tuple=config.get_number_of_configured_workers_and_api_keys(),
```
Kami, line 154: Technically I could also pass in the whole configuration object here, but to follow the existing API and be on the safe-side with potential cyclic dependencies, I went with this approach.

I think cyclic dependency should not be an issue even if I passed in the whole object since config + client gets re-created when config change is detected, but just in case.

czerwingithub: That sounds like the right call to me.

scalyr_agent/scalyr_client.py
```diff
@@ -248,6 +250,8 @@ def __init__(
         @param compression_level: An int containing the compression level of compression to use, from 1-9.  Defaults to 9 (max)
         @param enforce_monotonic_timestamps: A bool that indicates whether event timestamps in the same session
             should be monotonically increasing or not.  Defaults to False
+        @param workers_api_keys_tuple: Tuple containing a total number of configured workers and unique
```
czerwingithub, line 253: I would also list worker type here.  Saying tuple without saying three things creates a little confusion.

scalyr_agent/scalyr_client.py
```diff
@@ -937,6 +946,35 @@ def __get_user_agent(self, agent_version, fragments=None):
             # Indicates agent running as a regular user
             user_string = "admin-0"
 
+        sharded_copy_manager_string = ""
+
+        # Possible values for this header fragment:
+        # mw-0 - Sharded copy manager functionality is disabled
+        # mw-1|<num_workers>|<num_api_keys> - Functionality is enabled and there are <num_workers>
```
czerwingithub, line 953: Why use a `|` as a separator?  That feels more like it marks the beginning of new concept (i.e., we are transitioning now to information about K8s version).  My preference would just to use a hyphen to separate it.  But, I also won't fight for it.

czerwingithub (changes requested):

scalyr_agent/scalyr_client.py
Kami: Either way works for me.

I thought we used ``-`` for separating main fragment key and value for for "sub-values" I should use something else, but I can also just use ``-`` for those values as well.

Kami:
@czerwingithub Pushed tuple rename + docstrings change.

While at it, I also decided to push a small change to use shorter keys for some other options which are included with each request. Those are not backward compatible, but I don't think it should be too bad handling that in PowerQuery on the server side.

scalyr_agent/scalyr_client.py
czerwingithub: I think we didn't specify a good policy to begin with.  Main fragments are separated by semi-colons... and that's about all we have.  Eh, I don't mind either way.

czerwingithub (approved):

PR #678, opened 2020-12-22
Update docker monitor to log labels + container info when starting a logger for a new container

Kami:
This pull request updates Docker monitor to log some container + logger related information for newly started containers.

I had some problems with using labels as configs and having some log statements like that would make troubleshooting easier.

It's worth noting that this code only runs when a new container is found so it shouldn't add tons of overhead. I'm debating if I should use info log level with rate limit key being a container name (or perhaps log path / similar) or using debug level 1.

We probably can't just use straight up INFO without any rate limit since it would cause havoc in scenarios where a lot of short lived containers are started.

### DIFF
--- a/docker/Dockerfile.docker_monitor_testing_config
+++ b/docker/Dockerfile.docker_monitor_testing_config
@@ -21,7 +21,7 @@
 #
 # To run in:
 #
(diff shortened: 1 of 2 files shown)


czerwingithub (approved):
Logging at DEBUG_LEVEL_1 definitely feels right to me on this one.

PR #679, opened 2020-12-23
Add repos dynamically in the install script instead of repo packages.

ArthurKamalov:
This PR adds a new approach to install a Scalyr agent from an "easy" install script - `install-scalyr-agent-2.sh`. In this approach, the install script does not install the `scalyr-repo` package to configure the repository, instead of that, it adds the repository directly using the package manager.

The main steps:
1) Add the public key from the `keyserver.ubuntu.com` into the package manager's keyring.
2) Add the Scalyr repository to the package manager.
3) Update

To fallback to the previous behaviors, a new install script option `--use-bootstrap-packages` has been introduced. If it is set, the old approach with the "config" `scalyr-repo` package is used. By default, the new approach is used.

### DIFF
--- a/scripts/circleci/release/installScalyrAgentV2.sh
+++ b/scripts/circleci/release/installScalyrAgentV2.sh
@@ -411,7 +411,7 @@ if [[ $REPO_TYPE == "yum" ]]; then
       run_command "yum install -y yum-utils"
       echo "Adding the public key."
 
-      run_command "rpm --import https://${KEYSERVER_URL}/pks/lookup?op=get&search=0x${PUBLIC_KEY_FINGERPRINT}"
+      #run_command "rpm --import https://${KEYSERVER_URL}/pks/lookup?op=get&search=0x${PUBLIC_KEY_FINGERPRINT}"
       echo "Adding the scalyr agent repository."
       cat > /etc/yum.repos.d/scalyr.repo <<EOF
 [scalyr]
--- a/scripts/circleci/release/installScalyrAgentV2.sh
+++ b/scripts/circleci/release/installScalyrAgentV2.sh
@@ -469,7 +469,7 @@ else
     run_command "gpg --no-default-keyring --keyring gnupg-ring:/etc/apt/trusted.gpg.d/scalyr.gpg --keyserver ${KEYSERVER_URL} --recv ${PUBLIC_KEY_FINGERPRINT}"
 
     # change permissions for the gpg key.
-    chmod 666 /etc/apt/trusted.gpg.d/scalyr.gpg
+    chmod 660 /etc/apt/trusted.gpg.d/scalyr.gpg
 
     echo "Adding the scalyr agent repository."
     echo "deb ${REPOSITORY_URL}/apt scalyr main" > /etc/apt/sources.list.d/scalyr.list
--- a/scripts/circleci/release/installScalyrAgentV2.sh
+++ b/scripts/circleci/release/installScalyrAgentV2.sh
@@ -150,7 +150,7 @@ VERSION=
 # If non-empty, the scalyr server to use for uploading logs.
 SCALYR_SERVER=
 
-REPOSITORY_URL="https://scalyr-repo.s3.amazonaws.com/{ % REPLACE_REPOSITORY_TYPE % }" #// replace the placeholder and this comment from the 'create-agent-installer.sh'//
+REPOSITORY_URL="https://scalyr-repo.s3.amazonaws.com/{ % REPLACE_REPOSITORY_TYPE % }" # replace the placeholder and this comment from the 'create-agent-installer.sh'
 KEYSERVER_URL="keyserver.ubuntu.com"
 PUBLIC_KEY_FINGERPRINT="84AC559B5FB5463885CE0841F70CEEDB4AD7B6C6"
 USE_BOOTSTRAP_PACKAGES=false
--- a/scripts/circleci/release/installScalyrAgentV2.sh
+++ b/scripts/circleci/release/installScalyrAgentV2.sh
@@ -413,7 +413,17 @@ if [[ $REPO_TYPE == "yum" ]]; then
 
       run_command "rpm --import https://${KEYSERVER_URL}/pks/lookup?op=get&search=0x${PUBLIC_KEY_FINGERPRINT}"
       echo "Adding the scalyr agent repository."
-      run_command "yum-config-manager --add-repo ${REPOSITORY_URL}/yum/binaries/noarch"
+      cat > /etc/yum.repos.d/scalyr.repo <<EOF
+[scalyr]
+includepkgs=scalyr-agent-2
+name=Scalyr packages - noarch
+baseurl=${REPOSITORY_URL}/yum/binaries/noarch"
+mirror_expire=300
+metadata_expire=300
+enabled=1
+gpgcheck=1
+EOF
+      #run_command "yum-config-manager --add-repo ${REPOSITORY_URL}/yum/binaries/noarch"
   fi
 
   PACKAGE_NAME="scalyr-agent-2"
@@ -464,6 +474,7 @@ else
     # so we just manipulate key files directly through filesystem on newer versions.
     # see https://manpages.debian.org/testing/apt/apt-key.8.en.html#DESCRIPTION
 
+    # initialize gpg in case if it has been freshly installed.
     gpg --update-trustdb
 
     if [[ "$apt_version" < "1.8.2.1" ]]; then
--- a/scripts/circleci/release/installScalyrAgentV2.sh
+++ b/scripts/circleci/release/installScalyrAgentV2.sh
@@ -469,7 +469,7 @@ else
     run_command "gpg --no-default-keyring --keyring gnupg-ring:/etc/apt/trusted.gpg.d/scalyr.gpg --keyserver ${KEYSERVER_URL} --recv ${PUBLIC_KEY_FINGERPRINT}"
 
     # change permissions for the gpg key.
-    chmod 660 /etc/apt/trusted.gpg.d/scalyr.gpg
+    chmod 644 /etc/apt/trusted.gpg.d/scalyr.gpg
 
     echo "Adding the scalyr agent repository."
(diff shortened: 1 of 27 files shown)


scripts/circleci/release/installScalyrAgentV2.sh
```diff
@@ -405,24 +414,33 @@ else
...
+    lecho "  Refreshing available package list (may take several minutes)";
+    run_command "apt-get update" check_for_https_error;
+
+    lecho "  Updating Scalyr yum repository configuration files using repository.";
+    run_command "apt-get -y install scalyr-repo";
+  else
+    ls  $TMPDIR
+    run_command "apt-get update";
+    run_command "apt-get install -y software-properties-common gnupg"
+    run_command "apt-key add $TMPDIR/main.asc"
```
Kami, line 440: File with public key will be bundled with the installer script, right?

Kami:
I think this approach should work fine, but it does mean we can't change the key we use to sign the packages with "no interaction needed by the end user" anymore (which may be fine and in a lot of cases even preferred).

While at it, we could also update our build pipeline to push publish key to public keyserver such as mitm and ubuntu one (or perhaps doing it manually is fine since key is not something which changes very often).

scripts/circleci/release/installScalyrAgentV2.sh
ArthurKamalov: yes, I just used the existing ability to add files in the tar at the end of the script file. (The repo deb and rpm packages are bundled in the same way)

ArthurKamalov:
> I think this approach should work fine, but it does mean we can't change the key we use to sign the packages with "no interaction needed by the end user" anymore (which may be fine and in a lot of cases even preferred).

I guess in this case it should be enough to add a new public key (and, ideally, to remove the old one) and update the repositories right?

Kami:
@ArthurKamalov Yeah, I mostly meant the upgrade path though.

In this case user would need to add a new key if we change it. I personally think this should be fine and some may even argue that's it's preferred since the whole point of signing and key management is that you explicitly add and trust a specific key :)

ArthurKamalov:
> @ArthurKamalov Yeah, I mostly meant the upgrade path though.
> 
> In this case user would need to add a new key if we change it. I personally think this should be fine and some may even argue that's it's preferred since the whole point of signing and key management is that you explicitly add and trust a specific key :)

Agree with that. That may be somehow excusable it is a "convenience" script, but if it is a "step-by-step" installation, I think there must be a step with adding the public key explicitly, so the user can verify it.

scripts/circleci/release/installScalyrAgentV2.sh
```diff
@@ -147,6 +150,11 @@ VERSION=
 # If non-empty, the scalyr server to use for uploading logs.
 SCALYR_SERVER=
 
+REPOSITORY_URL="https://scalyr-repo.s3.amazonaws.com/{ % REPLACE_REPOSITORY_TYPE % }"
```
czerwingithub, line 153: I would add a comment that this should be replaced by the `create-agent-installer.sh` script.

scripts/circleci/release/installScalyrAgentV2.sh
```diff
@@ -158,6 +166,9 @@ while (( $# > 0)); do
     --extract-packages)
       echo "Extracting...";
       untar_tarball $0 || die "Failed to extract packages";
+      pwd
```
czerwingithub, line 169: Remember to remove this and the next few lines.  Looks like debugging output.

scripts/circleci/release/installScalyrAgentV2.sh
```diff
@@ -405,24 +430,51 @@ else
...
+    export DEBIAN_FRONTEND=noninteractive
+    run_command "apt-get install -y gnupg"
+
+
+    apt_version=$(dpkg -s apt | grep -i version | awk '{print $2}')
+
+    # the apt-key tool is deprecated since debian 10 and is removed in debian 11,
+    # so we just manipulate key files directly through filesystem on newer versions.
+    # see https://manpages.debian.org/testing/apt/apt-key.8.en.html#DESCRIPTION
+    if [[ "$apt_version" < "1.8.2.1" ]]; then
```
czerwingithub, line 466: Is there any benefit to use using the `apt-key` tool for the older versions?  Could we just use the same approach (modify the files directly) in both cases?  It seems to me, having different implementations for the different versions is just one more thing that can go wrong.

scripts/circleci/release/installScalyrAgentV2.sh
```diff
@@ -378,18 +393,28 @@ if [[ $REPO_TYPE == "yum" ]]; then
...
+    # way we found to make sure the repository configuration can be updated.
+    run_command "yum install -y scalyr-repo";
+  else
+      # Add tool that may be needed to add repository.
+      run_command "yum install -y yum-utils"
+      echo "Adding the public key."
+
+      run_command "rpm --import https://${KEYSERVER_URL}/pks/lookup?op=get&search=0x${PUBLIC_KEY_FINGERPRINT}"
+      echo "Adding the scalyr agent repository."
+      run_command "yum-config-manager --add-repo ${REPOSITORY_URL}/yum/binaries/noarch"
```
czerwingithub, line 416: I take it `yum-config-manager` is fairly standard?  We aren't going to have problems on older installations?

czerwingithub (commented):
In general, I think this looks fine.  Not installing the bootstrap repos shouldn't cause any problems beyond what you and Tomaz have discussed (we can't automatically update the signing keys).  As you two say though, that isn't the standard way to handle the keys anyway, so we shouldn't have been doing it in the first place.

I added a few comments but nothing really major.

scripts/circleci/release/installScalyrAgentV2.sh
ArthurKamalov: It should be a very common tool, but I've also managed how to add repo directly, without any tools.

scripts/circleci/release/installScalyrAgentV2.sh
ArthurKamalov: Yeah, Now I think so too. Done.

scripts/circleci/release/installScalyrAgentV2.sh
ArthurKamalov: Oops, thanks.

scripts/circleci/release/installScalyrAgentV2.sh
ArthurKamalov: Done.

Kami:
Let's also add upgrade notes entry on the "impact" to the existing users - technically it doesn't impact them since they already have the key installer, but perhaps we should still document how to remove the key and scalyr-repo package and add it manually if needed, or similar.

scripts/circleci/release/installScalyrAgentV2.sh
```diff
@@ -405,24 +434,49 @@ else
...
+
+
+    # initialize gpg in case if it has been freshly installed.
+    gpg --update-trustdb
+
+    echo "Adding the public key."
+    run_command "gpg --no-default-keyring --keyring gnupg-ring:/etc/apt/trusted.gpg.d/scalyr.gpg --keyserver ${KEYSERVER_URL} --recv ${PUBLIC_KEY_FINGERPRINT}"
+
+    # change permissions for the gpg key.
+    chmod 666 /etc/apt/trusted.gpg.d/scalyr.gpg
```
Kami, line 472: Does it need to be writable by everyone? I imagine 640 should be enough.

ArthurKamalov: Thanks for noticing that. Changed to 640.

scripts/circleci/release/installScalyrAgentV2.sh
Kami: On that note, to be on the safe side, we should probably add additional check to AMI tests which verifies installer file doesn't container this place holder string when AMI tests are operating with installer script file or URL and not directly with packages.

scripts/circleci/release/installScalyrAgentV2.sh
```diff
@@ -405,24 +434,49 @@ else
...
+
+
+    # initialize gpg in case if it has been freshly installed.
+    gpg --update-trustdb
+
+    echo "Adding the public key."
+    run_command "gpg --no-default-keyring --keyring gnupg-ring:/etc/apt/trusted.gpg.d/scalyr.gpg --keyserver ${KEYSERVER_URL} --recv ${PUBLIC_KEY_FINGERPRINT}"
+
+    # change permissions for the gpg key.
+    chmod 660 /etc/apt/trusted.gpg.d/scalyr.gpg
```
Kami, line 472: This is a public key so ``664`` or ``644`` should be fine (aka it can also be readable by others).

In fact, it seems that's what other software does:

```bash
$ ls -la /etc/apt/trusted.gpg.d
total 28
drwxr-xr-x 2 root root 4096 Dec 11 21:06 .
drwxr-xr-x 7 root root 4096 Dec 11 21:06 ..
-rw-r--r-- 1 root root 2796 Apr  9  2020 ubuntu-keyring-2012-archive.gpg
-rw-r--r-- 1 root root 2794 Apr  9  2020 ubuntu-keyring-2012-cdimage.gpg
-rw-r--r-- 1 root root 1733 Apr  9  2020 ubuntu-keyring-2018-archive.gpg
```

Kami: EDIT: Sorry, looks like I suggested this in the past - https://github.com/scalyr/scalyr-agent-2/pull/679#discussion_r549301213

I imagine not readable by others should also work, but since it's a public key, making it readable (but not writable) by others should be fine as well.

Kami:
I assume we also need some documentation changes to go with this PR? Aka how to add the key manually if not using the installer script.

ArthurKamalov:
@Kami 
> I assume we also need some documentation changes to go with this PR? Aka how to add the key manually if not using the installer script.
There is a PR to change to instructions in docs.

scripts/circleci/release/installScalyrAgentV2.sh
ArthurKamalov: Yeah, changed.

Kami (approved):

PR #680, opened 2020-12-24
Improve throughput of TCP syslog monitor with the new experimental config options

Kami:
This pull requests adds support for two new experimental syslog request parsers for syslog TCP monitor.

Both of those parsers should offer much better throughput than the current default parser which supports framed and line delimited syslog messages.

## Background, Context

We've done some benchmarking in the past and currently syslog monitor in TCP mode can only handle ~3 MB/s of data before it saturates all the CPU available to the syslog monitor thread.

I wanted to find the root cause for this poor performance so I've done some more profiling.

Based on the profiling results, the main bottleneck is calling ``handle_frame()`` function for each processed syslog message (frame / line).

Basically what that function does is calls scalyr logger method which writes data to a file on disk.

## Proposed Change

To be able to work around this poor performance, I introduced two new syslog TCP request parsers.

### 1) SyslogRawRequestParser

This parser doesn't handle framed or new line delimited messages, but it simply writes any data it reads from the socket directly to a file.

This one isn't really meant to be used in production, but it's mostly used for benchmarking purposes since in a sense it represents "best case" scenario where no processing is performed on read data.

Technically, it could still work in some scenarios where messages are line delimited and not framed with a caveat / limitation that we could potentially be writing partial line data to a file when logger method is called (depending on data we read from the socket).

This could perhaps be worked around with server-side parser definition, but it involve overhead on the server.

### 2) SyslogBatchedRequestParser

This parser works very similar to the current default parser. It can handle frame and line delimited data, only difference is that it doesn't call ``handle_frame()`` for each processed / parsed line / frame, but only after it has exhausted current buffer.

This results in much less calls to scalyr logger method which results in better efficiency and improved throughput.

As part of this parser, I also got rid of the ``message_size_can_exceed_tcp_buffer`` config option and made that the default behavior. Having said that, we may still potentially want to add a guard  which causes the handler to flush currently accumulated buffer if we haven't seen a complete frame / syslog message for X seconds.

With that change, I can easily get ~30 MB/s locally with our load gen (5 workers, 200 bytes message size, count 1000).

This basically means that the bottleneck will rarely be the syslog monitor itself, but actual agent code which is responsible for ingesting lines which are written to disk by syslog monitor up to Scalyr (and that's a familiar beast and we have many ways to mitigate that such as using multiple workers, etc.).

## Open Questions

I'm 95% this approach should work (aka we can write multiple lines at once to a log file on disk / call scalyr logger with multiple lines at once and file will be correctly ingested since it's no different than some other software writing arbitrary number of bytes to a file), but it would still be good if @czerwingithub can confirm that.

## TODO

- [x] Agree on the approach
- [x] Clean up and finish the code (remove duplicated code, etc).
- [x] Add tests

### DIFF
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -331,7 +331,7 @@ define_config_option(
     __monitor__,
     "tcp_incomplete_frame_timeout",
     "How long we wait (in seconds) for a complete frame / syslog message when running in TCP mode "
-    "with batched request parser before giving up and flushing what has accumulated in the buffer.",
+    "with batch request parser before giving up and flushing what has accumulated in the buffer.",
     default=5,
     min_value=0,
     max_value=600,
@@ -818,7 +818,7 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
                 max_buffer_size=self.server.tcp_buffer_size,
                 message_size_can_exceed_tcp_buffer=self.server.message_size_can_exceed_tcp_buffer,
             )
-        elif self.request_parser == "batched":
+        elif self.request_parser == "batch":
             request_stream = SyslogBatchedRequestParser(
                 socket=self.request,
                 max_buffer_size=self.server.tcp_buffer_size,
@@ -1493,7 +1493,7 @@ class SyslogServer(object):
                 )
                 request_parser = config.get("tcp_request_parser")
 
-                if request_parser not in ["default", "batched", "raw"]:
+                if request_parser not in ["default", "batch", "raw"]:
                     raise ValueError(
                         "Invalid tcp_request_parser value: %s" % (request_parser)
                     )
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -324,8 +324,11 @@ define_config_option(
     "Optional (defaults to default). Which TCP packet data request parser to use. Most users "
     "should leave this as is.",
     default="default",
-    convert_to=str,
+    convert_to=six.text_type,
 )
+
+# TODO: Make it a config option
+INCOMPLETE_FRAME_FLUSH_THRESHOLD_SECONDS = 60
 
 
 def _get_default_gateway():
@@ -644,8 +647,13 @@ class SyslogBatchedRequestParser(SyslogRequestParser):
 
         self._max_buffer_size = max_buffer_size
 
+        # Internal buffer of bytes remained to be processes
         self._remaining = bytearray()
+        # Current offset into the internal remaining buffer
         self._offset = 0
+
+        # Stores the timestamp of when we last called "handle_frame()"
+        self._last_handle_frame_call_time = int(time.time())
 
         self.is_closed = False
 
@@ -661,12 +669,13 @@ class SyslogBatchedRequestParser(SyslogRequestParser):
             )
             return
 
-        # append data to what we had remaining from the previous call (if any)
+        # Append data to what we had remaining from the previous call (if any)
         self._remaining += data
 
         size = len(self._remaining)
 
-        # process the buffer until we are out of bytes
+        # Process the buffer until we are out of bytes. Once we are out of bytes, flush processed
+        # data to file.
         frames_handled = 0
         data_to_write = bytearray()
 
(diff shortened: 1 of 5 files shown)


Kami:
Also of course open to better name for the "batched" request parser. Just something I came up while profiling things, but it's likely not the best name.

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -317,6 +318,26 @@
...
+    default="default",
+    convert_to=six.text_type,
+)
+
+define_config_option(
+    __monitor__,
+    "tcp_incomplete_frame_timeout",
+    "How long we wait (in seconds) for a complete frame / syslog message when running in TCP mode "
+    "with batched request parser before giving up and flushing what has accumulated in the buffer.",
+    default=60,
```
czerwingithub, line 335: I would do 5 seconds as the default.  We do something similar in the main part of the agent (we wait for a new line to send an `/addEvents` request, so we timeout after 5 seconds if we've gotten a partial line).  We haven't had any complaints about that to date.

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -584,27 +615,205 @@ def process(self, data, handle_frame):
...
+    def process(self, data, handle_frame):
+        handle_frame(data)
+
+
+class SyslogBatchedRequestParser(SyslogRequestParser):
+    """
+    This parser works in exactly the same manner as the default request parser (it supports framed
+    and line delimited data), but instead of calling scalyr logger class and writing each frame /
+    line as it's processed, it writes all the processed lines in a single batch at the end of
+    processing of a specific TCP packet.
```
czerwingithub, line 641: So, if I'm understanding this correctly, we are also not (really) increasing how much data we may drop if the Scalyr Agent shuts down unexpectantly, correct?  I could argue that if we write each line to disk one at time, then those lines become durable more quickly.  But, here, you are still guaranteeing all bytes are on disk when processing a single glob of data from TCP/IP.  We aren't waiting for another packet to be received, some time out to be exceeded (beyond the incomplete frame thing), etc that would significantly increase exposure.

Is that correct?

Heck, in syslog, we aren't really even acknowledging that we have received the bytes, are we?  We don't send a response.  So, syslog data could be lost in any point from buffering on the client's TCP/IP side to data in transit on the network to network buffering on the Scalyr Agent side.  Is that right?

czerwingithub (changes requested):
This looks really good overall.  I just have one minor tweak to a default value and then a question just to make sure I understand some points.

scalyr_agent/builtin_monitors/syslog_monitor.py
Kami: Correct, yeah.

Current code is not ideal in terms of the how much control we have of exactly when (e.g. after how many bytes / lines received) we actually flush data to disk / logger, but changing that would require bigger refactor and logic change.

And I don't think that change would bring us any substantial improvement in terms of throughput itself (with this change it's already good enough aka syslog monitor won't be the bottleneck), we would just get more control over "durability" and potentially the data loss window in a case of hard crash would be slightly smaller.

So for now, this change is likely fine, but going forward we may want to change the logic a bit to give us more control over the "flush" interval.

As far as actual numbers of how many times ``handle_frame`` was called before and with this change - previously, if we sent 13k messages it would be called 13k times (once per message) and now it will be called approximately total_data_sent_in_bytes / tcp_buffer_size which will be substantially less than 13k times.

czerwingithub: Ok, thanks for the explanation and yup, this approach looks good to me.  We can touch base on Monday morning, but yeah, we may very will use an internal build with this change for a customer looking for higher throughput from syslog.

Kami:
I've pushed a couple of more changes.

One of them is new ``tcp_message_delimiter`` option which specifies which character to use as a message delimiter. For most cases, default value of ``\n`` should work, but in some cases such as the Python syslogger log handler, we need to support custom values (Python syslog logging handler uses null character aka ``\000`` as the terminator / delimter and some other implementations use it as well).

This config option is only supported by the new "batched" parser. I think that in the future we should just get rid of the current "default" one and use "batched" one. I think it should be a safe change and would require us to maintain less duplicated / similar code.

czerwingithub (approved):

PR #681, opened 2020-12-24
replace "service" command in agent package *install scripts

ArthurKamalov:


Kami:
Yeah, this should work until we move to distro specific packaging.

Then we will need to support init.d, upstart, systemd... :D

Also, just curious where the ``service`` command didn't work? Since afaik, most recent distros have ``service`` command alias there for compatibility reasons and I didn't see any failures with AMI tests.

If we know which distro it was, we should add it to AMI tests to catch such issues automatically.

Kami:
I guess one option why that step didn't fain on AMI tests could also be, because we don't consider those command failures as fatal.

If we don't, we should update only the AMI part to consider failures in those steps as fatal to automatically catch such issues.

ArthurKamalov:
@Kami 
>Also, just curious where the service command didn't work? Since afaik, most recent distros have service command alias there for compatibility reasons and I didn't see any failures with AMI tests.

Things became much easier, the service command is not available on docker images. So that also should be one small step to working docker installation:)

PR #685, opened 2020-12-31
Add worker log status

ArthurKamalov:
This PR adds periodical stats logging for all workers in separate process'. 
The agent's `linux process monitor` also aggregates all metrics from the child worker process'. 

**Things to do/discuss:**

Create a separate app for the `linux process monitor` for each worker process to be able to monitor them separately.
In this case, we need to create a new instance if the `LinuxProcessMonitor` for each worker.

### DIFF
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -1674,7 +1674,7 @@ class ScalyrAgent(object):
             "copy_manager_status workers_number=%ld total_scan_iterations=%ld total_read_time=%lf total_compression_time=%lf total_waiting_time=%lf total_blocking_response_time=%lf "
             "total_request_time=%lf total_pipelined_requests=%ld avg_bytes_produced_rate=%lf avg_bytes_copied_rate=%lf"
             % (
-                stats.copying_manager_status.workers_number,
+                stats.num_workers,
                 stats.total_scan_iterations,
                 stats.total_read_time,
                 stats.total_compression_time,
@@ -1817,6 +1817,7 @@ class ScalyrAgent(object):
         result.num_copying_paths = copying_paths
         result.num_running_monitors = running_monitors
         result.num_dead_monitors = dead_monitors
+        result.num_workers = current_status.copying_manager_status.workers_number
 
         (
             result.user_cpu,
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -1671,9 +1671,10 @@ class ScalyrAgent(object):
         stats = overall_stats
 
         log.info(
-            "copy_manager_status total_scan_iterations=%ld total_read_time=%lf total_compression_time=%lf total_waiting_time=%lf total_blocking_response_time=%lf "
+            "copy_manager_status workers_number=%ld total_scan_iterations=%ld total_read_time=%lf total_compression_time=%lf total_waiting_time=%lf total_blocking_response_time=%lf "
             "total_request_time=%lf total_pipelined_requests=%ld avg_bytes_produced_rate=%lf avg_bytes_copied_rate=%lf"
             % (
+                stats.copying_manager_status.workers_number,
                 stats.total_scan_iterations,
                 stats.total_read_time,
                 stats.total_compression_time,
--- a/scalyr_agent/agent_main.py
+++ b/scalyr_agent/agent_main.py
@@ -1817,7 +1817,8 @@ class ScalyrAgent(object):
         result.num_copying_paths = copying_paths
         result.num_running_monitors = running_monitors
         result.num_dead_monitors = dead_monitors
-        result.num_workers = current_status.copying_manager_status.workers_number
+        if current_status.copying_manager_status is not None:
+            result.num_workers = current_status.copying_manager_status.workers_number
 
         (
             result.user_cpu,
--- a/scalyr_agent/agent_status.py
+++ b/scalyr_agent/agent_status.py
@@ -468,24 +468,22 @@ class CopyingManagerStatus(BaseAgentStatus):
 
         self._verify_workers_health_check()
 
-        # sum up all stats from workers if it is not a default config.
-        if not self.is_single_worker():
-            # sum up some worker stats to overall stats.
-            for worker_status in self._workers():
-                self.total_errors += worker_status.total_errors
-                self.total_bytes_uploaded += worker_status.total_bytes_uploaded
+        # sum up some worker stats to overall stats.
+        for worker_status in self._workers():
+            self.total_errors += worker_status.total_errors
+            self.total_bytes_uploaded += worker_status.total_bytes_uploaded
 
-                self.total_rate_limited_time = worker_status.total_rate_limited_time
-                self.total_read_time = worker_status.total_read_time
-                self.total_waiting_time = worker_status.total_waiting_time
-                self.total_blocking_response_time = (
-                    worker_status.total_blocking_response_time
-                )
-                self.total_request_time = worker_status.total_request_time
-                self.total_pipelined_requests = worker_status.total_pipelined_requests
-                self.rate_limited_time_since_last_status = (
-                    worker_status.rate_limited_time_since_last_status
-                )
+            self.total_rate_limited_time += worker_status.total_rate_limited_time
+            self.total_read_time += worker_status.total_read_time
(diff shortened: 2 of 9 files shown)


scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -1313,6 +1323,14 @@ def __select_processes(self):
                 for t_pid in self.__target_pids:
                     if t_pid == "$$":
                         t_pid = int(os.getpid())
+                        # if we use multiprocess workers we also search for the child processes if that workers by  matching the agent cmd line.
+                        if self._global_config.use_multiprocess_copying_workers is True:
+                            child_processes = set(
+                                ps.get_matches_commandline(ps.current_process["cmd"])
```
Kami, line 1329: It would probably se safer to retrieve child pids directly instead of matching on the command line arguments, right?

I believe we have an existing method for that (which may not be easily accessible here though).

Kami:
> Create a separate app for the linux process monitor for each worker process to be able to monitor them separately.
> In this case, we need to create a new instance if the LinuxProcessMonitor for each worker.

Yeah, as discussed in the past, I think it would be better to create a separate instance for each worker and not aggregate all the metrics against single main scalyr-agent-2 process.

The end goal is being able to see per worker process metrics in Scalyr dashboard. If needed, metrics for all the agent processes can still be aggregated together on the Scalyr server side, but vice-versa is not the case.

Basically under "process" dropdown in the "Linux Process metrics" monitor, we should be able to see all the worker processes and display metrics for them.

Screenshot from 2021-01-04 13-20-47

Right now the main process name is ``agent``, so for worker processes we should probably use this notation or similar - ``agent-worker-1``, ``agent-worker-2``, etc.

scalyr_agent/agent_main.py
```diff
@@ -1671,9 +1671,10 @@ def __log_copy_manager_stats(self, overall_stats):
         stats = overall_stats
 
         log.info(
-            "copy_manager_status total_scan_iterations=%ld total_read_time=%lf total_compression_time=%lf total_waiting_time=%lf total_blocking_response_time=%lf "
+            "copy_manager_status workers_number=%ld total_scan_iterations=%ld total_read_time=%lf total_compression_time=%lf total_waiting_time=%lf total_blocking_response_time=%lf "
```
Kami, line 1674: Just thinking out loud - I don't think this should break any PowerQueries or similar since I don't think any of that values ordering.

If it does / for just in case, we could add it at the end.

scalyr_agent/agent_status.py
```diff
@@ -468,24 +471,23 @@ def calculate_status(self):
...
-                self.total_blocking_response_time = (
-                    worker_status.total_blocking_response_time
-                )
-                self.total_request_time = worker_status.total_request_time
-                self.total_pipelined_requests = worker_status.total_pipelined_requests
-                self.rate_limited_time_since_last_status = (
-                    worker_status.rate_limited_time_since_last_status
-                )
+        # sum up some worker stats to overall stats.
+        for worker_status in self._workers():
```
Kami, line 475: This is just a "cosmetic" change, right?

scalyr_agent/configuration.py
```diff
@@ -2913,6 +2918,15 @@ def __verify_main_config_and_apply_defaults(
             min_value=1,
         )
 
+        self.__verify_or_set_optional_float(
+            config,
+            "default_worker_status_message_interval",
```
Kami, line 2923: :+1:

scalyr_agent/agent_status.py
ArthurKamalov: It was the assignment instead of incrementing and also I realized that it does not matter if it is a single worker or not we still can sum everything.

scalyr_agent/builtin_monitors/linux_process_metrics.py
ArthurKamalov: Yeah, for some reason, I did that to exclude the tcollector's process', but now I can not say why it should be excluded :)

Btw. since we try to introduce separate monitors for each worker, this aggregation was removed.

scalyr_agent/agent_main.py
ArthurKamalov: I am not sure, but it looks like each graph does its own line parsing, so we shouldn't worry about that.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -1307,17 +1320,42 @@ def __select_processes(self):
...
             self.__pids = pids
         return self.__pids
 
+    def set_pid(self, pid):  # type: (int) -> None
+        """
+        Set the PID of the process that was marked as $$TBD.
+        NOTE: this method is called from the a different thread, so all shared objects must be guarded.
+        :param pid: Process PID
+        """
+        with self._target_pids_lock:
```
Kami, line 1349: So it seems like this a read-only lock?

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -793,6 +807,33 @@ def run(self):
                 for worker_pool in self.__api_keys_worker_pools.values():
                     worker_pool.start_workers_and_block_until_copying()
 
+                # if the multi-process workers are enabled, then we have to set the worker PID's for the instances of the
+                # 'linux_process_monitor' that may be set to monitor the worker processes.
+                if self.__config.use_multiprocess_copying_workers:
```
Kami, line 812: I think we should add a config option / feature flag so this functionality (gathering metrics for worker process) can be disabled.

And for now it's probably better to default it to false / off.

Kami:
Some unit / end to end tests for status output functionality + "late / lazy pid population" would be good.

Once that's done, I think we should disable worker process metrics gather by default to be on the safe side.

Once the PR is merged, we should update version on qatesting and also enable worker process metrics collection functionality.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -1307,17 +1320,42 @@ def __select_processes(self):
...
+            target_pids = self.__target_pids[:]
+        for i in range(len(target_pids)):
+            if target_pids[i] == "$$TBD":
```
Kami, line 1352: So it seems this sets the pid on the first "$$TBD" value it sees, right?

How is the logic supposed to work? We set id in local ``target_pids`` list variable we create (which is a copy of ``self.__target_pids``, but local to this function) and we never update ``self.__target_pids`` itself.

Or am I missing something?

And if we will update ``self.__target_pids`` as part of this function, we will also need a read + write lock here since it can be called by multiple workers at the same time, right?

Kami:
As discussed - since worker status line change has bigger priority over "worker metrics" once - if worker metrics one is taking longer than expected, we should handle that in a separate PR and just merge agent status line change itself.

scalyr_agent/builtin_monitors/linux_process_metrics.py
ArthurKamalov: Oh right, I don't assign the temporary list back to `self.__target_pids`.  But anyway, as you mentioned above, it should be ok to not use the lock at all, since we only write the PID once for each monitor and the write itself should be atomic.

scalyr_agent/builtin_monitors/linux_process_metrics.py
```diff
@@ -1307,17 +1317,37 @@ def __select_processes(self):
...
                     pids.append(t_pid)
             self.__pids = pids
         return self.__pids
 
+    def set_pid(self, pid):  # type: (int) -> None
+        """
+        Set the PID of the process that was marked as $$TBD.
+        :param pid: Process PID
+        """
+        for i in range(len(self.__target_pids)):
```
Kami, line 1343: To confirm I'm understanding this correctly - there can only be a single entry with ``$$TBD`` in this list since we have this class instance per monitor, right?

Kami (approved):
LGTM.

Let's please get this change on qatesting with worker metrics gather enabled.

scalyr_agent/builtin_monitors/linux_process_metrics.py
ArthurKamalov: Right, the meaning is the same as for the agent process monitor with "$$" - we need to track only one particular process.

PR #687, opened 2021-01-08
CT-247 Add copy truncate log rotation support

oliverhsu77:
Made the following changes to support copy truncate log rotation:
1. Added `__find_copy_truncate_file` which located the copied copy/truncate file based on heuristics.  Docs for `__find_copy_truncate_file` describe the heuristics used.
2. Modified `__refresh_pending_files`, which compares if the node is the same to trigger copy/truncate logic.  Copy/truncate code modifies  `pending_files` to have `[copied file, original truncated file]`
3. Fixed test case issue where `_set_new_log_processor` was calling `self.write_file(self.__path, b"")`, which would truncate the log file causing the copy truncate log rotation code to be triggered.


czerwingithub:
@oliverhsu77  I'm just going to add comments to help unstick you on the two issues you raised.

> 
> 1. The copied file is read from the beginning instead of where the original truncated file left off.  Need some help to get the the copied file to read from where the original truncated last read left off.

I think you should be able to diagnosis all of this through use of a debugger.  I suggest you take a look at the `__read_file_chunk` method since that's what reads the actual bytes from disk.  In particular, pay attention to how `offset_in_file` is calculated and trace that back up to what you are doing.

> 2. possible issue: `test_max_existing_log_offset_size_over_max_existing_log_offset_size` and `test_max_log_offset_size_set_to_max_existing_log_offset_size_after_perform_processing` tests are triggering the copy/truncate code path.  Not quite sure if this is an issue.  The tests pass, but in real life if there are rotated log files, it may try to pick a rotated log file, even though it's not a copy/truncate log file.

Yeah, this doesn't sound like correct behavior to me.  I took a look at those test cases.  If, within them, we were truncating a log file, I might say that this would be expected.  But, I don't think that's what is happening here.

Have you been able to determine why the code is being triggered?

oliverhsu77:
Addressed the issues discussed in the WIP PR.  PR is ready for review now.

It's a bit unclear why the `black` lint is failing - need some pointers on this.  I've run `black`, checked in the changes, and it's passing locally:
```
black --check --config pyproject.toml scalyr_agent/log_processing.py tests/unit/log_processing_test.py
All done! ✨ 🍰 ✨
2 files would be left unchanged.
```

oliverhsu77:
Please also see the comments in the ticket for more context: https://scalyr.myjetbrains.com/youtrack/issue/CT-247

scalyr_agent/configuration.py
```diff
@@ -1600,6 +1600,13 @@ def enable_worker_process_metrics_gather(self):
         """
         return self.__get_config().get_bool("enable_worker_process_metrics_gather")
 
+    @property
+    def enable_copy_truncate_log_rotation(self):
```
czerwingithub, line 1604: I would rename this to `enable_copy_truncate_log_rotation_support` since otherwise it sounds like we are using copy truncate to rotate our own logs.

scalyr_agent/log_processing.py
```diff
@@ -1104,21 +1115,47 @@ def __refresh_pending_files(self, current_time):
...
+                            )
+                            truncated_log_file.position_end = (
+                                truncated_log_file.position_start + latest_size
+                            )
+                            truncated_log_file.last_known_size = latest_size
+                            truncated_log_file.is_log_file = True
+                            copied_log_file.is_log_file = False
+                        else:
+                            log.warning(
+                                "Could not find copied log file for copy-truncate log rotation file."
```
czerwingithub, line 1150: Let's modify this warning to be "It appears that the file was rotated using copy-truncate but the copied file could not be found.  This may mean some log lines are missing."

This will be emitted every time the file is rotated... but I guess we think that won't be that often where this log message will amount to many bytes.  Otherwise I'd suggest we rate limit the line.

czerwingithub (changes requested):

czerwingithub (approved):

tests/unit/log_processing_test.py
```diff
@@ -506,22 +506,100 @@ def test_losing_read_access(self):
...
+        self.assertEqual(copied_file, located_copy_truncate_file)
+
+    def test_find_copy_truncate_multiple_files(self):
+        # Test find_copy_truncate_file with multiple files - most recent file is used
+        filename = os.path.basename(self.__path)
+        copied_file_2 = os.path.join(self.__tempdir, filename + ".2")
+        copied_file_1 = os.path.join(self.__tempdir, filename + ".1")
+
+        self.write_file(copied_file_2, b"")
+        # No easy way to modify file ctime from Python, sleep to have later ctime
```
ArthurKamalov, line 548: What about `os.utime`?

ArthurKamalov: Oops sorry. It's only for python 3.

ArthurKamalov (approved):

PR #688, opened 2021-01-09
Rename sharded copying manager entities

ArthurKamalov:
This PR renames some copying manager entities and related configuration parameters.
`API key` now is `worker`
`worker now` is `session`
`default_workers_per_api_key` -> `default_sessions_per_worker`
 `use_multiprocess_copying_workers` -> `use_multiprocess_sessions`
 
Added the mechanism that allows setting "deprecated" config parameters for backward compatibility. Using this mechanism, the user still can set `use_multiprocess_copying_workers` or `default_workers_per_api_key` but it will also get the "deprecation" warning. See this commit https://github.com/scalyr/scalyr-agent-2/pull/688/commits/acca6eba889ef725853f9e09eeaaa28cdc471142


scalyr_agent/configuration.py
```diff
@@ -3054,6 +3066,60 @@ def __get_config_or_environment_val(
...
+        # the config param is not presented by its current name, look for a deprecated name.
+        if config_val is None and deprecated_names is not None:
+            for name in deprecated_names:
+                config_val = self.__get_config_val(config_object, name, param_type)
+                if config_val:
+                    config_object.put(param_name, config_val)
+                    del config_object[name]
+                    if self.__logger:
+                        self.__logger.warning(
+                            "The configuration option {0} is deprecated, use {1} instead.".format(
```
Kami, line 3117: I wonder if we should add rate limit here (e.g. once per day), since I believe this code will be executed each time config is re-read and parsed.

ArthurKamalov: Done. Yea, I guess once per day should be fine for such messages.

czerwingithub (changes requested):
Ok, I'm advocating for one name to be changed, both the variable and the name we use in the status line.

@ArthurKamalov Yeah, I'm sorry, in my mind I wasn't expecting us to change all of the internal variable names to match the config.  This turned out to be a larger PR than I envisioned.  But, yeah, that would have been some technical debt we were accruing so maybe it is better that you did this now.  I am just sorry that it ended up being larger than we meant it to be.

ArthurKamalov:
@czerwingithub yes, sorry there are a lot of mechanical renames. I've made a notion about the config options deprecation feature. It should've easier to review as a separate commit https://github.com/scalyr/scalyr-agent-2/commit/acca6eba889ef725853f9e09eeaaa28cdc471142.

scalyr_agent/configuration.py
```diff
@@ -3054,6 +3066,62 @@ def __get_config_or_environment_val(
...
```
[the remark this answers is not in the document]
ArthurKamalov: IIRC, this is only one valid way to use the `.format` function on python 2.6:)

czerwingithub (approved):

Kami:
Per my comment on Slack, there might be a bug with handling deprecated / changed option names.

I started the agent with the old config and it fails with this error:

```bash
2021-01-20 12:12:35.738Z WARNING [core] [configuration.py:524] Configured option max_log_offset_size is being overridden due to max_send_rate_enforcement setting.
Error reading configuration file: The required field use_multiprocess_workers was not found in object.
Terminating agent, please fix the configuration file and restart agent.
  File "scalyr_agent/agent_main.py", line 310, in main
    config_file_path, log_warnings=log_warnings
  File "scalyr_agent/agent_main.py", line 412, in __read_and_verify_config
    self.__verify_config(config)
  File "scalyr_agent/agent_main.py", line 460, in __verify_config
    monitors_manager = MonitorsManager(config, self.__controller)
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/monitors_manager.py", line 68, in __init__
    configuration, platform_controller
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/monitors_manager.py", line 277, in __create_monitors
    configuration, platform_controller
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/monitors_manager.py", line 410, in _get_all_monitor_configs
    for monitor in platform_controller.get_default_monitors(configuration):
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/platform_linux.py", line 128, in get_default_monitors
    config.use_multiprocess_workers
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/configuration.py", line 1607, in use_multiprocess_workers
    return self.__get_config().get_bool("use_multiprocess_workers")
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/json_lib/objects.py", line 227, in get_bool
    return self.__compute_missing_value(field, default_value, none_if_missing)
  File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/json_lib/objects.py", line 546, in __compute_missing_value
    "The required field %s was not found in object." % field_name
scalyr_agent.json_lib.exceptions.JsonMissingFieldException: The required field use_multiprocess_workers was not found in object.
```

Same config works with v2.1.15. Basically, this config would indicate I still want to use multiple sessions functionality, but I want to use thread and not mp based workers.

I assume it's something related to how we handle falsy / false value (likely treat that value as not being set instead of "false"), but not sure, didn't dig in yet.

Here are the relevant config options:

```json
....
    use_multiprocess_copying_workers: false,
    default_workers_per_api_key: 1,
"api_keys": [
{
    "id": "one",
        "api_key": "xxxx-",
        workers: 1
},
],
...
```

PR #689, opened 2021-01-11
Update shell monitor output to also include "duration" and "exit_code" fields

Kami:
This pull request updates ``shell_monitor`` so it includes two new fields:

1. ``duration`` - command execution duration in seconds.
2. ``exit_code`` - exit / return code of the command

Those two fields should make it easier to set up various alerts based on this monitor output.

It's worth noting that it does mean, we will include additional data with each emitted metric, but for the average case, that should be around 20 bytes so not sure if it's worth adding feature flag / config option for this.

## TODO

- [x] Changelog entry


ArthurKamalov (approved):
LGTM. But I can not say the final word about the increase of the emitted data, which you have also mentioned, but I also think that it should be fine.

Kami:
Thanks.

It's also worth pointing out that this doesn't affect default installation since no shell monitors are configured by default. It only affects installations which utilize shell monitors - and in those cases, it's quite likely that the actual shell command / script itself produces much more output than those ~20 bytes.

PR #691, opened 2021-01-12
Update the CHANGELOG of the Agent version 2.1.16

ArthurKamalov:


CHANGELOG.md
```diff
@@ -8,21 +8,35 @@ Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 31, 2020 14:00 -0800
...
-## 2.1.16 "Lasso" - December 23, 2020
+## 2.1.16 "Lasso" - January 13, 2021
 
 <!---
-Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 23, 2020 14:00 -0800
+Packaged by Arthur Kamalov <arthur@scalyr.com> on Jan 13, 2021 14:00 -0800
 --->
 
+Features:
+* Add copy truncate log rotation support. Use the ``enable_copy_truncate_log_rotation_support`` config option to enable it.
```
czerwingithub, line 28: It is actually enabled by default, so folks shouldn't have to do anything.  I would say something like:
```
Add copy truncate log rotation support.  This is enabled by default.  It does not support copy truncate with
compression unless the `delaycompress` option is used.  This feature can be disabled by setting `enable_copy_truncate_log_rotation_support` to false.
```

CHANGELOG.md
```diff
@@ -8,21 +8,35 @@ Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 31, 2020 14:00 -0800
...
 Misc:
 * On startup and when parsing a config file, agent now emits a warning if the config file is readable by others.
+* Add the config option ``enable_worker_process_metrics_gather`` to enable 'linux_process_metrics' monitor for each
+multiprocess worker.
+* Each session, which runs in a separate process, periodically writes its stats in the log file.
+The interval between writes can be changed by using the ``default_worker_session_status_message_interval``
+* Rename some of the configuration parameters: ``use_miltiprocess_copying_workers``  to ``use_multiprocess_workers``,
+``default_workers_per_api_key`` to ``default_sessions_per_api_key``. Previous option names are preserved for the
+backward compatibility but they are marked as deprecated.
+NOTE: The appropriate [environment variable names](https://app.scalyr.com/help/scalyr-agent-env-aware) are changed too.
```
czerwingithub, line 39: You should add something to mention the new release scripts.

czerwingithub (changes requested):

CHANGELOG.md
```diff
@@ -8,21 +8,35 @@ Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 31, 2020 14:00 -0800
 --->
 
 Improvements:
-* Add new ``tcp_request_parser`` and ``tcp_message_delimiter`` config option. Valid values for ``tcp_request_parser`` include ``default`` and ``batch``. New TCP recv batch oriented request parser is much more efficient than the default one and should be a preferred choice in most situations. For backward compatibility reasons, the default parsed hasn't been changed yet.
+* Add new ``tcp_request_parser`` and ``tcp_message_delimiter`` config option. Valid values for ``tcp_request_parser``
```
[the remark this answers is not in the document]
ArthurKamalov: I was just about to ask about this because I still was not 100% sure that Tomaz wanted to add this to the current release. Thanks for clarifying.

CHANGELOG.md
ArthurKamalov: Oh, thanks for the correction.

CHANGELOG.md
ArthurKamalov: Yes, I guess we should since the script is located in the agent's repository.

CHANGELOG.md
```diff
@@ -1,28 +1,36 @@
...
-
-## 2.1.16 "Lasso" - December 23, 2020
-
-<!---
-Packaged by Tomaz Muraus <tomaz@scalyr.com> on Dec 23, 2020 14:00 -0800
---->
-
-Misc:
-* On startup and when parsing a config file, agent now emits a warning if the config file is readable by others.
+* The agent install bash script now adds the Scalyr repositories directly without installing the ``scalyr-repo`` packages.
```
czerwingithub, line 33: I would also say this resolves problems with installing on some distributions (if you have the particular distribution from the ticket, I would add it here as well.)

czerwingithub (changes requested):
@ArthurKamalov Just one small change.

CHANGELOG.md
```diff
@@ -1,28 +1,38 @@
...
+This also eliminates errors caused by re-acquiring the package manager's lock file during the *pre/post* *install/uninstall* scripts.
+The issue occurred in both ``apt`` and ``rpm`` package managers.
```
czerwingithub, line 35: Because of the limitations of our packaging scripts, you can't line break lines these two lines.  They need to be part of the same line above starting with "The agent install bash script..."

czerwingithub (approved):
Ok, you just need to fix the line breaks.

CHANGELOG.md
ArthurKamalov: But those line breaks only for the source text, to follow the Markdown's "recommended" line length, and they have no effect in a rendered result.

https://github.com/scalyr/scalyr-agent-2/blob/160a0fb9d89de4f703bb2a31ac986f60e0c204dd/CHANGELOG.md

czerwingithub: The problem is that we have a custom script that converts this markdown into entries for the RPM and Debian packages changelog.  So, yeah, it is valid markdown, but it breaks our simple script.  We should fix it ... it's definitely one of those gotchas but we never got around to it.

See this for our custom script: https://github.com/scalyr/scalyr-agent-2/blob/c1218b2d0cc4f7a3217062af3e613d89c0e844a9/build_package.py#L1514

ArthurKamalov: Oh, I got in now. Thanks.

czerwingithub (approved):

PR #692, opened 2021-01-14
Fix debian change log build and small fixes

ArthurKamalov:
This PR fixes the changelog build warnings for the debian. It seems like those warnings were there for  a long time (the `dpkg-parsechangelog` throws them since agent version 2.0.1)
First issue: The first level entries(e.g. Features, Improvements) begin with only one space character. Because of that, those entries are not shown.
Second issue:  the release author entry also does not have extra space.

The changelog before the PR:
```
Source: scalyr-agent-2
Version: 2.1.16
Distribution: stable
Urgency: low
Changes:
 scalyr-agent-2 (2.1.16) stable; urgency=low
 .
    - Add copy truncate log rotation support. This is enabled by default.  It does not support copy truncate with compression unless the `delaycompress` option is used.  This feature can be disabled by setting `enable_copy_truncate_log_rotation_support` to false.
 .
    - Add new ``tcp_request_parser`` and ``tcp_message_delimiter`` config option. Valid values for ``tcp_request_parser`` include ``default`` and ``batch``. New TCP recv batch oriented request parser is much more efficient than the default one and should be a preferred choice in most situations.  For backward compatibility reasons, the default parser hasn't been changed yet.

```

after:
```
Source: scalyr-agent-2
Version: 2.1.16
Distribution: stable
Urgency: low
Maintainer: Arthur Kamalov <arthur@scalyr.com>
Timestamp: 1610575200
Date: Thu, 14 Jan 2021 01:00:00 +0300
Changes:
 scalyr-agent-2 (2.1.16) stable; urgency=low
 .
   * Features:
    - Add copy truncate log rotation support. This is enabled by default.  It does not support copy truncate with compression unless the `delaycompress` option is used.  This feature can be disabled by setting `enable_copy_truncate_log_rotation_support` to false.
 .
   * Improvements:
    - Add new ``tcp_request_parser`` and ``tcp_message_delimiter`` config option. Valid values for ``tcp_request_parser`` include ``default`` and ``batch``. New TCP recv batch oriented request parser is much more efficient than the default one and should be a preferred choice in most situations.  For backward compatibility reasons, the default parser hasn't been changed yet.

```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2246,6 +2246,8 @@ jobs:
       - run:
           name: Build deb and rpm packages
           command: |
+            set -e
+
             mkdir artifacts
             mkdir package
 
@@ -2255,9 +2257,19 @@ jobs:
             python ../build_package.py rpm
 
             dpkg -X "$(ls *.deb)" /tmp/deb_package
-            warnings=$(dpkg-parsechangelog --file /tmp/deb_package/usr/share/doc/scalyr-agent-2/changelog.gz 2>&1 1>/dev/null | grep warning || true)
+
+            set +e
+            stderr=$(dpkg-parsechangelog --file /tmp/deb_package/usr/share/doc/scalyr-agent-2/changelog.gz 2>&1 1>/dev/null)
+            if [[ "$?" -ne 0 ]]; then
+              >&2 echo "$stderr"
+              exit 1
+            fi
+            set -e
+            warnings=$(grep warning \<<< "$stderr" || true)
+
             if [[ -n "${warnings}" ]]; then
-              echo "The debian package changelog is built with warnings."
+              >&2 echo "The debian package changelog is built with warnings. This may be caused by splitting the text across miltiple lines."
+              >&2 echo "$warnings"
               exit 1
             fi
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2254,6 +2254,13 @@ jobs:
             python ../build_package.py deb
             python ../build_package.py rpm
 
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -2254,6 +2254,13 @@ jobs:
             python ../build_package.py deb
             python ../build_package.py rpm
 
+            dpkg -X "$(ls *.deb)" /tmp/deb_package
+            warnings=$(dpkg-parsechangelog --file /tmp/deb_package/usr/share/doc/scalyr-agent-2/changelog.gz 2>&1 1>/dev/null | grep warning || true)
```
Kami, line 2258: I wonder if we should use ``set +e`` for this section and then check exit code of the command where we grep for "warning" and in case it's non-zero exit with 1 and call the command again to print actual warnings.

The issue with current approach is that it won't fail in case ``/tmp/deb_package/usr/share/doc/scalyr-agent-2/changelog.gz`` file doesn't exist.

Which means that check could be incorrectly passing in situations where that file doesn't exist for whatever reason.

.circleci/config.yml
```diff
@@ -2254,6 +2254,13 @@ jobs:
...
+            if [[ -n "${warnings}" ]]; then
+              echo "The debian package changelog is built with warnings."
```
Kami, line 2260: And here we should also echo the actual warnings + some friendly text along the lines of "Text shouldn't be split / wrapped across multiple lines, etc.".

Kami:
Thanks for working on this and you are right - some of those warnings were also old and were there for quite a while.

czerwingithub (approved):
Thanks for taking of this Arthur.  That was great of you.

PR #693, opened 2021-01-14
Fix worker session missing log lines

ArthurKamalov:
This PR makes worker session agent logs act the same way as the main `agent.log` file does - to obtain its current size and use it as the initial position for the copying manager. This is needed to avoid skipping the first log lines before the copying manager has copied the log file.


scalyr_agent/agent_main.py
```diff
@@ -985,6 +994,61 @@ def __print_force_https_message(self, scalyr_server, raw_scalyr_server):
...
+            for worker_session_id in self.__config.get_session_ids_from_all_workers():
+                log_file_path = self.__config.get_worker_session_agent_log_path(
+                    worker_session_id
+                )
+                log_file_paths_to_check.append(log_file_path)
+
+        log_file_paths = []
+        for log_file_path in log_file_paths_to_check:
+            include_file = True
+            if os.path.exists(log_file_path):
```
Kami, line 1027: Technically and to be more correct, we could also use ``isfile``, but I think it's highly unlikely that directory with such name would exist.

scalyr_agent/configuration.py
```diff
@@ -699,6 +701,42 @@ def __get_default_hostname(self):
...
+    def get_session_ids_of_the_worker(worker_config):  # type: (Dict) -> List
+        """
+            Generate the list of IDs of all sessions for the specified worker.
+            :param worker_config: config entry for the worker.
+            :return: List of worker session IDs.
+        """
+        result = []
+        for i in range(worker_config["sessions"]):
+            # combine the id of the worker and session's position in the list to get a session id.
+            worker_session_id = "%s_%s" % (worker_config["id"], i)
```
Kami, line 714: Do we have some kind of regex validation for valid characters in the worker id config option?

If not, it's probably a good idea to only allow ``a-zA-Z-`` or similar.

Technically, I believe we use ``os.path.join`` everywhere so using arbitrary characters should still be safe from command / shell injections perspective, but it's probably still better we limit it (at the very list it could cause issues if we split the string and the id contains ``_`` character).

scalyr_agent/agent_main.py
```diff
@@ -1007,6 +1016,61 @@ def __print_force_https_message(self, scalyr_server, raw_scalyr_server):
...
+
+        # we also all log files for all worker session if they are in multiprocess configuration.
+        if self.__config.use_multiprocess_workers:
+            for worker_session_id in self.__config.get_session_ids_from_all_workers():
+                log_file_path = self.__config.get_worker_session_agent_log_path(
+                    worker_session_id
+                )
+                log_file_paths_to_check.append(log_file_path)
+
+        log_file_paths = []
```
Kami, line 1046: I assume this will always contain unique file names so we don't need to use a set, right?

Kami (approved):
Please add a changelog entries, besides that, LGTM.

scalyr_agent/agent_main.py
ArthurKamalov: Yes. As I see, there has to be a "main" `agent.log` file and log files for each worker session, which also have to be unique.

PR #694, opened 2021-01-15
Also validate changelog syntax / formatting in the RPM package

Kami:
This pull request adds small change on top of #692 so we also validate syntax / conformance of changelog file / content which is embedded into the builtrpm package.

RPM seems to be much less strict with changelog formatting than deb, but it also doesn't hurt to have this validation.

In addition to that change, I also moved validation logic into a standalone script file.

### DIFF
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -712,8 +712,8 @@ class SyslogBatchedRequestParser(SyslogRequestParser):
 
         while self._offset < size:
             # 2->TODO use slicing to get bytes in both python versions.
-            # TODO: This is not really robust, we should make it an explicit config option if the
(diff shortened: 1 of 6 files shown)


scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -709,6 +712,8 @@ def process(self, data, handle_frame):
 
         while self._offset < size:
             # 2->TODO use slicing to get bytes in both python versions.
+            # TODO: This is not really robust, we should make it an explicit config option if the
```
ArthurKamalov, line 715: It looks like the leftover "the".

ArthurKamalov (approved):

scalyr_agent/builtin_monitors/syslog_monitor.py
Kami: Thanks, will fix it.

PR #698, opened 2021-01-16
Include syslog_monitor changes (multiple test runs testing changes)

Kami:
PR which will run smoke-docker-syslog tests 100 times to try to provide us more confidence with some of the changes (to determine if there is an issue / race on the server / local side, etc.).

This PR includes syslog_monitor.py as is in v2.1.16 release (aka no changes).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2361,101 +2361,12 @@ jobs:
 
 workflows:
  version: 2
- static-analysis:
-   <<: *skip_release_build_branch_push
-   jobs:
-     - lint-checks:
-         context: scalyr-agent
- unit-tests:
-   <<: *skip_release_build_branch_push
-   # NOTE: To avoid massive and inflated build matrix, we currently don't run tests under multiple
-   # Python 3 versions
-   jobs:
-     - unittest-26:
-         context: scalyr-agent
-     - unittest-27:
-         context: scalyr-agent
-     - unittest-35:
-         context: scalyr-agent
-     - unittest-35-osx:
-         context: scalyr-agent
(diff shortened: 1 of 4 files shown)


Kami:
Here is the summary after 50 runs of those tests (https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/5053/workflows/b405d321-9709-470f-8694-57bcd593ac5f/jobs/69950).

> 
> 
> Status after 50 of command runs
> Successful runs: 50
> Failed runs: 0

Same story with #699.

I will re-run it for another 50 runs, but at this point I'm 97% there is something else going on which is causing those occasional failures - likely something to do with server side ingest + query delay perhaps combined with some local docker log gen stuff.

As I mentioned in the past, we should also check with the backend team and see if any potentially related changes were deployed recently and also check ingest delay and status for our account when those failures occur.

It seems to be there are more frequent during PST time which could mean that the shard our cloudtech testing is on may be seeing increased usage and load and this could cause delays...

@czerwingithub If it passes again after another re-run with 50 runs, I'm really just leaning towards including those changes in v2.1.17 so we don't need to go back and forth twice.

Kami:
It passed yet again for another 50 iterations without any failures - https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/5053/workflows/4e082428-bd36-499e-bdb1-6b53c67c0a91/jobs/69952.

Kami:
I also pushed a change which tails docker container + agent_syslog.log, docker_monitor.log file at the end.

This may help us troubleshoot this issue in the future.

PR #700, opened 2021-01-16
v2.1.18 changes

Kami:
Reverts changelog + syslog_monitor change and cherry picks testing related changes from #698.


czerwingithub (approved):

Kami:
OK, it looks like my additional debugging change finally after gazillion re-runs bore fruit!

Earlier, the job passed 150+ times without an error, but an hour or so ago during the release process, we've seen another failure.

Sadly my testing change to include more context on test failure needed some more work. After fixing the testing change and re-running it ~6 times, it finally failed again (https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/5064/workflows/601c64de-0490-4583-9a23-0ea33f2404bd/jobs/70090/parallel-runs/0/steps/0-106) and we finally managed to get the stack trace / more context!

Screenshot from 2021-01-16 22-24-03

It looks like the issue lies in the docker_monitor itself, some interaction / race going on with other threads.

That explains why test failed on verifying that docker_monitor.log was ingested - docker monitor crashed on ``run()`` so that file was never created and ingested.

It looks like this issue / race has been in the code for a long time already. We were likely just unlucky that we started to hit it more recently.

With having said all that, I'm now 99% sure that those test failures are related to that docker monitor issue and nothing to do with syslog monitor itself.

Going forward, I would just propose adding ``try / except / pass`` around that ``time.strptime`` call, merge that PR and see what happens going forward (aka if we still see this race every now and then).

Kami:
Added a workaround which may help with strptime race - ac113d2d2163b0980d4f6b2ee6c2ba61e110d3d7.

At this point, I would say we should just merge this PR into master and if that issue / test failure ever occurs again, we can look into a different workaround for it.

PR #706, opened 2021-01-20
Warn if "k8s_logs" config option is configured, but Kubernetes monitor is not configured / enabled

Kami:
This pull request updates config parsing code to warn user if ``k8s_logs`` config option is configured, but Kubernetes monitor is not enabled / configured.

## Background, Context

This "issue" was originally reported by Wei Liu. I believe he configured ``k8s_logs`` config option, but not the Kubernetes monitor itself so the ``k8s_logs`` config option had no affect.

I do think this config option is somewhat confusing - it's a top level config option, but (afaik, please correct me if I'm wrong), it's used by Kubernetes monitor so it has no affect without that monitor being configured. I assume we made it a top-level option for "simplicity" reasons, but dunno.

In other cases we don't have such issues since monitor-specific config options go under that monitor config entry.

I also asked @weilliu to confirm this and if that's indeed an issue also update documentation to make that very explicit and obvious to the end users.

### DIFF
--- a/scalyr_agent/util.py
+++ b/scalyr_agent/util.py
@@ -1316,7 +1316,12 @@ class StoppableThread(threading.Thread):
             indefinitely.
         @type timeout: float|None
         """
-        threading.Thread.join(self, timeout)
+        try:
+            threading.Thread.join(self, timeout)
+        except RuntimeError:
+            # Calling .join() on thread which is not started should not be considered fatal
+            pass
+
         if not self.isAlive() and self.__exception_info is not None:
             six.reraise(
                 self.__exception_info[0],
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1608,7 +1608,7 @@ jobs:
     docker:
       - image: circleci/python:3.6-jessie
     environment:
-      VERSION: "4.1.0.1829"
+      VERSION: "4.6.0.2311"
       SCANNER_DIRECTORY: "/tmp/cache/scanner"
       SONAR_USER_HOME: "/tmp/cache/scanner/.sonar"
       OS: "linux"
@@ -1621,7 +1621,7 @@ jobs:
           name: Create cache directory if it doesn't exist
           command: mkdir -p /tmp/cache/scanner
       - restore_cache:
-          key: v2-sonarcloud-scanner-4.1.0.1829
+          key: v2-sonarcloud-scanner-4.6.0.2311
       - run:
           name: Download Sonar Scanner
           command: |
@@ -1635,7 +1635,7 @@ jobs:
             chmod +x $SCANNER_DIRECTORY/sonar-scanner-$VERSION-$OS/bin/sonar-scanner
             chmod +x $SCANNER_DIRECTORY/sonar-scanner-$VERSION-$OS/jre/bin/java
       - save_cache:
-          key: v2sonarcloud-scanner-4.1.0.1829
+          key: v2sonarcloud-scanner-4.6.0.2311
           paths:
             - /tmp/cache/scanner
       - run:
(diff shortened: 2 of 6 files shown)


Kami:
I also noticed this error in the agent log when testing this so I pushed a fix in - 4abe38c1653e2c0d4455b3e19c371ebfb76fd2b0.

```bash
2021-01-20 12:34:32.710Z ERROR [core] [scalyr_logging.py:618] Failed to stop the metric log due to an exception :stack_trace:
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/kubernetes_monitor.py", line 4254, in stop
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/kubernetes_monitor.py", line 2788, in stop
      self.__thread.stop(wait_on_join=wait_on_join, join_timeout=join_timeout)
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 1297, in stop
      self.join(join_timeout)
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 1319, in join
      threading.Thread.join(self, timeout)
    File "/home/kami/.pythonz/pythons/CPython-3.6.9/lib/python3.6/threading.py", line 1051, in join
      raise RuntimeError("cannot join thread before it is started")
  RuntimeError: cannot join thread before it is started
```

In short, under some edge cases in Docker / Kubernetes monitor, ``.join()`` will be called before thread has been started which will result in a runtime exception.

We should ignore that error and not propagate it to the user since it's not fatal in that context.

scalyr_agent/util.py
```diff
@@ -1316,7 +1316,12 @@ def join(self, timeout=None):
             indefinitely.
         @type timeout: float|None
         """
-        threading.Thread.join(self, timeout)
+        try:
+            threading.Thread.join(self, timeout)
+        except RuntimeError:
+            # Calling .join() on thread which is not started should not be considered fatal
```
ArthurKamalov, line 1322: I guess I am missing something but I cannot find the place in this PR where this error suppression may be needed.

Kami: It's not related directly to this PR, but I encountered this issue multiple times when testing those changes.

In short - this edge case was there for a while already.

Kami: Here is an example:

```bash
2021-01-20 12:34:32.710Z ERROR [core] [scalyr_logging.py:618] Failed to stop the metric log due to an exception :stack_trace:
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/kubernetes_monitor.py", line 4254, in stop
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/builtin_monitors/kubernetes_monitor.py", line 2788, in stop
      self.__thread.stop(wait_on_join=wait_on_join, join_timeout=join_timeout)
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 1297, in stop
      self.join(join_timeout)
    File "/home/kami/w/scalyr/scalyr-agent-2/scalyr_agent/util.py", line 1319, in join
      threading.Thread.join(self, timeout)
    File "/home/kami/.pythonz/pythons/CPython-3.6.9/lib/python3.6/threading.py", line 1051, in join
      raise RuntimeError("cannot join thread before it is started")
  RuntimeError: cannot join thread before it is started
```

Technically, we could also try to re-order the code so we never call ``.join()`` on a thread which has been stopped / not started yet, but that would be a larger and more risky change.

ArthurKamalov: oh, sorry I haven't noticed your comment above. I think it's fine if we are sure that we do not also suppress any other errors, first of all - the timeout error. But I am pretty sure that there should be another exception type.

ArthurKamalov (approved):

PR #707, opened 2021-01-20
replace "falsy" check with None check in deprecated option check.

ArthurKamalov:


scalyr_agent/configuration.py
```diff
@@ -3124,7 +3124,7 @@ def __get_config_or_environment_val(
         if config_val is None and deprecated_names is not None:
             for name in deprecated_names:
                 config_val = self.__get_config_val(config_object, name, param_type)
-                if config_val:
+                if config_val is not None:
```
Kami, line 3127: Heh, assumed it's something like that :)

Kami (approved):
LGTM.

Can you please also add a changelog entry?

tests/unit/configuration_test.py
```diff
@@ -2184,7 +2184,8 @@ def test_deprecated_param_names(self):
                         "workers": 4
                     }
                 ],
-                "default_workers_per_api_key": 2
+                "default_workers_per_api_key": 2,
+                "use_multiprocess_copying_workers": false
```
Kami, line 2188: Perhaps also a good idea to add a comment here / similar adding a note that we are explicitly testing this scenario.

PR #711, opened 2021-01-21
Various Windows fixes and improvements

Kami:
This pull request fixes a bug which would result in an exception being thrown if user tried to start agent on Windows with a non-admin user.

An exception would be thrown during permissions escalation phase due to the code not handling that scenario correctly.

While troubleshooting this issue, I also encountered various other scenarios related to incorrect permissions which result in non-user friendly exceptions being thrown.

I tried to update all of those places to throw more explicit and user-friendly error messages.

## Notes on tests

I already mentioned this in YT issue - our current AMI based Windows end to end tests run in shell / non-gui mode which means there is no easy way to write automated test which would have caught the permission escalation issue since that code path depends on GUI.

Technically It's possible to write some kind of GUI based tests for that, but we have no existing framework and code in place for that which means that it would likely be multiple weeks long effort.

### DIFF
--- a/scalyr_agent/platform_windows.py
+++ b/scalyr_agent/platform_windows.py
@@ -926,7 +926,11 @@ class PipeRedirectorClient(RedirectorClient):
                 self.__pipe_handle = None
 
 
-if __name__ == "__main__":
+# Workaround to make sure we don't run this as part of tests on import since it breaks
+# things
+if __name__ == "__main__" and not not compat.os_environ_unicode.get(
+    "INSIDE_WINDOWS_TESTS_PYTEST", False
+):
     if len(sys.argv) == 1:
         servicemanager.Initialize()
         servicemanager.PrepareToHostSingle(ScalyrAgentService)
--- a/scalyr_agent/platform_windows.py
+++ b/scalyr_agent/platform_windows.py
@@ -126,6 +126,16 @@ _SERVICE_DESCRIPTION_ = "Collects logs and metrics and forwards them to Scalyr.c
 # A custom control message that is used to signal the agent should generate a detailed status report.
 _SERVICE_CONTROL_DETAILED_REPORT_ = win32service.SERVICE_USER_DEFINED_CONTROL - 1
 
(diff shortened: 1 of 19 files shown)


scalyr_agent/platform_windows.py
```diff
@@ -896,7 +926,11 @@ def close(self):
                 self.__pipe_handle = None
 
 
-if __name__ == "__main__":
+# Workaround to make sure we don't run this as part of tests on import since it breaks
+# things
+if __name__ == "__main__" and not not compat.os_environ_unicode.get(
+    "INSIDE_WINDOWS_TESTS_PYTEST", False
```
ArthurKamalov, line 932: Will this have an effect when the `platform_windows` module is imported from somewhere else and is not an "entry point" for the interpreter?

Kami: It shouldn't - I tried to design it so the second part of if will only be False during tests (that's also the reason why I tried to a use a unique environment variable name and not something just like ``TEST`` which should never appear in real-life)

The reason for that is that this module has side effect on import which breaks things which means we can't unit test it without a workaround like that.

Kami: Managed to change tests up a bit so we don't need this workaround anymore - now we run platform_windows tests in a standalone fashion.

tests/unit/util/json_util_test.py
```diff
@@ -83,10 +83,10 @@ def test_negative_long(self):
         self.__test_encode_decode(r"-1234567890123456789", -1234567890123456789)
 
     def test_64_bit_int(self):
-        if six.PY3:
-            # this should throw an error because orjson can not serialize more than signed 64 integer.
-            with self.assertRaises(orjson.JSONEncodeError):
-                orjson.dumps(18446744073709551615)
+        # NOTE: latest version of orjson available for Python >= 3.6 can also serialize large
+        # igned and unsigned 64 bit ints now
```
ArthurKamalov, line 87: typo

ArthurKamalov (approved):
Everything is good to me.

PR #712, opened 2021-01-22
Windows package transition to python 3.  

ArthurKamalov:
Since the EOL of the Python 2 was ended and we already started experiencing problems with using Python 2, it was decided to start using Python 3 for the Windows frozen binaries.


.circleci/config.yml
```diff
@@ -2129,48 +2125,26 @@ jobs:
...
             $ProgressPreference = "SilentlyContinue"
             mkdir python_installer -Force
-            if (!(Test-Path .\python_installer\python-2.7.18.amd64.msi -PathType Leaf)) {
-                wget -O .\python_installer\python-2.7.18.amd64.msi https://www.python.org/ftp/python/2.7.18/python-2.7.18.amd64.msi
+            if (!(Test-Path .\python_installer\python-3.8.7-amd64.exe -PathType Leaf)) {
+                wget -O .\python_installer\python-3.8.7-amd64.exe https://www.python.org/ftp/python/3.8.7/python-3.8.7-amd64.exe
             }
-            Start-Process msiexec -ArgumentList "/i python_installer\python-2.7.18.amd64.msi /qn" -wait
-            Start-Process msiexec -ArgumentList "/i python_installer\python-2.7.18.amd64.msi /qn REMOVE=TclTk" -wait
+            python_installer\python-3.8.7-amd64.exe /quiet
```
Kami, line 2131: It's worth noting that we use Nuget for installing Python in windows unit tests job so in theory we could also use that here for consistency :)

.circleci/config.yml
```diff
@@ -2129,48 +2125,26 @@ jobs:
...
 
-      - run:
-          name: Download patched win32evtlog.pyd
-          command: |
-            $ProgressPreference = "SilentlyContinue"
-            mkdir win32evtlog-cache -Force
-            if (!(Test-Path .\win32evtlog-cache\win32evtlog.pyd -PathType Leaf)) {
```
Kami, line 2143: :wave:

.circleci/config.yml
```diff
@@ -2129,48 +2125,26 @@ jobs:
           command: |
             $ProgressPreference = "SilentlyContinue"
             mkdir python_installer -Force
-            if (!(Test-Path .\python_installer\python-2.7.18.amd64.msi -PathType Leaf)) {
-                wget -O .\python_installer\python-2.7.18.amd64.msi https://www.python.org/ftp/python/2.7.18/python-2.7.18.amd64.msi
+            if (!(Test-Path .\python_installer\python-3.8.7-amd64.exe -PathType Leaf)) {
+                wget -O .\python_installer\python-3.8.7-amd64.exe https://www.python.org/ftp/python/3.8.7/python-3.8.7-amd64.exe
```
Kami, line 2129: Just curious how we have decided which Python 3 version to use?

Since we have freedom which one to use, perhaps we could just use the latest one (3.9.x)? Or some modules we depend on are not compatible with it yet?

Kami:
So far, LGTM.

scalyr_agent/platform_windows.py
```diff
@@ -598,12 +599,14 @@ def stop_agent_service(self, quiet):
             win32serviceutil.StopService(_SCALYR_AGENT_SERVICE_)
             if not quiet:
                 print("Agent service has stopped.")
+
+            return 0
         except win32api.error as e:
-            if e[0] == winerror.ERROR_SERVICE_NOT_ACTIVE:
+            if e.winerror == winerror.ERROR_SERVICE_NOT_ACTIVE:
```
Kami, line 605: I assume this change was done in a recent version of pywin32, or?

Kami:
Another thing - now that we are using Python 3, we could also bundle and use ``orjson`` json library (https://pypi.org/project/orjson/).

I believe existing pre-built WIndows wheels should be available.

We already run unit tests with orjson under WIndows and Linux.

@ArthurKamalov wdyt?

Kami:
As discussed, I also pushed a change in #712 (3d5e9740) to run unit tests on Windows under the same Python version we use for pre-built binaries.

Kami:
Also, fyi - 51185e45deee2deee4d1238154223f69a1dd7cc0.

You may encounter similar issues if you try to build / install orjson with old version of pip.

scalyr_agent/platform_windows.py
ArthurKamalov: No, it has to be the change in exception objects in python 3 that does not allow get its attributes through indices.

ArthurKamalov:
@Kami According to the test deployment to the Windows VM, it looks like the `orjson`  is bundled fine with windows frozen binary and it also seems that it doesn't have any issues with an older version of `pip`

.circleci/config.yml
```diff
@@ -2415,161 +2386,16 @@ jobs:
 
 workflows:
  version: 2
- static-analysis:
```
Kami, line 2418: I assume you will add tests back before merging :)

.circleci/config.yml
```diff
@@ -2415,161 +2386,16 @@ jobs:
...
-         context: scalyr-agent
-         requires:
-           - package-test-deb
-         <<: *master_and_release_only
-     - smoke-deb-package-py3:
-         context: scalyr-agent
-         requires:
-           - package-test-deb
-         <<: *master_and_release_only
-#     - ami-tests-development:
```
Kami, line 2505: And also uncomment  that before merging :)

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -599,9 +608,7 @@ def log_event(self, event):
 
         if "TimeCreated" in vals:
             time_format = "%Y-%m-%d %H:%M:%SZ"
-            vals["TimeCreated"] = time.strftime(
-                time_format, time.gmtime(int(vals["TimeCreated"]))
-            )
+            vals["TimeCreated"] = vals["TimeCreated"].strftime(time_format)
```
Kami, line 611: To confirm, ``vals["TimeCreated"]`` is a datetime object, right?

Kami: Is this a change in recent pywin32 version, or? (since it looks like it used to be a timestamp / number, right?)

scalyr_agent/platform_windows.py
Kami: ack, thanks.

Kami (approved):
Couple of small comments / questions, but otherwise, LGTM.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
ArthurKamalov: It should be a change in the `pywin32`. Yes, it used to be an int but it seems that they introduced their own datetime object and now it is used there.

PR #715, opened 2021-01-27
Fix stdout isatty check, update win32_max_open_fds logic to be more robust

ArthurKamalov:
Add a missing negation for the tty check.


Kami:
Alright, I just pushed unit tests for isatty logic for all the possible scenarios to make sure we handle it correctly.

In addition to that, I'm also just finalizing win32_max_open_fds fix :)

Kami:
Alright, I also pushed a fix for win32_max_open_fds issue.

The approach I used is the same one we use for ``json_library`` - we print configured + actual value.

Other possible solutions are much more complicated due to when and how global config options are applied.

ArthurKamalov:
Thanks a lot!

PR #716, opened 2021-01-28
rename worker checkpoint files.

ArthurKamalov:
This PR fixes worker session checkpoints consolidation bug, where the copying manager also was deleting the monitor checkpoint files.

The worker session checkpoint files now have the format - `checkpoints-worker-<worker_id>`. 
THe consolidation logic is also changed to use the glob that only matches the checkpoint files from the worker sessions.

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -3603,6 +3603,19 @@ class Configuration(object):
                 worker_entry, "api_key", self.api_key, description % entry_index
             )
         else:
+
+            # validate the worker id. It should consist only from this set of characters.
+            allowed_worker_id_characters_pattern = "a-zA-Z0-9_"
+            if re.search(
+                r"[^{0}]+".format(allowed_worker_id_characters_pattern), worker_id
+            ):
+                raise BadConfiguration(
+                    "The worker id '{0}' contains an invalid character. Please use only '{1}'".format(
+                        worker_id, allowed_worker_id_characters_pattern
+                    ),
+                    "workers",
+                    "invalidWorkerId",
+                )
             self.__verify_required_string(
                 worker_entry, "api_key", description % entry_index
             )
--- a/scalyr_agent/copying_manager/copying_manager.py
+++ b/scalyr_agent/copying_manager/copying_manager.py
@@ -48,7 +48,7 @@ from scalyr_agent.copying_manager.worker import (
     CopyingManagerWorkerSession,
     create_shared_object_manager,
     CopyingManagerWorkerSessionProxy,
-    WORKER_SESSION_CHECKPOINT_FILENAME_PREFIX,
+    WORKER_SESSION_CHECKPOINT_FILENAME_GLOB,
 )
 from scalyr_agent.log_processing import LogMatcher, LogFileProcessor
 from scalyr_agent.log_watcher import LogWatcher
@@ -75,7 +75,7 @@ def get_worker_session_ids(worker_config):  # type: (Dict) -> List
     result = []
     for i in range(worker_config["sessions"]):
         # combine the id of the worker and session's position in the list to get a session id.
-        worker_session_id = "%s_%s" % (worker_config["id"], i)
+        worker_session_id = "%s-%s" % (worker_config["id"], i)
         result.append(worker_session_id)
     return result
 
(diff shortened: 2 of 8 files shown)


scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1338,15 +1339,21 @@ def __consolidate_checkpoints(self, checkpoints):
...
-            if path == consolidated_checkpoints_path:
-                continue
             os.unlink(path)
 
+            # also delete active checkpoint file.
+            checkpoint_file_name = os.path.basename(path)
+            active_checkpoint_path = os.path.join(
+                os.path.dirname(path), "active-{0}".format(checkpoint_file_name),
+            )
+            if os.path.exists(active_checkpoint_path):
```
Kami, line 1354: May be better to use ``os.path.isfile`` instead.

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1338,15 +1339,21 @@ def __consolidate_checkpoints(self, checkpoints):
 
         # clear data folder by removing all worker session checkpoint files.
         checkpoints_glob = os.path.join(
-            self.__config.agent_data_path, "*checkpoints*.json"
+            self.__config.agent_data_path,
+            "{0}*.json".format(WORKER_SESSION_CHECKPOINT_FILENAME_PREFIX),
         )
 
         for path in scalyr_util.match_glob(checkpoints_glob):
-            # do not delete consolidated checkpoint file
```
Kami, line 1345: Probably also wouldn't hurt to add additional assert here (e.g. at the very least assert ``assert path.contains(WORKER_SESSION_CHECKPOINT_FILENAME_PREFIX)`` or similar).

Kami:
I had a somewhat related comment in another pr - https://github.com/scalyr/scalyr-agent-2/pull/693#discussion_r565965024.

Not sure if we want to address this here or not.

tests/unit/copying_manager_tests/copying_manager_new_test.py
```diff
@@ -301,10 +321,12 @@ def test_checkpoints(self):
         test_file.append_lines("Line7")
         test_file.append_lines("Line8")
 
-        for checkpoint_path in pathlib.Path(
-            self._env_builder.config.agent_data_path
-        ).glob("*checkpoints*.json"):
-            checkpoint_path.unlink()
+        # make sure that all worker session checkpoint files are consolidated and removed.
+        for worker_session in manager.worker_sessions:
```
Kami, line 325: Can we please also add scenario where per worker active-checkpoints file is also created and assertion that this file is deleted as well?

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1338,15 +1339,21 @@ def __consolidate_checkpoints(self, checkpoints):
 
         # clear data folder by removing all worker session checkpoint files.
         checkpoints_glob = os.path.join(
-            self.__config.agent_data_path, "*checkpoints*.json"
+            self.__config.agent_data_path,
+            "{0}*.json".format(WORKER_SESSION_CHECKPOINT_FILENAME_PREFIX),
```
Kami, line 1343: To be even on a safer side, we could also use even stricter glob since we know the file format will always be ``checkpoints-worker-<session id>-<worker id>.json`` - e.g. ``checkpoints-worker-*-*.json``.

But if we do that, we do need to address worker session name related comment from here (https://github.com/scalyr/scalyr-agent-2/pull/693#discussion_r565965024).

This should work as long as we never change the naming format.

Kami:

scalyr_agent/copying_manager/copying_manager.py
ArthurKamalov: Done. I've also added the worker id validation.

scalyr_agent/configuration.py
```diff
@@ -3603,6 +3603,19 @@ def __verify_workers_entry_and_set_defaults(self, worker_entry, entry_index=None
                 worker_entry, "api_key", self.api_key, description % entry_index
             )
         else:
+
+            # validate the worker id. It should consist only from this set of characters.
+            allowed_worker_id_characters_pattern = "a-zA-Z0-9_"
+            if re.search(
+                r"[^{0}]+".format(allowed_worker_id_characters_pattern), worker_id
```
Kami, line 3610: LGTM.

Eventually we could add this as a native "feature" for any string option, but this works for now.

Kami (approved):
LGTM.

scalyr_agent/copying_manager/copying_manager.py
ArthurKamalov: According to your suggestion there https://github.com/scalyr/scalyr-agent-2/pull/716#discussion_r566150203, the glob has to become safer, so maybe there is no need to add this extra check?

PR #717, opened 2021-01-28
Refactor install script

ArthurKamalov:
the most important changes from this PR:
* creates a new `scalyr.repo` file for the `yum` users. This file is also embedded in the install script.
* embeds a Scalyr repo public key into the install script.
* move bash code from the circleci config file into a separate script for the job that builds Linux related artifacts.


scripts/circleci/release/create-agent-installer.sh
```diff
@@ -358,12 +358,50 @@ create_alt_yum_repo_packages;
...
+YUM_REPO_SPEC=$(cat << EOM
+[scalyr]
+includepkgs=scalyr-agent,scalyr-agent-2,scalyr-repo
+name=Scalyr packages - noarch
+baseurl=https://scalyr-repo.s3.amazonaws.com/$RELEASE_REPO_BASE_URL/yum/binaries/noarch
+mirror_expire=300
+metadata_expire=300
+enabled=1
+gpgcheck=1
+gpgkey=${PUBLIC_KEY_URL}
```
Kami, line 377: One option would also be to host that key file on our website / s3 bucket, wdyt?

scripts/circleci/release/create-agent-installer.sh
```diff
@@ -358,12 +358,50 @@ create_alt_yum_repo_packages;
...
+EOM
+)
+
+echo "Create Scalyr yum repo spec file."
+echo "${YUM_REPO_SPEC}" > "scalyr.repo"
```
Kami, line 382: As discussed in the past, we now use this spec file for all our AMI tests, right? (to ensure it works with all the various versions of RPM based distros we support)

scripts/circleci/release/create-agent-installer.sh
Kami: But using ubuntu keyserver should also work for now.

.circleci/config.yml
```diff
@@ -2480,66 +2410,66 @@ workflows:
...
-   triggers:
-     - schedule:
-         cron: "0 0 * * *"
-         filters:
-           branches:
-             only:
-               - master
-   jobs:
-     - ami-tests-stable
+# benchmarks:
```
Kami, line 2413: Let's make sure all those are re-enabled and pass before merging.

scripts/circleci/release/build-linux-packages-and-installer-script.sh
```diff
@@ -0,0 +1,66 @@
+
```
Kami, line 1: Please also add a shebang + license header.

Kami: Also, probably not a bad idea to also use ``set -e``?

scripts/circleci/release/create-agent-installer.sh
ArthurKamalov: In this PR we embed the spec file into the install script, so it is also used in ami tests, since they also use the install script.

Kami (approved):
LGTM.

Do we need any corresponding docs change for this PR?

ArthurKamalov:
@Kami Yes, we need to modify a manual installation instruction.

PR #718, opened 2021-01-29
CT-77 capture k8s container names

dlobue:
This PR adds support for capturing k8s container names in log event metadata. I implemented Imron's suggestion of providing a generic way to add any k8s data we have. In order to add k8s container names to log event metadata the following config is required:

	"k8s_logs": [
	  {
	    "attributes": { "container_name": "${k8s_container_name}" }
	  }
	]

### DIFF
--- a/.circleci/docker_unified_smoke_unit/smoketest/smoketest.py
+++ b/.circleci/docker_unified_smoke_unit/smoketest/smoketest.py
@@ -1026,6 +1026,7 @@ class K8sActor(DockerSmokeTestActor):
                 att.get("stream") in stream_name,
                 att.get("monitor") == "agentKubernetes",
                 process_name in att.get("pod_name"),
+                att.get("container_name") == "scalyr-agent",
             ]
         ):
             return False
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -833,7 +833,7 @@ _CONTROLLER_KEYS = {
 # A regex for splitting a container id and runtime
 _CID_RE = re.compile("^(.+)://(.+)$")
 # A regex to determine whether a string contains template directives
-_TEMPLATE_RE = re.compile("\${[^}]+}")
+_TEMPLATE_RE = re.compile(r"\${[^}]+}")
 
 
 class K8sInitException(Exception):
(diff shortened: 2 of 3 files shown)


yanscalyr (approved):
LGTM

Kami:
I had a very quick and brief look and it looks good to me.

It would also be good / nice to add some smoke tests for this functionality, but this may take quite some effort (especially if you are not familiar with our various testing frameworks).

dlobue:
> LGTM



> I had a very quick and brief look and it looks good to me.
> 
> It would also be good / nice to add some smoke tests for this functionality, but this may take quite some effort (especially if you are not familiar with our various testing frameworks).

I'd also like smoke tests to verify functionality. Pretty sure I won't be able to get them done in time though. Where can I see some examples of existing smoke tests? And how do I run them in my local?

Kami:
@dlobue Sorry for the delay.

Last step we need to get this merged is some basic smoke / end to end tests.

Here is an example setup for k8s end to end / smoke tests we currently use:

* https://github.com/scalyr/scalyr-agent-2/blob/master/.circleci/smoketest_k8s.sh - how we setup kubernetes + verifier image for smoke tests on Circle CI
* https://github.com/scalyr/scalyr-agent-2/blob/master/.circleci/docker_unified_smoke_unit/smoketest/smoketest.py - script which queries scalyr which verifies that the data has been ingested (correctly)
* https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/5385/workflows/ddc1ff64-5f4f-46e7-9d7d-0fd38ebdf254/jobs/74622 - example Circle CI run for that job

I believe you will need to update ``smoketest_k8s.sh`` setup to use config which utilizes this new attribute (may be able to do it using env variable, etc.).

And then you will need to add additional assertions to the corresponding place in ``smoketest.py`` to validate that the attribute is present and contains the correct value.

Sadly right now I don't think it's easy to test / run this setup locally, so it means you might need to develop / experiment on Circle CI (I suggest disabling all other jobs besides ``smoke-k8s`` in circle config to speed things up). And yes, sadly Circle CI development flow is quite slow since you need to make a change, push wait on build, rinse and repeat. But it may still be faster than trying to set everything up to run it locally for that one time.

smoketest_k8s.sh also hasn't been touched for a while so you will likely need to update it to use correct (smoketest.py script from corresponding branch) - https://github.com/scalyr/scalyr-agent-2/blob/master/.circleci/smoketest_k8s.sh#L40, https://github.com/scalyr/scalyr-agent-2/blob/master/.circleci/smoketest_docker.sh#L38). By default it uses some ancient version of that file bundled in that not so frequently updated Docker images.

And yes, I know that setup is not ideal (at least for development, etc.), but sadly refactoring and improving it was never a big priority.

Kami (approved):
Thanks for adding the smoke tests :+1:

PR #719, opened 2021-01-29
Add additional safety checks to the ami-tests-stable job

Kami:
This pull request updates ``ami-tests-stable`` job to ensure that the URL in the installer script indeed points to the correct stable repo.

In the past we had a bug when it could incorrectly point to internal instead of stable repo.


Kami:
Confirmed stable job is now passing with this additional check.

ArthurKamalov (approved):

PR #721, opened 2021-02-03
get checkpoint stats for the dynamic matchers directly from the workers.

ArthurKamalov:
### The issue background:
This issue was introduced with the new "multy-worker" copying manager feature. There is a race condition between the agent's main thread and the worker session threads when they try to access the worker checkpoint files. On the Windows that even leads to a file access error.

### Solution 
The solution is to guard critical code parts with a lock and also fetch the worker checkpoints directly from workers during the agent run without reading their checkpoint files.


scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1340,7 +1341,33 @@ def __consolidate_checkpoints(self, checkpoints):
...
+        previous_checkpoints = self.__read_checkpoint_state(
+            consolidated_checkpoint_path
+        )
+        if previous_checkpoints:
+            checkpoints.update(previous_checkpoints["checkpoints"])
+
+        # get all checkpoints from all files that are currently processed by all workers.
+        for worker in self.workers.values():
+            for worker_session in worker.sessions:
+                checkpoints.update(worker_session.get_full_checkpoints())
```
Kami, line 1366: Is there any chance this could throw?

I guess it's unlikely, probably only if a worker was killed or similat.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1032,18 +1057,30 @@ def __write_checkpoint_state(
...
+                checkpoints_to_write[log_path] = processor_checkpoint
 
             if full_checkpoint:
                 processor.set_inactive()
 
+            # we also save the full collection of the checkpoints in order to share this information
+            # with the copying manager.
+            all_checkpoints[log_path] = processor_checkpoint
+
+        with self._checkpoints_lock:
```
Kami, line 1077: Just thinking out loud - is there a chance for race - e.g. ``get_full_checkpoints`` is called from another thread when ``__write_checkpoint_state`` method is running, but before ``self._full_checkpoints`` is updated?

I imagine it's probably a very short time windows and likely not worth locking on the whole method level.

Kami:
Any chance we could add AMI regression test for Windows issue - aka test which reproduces the original issue / bugs and throws without this change?

scalyr_agent/copying_manager/worker.py
```diff
@@ -1021,6 +1036,16 @@ def __scan_for_new_bytes(self, current_time=None):
...
 
+    def get_full_checkpoints(self):  # type: () -> Dict[six.text_type, Any]
+        """
+        This method returns the full collection of the current log processor checkpoint states.
+        It also mainly designed to be called from the outside threads, so the checkpoint object should be protected
+        by the lock.
+        :return: the dict of the checkpoint states of the current log processors.
+        """
+        with self._checkpoints_lock:
+            return self._full_checkpoints
```
Kami, line 1047: Could also add a comment which says that this variable is updated when writing checkpoints in ``__write_checkpoint_state`` method.

Kami:

scalyr_agent/copying_manager/worker.py
ArthurKamalov: I am also not 100% sure about that.
The chance of this race should be the same as the chance of the original "File is busy by another process" error on windows. So, as we could see from the original [issue](https://scalyr.myjetbrains.com/youtrack/issue/CT-274) it may happen pretty often. The main question is - what is an impact on the log reading process if the `get_full_checkpoints` has returned a bit outdated checkpoint states? I think, in theory, this may lead to the duplicated ingestion. Now I am starting to think that it may be better to wrap it all by the lock...

scalyr_agent/copying_manager/copying_manager.py
ArthurKamalov: Yeah, something really bad should happen with the worker or with the `multiprocessing.Manager`(which provides the IPC with the worker methods). Something, that we cannot handle properly and in this case, the only thing that we can do is to indicate a bad health check of the agent.

Kami (approved):
LGTM, but it would be good to have at least one other person to review it.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: @czerwingithub what do you think about that?

ArthurKamalov:
> LGTM, but it would be good to have at least one other person to review it.

@czerwingithub could you please also have a look at it? (Especially at the https://github.com/scalyr/scalyr-agent-2/pull/721#discussion_r570175809 )

scalyr_agent/copying_manager/copying_manager.py
czerwingithub, line 1366: What happens if there are multiple checkpoints for a given file?  (For example, the consolidated checkpoint file has one checkpoint, but the worker has a newer one.)  It looks like from the sequence, we will always take whatever the worker gives us as the latest version.  However, I wonder if we should explicitly make sure that the timestamp in that checkpoint is newer than what was in the consolidated one.

Would doing that check be useful?  I guess because we always consolidate before we start the workers, there is no real way the worker checkpoint could be older than the consolidated one.  Is that true?

scalyr_agent/copying_manager/worker.py
```diff
@@ -288,6 +288,16 @@ def log_worker_session_status(self):
         """
         pass
 
+    @abstractmethod
+    def get_full_checkpoints(self):  # type: () -> Dict
+        """
+        This method returns the full collection of the current log processor checkpoint states.
```
czerwingithub, line 294: There's a subtle distinction that I think we need to dig into.  Should this return only the checkpoints for what the worker is currently processing, or all checkpoints from its checkpoint file?  From my understanding, the checkpoints here will be used when opening a file to start processing it, so we want to make sure we pick up from where we previously left off.

Here's a concrete scenario to think about:

Suppose we have `file_a.log`:
1.  The main agent process tells worker 1 to copy `file_a.log`.
2. Worker 1 starts copying `file_a.log`, updating the file's position in its own checkpoint file.
3. The main agent process tells worker 1 to stop copying `file_a.log` (there's multiple ways this could happen such as a configuration change removing `file_a.log` from the files that are being copied to a monitor removing it from the files that should be copied.)
4. Worker 1 stops copying `file_a.log` and removes it from its processing list.
5. The main agent process tells worker 1 (or maybe even another worker) to start copying `file_a.log` again.
6. When the main agent process goes to get the list of all checkpoints, worker 1's updates are not included in the returned values from this `get_full_checkpoints` call because `file_a.log` is no longer currently being processed.
7. The main agent process will then just use the latest checkpoint from the consolidated checkpoint file, which will not reflect the place where worker 1 had previously left off from.
8. So, when `file_a.log` starts being copied again, we could try to re-upload old bytes.

I think this is possible, correct?  We only consolidate checkpoints at specific points in time (like start up and shutdown) and I don't think all of those points cover when workers could be start and stopping the log files they are processing.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1021,6 +1036,17 @@ def __scan_for_new_bytes(self, current_time=None):
...
+    def get_full_checkpoints(self):  # type: () -> Dict[six.text_type, Any]
+        """
+        This method returns the full collection of the current log processor checkpoint states.
+        It also mainly designed to be called from the outside threads, so the checkpoint object should be protected
+        by the lock.
+        :return: the dict of the checkpoint states of the current log processors.
+        """
+        with self._checkpoints_lock:
+            # this variable is updated when the '__write_checkpoint_state' method is writing to the checkpoint file.
+            return self._full_checkpoints
```
czerwingithub, line 1048: I think we also need to add a comment either in `__write_checkpoint_state` or where `_full_checkpoints` is declared that the you can only modify `_full_checkpoints` by replacing a reference to it.  I say that because we are just getting a reference to the dictionary while holding the lock.  Once we release the lock, someone could modify the dictionary pointed to by `_full_checkpoints` and screw us up.

The safer approach is to just return a clone of `_full_checkpoints` here.  Which, maybe we should do.  But, it's more copying.

scalyr_agent/copying_manager/worker.py
czerwingithub: I think the race really only leads to bad things in the case where we are about to start copying a log file.  We want to be sure when we start copying `file_a.log` that we are picking up from the latest place we processed.  If that's the case though, then no worker is actively processing `file_a.log` as it is.  So, if no one is currently processing it, then no one is going to be racing here, and we aren't going to get a slightly stale value for the checkpoint.

It does make me think we need to make sure we understand the different regimes here and where the latest checkpoint state for any given log file could be.  I made that point earlier about what happens when a worker has stopped processing a log file but we haven't consolidated all of the checkpoint files again.

In my mental model:

Regime 1: At the start of the agent process, all checkpoints are consolidated into a central checkpoint file.
Regime 2: A worker begins processing a log file.  At this point, the checkpoint state is stored just in the worker's checkpoint file.  More generally, I think whenever a log file has been processed by the agent, we have to assume it's checkpoint state could be in any of the worker's checkpoint files (because the file has moved between workers over time).
Regime 3: A log file has been processed by the agent, but is no longer being processed.  I think this just boils down to regime 2 again... the latest checkpoint state could be held in any of the workers checkpoint files.

All of this is making me think there are more holes in how we are handling checkpoints across the workers.  For example, what happens when a file has been processed by multiple workers and we go to consolidate all of the worker's checkpoint files?  If `file_a.log` exists in multiple worker checkpoints, there is no guarantee that the consolidated version will actually use the latest checkpoint.

czerwingithub (changes requested):
Ok, my apologies, but thinking through this, I fear there are some design problems we have to address.  Take a look at my comments and please see if my mental model is wrong.  I'm assuming that a given log file can be processed by different workers over the course of its lifetime without all checkpoints being consolidated in a safe way.

We should probably jump on a call and discuss this if you and @Kami do agree what I'm suggesting is a problem.

scalyr_agent/copying_manager/copying_manager.py
ArthurKamalov: Yes, I think that's true. The new checkpoint object is always created from the state of the current log processor. Even if the consolidated checkpoint file has already had the saved state for the same log file, this state should be used as the initial position for the log processor when it has been created, so the new checkpoint from the worker should contain at least the same position(in case if a log file hasn't been written) as the consolidated log file.

scalyr_agent/copying_manager/worker.py
Kami: Thanks for bringing this up.

I assumed that checkpoints are also consolidated when a worker itself is stopped, but this may actually not be the case (need to dig in again - I know we do it on the whole agent process start up and shutdown).

And even if we do use this logic - there is still a chance for a race condition in case correct locking is not place.

In any case, good catch and it indeed sounds like we have more work to do. We should probably start with tests for those scenarios and then decide on the approach.

One option would be to also store timestamp in the checkpoint file (iirc, we already do that, but need to double check) and then during consolidation, also check the timestamps and use entry for a particular file with the most recent timestamp. We would of course still need to use a lock to avoid race conditions.

czerwingithub: > I assumed that checkpoints are also consolidated when a worker itself is stopped, but this may actually not be the case (need to dig in again - I know we do it on the whole agent process start up and shutdown).

With the current approach, I think we even have to consolidate any time a log processor is stopped on a worker, not just when an entire worker is stopped.

Yeah, we could go to per-entry timestamps.  That's where I went as well.  That could get messy and will increase the size of the checkpoint file.

The other approach I was contemplating was to just enforce a few key invariants that make reasoning about everything else easier.  For example, the workers only ever write checkpoints for files it is currently processing.  Once it stops processing a file, it will no longer include it in its checkpoint state.  The reasonability for persisting checkpoints for  files no longer being actively processed falls on the main agent process.  Of course, how we handle handing off that state in a safe manner might be tricky, but it might be worth it because it gives us easier invariants to reason with everywhere else.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: > 3. The main agent process tells worker 1 to stop copying file_a.log (there's multiple ways this could happen such as a configuration change removing file_a.log from the files that are being copied to a monitor removing it from the files that should be copied.)

Indeed, there are multiple ways to stop a log processor.
- set a `ignore_stale_files` option for a log file so it will be closed if it will be untouched for a while.
- add and then remove the log matcher dynamically, so it also will close all its current log processors.
- 
In each of those cases, we do not mean to save the checkpoint states for the log processors that are closed.  In other words, if we remove the file and then add it again, we won't start copying from the position where it was removed but from the current position. I was trying to double-check everything and even checked this on the "pre-sharded copying manager" version and it works the same way. 

It seems that the current behavior (and behavior that was also in previous versions) may contradict the logic that you have described above, what do you think?

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -424,6 +444,9 @@ def __init__(self, configuration, monitors):
             self.__config, self.__monitors
         )  # type: List[LogMatcher]
 
+        # this dict stores all checkpoint states for all current and closed files
+        self._checkpoints = {}
```
czerwingithub, line 448: I believe this means our memory usage will grow every time if we start copying a new file.  In the case of K8s, where pods keep coming up and down, this list will grow unbounded.

I would like us to try to make sure our memory usage is only proportional to the number of files we are currently copying.  Previously, we just made sure we read from disk all checkpoints (including files we are no longer processing) when we were looking for old checkpoints.  So, there was only a limited time that we held onto that memory.  I'd like you to think a little more if there's some way we can avoid holding onto all of these checkpoints all of the time.

scalyr_agent/copying_manager/worker.py
```diff
@@ -490,6 +490,7 @@ def run(self):
                     )
                     current_time = time.time()
                     pipeline_time = 0.0
+                    write_active_checkpoints = False
```
czerwingithub, line 493: Add comment for what this var is.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1036,21 +1043,58 @@ def __scan_for_new_bytes(self, current_time=None):
...
+        Update the checkpoints collection from the current state of the log processors.
+        :param current_time:
+        :return:
+        """
+        # get all log processors, including those that need to be removed.
+        log_processors = self.__log_processors
+
+        with self._checkpoints_lock:
+            # remove stale checkpoints using the same config option as for the whole file staleness check.
+            for path, checkpoint in list(self._checkpoints.items()):
```
czerwingithub, line 1057: So, we are changing the workers to also track the checkpoints for files that they have recently processed.  Is that correct?  And we just remove them when they exceed a maximum checkpoint age.

But, we really only need them to hold onto these old checkpoints until the main agent process retrieves this checkpoint and writes it down to disk.  Is that right?

scalyr_agent/copying_manager/worker.py
```diff
@@ -1036,21 +1043,58 @@ def __scan_for_new_bytes(self, current_time=None):
...
+                if checkpoint_time:
+                    if (
+                        current_time
+                        > checkpoint_time + self.__config.max_allowed_checkpoint_age
+                    ):
+                        # remove stale checkpoint states.
+                        del self._checkpoints[path]
+
+            for processor in log_processors:
+                processor_checkpoint = processor.get_checkpoint()
```
czerwingithub, line 1068: It seems like we should do a deep copy here and not work with the object returned by the `get_checkpoint` method.

czerwingithub (commented):
I have a few questions about the design / approach.  Take a look at my comments and let's have a discussion in Office Hours.

PR #722, opened 2021-02-06
Update agent status output to include pids, fix / remove warnings logged in status output

Kami:
This pull request includes two changes.

## 1. Update agent status to include process id of the main agent process and also children worker processes (when configured with multiple workers)

This should make building various 3rd party monitoring / similar scripts on top of agent status output easier (technically pid file for main process is available via other means, but it's located in /var/log/scalyr-agent-2/agent.pid which is really not a standard location for pid files, pid files should really live in /var/run or similar).

Before:

```bash
Scalyr Agent status.  See https://www.scalyr.com/help/scalyr-agent-2 for help

Current time:            Fri Sep  5 23:14:13 2014 UTC
Agent started at:        Thu Sep  4 23:14:13 2014 UTC
Version:                 2.0.0.beta.7
VCS revision:            git revision
Python version:          3.6.8
....
````

After:

```bash
Scalyr Agent status.  See https://www.scalyr.com/help/scalyr-agent-2 for help

Current time:            Fri Sep  5 23:14:13 2014 UTC
Agent started at:        Thu Sep  4 23:14:13 2014 UTC
Main process pid:        1557010
Worker processes pids:   3, 4
Version:                 2.0.0.beta.7
VCS revision:            git revision
Python version:          3.6.8
```

## 2. Fix a bug and ensure we don't emit warnings in the agent status output

We had similar "regression" many times in the past already. Part of the problem is that the whole logic to determine if warnings should be emitted or not when config is parsed is not the best and it's also hard to add test cases which would catch all those scenarios (since a lot of those warnings are only emitted when specific config options are set).

This pull request fixes that and makes sure two of the recently added warnings are not included in the agent status output (status should be machinable parsable and emitting warnings breaks that - it was also included in the JSON output).

Before:

```bash
2021-02-06 12:53:33.856Z WARNING [core] [configuration.py:3234] The configuration option use_multiprocess_copying_workers is deprecated, use use_multiprocess_workers instead.
2021-02-06 12:53:33.857Z WARNING [core] [configuration.py:531] "k8s_logs" config options is defined, but Kubernetes monitor is not configured / enabled. That config option applies to Kubernetes monitor so for it to have an affect, Kubernetes monitor needs to be enabled and configured
Scalyr Agent status.  See https://www.scalyr.com/help/scalyr-agent-2 for help

Current time:            Sat Feb  6 12:53:33 2021 UTC
Agent started at:        Sat Feb  6 12:53:33 2021 UTC
```

After:

```bash
Scalyr Agent status.  See https://www.scalyr.com/help/scalyr-agent-2 for help

Current time:            Sat Feb  6 12:53:33 2021 UTC
Agent started at:        Sat Feb  6 12:53:33 2021 UTC
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Packaged by Arthur Kamalov <arthur@scalyr.com> on Jan 29, 2021 14:00 -0800
 
 Improvements:
 * Add ``Endpoint`` to the default ``event_object_filter`` values for the Kubernetes Event Monitor.
+* Update ``scalyr-agent-2 status -v`` output to also include process id (pid) of the main agent process and children worker processes in scenarios where the agent is configured with multiple worker processes.
+
+Bug fixes:
+* Fix ``scalyr-agent-status -v`` to not emit / print warnings under some edge cases.
 
 ## 2.1.18 "Ravis" - January 29, 2021
 
--- a/scalyr_agent/agent_status.py
+++ b/scalyr_agent/agent_status.py
@@ -610,7 +610,6 @@ def report_status(output, status, current_time):
         "Agent started at:        %s" % scalyr_util.format_time(status.launch_time),
         file=output,
     )
-    print(status)
 
     parent_process_pid = os.getpid()
     print("Main process pid:        %s" % (os.getpid()), file=output)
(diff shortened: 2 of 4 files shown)


ArthurKamalov (approved):
LGTM

PR #723, opened 2021-02-09
Add sample systemd service config file for the agent

Kami:
This pull request adds a sample systemd service config file for the agent with information on how to install / use it.

The "setup process" does include some additional steps which need to be taken care of after agent install and upgrade (aka rc.d symlinks need to be manually removed on install / upgrade because they are created automatically in ``postinst`` script).

We do plan to work on distribution specific packages in the future, but that will take a while and this example config is to service as a stop gap in the mean time.

I tested various situations, including multiple worker processes and verified it works correctly.

## Proposal for a small improvement for manual init.d rc.d symlink removal on install / upgrade

As mentioned above, we do plan to ship packages which will utilize distro specific service managers such as systemd in the future, but this will take a while.

To remove burden on the end user of needing to manually clean up ``init.d`` symlinks so service is not managed twice (once by init.d and once by systemd), I propose a small change to the ``postinstall`` script - only create init.d boot symlinks if ``/etc/scalyr-agent-2/systemd_managed`` file is not present (we would then document this behavior and tell users how to use systemd if they wish to).

That change would be easy to implement and I think would serve as a good enough workaround to reduce post install cleanup burden on the end user until proper distro specific stuff is in place.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -745,6 +745,13 @@ commands:
             i=0
             until [ "$i" -ge ${max_attempts} ]; do
                 echo "Running tests command"
+
+                # On subsequent runs we enable VERBOSE flag to perhaps make
+                # troubleshooting easier
+                if [ "${i}" -ge 1 ]; then
+                    export VERBOSE="true"
+                fi
+
                 ./scripts/circleci/run-ami-tests-in-bg.sh << parameters.test_type >>
                 exit_code=$?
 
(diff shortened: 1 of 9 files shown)


Kami:
@czerwingithub Thanks.

Yeah, I like the ``scalyr-agent-2-config`` idea.

I was thinking of us shipping a bash script or similar to simplify that, but I totally forgot we could also use an existing ``scalyr-agent-2-config`` for that (and it should make it even easier since it's Python).

Kami:
@czerwingithub I pushed ``scalyr-agent-2-config`` approach.

Based on our other priorities and timelines, I think that's about as much as it makes sense to do at this point.

If there are no objections, I will merge this, start beta build and share beta artifacts with Mark so the end user can try it out.

czerwingithub (approved):

PR #724, opened 2021-02-23
Move worker session agent log file settings into a separate log config.

ArthurKamalov:
This pull request fixes the issue that was caused by to broad glob for the agent log file paths. Because of that, the glob also matched another log files such as `agent_syslog`. 

**Solution:**
As a solution, the path in the agent log config is not glob any more, so it matches only the agent log file. For the worker session log files, we introduce another log config that is supposed to match only worker log files by using more strict glob pattern. We also  rename agent worker log files for the same reason. They we renamed from `agent_<worker_session_id>.log` to `agent-worker-session-<worker_session_id>`.log. In theory, this renaming may cause some agent logs skipping in some edge cases, but I think that we still have to do that, since the multi-process  copying manager in still in beta and it hasn't been so widely adopted yet.


scalyr_agent/configuration.py
```diff
@@ -359,23 +362,39 @@ def parse(self):
...
             if self.implicit_agent_log_collection:
                 # set path as glob to handle log files from the multiprocess worker sessions.
-                config = JsonObject(
-                    path="agent*.log",
-                    exclude=JsonArray("*_debug.log"),
+                config = JsonObject(path="agent.log", parser="scalyrAgentLog",)
+
+                # add log config for the worker session agent log files.
+                worker_session_logs_config = JsonObject(
+                    path="%s*.log" % AGENT_WORKER_SESSION_LOG_NAME_PREFIX,
```
Kami, line 374: Yeah, I think this more strict glob approach is indeed much safer and better.

scalyr_agent/configuration.py
```diff
@@ -359,23 +362,39 @@ def parse(self):
...
+            worker_session_agent_logs = None
             if self.implicit_agent_log_collection:
                 # set path as glob to handle log files from the multiprocess worker sessions.
-                config = JsonObject(
-                    path="agent*.log",
-                    exclude=JsonArray("*_debug.log"),
+                config = JsonObject(path="agent.log", parser="scalyrAgentLog",)
+
+                # add log config for the worker session agent log files.
+                worker_session_logs_config = JsonObject(
```
Kami, line 373: I assume we want to add this one even if multiple sessions are not configured, right?

(in case user changes the configuration, to make sure old / previous logs are still picked up)

scalyr_agent/configuration.py
```diff
@@ -359,23 +362,39 @@ def parse(self):
...
+                    path="%s*.log" % AGENT_WORKER_SESSION_LOG_NAME_PREFIX,
+                    # exclude debug files for the worker session debug logs.
+                    # NOTE: the glob for the excluded files has to be restrictive enough,
+                    # so it matches only worker session debug logs.
+                    exclude=JsonArray(
+                        "*%s*_debug.log" % AGENT_WORKER_SESSION_LOG_NAME_PREFIX
```
Kami, line 379: IIRC, if we want actual debug logs to be ingested, user needs to add a manual entry to the config, right (aka we don't have a specific config flag for it, right)?

ArthurKamalov: Right. A new log matcher should be created so there won't be those excluded paths. If i am not mistaken, we are already doing that for the `agent_debug.log` file in smoke tests.

ArthurKamalov: Btw, do you think it would be better to introduce this config option?

scalyr_agent/configuration.py
ArthurKamalov: Correct. We create a new log matcher for the worker logs anyway.

scalyr_agent/configuration.py
czerwingithub: A config option to turn on collection of debug logs?  I'm not sure.  Debug logs can add a fair bit of noise and the users get charged for it.  So, we haven't made it easy for them to accidentally turn them on.  I think we can hold with that for now.

czerwingithub (commented):
This looks like a good approach to me, thanks Arthur for doing it.  I don't have any comments on the coding that Tomaz hasn't already brought up.

scalyr_agent/configuration.py
Kami: Yeah, fair point, agreed with Steven, not needed at this point.

Kami:
So far LGTM, just some more (end to end) tests would be nice.

When merging please also add a changelog entry and reference it fixes regression with sylog monitor log file ingestion.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Improvements:
 
 Bug fixes:
 * Fix ``scalyr-agent-status -v`` to not emit / print warnings under some edge cases.
+* Fix regression with syslog monitor which caused the ingestion of the file with a wrong parser.
```
Kami, line 19: You could perhaps add more details, but I guess this version also works.

Kami (approved):
LGTM :+1:

ArthurKamalov:
Please ignore the last commit name, it just fixes the changelog.

PR #725, opened 2021-02-25
CT-302: K8s metrics on CRI

yanscalyr:
Newer Kubernetes deployments are beginning to default to running in a CRI runtime rather than Docker, we don't gather any metrics or stats from CRI runtimes yet, this PR implements some metrics gathering in CRI.

Container metrics we collect in Docker don't seem to have a good equivalent in any other API we have access to (that I'm aware of at least), commit `d013443` implements grabbing some stats from the Kubelet using the `/stats/summary` endpoint we already use. I don't know if we want these new metrics or not.

TODO:
Smoketests


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -3877,6 +3950,48 @@ def __gather_metrics_from_api(self, containers, cluster_name):
...
+                    k8s_extra = self.__get_k8s_controller_info(info)
+                    if k8s_extra is not None:
+                        k8s_extra.update(cluster_info)
+                        k8s_extra.update({"pod_uid": info["name"]})
+                        all_k8s_extra[info["name"]] = k8s_extra
+
+            pods = stats.get("pods", [])
+
+            for pod in pods:
+                container_stats = pod.get("containers", {})
```
Kami, line 3973: Is this attribute value a dictionary or list of dictionaries?

Based on the code, it looks like it's probably a list of dicts, but then default value we pass to ``pod.get`` is likely wrong.

Kami:
If not already, I think it would be good to do some research on how long querying metrics takes for deployments with many containers.

Based on the code (https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/server/stats/handler.go#L236), it looks like it does some caching, but we should still test it to avoid surprises.

And depending on the metrics we need, we may be able to retrieve less data (https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/server/stats/handler.go#L234), but even if we can do that, would still need to time / benchmark it to confirm it's actually faster than retrieving all the stats.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -4010,39 +4126,46 @@ def gather_sample(self):
...
+                    k8s_cache=k8s_cache,
+                    k8s_include_by_default=self.__include_all,
+                    k8s_namespaces_to_include=self.__namespaces_to_include,
+                    controlled_warmer=self.__metrics_controlled_warmer,
+                )
+            else:
+                containers = self.__container_checker.get_containers()
         try:
             if containers:
+                stats = None
```
Kami, line 4150: I'm having some troubles understanding this piece of code which populates stats variable

Kami:
What's our short and long term testing strategy for that?

I think it would be good to try to run existing end to end tests with a couple of different Kubernetes version (at least the old one which utilizes Docker socket and the new one which utilizes /stats/summary API).

Hopefully that shouldn't be too hard with the existing testing setup.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
Yarnsh: It is a dictionary, the same process of iterating over `pods` and then doing `pod.get()` was already done in the existing `__gather_k8s_metrics_from_kubelet` method, and it works properly from my testing.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -3877,6 +3950,48 @@ def __gather_metrics_from_api(self, containers, cluster_name):
...
+            for pod in pods:
+                container_stats = pod.get("containers", {})
+                for container_stat in container_stats:
+                    container_name = container_stat.get("name", "<unknown>")
+                    if container_name in containers_to_check:
+                        self.__log_cri_container_metrics(container_name, container_stat, all_k8s_extra.get(container_name, {}))
+
+            return stats
+
+        except ConnectionError as e:
```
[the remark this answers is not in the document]
Yarnsh: Do you think we should also make this change in `__gather_k8s_metrics_from_kubelet`? It did the try catch the same way but I figure it wouldn't hurt to clean things up a bit.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
Yarnsh: Yes, `__gather_k8s_metrics_from_kubelet` uses the /stats/summary results regardless of which runtime we are in, before my changes it always called that endpoint itself. In CRI we call that endpoint in my new code as well so I figure there was no reason calling the endpoint twice if we already had the result.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
yanscalyr: It is a dictionary, the same process of iterating over pods and then doing `pod.get()` was already done in the existing `__gather_k8s_metrics_from_kubelet` method, and it works properly from my testing.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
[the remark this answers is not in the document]
yanscalyr: Do you think we should also make this change in `__gather_k8s_metrics_from_kubelet`? It did the try catch the same way but I figure it wouldn't hurt to clean things up a bit.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
yanscalyr: Yes, `__gather_k8s_metrics_from_kubelet` uses the /stats/summary results regardless of which runtime we are in, before my changes it always called that endpoint itself. In CRI we call that endpoint in my new code as well so I figure there was no reason calling the endpoint twice if we already had the result.

yanscalyr:
Regarding testing, I don't think we have any existing end to end tests that actually check anything related to the metrics code. Putting something like that in would be nice but from my limited knowledge of our end to end tests it seems like it would be a somewhat significant effort to do that, and we would end up with pretty lenient tests.

Suggestions for unit tests and such would also be great appreciated.

yanscalyr:
Regarding smoke tests:
Having Minikube run with a CRI runtime in our current setup is throwing me for a loop. Definitely need to upgrade the version we use, and on top of that need to install `cri-o` or `containerd` if we keep running with `--driver=none` (which I think is necessary). My attempts at installing either failed but I'm really not familiar with things like this, I assume there is a relatively simple way I'm missing that isn't building it from source.

Also, I totally forgot that our smokiest image needs to be rebuilt after changes to `smoktest.py`, at the very least there was the container name change to that recently which didn't get built into a new version. Or am I misunderstanding some part of this pipeline?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2279,6 +2286,10 @@ def _get_containers(
                 # add this container to the list of results
                 result[cid] = {
                     "name": cname,
+                    # construct a pod name that is similar to what we see from labels in a docker runtime
+                    "metrics_name": construct_metrics_container_name(
+                        cname, pod_name, pod_namespace, k8s_info.get("pod_uid", "")
```
Kami, line 2291: So if pod_uid is not available, metric name will be ``...__0``, is that desired?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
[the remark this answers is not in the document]
Kami: Yeah, I think it would be better.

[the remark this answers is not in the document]
Kami: EDIT: nvm, I see you already made that change.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -4010,39 +4191,48 @@ def gather_sample(self):
...
-                controlled_warmer=self.__metrics_controlled_warmer,
-            )
+        if self.__report_container_metrics:
+            if self.__client:
+                if (
+                    self.__metrics_controlled_warmer is not None
+                    and not self.__metrics_controlled_warmer.is_running()
+                ):
+                    self.__metrics_controlled_warmer.set_k8s_cache(k8s_cache)
+                    self.__metrics_controlled_warmer.start()
```
Kami, line 4201: Should we perhaps log some rate limited debug log message in both of those if and else branches?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -4010,39 +4191,48 @@ def gather_sample(self):
...
             if containers:
+                stats = None
                 if self.__report_container_metrics:
                     self._logger.log(
                         scalyr_logging.DEBUG_LEVEL_3,
                         "Attempting to retrieve metrics for %d containers"
                         % len(containers),
                     )
-                    self.__gather_metrics_from_api(containers, cluster_name)
+                    if self.__metric_fetcher:
```
Kami, line 4222: Probably also a good idea to log some rate limited info message here (e.g. once per day) something along the lines of - retrieving container metrics from Docker socket / kubeletes API.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -411,10 +411,10 @@ def __init__(
...
         # any of the pod info has changed
         md5 = hashlib.md5()
-        md5.update(name)
-        md5.update(namespace)
-        md5.update(uid)
-        md5.update(node_name)
+        md5.update(name.encode("latin1"))
+        md5.update(namespace.encode("latin1"))
+        md5.update(uid.encode("latin1"))
+        md5.update(node_name.encode("latin1"))
```
Kami, line 417: Hm, can any of those values contain unicode? If it can, we should use ``utf-8`` encoding to avoid surprises.

tests/unit/builtin_monitors/kubernetes_monitor_test.py
```diff
@@ -918,3 +921,185 @@ def _pod(pod_namespace, pod_name, **kwargs):
...
..., "capacityBytes": 1043087360, "availableBytes": 1043075072, "usedBytes": 12288, "time": "2021-03-05T03:39:07Z", "inodes": 254660, "inodesUsed": 9}], "process_stats": {"process_count": 0}, "startTime": "2021-03-05T03:38:26Z", "memory": {"pageFaults": 0, "rssBytes": 10645504, "majorPageFaults": 0, "workingSetBytes": 17891328, "time": "2021-03-05T03:57:22Z", "usageBytes": 25665536}, "cpu": {"usageCoreNanoSeconds": 1464616252, "usageNanoCores": 240928, "time": "2021-03-05T03:57:22Z"}, "containers": [{"logs": {"inodesFree": 3765645, "capacityBytes": 62725623808, "availableBytes": 49265778688, "usedBytes": 8192, "time": "2021-03-05T03:57:30Z", "inodes": 3907584, "inodesUsed": 2}, "name": "kube-proxy", "startTime": "2021-03-05T03:38:27Z", "memory": {"pageFaults": 23067, "rssBytes": 10641408, "majorPageFaults": 1782, "workingSetBytes": 17551360, "time": "2021-03-05T03:57:20Z", "usageBytes": 25288704}, "cpu": {"usageCoreNanoSeconds": 1437692295, "usageNanoCores": 232273, "time": "2021-03-05T03:57:20Z"}, "rootfs": {"inodesFree": 3765645, "capacityBytes": 62725623808, "availableBytes": 49265778688, "usedBytes": 63734, "time": "2021-03-05T03:57:27Z", "inodes": 3907584, "inodesUsed": 25}}]}]}'
```
Kami, line 960: That's OK for now, but in the future you could also store it in a fixture file.

.circleci/config.yml
```diff
@@ -1504,6 +1504,120 @@ jobs:
...
+            kubectl create clusterrolebinding add-on-cluster-admin --clusterrole cluster-admin --serviceaccount=kube-system:default
+      - run:
+          name: dump cluster-info
+          command: |
+            kubectl cluster-info
+            kubectl get po --all-namespaces
+      - run:
+          name: Run Tests (with coverage)
+          command: |
+            source ./.circleci/smoketest_k8s.sh scalyr/scalyr-agent-ci-unittest:4 150 no_delete_existing_k8s_objs
```
Kami, line 1595: Hm, does our k8s verification script already verify that all the corresponding metrics are indeed retrieved and ingested?

If not, we should modify it to check for the presence of at least the basic metrics (ones which are present with Docker and CRI runtime).

Kami:
LGTM so far, just need to wrap up end to end tests :+1:

scalyr_agent/monitor_utils/k8s.py
yanscalyr: The reason I went with "latin1" is because it is supposed to be equivalent to what `md5.update()` does in python2 when passed in a raw string. I'm not actually 100% sure if "latin1" is always the default or if it infers the encoding somehow, utf-8 is probably safer though, I hope.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
yanscalyr: Not exactly desired, but if we are missing expected parts of the `k8s_info` I don't think there is any amount of massaging that will make it look correct. It's really a case we should never hit but maybe something like `<unknown>` would be better, just to make it clear something is wrong.

yanscalyr:
The modifications to `smoketest.py` won't affect the tests until I or someone rebuilds the docker image it is used inside, and I can't do that on phone internet because the downloads of things like gcc all fail. Assuming my internet comes back at some point I will test this change (plus add some more expected metrics) and keep it in.

My attempts at getting minikube running a CRI runtime in CircleCI are still fruitless and I don't think I'm going to get anywhere anytime soon, so I will leave that for another PR.

yanscalyr:
I have created a ticket for myself to modify the smoketests: https://scalyr.myjetbrains.com/youtrack/issue/CT-372
I can't even build the smoketest image without modifications because of how unsupported older python versions are at this point, so it is really becoming its own set of work.

.circleci/config.yml
yanscalyr: Our script does not check for metrics, and on top of that I can't build the image that runs the script anymore. I'm going to split off the smoketest work into its own PR/ticket because there is much more to it than I expected at this point.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
Kami: Yeah, agreed, unknown would be better.

Kami:
> The modifications to smoketest.py won't affect the tests until I or someone rebuilds the docker image it is used inside, and I can't do that on phone internet because the downloads of things like gcc all fail. Assuming my internet comes back at some point I will test this change (plus add some more expected metrics) and keep it in.

I added a quick "hack" for that a while ago since we didn't have any CI in place for updating existing image, etc. - https://github.com/scalyr/scalyr-agent-2/blob/master/.circleci/smoketest_standalone.sh#L53.

It's not great, but it works. Perhaps you can do the same thing for k8s tests.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -424,11 +424,11 @@ def __init__(
         for k, v in six.iteritems(labels):
```
Kami, line 424: This is not something to do with this PR, but just thinking out loud - if we call this ``PodInfo`` code relatively often and on many pods it may actually be beneficial to optimize this hash calculation approach a bit.

Kami: So after checking the code again, I'm not even sure it always works correctly / as expected.

We iterate over and dictionaries and lists before first sorting them. For lists that's probably fine, but for dicts it's not since sorting and as such also iteration order is not guaranteed in Python (and we are not using ``collections.OrderedDict`` either).

Which means that when retrieving and deserializating pod information from Kubernetes API, this digest value may indicate that the pod information has changed, but it actually hasn't and the change is due to the different order of the keys in the dictionary on iteration which of course results in a different hash.

So this could cause unnecessary churn and updating the log configs for the "affected" pods.

Kami (approved):
LGTM, but let's please also check if we can add some basic smoke test without actually updating those Docker image utilizing that "wget hack".

scalyr_agent/monitor_utils/k8s.py
Kami: #733

yanscalyr:
> LGTM, but let's please also check if we can add some basic smoke test without actually updating those Docker image utilizing that "wget hack".

The wget hack does work for running an updated smoketest.py, but our latest smoketest.py doesn't actually pass because we check for container name, but we didn't configure the agent to actually log that. We can most likely hack that in as well, but I really should get this preview build out sooner rather than later, I will handle updates to how we handle smoketests in a separate PR I think.

PR #727, opened 2021-03-09
Release naming and notes for 2.1.19

yanscalyr:
Add k8s container name change to release notes

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -23,6 +23,7 @@
   If you want to apply this configuration to all pods the Agent tracks you will need to add the configuration to the
   Agent deployment, instructions for this can be found [here](https://gist.github.com/yanscalyr/6f986f0475337509894bce03d0a81c11).
   
+  
   Note, there will be charges for the extra bytes sent due to attaching the container name attribute.
 
 ## 2.1.16 "Lasso" - December 23, 2020
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,13 +10,13 @@
 
   Through CLI:
   
-      kubectl annotate pod <pod name> --overwrite log.config.scalyr.com/attributes.container_name=${k8s_container_name}
+      kubectl annotate pod <pod name> --overwrite log.config.scalyr.com/attributes.container_name=\$\{k8s_container_name\}
   
   Through YAML:
   
       metadata:
          annotations:
-            "log.config.scalyr.com/attributes.container_name": "${k8s_container_name}"
+            "log.config.scalyr.com/attributes.container_name": "\$\{k8s_container_name\}"
   
   Whichever pod has this annotation applied will then have the container name to the "container_name" field attached to 
   each log line.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,6 @@
 # Release Notes
 
-## 2.1.19 "TBD" - TBD
+## 2.1.19 "StarTram" - March 9, 2021
 
 * This release adds support for capturing Kubernetes container names to the kubernetes monitor via log line attributes.
   You can configure a container name as a Kubernetes log attribute using the ``${k8s_container_name}`` config template 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,23 @@
 # Release Notes
 
+## 2.1.19 "TBD" - TBD
+
+* This release adds support for capturing Kubernetes container names to the kubernetes monitor via log line attributes.
+  You can configure a container name as a Kubernetes log attribute using the ``${k8s_container_name}`` config template 
+  syntax.
+  
+  ```
+  "k8s_logs": [
+    {
+      "attributes": { "container_name": "${k8s_container_name}" }
+    }
+  ]
+  ```
+  
+  Instructions for how to add configuration to the Kubernetes agent can be found in the documentation [here.](https://app.scalyr.com/help/scalyr-agent-k8s?teamToken=7bn1PMhjS%2F8335sB6vgDcQ--)
+  
+  Note, there will be charges for the extra bytes sent due to attaching the container name attribute.
+
 ## 2.1.16 "Lasso" - December 23, 2020
 
 * This release fixes default permissions for the ``agent.json`` file and ``agent.d/`` directory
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -1,5 +1,23 @@
 # Release Notes
 
+## 2.1.19 "TBD" - TBD
+
+* This release adds support for capturing Kubernetes container names to the kubernetes monitor via log line attributes.
```
czerwingithub, line 5: "This release add support for adding the Kubernetes container name as an attribute to all log lines ingested via the Kubernetes monitor.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,23 @@
...
+  You can configure a container name as a Kubernetes log attribute using the ``${k8s_container_name}`` config template
```
czerwingithub, line 6: You can include the container name as a log line attribute using the ``${k8s_container_name}`` config template syntax:

RELEASE_NOTES.md
```diff
@@ -1,5 +1,23 @@
...
+  
+  ```
+  "k8s_logs": [
+    {
+      "attributes": { "container_name": "${k8s_container_name}" }
+    }
+  ]
+  ```
+  
+  Instructions for how to add configuration to the Kubernetes agent can be found in the documentation [here.](https://app.scalyr.com/help/scalyr-agent-k8s?teamToken=7bn1PMhjS%2F8335sB6vgDcQ--)
```
czerwingithub, line 17: Ok, you should remove the `teamToken` URL parameter here because that maps to the account you were logged in.

czerwingithub, line 17: The other issue is this link by itself doesn't really quickly tell people how to do the configuration.  They have to read through it fairly thoroughly to understand it.  At the very least, we should link to a specific section of the help.

But, after looking over that page, I think we need to add another section to it to show the easiest way to do this. using the `extra_config_directory` approach.  Of course though, it's going to take some time to get that documentation live on the Scalyr site.

Could you create a gist that describes how to do this, with the specific example of creating a config file for holding `k8s_logs`?  I know that's extra work, but I do think we need to create this documentation, even beyond this feature.  

The general approach would be:
* Create a config map that holds this single json file.  Since we will be creating this for the `extra_config_directory`, we do not need to copy in the config files in `agent.d`.  That's the nice thing about this approach.
* Modify the manifest to map in that config map to the directory `/etc/scalyr-agent-2/agent-extra.d`.
* Modify the manifest to define the `SCALYR_K8S_EXTRA_CONFIG_DIRECTORY` env variable to be `/etc/scalyr-agent-2/agent-extra.d`.  (I think that's the right env variable name but you might need to check that)
* Start up the agent.

I think in the near future, we are going to be telling folks to define `k8s_log` more and more... so having an easy way to do that would be good for them.

Can you track down right those instructions up?  Let me know if you have trouble or don't understand what I'm describing -- I know it is a bit abbreviated.

czerwingithub (changes requested):

RELEASE_NOTES.md
yanscalyr: I did start on docs on the site side of things and actually you can configure this with annotations, which might be an easier thing to point people at. In my PR I added a short bit on annotating specifically this, so I could point towards that section, what do you think?

czerwingithub: Yeah, if you can do it with annotations and it's fairly short, let's point them in that direction.  For now, I think it's better to have the full instructions in a gist and then have you copy that into a permanent section of the Kuberenetes documenation.

yanscalyr: Well I planned to write out both approaches in the gist, since the annotation approach would need to be added to every pod you care about (I assume there isn't an annotation way to just set it for all logs, but I might be wrong), but my minikube is dead so I can't test my instructions for the config map approach...
I'll make the gist for the already tested annotation approach for now and see if I can fix minikube I guess.

yanscalyr: I realized the gist for the annotation approach was so short that it would fit very easily into the actual release notes, so I did that.

czerwingithub: Ok, but yeah, this is a bit limited in that you have to apply it to all deployments you want container names.  I'm hoping you can get your minikube fixed and work through the extra config approach.

yanscalyr: Got everything working and tested, instructions are in this gist: https://gist.github.com/yanscalyr/6f986f0475337509894bce03d0a81c11

Kami (approved):
LGTM :+1:

RELEASE_NOTES.md
```diff
@@ -1,5 +1,30 @@
...
+  
+  Through YAML:
+  
+      metadata:
+         annotations:
+            "log.config.scalyr.com/attributes.container_name": "\$\{k8s_container_name\}"
+  
+  Whichever pod has this annotation applied will then have the container name to the "container_name" field attached to 
+  each log line.
+  If you want to apply this configuration to all pods the Agent tracks you will need to add the configuration to the
```
czerwingithub, line 23: put another newline here to break this into its own separate paragraph.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,30 @@
...
+  Agent deployment, instructions for this can be found [here](https://gist.github.com/yanscalyr/6f986f0475337509894bce03d0a81c11).
```
czerwingithub, line 24: I added a comment to this gist with a couple of minor tweaks.

czerwingithub (commented):
I just have one minor comment for this PR and then I made a couple of suggestions in the gist.  Thanks very much for verifying the steps for that extra config map approach.

czerwingithub (approved):

yanscalyr:
Looks like a `mysql` test is failing, which I really doubt is due to this change. Seems like SonarCloud refuses to finish without passing smoke tests though. Is there an easy way to get around this, or should we wait for the test to stabilize?

czerwingithub:
Have you tried re-running it?

yanscalyr:
> Have you tried re-running it?
> […](#)

Yup, a couple of times, it's the same test failing every time.

czerwingithub:
It looks like the mysql server itself is having trouble starting up.  Argh,
yeah, we could disable that test but I'd hate to release something that
could break this monitor that we don't test regularly.

Can you look into running this test locally and see what is causing the
mysql server from starting?  I think Arthur might be able to provide some
guidance on how to run it locally.

PR #728, opened 2021-03-11
Improve "skipping copying log lines" warning edge case handling

Kami:
I noticed that agent would in some conditions log "Warning, skipping copying log lines" warning when that is not really desired or correct.

An example of that is if agent has skipped copied some data due to staleness reasons (e.g. skipForStaleness). In this case, we should not log this warning since we already log a different warning which includes ``skipForStaleness``).

User would see something like this:

> 2021-03-10 22:20:29.390Z WARNING [core] [scalyr-agent-2:1971] Warning, skipping copying log lines.  Only copied 0.0 MB/s log bytes when 0.0 MB/s were generated over the last 5.0 minutes.  This may be desired (due to excessive bytes from a problematic log file).  Please contact support@scalyr.com for additional help.

This PR attempts to fix this.

I believe there are actually some more edge cases where we don't handle this correctly (seen another one in the past, but forget the details) so we should probably audit this code a bit more and ensure current approach is correct / robust enough.


Kami:
@yanscalyr Since you originally worked on that, can you please audit the original code logic for printing this message (and my changes) and ensure it's correctly.

I'm actually not 100% sure my change is the most appropriate one to handle that edge case (would need to dig in some more).

yanscalyr (approved):
LGTM

PR #731, opened 2021-03-17
Debug build for the Kubernetes

ArthurKamalov:


scalyr_agent/log_line_debug.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # global config.
+        self._config = None
+
+    def set_debug_attributes(self, log_file_path, config):
+        self._debug_log_path = log_file_path
+        self._config = config
+
+    @property
+    def is_not_debug_file(self):
```
czerwingithub, line 110: Why not write this as `is_debug_file`?

czerwingithub (commented):
Ok, for the purposes of the debug build, I think this is good.  I think we do not merge this into the master branch yet.  You can just do a debug build from this branch.

I do have a bunch of comments I want to make before we merge it into master, mainly around documentation and maybe altering your approach in some places.  I do think you did a good job of minimizing how much code you had to tweak to get this debugging in there, so thank you for that.

scalyr_agent/log_line_debug.py
ArthurKamalov: It's just avoid using `not` on every property access 😅  The main purpose of this property is to verify if the current log file is not the log file that we want, so we can exit immediately.

https://github.com/scalyr/scalyr-agent-2/blob/1ac1ca901c5f5c1ce7da7a04239917ed47c24332/scalyr_agent/log_line_debug.py#L128-L132

ArthurKamalov:
@czerwingithub Yes, I made a PR just to ease the review and I don't think that we need to merge this too.

czerwingithub:
Ok, let's go ahead and do the debug build.

ArthurKamalov:
@czerwingithub As you may notice, the log file for the AddEventRequest payloads is located in the `/tmp` folder in the Scaly agent pod's container.  I think we shouldn't face any difficulties in fetching this file during the upcoming meeting, WDYT?

scalyr_agent/log_line_debug.py
```diff
@@ -20,6 +26,24 @@
 # the kubernetes monitor will write the id of the needed contained in this variable.
 CONTAINER_ID = None
 
+file_logger = None
+
+original_set_log_destination = scalyr_logging.set_log_destination
```
czerwingithub, line 31: This seems a little brittle and depends on what is importing what (i.e., this requires `log_line_debug` is imported before the call to `set_log_destination`.  In the short term, for this build, we can guarantee that.  However, I don't think this is a good long term approach.

For now, just add a TODO to move this initialization to a better place.

scalyr_agent/log_line_debug.py
```diff
@@ -20,6 +26,24 @@
...
+def scalyr_logging_set_log_destination(*args, **kwargs):
+    global file_logger
+    original_set_log_destination(*args, **kwargs)
+    file_logger = logging.getLogger("PAYLOAD")
+
+    path = os.path.join(
+        tempfile.gettempdir(),
+        "debug_payload.log"
+    )
+    file_handler = logging.FileHandler(filename=path)
```
czerwingithub, line 42: I don't think, by default, that we get a rotating log file handler.  Can you change to use rotation please?  Let's keep 3 rotations and cap each file at 50MB.

scalyr_agent/log_line_debug.py
```diff
@@ -243,6 +267,29 @@ def serialize(self, output_buffer):
...
+        if CONTAINER_ID is None:
+            return result
+
+        m = type(self).SEARCH_PATTERN.search(payload)
+        if m:
+            threads_string, = m.groups(1)
+
+            if CONTAINER_ID in threads_string:
+                if file_logger:
+                    log_message(
```
czerwingithub, line 291: I want to be able to quick extract a single addEvents request from the log.  Let's add marker lines before and after you emit this request.  So something like "REQUEST BEGINS" and "REQUEST END" each on newlines by themselves.

czerwingithub (approved):

PR #734, opened 2021-03-18
save checkpoints for the closed log processors

ArthurKamalov:


scalyr_agent/copying_manager/common.py
```diff
@@ -16,27 +16,83 @@
...
 from scalyr_agent import util as scalyr_util
+from scalyr_agent.scalyr_logging import AgentLogger
+from scalyr_agent.configuration import Configuration
 
 import six
 
 
-def write_checkpoint_state_to_file(checkpoints, file_path, current_time):
-    # type: (Dict, six.text_type, float) -> None
+def read_checkpoint_state_from_file(
```
czerwingithub, line 32: Can you add documentation to each of these methods about who is allowed to invoke them?  For example, only the worker process should read the file containing it's checkpoint.  Same with update and write below.

We need some place to document the overall checkpointing strategy.  Who owns what, what the main process does, what the worker processes do, etc.

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1083,6 +1086,28 @@ def __purge_finished_log_matchers(self):
                 self.remove_log_path(SCHEDULED_DELETION, path)
                 self.__dynamic_matchers.pop(path, None)
 
+    def _merge_checkpoints(self, checkpoints_to_merge):  # type: (List[Dict]) -> Dict
```
czerwingithub, line 1089: Add documentation

scalyr_agent/copying_manager/worker.py
```diff
@@ -1021,29 +1033,53 @@ def __scan_for_new_bytes(self, current_time=None):
...
-        # Create the format that is expected.  An overall dict with the time when the file was written,
-        # and then an entry for each file path.
-        checkpoints = {}
 
-        for processor in log_processors:
-            if full_checkpoint or processor.is_active:
-                checkpoints[processor.get_log_path()] = processor.get_checkpoint()
+        with self._closed_files_checkpoints_lock:
+            result = self._closed_files_checkpoints
+            self._closed_files_checkpoints = {}
```
czerwingithub, line 1063: Don't we want to clear out `_closed_log_processors` here as well?  I don't think I see where we are clearing that list out.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1021,29 +1033,53 @@ def __scan_for_new_bytes(self, current_time=None):
...
-        # type: (List[LogFileProcessor], six.text_type, float, bool) -> None
-        """Writes the current checkpoint state to disk.
+    def _remove_closed_log_processors_and_save_checkpoints(self):
+        new_log_processors = []
 
-        This must be done periodically to ensure that if the agent process stops and starts up again, we pick up
-        from where we left off copying each file.
+        for processor in self.__log_processors:
+            if processor.is_closed():
+                self._closed_log_processors.append(processor)
```
czerwingithub, line 1041: This should be protected by the lock, correct?  I would build a local variable with all of the closed log processors you need to add, and then add them at the end.

scalyr_agent/copying_manager/worker.py
```diff
@@ -392,6 +395,11 @@ def __init__(self, configuration, worker_config_entry, session_id, is_daemon=Fal
         self.__last_total_bytes_copied = 0
         self.__last_total_bytes_pending = 0
 
+        self._closed_files_checkpoints = {}  # type: Dict
+        self._closed_files_checkpoints_lock = threading.Lock()
+
+        self._closed_log_processors = []  # type: List
```
czerwingithub, line 401: Why is this needed?  Does it also require the lock to read/write it?

scalyr_agent/copying_manager/worker.py
```diff
@@ -1052,22 +1088,49 @@ def _write_full_checkpoint_state(self, current_time):
...
-            "%s%s.json" % (WORKER_SESSION_CHECKPOINT_FILENAME_PREFIX, self._id),
+
+        log_processors = self.__log_processors + self._closed_log_processors
+
+        checkpoints = self._get_log_processors_checkpoints(log_processors, current_time)
+
+        for processor in self.__log_processors:
+            # set this log processor inactive, so it won't be included into the "active" checkpoints
+            # until it is marked as active again.
+            processor.set_inactive()
```
czerwingithub, line 1099: If a processor has been closed, is it automatically considered inactive?  I just ask since this is only iterating over `__log_processors` and not `log_processors`.

scalyr_agent/log_processing.py
```diff
@@ -2637,8 +2639,9 @@ def completion_callback(result):
                             close = True
 
                         if close:
-                            self.__log_file_iterator.close()
-                            self.__is_closed = True
+                            self._close()
+                            # self.__log_file_iterator.close()
```
czerwingithub, line 2643: Delete dead code.

scalyr_agent/log_processing.py
```diff
@@ -2811,20 +2814,33 @@ def total_bytes_pending(self):
         self.__lock.release()
         return result
 
+    def _close(self):
```
czerwingithub, line 2817: Document why this and `close` are needed.

scalyr_agent/copying_manager/worker.py
```diff
@@ -1021,29 +1033,53 @@ def __scan_for_new_bytes(self, current_time=None):
...
+
+        self.__log_processors = new_log_processors
+
+        checkpoints = self._get_log_processors_checkpoints(
+            self._closed_log_processors, time.time()
+        )
+        with self._closed_files_checkpoints_lock:
+            self._closed_files_checkpoints.update(checkpoints)
+
+    def get_closed_files_checkpoints(self):  # type: () -> Dict[six.text_type, Any]
```
czerwingithub, line 1053: We should have a different name for this method since this mutates state.  Maybe `get_and_reset_closed_files_checkpoints`?  Or something like that.

czerwingithub (commented):
Ok, some comments on your WIP so far.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: Yes, there has to be a `log_processors ` variable. Thanks.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: I removed this variable, so now only the `_closed_files_checkpoints ` remains.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: This variable was removed https://github.com/scalyr/scalyr-agent-2/pull/734#discussion_r607252502

scalyr_agent/copying_manager/checkpoints.py
```diff
@@ -0,0 +1,122 @@
...
+Since the log copying activity is distributed among worker sessions, those sessions have their own checkpoint files.
+The overall state of the particular worker session is periodically written to the checkpoint file
+named "checkpoints-worker-<WORKER_SESSION_ID>.json". This file contains the checkpoint states for every log file processor
+in the worker session.
+
+Besides that file, there is also the "active-checkpoints" file. This file contains only checkpoint states for the log processors
+that have been active during the current worker session iteration. This is needed to save the latest progress as soon as
+possible to minimize the chance of data loss if something unexpected happens.
+
+Before the agent starts, it looks for all checkpoint files that may be left from the previous run, reads and writes them
```
czerwingithub, line 44: I would rephrase this sentence as "When the agent starts but before the individual workers start executing, the agent looks for all checkpoint files..."

scalyr_agent/copying_manager/checkpoints.py
```diff
@@ -0,0 +1,122 @@
...
+    # read existing checkpoint files (if presented).
+    file_data = read_checkpoint_state_from_file(file_path, logger)
+
+    if file_data:
+        checkpoints = file_data["checkpoints"]
+
+    # remove stale checkpoint states.
+    for path, state in list(checkpoints.items()):
+        state_timestamp = state.get("time")
+        if state_timestamp is None:
```
czerwingithub, line 100: Add a comment explaining what this line is for.  i.e., just to be safe, we will keep checkpoints if they do not have a timestamp.

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1083,6 +1086,44 @@ def __purge_finished_log_matchers(self):
                 self.remove_log_path(SCHEDULED_DELETION, path)
                 self.__dynamic_matchers.pop(path, None)
 
+    @staticmethod
+    def _merge_checkpoints(checkpoints_to_merge):  # type: (List[Dict]) -> Dict
+        """
+        Merge multiple collections of the checkpoints, which are given in the *checkpoints_to_merge* listand return
```
czerwingithub, line 1092: space between `list` and `and`

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1083,6 +1086,44 @@ def __purge_finished_log_matchers(self):
...
+
+                if state_time > result_state_time:
+                    result[path] = state
+
+        return result
+
+    def _get_closed_log_processor_checkpoints(self):
+        """
+        Get all checkpoint states for all closed log processors from all workers.
+        NOTE: In case if different worker sessions has checkpoints for the same log file processors, then the mosr
```
czerwingithub, line 1112: Misspelled `most`

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1105,7 +1146,24 @@ def __scan_for_pending_log_files(self):
         for matcher in log_matchers:
             self.__dynamic_matchers[matcher.log_path] = matcher
 
-        checkpoints = self.__find_and_read_checkpoints()
+        # Before we scan for the new log files for dynamically added log matchers, we have to prepare the checkpoints.
+        # Read previously saved checkpoints.
```
czerwingithub, line 1150: Add a little more detail here.  Example what bug this avoids.

scalyr_agent/copying_manager/worker.py
```diff
@@ -708,6 +722,9 @@ def run(self):
                         self._write_full_checkpoint_state(current_time)
                         last_full_checkpoint_write = current_time
 
+                    # remove all closed log processors
```
czerwingithub, line 725: Can you add any comment about why this should be done here?

scalyr_agent/copying_manager/worker.py
```diff
@@ -964,15 +978,13 @@ def handle_completed_callback(result):
...
-            self.__log_processors = []
-
             add_events_request.close()
 
             total_bytes_copied = 0
 
-            for processor in processor_list:
+            self._closed_files_checkpoints_lock.acquire()
+
+            for processor in self.__log_processors:
```
czerwingithub, line 987: For safety, I would surround this with a `try ... finally` where the lock release is done in the `finally` clause.

czerwingithub (changes requested):

scalyr_agent/copying_manager/worker.py
ArthurKamalov: Wow, that's a huge oversight, thanks.

scalyr_agent/copying_manager/worker.py
```diff
@@ -964,31 +975,33 @@ def handle_completed_callback(result):
...
-                    keep_it = not closed_processor
-                    total_bytes_copied += bytes_copied
-                else:
-                    keep_it = True
-
-                if keep_it:
-                    self.__log_processors.append(processor)
-                else:
-                    processor.close()
+            with self._closed_files_checkpoints_lock:
```
czerwingithub, line 982: I worry that we are holding onto this lock for too long.  Can any of the calls you are making while this lock is held block?  Is there a way to structure this so that the lock can be held for a much shorter time?  For example, while we are going through this loop, store which the checkpoints for the closed processors in a temporary variable.  Once out of the loop, copy it back into the `_closed_files_checkpoints` while you hold the lock.

czerwingithub (commented):
Just one more question and maybe one more change from that.

scalyr_agent/copying_manager/worker.py
ArthurKamalov: Oh, sure. Yeah, there's no need to hold the lock during the whole loop, thanks.

czerwingithub (approved):

PR #741, opened 2021-03-29
Configure journald with globs on multiple fields

imron:
The current journald monitor only allows defining separate log
configurations based on the `unit` field.

This commit makes it possible to define log configurations based on
globs of multiple journald fields.  This is useful when docker has been
configured to use the journald log driver, because in this case the
`unit` will always be `docker`, and it is therefore preferable to match
configuration stanzas based on other fields.

The way this is handled in this commit is through the `journald_globs`
field of the journald log configuration stanzas.

`journald_globs` is a dict of globs, keyed by journald fields.  Only
fields defined in the journald monitor option `journald_fields` are
available for glob matching.

In order to match the configuration stanza, all fields defined in
`journald_globs` must be present in the journald record, and all globs
must match the value of the field.

It is an error to define both `journald_globs` and `journald_unit` for a
single configuration stanza.  It is ok to use both of them individually
in separate configuration stanzas.


Kami:
I guess some people may argue that mixing regexes (for units) and glob patterns for field value filters may be confusing / inconsistent.

I'm personally fine with this approach though and in general prefer using simpler matching methods than regexes to avoid various potential issues related to regexes (performance, pathological cases, etc.).

tests/unit/configuration_test.py
```diff
@@ -2683,6 +2683,69 @@ def test_regex_config(self):
...
+        config.parse()
+
+        lcm = LogConfigManager(config, None)
+        # test matches both
+        matched_config = lcm.get_config({"unit": "testtest", "container": "frobar"})
+        self.assertEqual("TestParser", matched_config["parser"])
+
+        # test when matches one glob but not the other
+        matched_config = lcm.get_config({"unit": "testtest", "container": "foobaz"})
+        self.assertNotEqual("TestParser", matched_config["parser"])
```
Kami, line 2727: I know that match function returns ``matched_config`` instance on match and ``None`` on no match, but perhaps to be on the safe side we could also assert for negatives scenarios that the returned value is ``None``.

scalyr_agent/configuration.py
```diff
@@ -3329,12 +3329,26 @@ def __verify_logs_and_monitors_configs_and_apply_defaults(self, config, file_pat
...
+            try:
+                self.__verify_log_entry_with_key_and_set_defaults(
+                    log_entry,
+                    key="journald_unit",
+                    config_file_path=file_path,
+                    entry_index=i,
+                )
+            except BadConfiguration:
+                bad_config = True
+
```
Kami, line 3342: So it looks like we fall back to trying to find ``journald_globs`` entry if ``journald_unit`` fails validation (and in such case we would also use default value for ``journald_unit`` in the monitor, right?).

I think that's probably fine in most scenarios, but I guess we could also be masking some of the potential issues (e.g. unintentional typo / wrong field value for ``journald_unit`` config option).

So I wonder how we could handle that better.

Perhaps store ``BadConfiguration` error for ``journald_unit`` and re-throw that in case ``journald_globs`` code path returns empty / default value?

I guess that would be better than nothing, but we could still be masking one error scenario (when user tries to define both options, but the first one contains a typo / similar - or would that be handled correctly later on in the monitor code?).

scalyr_agent/configuration.py
```diff
@@ -3468,9 +3485,13 @@ def __verify_log_entry_with_key_and_set_defaults(
             )
         log = None
         if key is not None:
-            # Verify it has a `key` entry that is a string.
-            self.__verify_required_string(log_entry, key, description)
-            log = log_entry.get_string(key)
+            if key_type == "object":
```
Kami, line 3488: Some direct unit tests for the changed functionality in two functions in this module would be nice :)

Kami:
Thanks for the contribution and sorry for the delay.

Overall the change looks good to me, I just had a couple of small questions and comments.

scalyr_agent/builtin_monitors/journald_utils.py
```diff
@@ -82,47 +85,101 @@ def create_config_matcher(self, conf):
...
+        # everything
+        if journald_unit is None and journald_globs is None:
+            journald_unit = ".*"
+            config["journald_unit"] = journald_unit
+
+        match_hash = "monitor"
+        regex = None
+        if journald_unit is not None:
+            match_hash = six.text_type(hash(journald_unit))
+            if journald_unit == ".*":
```
Kami, line 111: This is just thinking out loud and probably micro optimization we don't need (since it's an existing behavior), but we could also return a special version of the matcher  function which just always return ``True`` in case ``journald_unit == ".*"`` to avoid regex match operation (Which I believe should be very fast anyway due to the nature of the regex itself. I would assume Python C regex code already has some special case to bail out early in such scenarios, but not 100% especially since there are some edge case with new line matching where even if ``.*`` is used, whole input string would still need to be checked).

imron:
> I guess some people may argue that mixing regexes (for units) and glob patterns for field value filters may be confusing / inconsistent.
> 
> I'm personally fine with this approach though and in general prefer using simpler matching methods than regexes to avoid various potential issues related to regexes (performance, pathological cases, etc.).

Yeah, this was my main thinking - our use case is solved entirely with globs so wanted to use them, but then needed to keep existing behaviour also, which uses regexes.  Having said that though, I might look at re-doing this with regexes, as it might be useful to use a capture expression and replace parts of the config with that.  That might be more hassle than it's worth though.

tests/unit/configuration_test.py
imron: Will do.

scalyr_agent/configuration.py
imron: I'll have a think to see if there's a better way to solve this.

scalyr_agent/configuration.py
imron: Fair point :-)

scalyr_agent/builtin_monitors/journald_utils.py
imron: Yep, that's an easy thing to do.

Kami (approved):
This has been opened for a while already so I made small changes mentioned in the review and I will go ahead and merge it into master.

scalyr_agent/configuration.py
Kami: For now I made a small change to just always log a message in case we end up here.

It's not the most ideal thing, but it's probably better than nothing. Probably an improvement would be to just log a messages in case the validation fails and the value is also empty.

PR #742, opened 2021-04-06
CT-384: Downgrade missing pod exception to a debug log

yanscalyr:
This error occurs if the Kubernetes events API returns an event for a pod that is already gone, common if the agent is restarted, this does not affect us in any meaningful way and goes away on its own but makes a big red error for the user. This PR will make that error only get logged as a debug log to not spook anybody.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
@@ -830,7 +830,11 @@ This monitor was released and enabled by default in Scalyr Agent version `2.0.43
                                         global_log.log(
                                             scalyr_logging.DEBUG_LEVEL_1,
                                             "Failed to process single k8s event line due to following exception: %s, %s, %s"
-                                            % (repr(e), six.text_type(e), traceback.format_exc()),
+                                            % (
+                                                repr(e),
+                                                six.text_type(e),
+                                                traceback.format_exc(),
+                                            ),
                                         )
                                         continue
                                     if pod and pod.controller:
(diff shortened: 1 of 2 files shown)


scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -819,12 +819,24 @@ def run(self):
...
-                                        query_options=k8s_events_query_options,
-                                    )
+                                    try:
+                                        pod = k8s_cache.pod(
+                                            namespace,
+                                            name,
+                                            current_time,
+                                            query_options=k8s_events_query_options,
+                                        )
+                                    except k8s_utils.K8sApiNotFoundException as e:
```
Kami, line 829: Change looks reasonable to me as long as we don't potentially mask some other errors which should actually be logged under info.

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -819,12 +819,24 @@ def run(self):
...
+                                        global_log.log(
+                                            scalyr_logging.DEBUG_LEVEL_1,
+                                            "Failed to process single k8s event line due to following exception: %s, %s, %s"
```
Kami, line 832: Perhaps we should also log other info (namespace, name, query_options)?

scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
Kami: Also, do you perhaps have access to some error message which arises in such situation?

Just wondering if we can scope this except some more based on the error message..

yanscalyr: ```K8sApiNotFoundException: The resource at location `/api/v1/namespaces/scalyr/pods/<pod name>` was not found```

Not sure if there is any meaningful way we can scope this down further, since this exception is about a queried resource being missing, and the `try` is around the code where we attempt to query a pod and nothing else.

Kami (approved):
LGTM 👍 

Please just add a changelog entry.

PR #744, opened 2021-04-14
CT-377: Support SSL in the MySQL monitor

yanscalyr:
Concern: I exposed `ca_file` as a config option for this monitor, which tells pymysql which file to use when using ssl, but it seems to do nothing. Even when passed a non-existent file SSL connects with no issues. Is this just a mac thing maybe?


Kami:
> Concern: I exposed ca_file as a config option for this monitor, which tells pymysql which file to use when using ssl, but it seems to do nothing. Even when passed a non-existent file SSL connects with no issues. Is this just a mac thing maybe?

It seems like the constructor also takes other arguments such as ``ssl_ca``, etc. (https://pymysql.readthedocs.io/en/latest/modules/connections.html#module-pymysql.connections), so maybe we need to use that?

Or maybe we are bundling some older versions of the library? In this case, will probably need to update it.

And we maybe also want to set some other options such as very mode - it could be that the code is working even with an invalid path since ca verify mode is set to cert optional / none / similar.

Kami:
EDIT: https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/connections.py#L266

Looks like those arguments are just ``convenience`` wrappers around ``ssl`` dict. Having said that, might still not be a bad idea to update bundled library just in case.

yanscalyr:
I attempted to expose more of the SSL options as one thing I saw in my digging is that some options are ignored unless others are also set, but even with that none of them seem to have any effect at all.

From what I can tell basic SSL support doesn't actually require any of these options, not exposing them would likely cover a lot of customer use cases (it would work for the case in the ticket I believe, unless some info was left out). Would it be worth it to merge this PR without those options and possibly loop back around to this for a deeper dive when we all have more time on our hands?

yanscalyr:
Turns out I was passing the wrong thing to the `ssl` parameter. Now the options are sent along correctly, I tested and specifying a CA and cert and key for x509 authentication both work as expected as far as I can tell.

yanscalyr:
Because x509 client authentication still applies over a file socket I also added SSL to that path.

docs/monitors/mysql_monitor.md
```diff
@@ -74,6 +74,10 @@ You can also use this data in [Dashboards](/help/dashboards) and [Alerts](/help/
 |||# ``database_hostport``||| Hostname (or IP address) and port number of the MySQL server, e.g. ``dbserver:3306``, or \
                               simply ``3306`` when connecting to the local machine. You should specify one of \
                               ``database_socket`` or ``database_hostport``, but not both.
+|||# ``use_ssl``          ||| Whether or not to use SSL when connecting to the MySQL server.
```
Kami, line 77: What happens if this value is not provided. Does it use default system ca bundle? Or certifi one? Or something else?

tests/smoke_tests/monitors_test/mysql_test.py
```diff
@@ -103,10 +112,10 @@ class MySqlLogReader(LogMetricReader):
     LINE_PATTERN = r"\s*(?P<timestamp>\d{4}-\d{2}-\d{2}\s+\d{2}:\d{2}:\d{2}.\d+Z)\s\[mysql_monitor\((?P<instance_id>[^\]]+)\)\]\s(?P<metric_name>[^\s]+)\s(?P<metric_value>.+)"
 
 
-def _test(request, python_version):
+def _test(request, python_version, use_socket=True, use_ssl=False, ca_file=None):
```
Kami, line 115: Can we add some tests which verify code throws on failed hostname check and also on invalid CA bundle?

docs/monitors/mysql_monitor.md
yanscalyr: With no `use_ssl` it defaults to False and nothing is used. With no ca bundle defined it doesn't use one at all, it doesn't validate the server. Thats how mysql works for whatever reason but yeah it would be a good idea to clarify that here just in case people aren't aware.

docs/monitors/mysql_monitor.md
```diff
@@ -74,6 +74,11 @@ You can also use this data in [Dashboards](/help/dashboards) and [Alerts](/help/
 |||# ``database_hostport``||| Hostname (or IP address) and port number of the MySQL server, e.g. ``dbserver:3306``, or \
                               simply ``3306`` when connecting to the local machine. You should specify one of \
                               ``database_socket`` or ``database_hostport``, but not both.
+|||# ``use_ssl``          ||| Whether or not to use SSL when connecting to the MySQL server. Defaults to False.
```
Kami, line 77: Per test cases, if this config option is set, specified hostname will also be verified against the hostname in the cert, right?

scalyr_agent/builtin_monitors/mysql_monitor.py
```diff
@@ -96,6 +96,31 @@
...
+define_config_option(
+    __monitor__,
+    "use_ssl",
+    "Whether or not to use SSL when connecting to the MySQL server. Defaults to False.",
+    convert_to=bool,
+)
+define_config_option(
+    __monitor__,
+    "ca_file",
+    "Location of the ca file to use for the SSL connection. Defaults to None, which means the validity of the server "
```
Kami, line 108: Perhaps should say ``server certificate verification won't be performed`` or similar since "validity of the server" sounds a bit weird.

docs/monitors/mysql_monitor.md
Kami: And minor thing - we should probably emit a warning (with a rate limit, e.g. couple of times per day) in case ``use_ssl`` is true, but ``ca_file`` is not specified.

IIRC, we already do the same thing in a couple of places.

yanscalyr: Only if `use_ssl` and also `ca_file` are specified will we verify the hostname. Just the `use_ssl` option alone does nothing except satisfy a mysql server configured with `require_secure_transport`. This is technically a valid use case but warning about it is probably smart.

Kami (approved):
LGTM after the warning change + minor rewording is added.

PR #748, opened 2021-04-20
Remove support (automated tests and checks) for Python 2.6

Kami:
This pull request removes support for Python 2.6 - it removes all the automatic CI checks and tests we run under Python 2.6.

2.6 has been EOL for a long time now (since 2013) and it's getting increasingly hard to run all of our various checks and tests under Python 2.6 due to that.

Code was also updated to emit a warning if agent is running under Python 2.6 - this should be fine since code in the next release will still work under 2.6, but beyond that, we will offer no support and guarantees anymore.


Kami:
Note: It's normal that the coverage will drop a bit until we remove all the Python 2.6 compatibility code (we should do that and remove all the Python 2 related compact code post frozen binaries release).

.circleci/config.yml
```diff
@@ -2351,18 +2301,12 @@ workflows:
...
-         context: scalyr-agent
      - smoke-standalone-27:
          context: scalyr-agent
      - smoke-standalone-35:
          context: scalyr-agent
      - smoke-standalone-35-rate-limit:
          context: scalyr-agent
-     - smoke-standalone-26-tls12:
-         context: scalyr-agent
-     - smoke-standalone-27-tls12:
```
ArthurKamalov, line 2364: Should the 2.7 version also be removed?

ArthurKamalov (approved):
:+1:

.circleci/config.yml
Kami: Sadly we can't do that just yet - we will be able to do that once we actually officially remove support for it.

ArthurKamalov: I mean particularly the `smoke-standalone-27-tls12` :)

PR #749, opened 2021-04-26
CT-394, CT-395: Fix repeating and missing logs when get_containers is unstable

yanscalyr:
This PR does two things:
1. Changes the return value of get_containers in the CRIEnumerator on error to `None`, which matches the Docker enumerator and is something we expect for error handling.
2. Enables `force_add` in the k8s monitor when adding new tracked logs. This is needed in case we lose and find logs in rapid succession which may happen with when get_containers fails. (With change #1 I guess it actually won't happen, it would be an extra layer of protection in case there are other routes to reach this scenario.)

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_monitor.py
@@ -2300,6 +2300,7 @@ class CRIEnumerator(ContainerEnumerator):
                 limit_once_per_x_secs=300,
                 limit_key="query-cri-containers",
             )
+            return None
         return result
 
     def _get_containers_from_filesystem(self, k8s_namespaces_to_include):
@@ -3032,7 +3033,7 @@ class ContainerChecker(object):
         for log in docker_logs:
             if self.__log_watcher:
                 log["log_config"] = self.__log_watcher.add_log_config(
-                    self.__module.module_name, log["log_config"]
+                    self.__module.module_name, log["log_config"], force_add=True
                 )
 
             self.raw_logs.append(log)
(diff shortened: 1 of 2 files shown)


czerwingithub (approved):

czerwingithub:
@yanscalyr Argh, forgot.  Add in something to the change log about the bug you are fixing.

czerwingithub (approved):

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -3032,7 +3033,7 @@ def __start_docker_logs(self, docker_logs):
         for log in docker_logs:
             if self.__log_watcher:
                 log["log_config"] = self.__log_watcher.add_log_config(
-                    self.__module.module_name, log["log_config"]
+                    self.__module.module_name, log["log_config"], force_add=True
```
Kami, line 3036: When I added that change a while back, both @czerwingithub and I agreed it's probably safe to make ``force_add=True`` a default behavior, but to me on the safe side and avoid potential issues, I decided to just set it to True in the Docker monitor where a race / issue was observed.

Having said that, I'm not too surprised that a similar issue exists in the Kubernetes monitor since there is quite a lot of duplicated and similar code in those two monitors.

Kami: EDIT: Yeah, looks like that the issue was present in the same method (https://github.com/scalyr/scalyr-agent-2/blob/master/scalyr_agent/builtin_monitors/docker_monitor.py#L1259) so very likely same root cause for both issues.

PR #752, opened 2021-05-12
Allow daemonset to tolerate all taints, not just master role

dparker-scalyr:
With the current behavior if a  worker has _any_ taint the scalyr-agent is not scheduled. This change allow the scalyr-agent-2 daemonset to tolerate all K8s taints, not just that of the master role. `- operator: Exists` essentially acts as a toleration wildcard.


oliverhsu77 (approved):
LGTM

PR #753, opened 2021-05-27
CT-334 Fix copy truncate's wrong file selection

ArthurKamalov:
### A small background:
If the agent detects that the file was "copy-truncated", then it looks for that copied file. This search process is not so trivial because the `logrotate` supports multiple rotated log filename formats.  

The previous searching method checked if the candidate file name starts with the same prefix as the "copy-truncated"  and might pick the wrong one. For example, for the file `webapp.log` it could pick `webapp_access.log`. 

The solution is to strictly compare file name prefixes instead of just doing the `.startswith` check.


scalyr_agent/log_processing.py
```diff
@@ -1216,25 +1218,37 @@ def __find_copy_truncate_file(self):
...
-        )
+        for other_file in self.__file_system.list_files(dir_path):
+
+            if other_file.endswith(".gz"):
+                continue
+            if other_file == self.__path:
+                continue
+
+            basename = os.path.basename(other_file)
+            parts = basename.split(".")
```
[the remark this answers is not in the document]
oliverhsu77: Any reason not to write this as a `lambda` in the `filter` function?  Functional code is typically easier to understand and less error-prone than all the `continues` in the loop.

oliverhsu77 (changes requested):
Thanks for making this fix.  Comments on handing filenames with multiple `.` in the filename.  Please also see the lint and test failures.

scalyr_agent/log_processing.py
```diff
@@ -1095,6 +1095,8 @@ def __refresh_pending_files(self, current_time):
 
             # See if it is rotated by checking out the file handle we last opened to this file path.
             if current_log_file is not None:
+                if current_log_file.last_known_size > latest_size:
```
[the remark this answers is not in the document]
ArthurKamalov: whoops, the debug leftover. Facepalm.

Thanks.

scalyr_agent/log_processing.py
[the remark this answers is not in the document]
ArthurKamalov: No any serious reason, I just wanted to debug it step by step :)

ArthurKamalov:
@oliverhsu77 I've just added three simple regexp matches to find the rotated files. 
Two of them to find the default formats - the number (`foo.log.1`) and the date (`foo.log-200101`) 
And the third, when the number is located before the ext (`foo.1.log`)

scalyr_agent/log_processing.py
```diff
@@ -1216,18 +1216,39 @@ def __find_copy_truncate_file(self):
...
+                # Since it's not a trivial task to handle all possible variants of the logrotate configuration,
+                # we just looking for the files which have been rotated by the default logrotate settings.
+                # TODO: Add ability to specify the pattern of the rotated files in the agent's config.
+                and (
+                    # if rotated file name ends with a new extension with number (.e.g foo.log.1).
+                    re.match(
+                        r"{0}\.\d+".format(re.escape(file_name)), os.path.basename(f)
+                    )
+                    or
+                    # the same but the number is located before  the extension ("extension" option in logrotate)
```
oliverhsu77, line 1240: NIT: add `(e.g. foo.1.log)`

scalyr_agent/log_processing.py
```diff
@@ -1216,18 +1216,39 @@ def __find_copy_truncate_file(self):
...
-                and not f.endswith(".gz")
+                lambda f: not f.endswith(".gz")
                 and f != self.__path
-                and (int(time.time()) - self.__file_system.stat(f).st_ctime < 300),
+                and (int(time.time()) - self.__file_system.stat(f).st_ctime < 300)
+                # Since it's not a trivial task to handle all possible variants of the logrotate configuration,
+                # we just looking for the files which have been rotated by the default logrotate settings.
+                # TODO: Add ability to specify the pattern of the rotated files in the agent's config.
+                and (
+                    # if rotated file name ends with a new extension with number (.e.g foo.log.1).
```
oliverhsu77, line 1235: NIT: s/.e.g/e.g./

oliverhsu77 (approved):
LGTM.  A couple nit comment suggestions if you would like to make that change.

scalyr_agent/log_processing.py
```diff
@@ -1216,18 +1216,40 @@ def __find_copy_truncate_file(self):
         @return Most recent file that matches copy truncate filename heuristic
         """
         dir_path = os.path.dirname(self.__path)
-        file_prefix = os.path.basename(self.__path)
-        file_prefix = file_prefix[:-4] if file_prefix.endswith(".log") else file_prefix
+        file_name = os.path.basename(self.__path)
+        file_prefix, file_ext = os.path.splitext(file_name)
```
Kami, line 1220: :+1:

scalyr_agent/log_processing.py
```diff
@@ -1216,18 +1216,40 @@ def __find_copy_truncate_file(self):
...
+                            re.escape(file_prefix), re.escape(file_ext)
+                        ),
+                        os.path.basename(f),
+                    )
+                    or
+                    # if rotated file name ends with the date without additional extension (e.g. foo.log-20210527),
+                    # then the only thing that we can check is that the extension starts with original file's extension.
+                    re.match(r"{0}.+".format(re.escape(file_name)), os.path.basename(f))
+                ),
                 self.__file_system.list_files(dir_path),
```
Kami, line 1253: Should we add some debug log there to make it easier to troubleshoot some edge cases? Or is this not needed since we already log elsewhere or similar?

scalyr_agent/log_processing.py
```diff
@@ -1216,18 +1216,40 @@ def __find_copy_truncate_file(self):
...
+                    # the same but the number is located before the extension ("extension" option in logrotate)
+                    # (e.g. foo.1.log)
+                    re.match(
+                        r"{0}\.\d+{1}".format(
+                            re.escape(file_prefix), re.escape(file_ext)
+                        ),
+                        os.path.basename(f),
+                    )
+                    or
+                    # if rotated file name ends with the date without additional extension (e.g. foo.log-20210527),
```
Kami, line 1249: IIRC, depending on the configuration, date can also be placed in the middle (e.g. ``foo-date.log``), but I guess we don't care about it since we don't support custom formats yet?

scalyr_agent/log_processing.py
```diff
@@ -1216,18 +1216,40 @@ def __find_copy_truncate_file(self):
...
+
         possible_copy_filepaths = list(
             six.moves.filter(
-                lambda f: os.path.basename(f).startswith(file_prefix)
-                and not f.endswith(".gz")
+                lambda f: not f.endswith(".gz")
                 and f != self.__path
-                and (int(time.time()) - self.__file_system.stat(f).st_ctime < 300),
+                and (int(time.time()) - self.__file_system.stat(f).st_ctime < 300)
+                # Since it's not a trivial task to handle all possible variants of the logrotate configuration,
```
Kami, line 1231: IIRC, we already have some docs on log rotate, but if not there already, it would be good to also add example file names and formats which are supported.

scalyr_agent/log_processing.py
ArthurKamalov: Yes, for now, we just care about the default logrotate settings, and yes, doing `{0}\d{8}`  definitely should be safer. 
I'll create a new PR for that.

scalyr_agent/log_processing.py
ArthurKamalov: If nothing is found, then the warning message is logged after this function.

tests/smoke_tests/monitors_test/mysql_test.py
```diff
@@ -280,7 +280,7 @@ def test_mysql_python2_ssl_bad_hostname(request):
         use_socket=False,
         use_ssl=True,
         ca_file="/var/lib/mysql/ca.pem",
-        expected_exception="hostname '127.0.0.1' doesn't match 'MySQL_Server_5.7.33_Auto_Generated_Server_Certificate'",
```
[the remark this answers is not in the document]
ArthurKamalov: This is not a special docker image for MySQL, it's just ubuntu:18.04 which installs mysql-server. So I think that was an update in the package manager's repository.

scalyr_agent/log_processing.py
Kami: ack, thanks

tests/smoke_tests/monitors_test/mysql_test.py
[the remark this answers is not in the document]
Kami: ack, thanks for clarification

PR #754, opened 2021-06-01
Small copy-truncate fixes and release info update

ArthurKamalov:
Since we already trying to handle the `extension` option for numbers(e.g. foo.1.log), so it should be also possible to handle the same with the `dateext` option with the default format.


Kami (approved):
LGTM.

Can you please also add some examples of the supported file names and formats to the docs as per https://github.com/scalyr/scalyr-agent-2/pull/753#discussion_r642316315 (I know we already have some docs, but I believe there are no explicit examples).

ArthurKamalov:
> LGTM.
> 
> Can you please also add some examples of the supported file names and formats to the docs as per [#753 (comment)](https://github.com/scalyr/scalyr-agent-2/pull/753#discussion_r642316315) (I know we already have some docs, but I believe there are no explicit examples).

Sorry for missing that. The existing docs should be fine since we haven't added anything more, just fixed the bug.

PR #755, opened 2021-06-11
Adding Docker buildx to support arm64 docker builds

jonathanmccracken:
- Uses docker buildx 
- Publish to both linux/arm64 and linux/amd64 
- Noticed about a 60-70 second build increase time when a buildx builder e.g. (`docker buildx create`), but if the Jenkins agent has a multi-architecture builder installed and reused this build time goes down to 20-30 seconds, so that increase in build time should only happen the first time an Jenkins agent runs this build

### DIFF
--- a/docker/scripts/container_builder_base.sh
+++ b/docker/scripts/container_builder_base.sh
@@ -144,15 +144,12 @@ do
    TAG_OPTIONS="$TAG_OPTIONS -t $x"
 done
 
-run_docker_command "build $TAG_OPTIONS ." "$QUIET" || die "Failed to build the container image"
-
 if [ ! -z "$PUBLISH" ]; then
   report_progress "Publishing image(s)." "$QUIET";
-
-  for x in "${IMAGES[@]}"
-  do
-    run_docker_command "push $x" "$QUIET";
-  done
+  run_docker_command "buildx build --push --platform linux/arm64,linux/amd64 $TAG_OPTIONS ." "$QUIET" || die "Failed to build the container image"
+else
+  report_progress "Building to image(s) to local cache." "$QUIET";
+  run_docker_command "buildx build -o type=image --platform linux/arm64,linux/amd64 $TAG_OPTIONS ." "$QUIET" || die "Failed to build the container image"
 fi
 
 report_progress "Success." "$QUIET";
--- a/docker/scripts/container_builder_base.sh
+++ b/docker/scripts/container_builder_base.sh
@@ -144,6 +144,15 @@ do
    TAG_OPTIONS="$TAG_OPTIONS -t $x"
 done
 
+# Look for prescense of docker buildx instance, otherwise create one
+HAS_BUILD_X=$(docker buildx ls | grep 'docker-container')
+
+if [ -z "$HAS_BUILD_X" ]; then
+  report_progress "Adding Docker buildx instance" "$QUIET";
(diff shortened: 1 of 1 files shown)


docker/scripts/container_builder_base.sh
```diff
@@ -144,15 +144,25 @@ do
    TAG_OPTIONS="$TAG_OPTIONS -t $x"
 done
 
-run_docker_command "build $TAG_OPTIONS ." "$QUIET" || die "Failed to build the container image"
+# Look for prescense of docker buildx instance, otherwise create one
+# 'docker-container' is what the driver is called:
+# https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images
+HAS_BUILD_X=$(docker buildx ls | grep 'docker-container' | cut -d ' ' -f 1)
```
czerwingithub, line 150: Theoretically, our customers can use these scripts to create their own images as well.  Does `buildx` come in `docker` by default or do they have to install additional dependencies?  If it is supported by default, how long has that support been there?  (I want to understand the minimum docker requirements for the customer)

czerwingithub (commented):
Thanks Jonathan for taking this work on.  I just have some questions below.

docker/scripts/container_builder_base.sh
jonathanmccracken: Yes buildx is bundled with docker desktop for windows, mac and Linux (rpm or deb) - https://docs.docker.com/buildx/working-with-buildx/

But it can be installed separately if a customer is missing it for some reason (using an older version of docker) - https://github.com/docker/buildx/

Looks like it's been bundled with docker desktop community edition since August 2019 - https://docs.docker.com/docker-for-windows/release-notes/#docker-desktop-community-2101

czerwingithub: Ok, it's been supported for almost two years.  It seems reasonable to rely on customers having it.  Ok, in general this looks good to me, but I'll defer to Dave on any script-fu related stuff.  Thanks Jonathan.

docker/scripts/container_builder_base.sh
```diff
@@ -144,15 +144,25 @@ do
...
 if [ ! -z "$PUBLISH" ]; then
   report_progress "Publishing image(s)." "$QUIET";
-
-  for x in "${IMAGES[@]}"
-  do
-    run_docker_command "push $x" "$QUIET";
-  done
+  run_docker_command "buildx build --push --platform linux/arm64,linux/amd64 $TAG_OPTIONS ." "$QUIET" || die "Failed to build the container image"
+else
+  report_progress "Building to image(s) to local cache." "$QUIET";
```
dparker-scalyr, line 164: Extra "to", should just be `"Building image(s) to local cache."`

docker/scripts/container_builder_base.sh
```diff
@@ -144,15 +144,25 @@ do
    TAG_OPTIONS="$TAG_OPTIONS -t $x"
 done
 
-run_docker_command "build $TAG_OPTIONS ." "$QUIET" || die "Failed to build the container image"
+# Look for prescense of docker buildx instance, otherwise create one
```
dparker-scalyr, line 147: Minor typo: `'s/prescense/presence/'`

dparker-scalyr (approved):
Two minor typos, but LGTM.

PR #760, opened 2021-07-13
Add Frozen binaries packaging

ArthurKamalov:


Kami:
General comment / question on handling 3rd party dependencies.

We won't use vendored / bundled dependencies anymore, but install them from pip, right?

If so, we should ensure that we pin each dependency to a specific version to ensure reproducable builds.

Ideally we would also specify and verify GPG signatures or at the very least SHA signatures (e.g. https://github.com/envoyproxy/envoy/issues/12951).

On a licensing note - we should likely still have some text file in the repo (e.g. ``3rd_party_deps_licensing_info.md`` or similar where we specify licensing information for all the first level (non-transitive dependencies) - since we will provide packages with all the dependencies bundled we need to be careful licensing wise.

.github/workflows/packages-build.yml
```diff
@@ -0,0 +1,141 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
+        package_type: ["deb", "rpm"]
+        python-version: [3.8]
+        include:
+          - os: windows-2019
+            package_type: "msi"
+            python-version: "3.8"
```
Kami, line 16: Do we need to also specify a specific patch Python version for package reproducibility purposes (e.g. 3.8.x) or it doesn't matter?

scalyr_agent/agent_main.py
```diff
@@ -2152,11 +2145,31 @@ def stop(self):
...
+                 "to signal that the frozen binary is able to run on the current system. If the command fails, "
+                 "then, for some reason, the current system is not able to execute the frozen binary."
+        )
+
     my_controller.add_options(parser)
 
     (options, args) = parser.parse_args()
     my_controller.consume_options(options)
 
+    if __scalyr__.INSTALL_TYPE == __scalyr__.InstallType.TARBALL_INSTALL:
```
Kami, line 2169: Should we add explicit ``else`` with with non-zero exit and print "useful" error message? I know we will already exit with non-zero even without the else clause because of the code below, but it may be confusing since "unrelated" error message will be printed.

PR #765, opened 2021-08-02
Update Windows Event Log monitor to emit a warning when non-suppotred config option is used with new event API

Kami:
This pull request updates Windows Event Log monitor to log a warning in case ``maximum_records_per_source`` config option is used in combination with the new event API. 

This config option is only applicable to the old event log API (should rarely, never? be used these days with frozen binaries and recent versions of Windows) and it's not applicable to the new one so it makes sense to emit a warning to reduce confusion.

Having said that, we should consider if / when old API is still used. If the answer is "never" or "very rarely" we should consider removing that code. If needed, we may need to add additional instrumentation to the monitor to include which version of the API is used in the User-Agent string.


scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -814,6 +816,19 @@ def run(self):
         if isinstance(self.__api, OldApi):
             self._logger.info("Evt API not detected.  Using older EventLog API")
 
+            maximum_records = self._config.get("maximum_records_per_source")
+            if (
+                maximum_records is not None
+                and maximum_records != DEFAULT_MAXIMUM_RECORDS_PER_SOURCE
+            ):
+                # This option is only applicable when using old API so we emit a warning is this
```
ArthurKamalov, line 824: typo: 
...emit a warning **_if_** this ...

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -814,6 +816,19 @@ def run(self):
...
+                # config option is set to a non default value when using a new API
+                self._logger.warn(
```
ArthurKamalov, line 826: Maybe this message should be shown with a new API then?

Kami: Thanks, good catch.

That was an error on my side -  I put it in a wrong branch. I wanted to add unit tests, but it's a bit of PITA.

ArthurKamalov (approved):
LGTM

PR #770, opened 2021-08-13
Do not require buildx to build container images

czerwingithub:
This adds back in the code to build the container images using
normal `docker build` instead of the `docker buildx` approach
that was recently added.  To support both modes, we check to see
if `docker buildx` is supported.  If not, then we use the legacy,
otherwise buildx.

This is necessary to address issues with Jenkins in the Scalyr
build pipeline.


Yarnsh (approved):
Not too familiar with this script but I don't notice any issues, LGTM.

yanscalyr (approved):
And once again with the correct github account...

PR #771, opened 2021-08-14
Frozen binaries/k8s and docker builder

ArthurKamalov:
This pull request adds a new build process for the Kubernetes and Docker images of the Scalyr Agent.

**Important notion:** Kubernetes and docker images still rely on the source code. Due to the nature of the docker images, we can bundle all dependencies during the image build so we don't have to use frozen binaries. This, in turn, should help to support multi-architecture docker builds easier, without building architecture-specific frozen binaries.

### DIFF
--- a/.github/workflows/packages-build.yml
+++ b/.github/workflows/packages-build.yml
@@ -9,28 +9,28 @@ jobs:
       matrix:
         include:
           - package_type: "deb"
-            package_name: "scalyr-agent-2_*.*.*_all.deb"
+            package_filename_glob: "scalyr-agent-2_*.*.*_all.deb"
             os: ubuntu-20.04
             python-version: "3.8"
           - package_type: "rpm"
-            package_name: "scalyr-agent-2-*.*.*-*.noarch.rpm"
+            package_filename_glob: "scalyr-agent-2-*.*.*-*.noarch.rpm"
             os: ubuntu-20.04
             python-version: "3.8"
           - package_type: "tar"
-            package_name: "scalyr-agent-*.*.*.tar.gz"
+            package_filename_glob: "scalyr-agent-*.*.*.tar.gz"
             os: ubuntu-20.04
             python-version: "3.8"
           - package_type: "msi"
-            package_name: "ScalyrAgentInstaller-*.*.*.msi"
+            package_filename_glob: "ScalyrAgentInstaller-*.*.*.msi"
             os: windows-2019
             python-version: "3.8"
           - package_type: "k8s"
             os: ubuntu-20.04
-            package_name: "scalyr-agent-k8s-*.*.*"
+            package_filename_glob: "scalyr-agent-k8s-*.*.*"
             python-version: "3.8"
           - package_type: "docker-json"
             os: ubuntu-20.04
-            package_name: "scalyr-agent-docker-json-*.*.*"
+            package_filename_glob: "scalyr-agent-docker-json-*.*.*"
             python-version: "3.8"
     steps:
       - name: Checkout repository
@@ -77,7 +77,7 @@ jobs:
         with:
           name: "package-${{ matrix.package_type }}"
           path: |
-            agent-build-output/${{ matrix.package_name }}*
+            agent-build-output/${{ matrix.package_filename_glob }}*
 
 
 
@@ -87,7 +87,7 @@ jobs:
         with:
           name: "package-test-${{ matrix.package_type }}"
           path: |
-            agent-build-output/package_test_frozen_binary/package_test*
+            agent-build-output/package_test_frozen_binary/package_test_runner*
 
 
 
@@ -161,7 +161,7 @@ jobs:
       - name: Make package test frozen binaries executable.
         if: matrix.package_type != 'k8s' && matrix.package_type != 'docker-json'
         run: |
-          chmod +x ~/package_test/package_test*
+          chmod +x ~/package_test/package_test_runner*
 
       - name: Test.
         shell: bash
@@ -169,7 +169,7 @@ jobs:
           SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN: ${{ secrets.SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN }}
         run: |
           # Define the package_test_path option only if package test file is presented in artifacts.
-          package_test_path="$(ls ~/package_test/package_test* || true)"
+          package_test_path="$(ls ~/package_test/package_test_runner* || true)"
           if [ -f "${package_test_path}" ]; then
             package_test_path_option="--package-test-path ${package_test_path}"
           else
--- a/.github/workflows/packages-build.yml
+++ b/.github/workflows/packages-build.yml
@@ -145,7 +145,7 @@ jobs:
           name: "package-test-${{ matrix.package_type }}"
           path: ~/package_test
 
-      - name: Start minikube for the test of the kuberneter build.
+      - name: Start minikube for the test of the kubernetes build.
         if: ${{ matrix.package_type == 'k8s' }}
         shell: bash
         run: |
--- a/.github/workflows/packages-build.yml
+++ b/.github/workflows/packages-build.yml
@@ -7,17 +7,31 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        package_type: ["deb", "rpm"]
-        python-version: [3.8]
-        os: [ubuntu-20.04]
         include:
-          - os: ubuntu-20.04
-            package_type: "tar"
+          - package_type: "deb"
+            package_name: "scalyr-agent-2_*.*.*_all.deb"
+            os: ubuntu-20.04
             python-version: "3.8"
-          - os: windows-2019
-            package_type: "msi"
+          - package_type: "rpm"
+            package_name: "scalyr-agent-2-*.*.*-*.noarch.rpm"
+            os: ubuntu-20.04
             python-version: "3.8"
-
+          - package_type: "tar"
+            package_name: "scalyr-agent-*.*.*.tar.gz"
+            os: ubuntu-20.04
+            python-version: "3.8"
+          - package_type: "msi"
+            package_name: "ScalyrAgentInstaller-*.*.*.msi"
+            os: windows-2019
+            python-version: "3.8"
+          - package_type: "k8s"
+            os: ubuntu-20.04
+            package_name: "scalyr-agent-k8s-*.*.*"
+            python-version: "3.8"
+          - package_type: "docker-json"
+            os: ubuntu-20.04
+            package_name: "scalyr-agent-docker-json-*.*.*"
+            python-version: "3.8"
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
@@ -57,24 +71,23 @@ jobs:
         shell: bash
         run: |
           python agent_build/build_package.py ${{ matrix.package_type }} build --output-dir agent-build-output
-          ls -al agent-build-output/${{ matrix.package_type }}/*.${{ matrix.package_type }}*
-
 
       - name: Save the package.
         uses: actions/upload-artifact@v2
         with:
           name: "package-${{ matrix.package_type }}"
           path: |
-            agent-build-output/${{ matrix.package_type }}/*.${{ matrix.package_type }}*
+            agent-build-output/${{ matrix.package_name }}*
 
 
 
       - name: Save the frozen test script.
         uses: actions/upload-artifact@v2
+        if: matrix.package_type != 'k8s' && matrix.package_type != 'docker-json'
         with:
           name: "package-test-${{ matrix.package_type }}"
           path: |
-            agent-build-output/${{ matrix.package_type }}/frozen_binary_test/package_test*
+            agent-build-output/package_test_frozen_binary/package_test*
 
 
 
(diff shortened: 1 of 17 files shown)


agent_build/common.py
```diff
@@ -66,6 +67,46 @@ def recursively_delete_files_by_name(
...
+            for matcher in matchers:
+                if matcher.match(dir_path):
+                    remove_it = True
+            if remove_it:
+                to_remove.append(dir_path)
+
+        # Go back and delete it.  Also, remove it from dirs so that we don't try to walk down it.
+        for remove_dir_path in to_remove:
+            shutil.rmtree(os.path.join(root, remove_dir_path))
+            dirs.remove(remove_dir_path)
```
Kami, line 107: I guess this works fine and doesn't throw "changed array during" iteration or similar error?

tests/package_tests/docker_test.py
```diff
@@ -0,0 +1,116 @@
...
+        ["docker", "exec", "-i", _AGENT_CONTAINER_NAME, "tail", "-f", "-n+1", "/var/log/scalyr-agent-2/agent.log"],
+        stdout=subprocess.PIPE
+    )
+
+    # Read lines from agent.log.
+    pipe_reader = PipeReader(pipe=agent_log_tail_process.stdout)
+
+    timeout_time = datetime.datetime.now() + datetime.timedelta(seconds=COMMON_TIMEOUT)
+    try:
+        while True:
```
Kami, line 72: Do we need some kind of maximum time limit / timeout here instead of while true to make sure we terminate in some reasonable amount of time even if we can't find stats line due to some error / similar?

tests/package_tests/k8s_test.py
```diff
@@ -0,0 +1,207 @@
...
+    )
+    found_errors = []
+
+    # Read lines from agent.log.
+    pipe_reader = PipeReader(pipe=agent_log_tail_process.stdout)
+
+    timeout_time = datetime.datetime.now() + datetime.timedelta(seconds=COMMON_TIMEOUT)
+    error_lines = []
+    try:
+        while True:
```
Kami, line 158: Same here, we probably also need a timeout here so we don't end up in infinite ``while True`` loop in case of an error or similar.

tests/package_tests/k8s_test.py
```diff
@@ -0,0 +1,207 @@
...
+
+            # TODO: add more checks.
+
+            if check_agent_log_request_stats_in_line(line):
+                # The requests status message is found. Stop the loop.
+                break
+    finally:
+        agent_log_tail_process.terminate()
+
+    agent_log_tail_process.communicate()
```
Kami, line 175: Do we also want to stop ``PipeReader`` thread here?

And I guess it's probably safer to run all the "clean up" routines inside ``finally`` block?

tests/package_tests/k8s_test.py
```diff
@@ -0,0 +1,207 @@
...
+    error_lines = []
+    try:
+        while True:
+
+            seconds_until_timeout = timeout_time - datetime.datetime.now()
+            line = pipe_reader.next_line(timeout=seconds_until_timeout.seconds)
+            print(line)
+            # Look for any ERROR message.
+            if check_if_line_an_error(line):
+                error_lines.append(line)
```
Kami, line 165: It appears we don't do anything with those error lines?

Kami: Edit: Looks like ``check_if_line_an_error`` throws. 

Perhaps better name should be ``assert_and_throw_if_line_an_error`` or similar and get rid of if + ``error_lines`` array.

tests/package_tests/common.py
```diff
@@ -0,0 +1,93 @@
...
+        self._queue = queue.Queue()
+        self._started = False
+        self._thread = None
+
+    def read_pipe(self):
+        """
+        Reads lines from pipe. Runs in a separate thread.
+        """
+        while True:
+            line = self._pipe.readline()
```
Kami, line 53: Another approach would be to use ``select.poll`` with a timeout, but threading approach is also fine.

agent_build/common.py
ArthurKamalov: This code just has been copied from the current `build_package.py` script, so it should work is it did before :)

tests/package_tests/docker_test.py
ArthurKamalov: Done. Previously, I relied on the `pipe-reader.next_line` to throw the timeout error, but now I agree that it's better to explicitly do that here.

tests/package_tests/k8s_test.py
ArthurKamalov: I changed the pipe reader, so it doesn't create a thread anymore.

tests/package_tests/k8s_test.py
ArthurKamalov: > It appears we don't do anything with those error lines?

Oops, this is a leftover from my previous code, removed.

.github/workflows/packages-build.yml
```diff
@@ -7,17 +7,31 @@ jobs:
...
     strategy:
       matrix:
-        package_type: ["deb", "rpm"]
-        python-version: [3.8]
-        os: [ubuntu-20.04]
         include:
-          - os: ubuntu-20.04
-            package_type: "tar"
+          - package_type: "deb"
+            package_name: "scalyr-agent-2_*.*.*_all.deb"
```
Kami, line 12: Minor thing - might be better to call this ``package_filename_glob`` to make it obvious it expects / can take a glob pattern.

docker/Dockerfile
```diff
@@ -1,21 +1,23 @@
 # base image that creates all necessary dependencies for
 # the scalyr-agent
 # NOTE: multi-stage builds require Docker 17.05 or greater
-FROM python:2.7-slim as scalyr-dependencies
+FROM python:3.8-slim as scalyr-dependencies
```
Kami, line 4: 👍

tests/package_tests/common.py
```diff
@@ -0,0 +1,135 @@
...
+
+            # Update the remaining data.
+            self._remaining_data = self._remaining_data[first_new_line_index+1:]
+
+            return line.decode()
+
+    def close(self):
+        self._selector.close()
+
+
```
Kami, line 96: I guess all those extra blank lines are not needed?

tests/package_tests/common.py
```diff
@@ -0,0 +1,135 @@
...
+        # create a selector which will poll the pipe.
+        self._selector = selectors.DefaultSelector()
+
+        # register the needed events, for us, it's just all read events.
+        self._selector.register(pipe, selectors.EVENT_READ)
+
+        # Buffer with the part of the incomplete line which is remaining from a previous line read.
+        self._remaining_data = b""
+
+    def next_line(self, timeout: int):
```
Kami, line 56: Can also add type annotation for that that function returns aka ``-> str``.

tests/package_tests/docker_test.py
```diff
@@ -0,0 +1,119 @@
...
+            seconds_until_timeout = (timeout_time - datetime.datetime.now()).seconds
+            if seconds_until_timeout <= 0:
+                raise TimeoutError("Test has timed out.")
+            line = pipe_reader.next_line(timeout=seconds_until_timeout)
+            # Look for any ERROR message.
+            assert_and_throw_if_line_an_error(line)
+
+            # TODO: add more checks.
+
+            if check_agent_log_request_stats_in_line(line):
```
Kami, line 86: To make it easier to troubleshoot (especially scenarios when there is no data for a while), it would probably be good to log something during each poll interval - e.g. "Polling pipe for new line and waiting for x seconds".

tests/package_tests/common.py
```diff
@@ -0,0 +1,135 @@
...
+
+                continue
+
+            # Get the first line.
+            line = self._remaining_data[:first_new_line_index + 1]
+
+            # Update the remaining data.
+            self._remaining_data = self._remaining_data[first_new_line_index+1:]
+
+            return line.decode()
```
Kami, line 91: I guess technically you could also make this method a generator and use yield, but current approach is also fine.

tests/package_tests/docker_test.py
```diff
@@ -0,0 +1,119 @@
...
+    # Read lines from agent.log. Create pipe reader to read lines from the previously created tail process.
+    # Also set the mode for the process' std descriptor as non-blocking.
+    os.set_blocking(agent_log_tail_process.stdout.fileno(), False)
+    pipe_reader = PipeReader(pipe=agent_log_tail_process.stdout)
+
+    timeout_time = datetime.datetime.now() + datetime.timedelta(seconds=COMMON_TIMEOUT)
+    try:
+        while True:
+
+            seconds_until_timeout = (timeout_time - datetime.datetime.now()).seconds
```
Kami, line 77: I'm somewhat confused here - we appear to use the same timeout when waiting for a single line and for the global while loop one.

Is that desired? I would imagine that we would want to use lower timeout for the actual next_line / read timeout (since we expect multiple lines), but dunno.

ArthurKamalov: I thought that there were no significant actions between calculating the timeout and calling the next_line, so the same timeout should be still relevant.

There's only one line for each loop iteration, so the timeout is updated for each new line.

tests/package_tests/common.py
```diff
@@ -0,0 +1,135 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import datetime
+import re
+import selectors
+from typing import IO
+
+AGENT_LOG_LINE_TIMESTAMP = r"\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d+Z"
```
czerwingithub, line 19: Put an example of the timestamp in a comment above this.  Makes reading the regex easier.

tests/package_tests/common.py
```diff
@@ -0,0 +1,135 @@
...
+COMMON_TIMEOUT = 5 * 60
+
+class TestFail(Exception):
+    pass
+
+
+class PipeReader:
+    """
+    The reader for the file-like objects (for now, just for pipes), which uses the 'selectors' module to poll the
+    pip and to read from it in a non-blocking fashion.
```
czerwingithub, line 39: s/pip/pipe/

tests/package_tests/common.py
```diff
@@ -0,0 +1,135 @@
...
+        if bytes_sent <= 0 and requests_sent <= 0:
+            raise TestFail(
+                "Agent log says that during the run the agent has sent zero bytes or requests."
+            )
+        return True
+    else:
+        return False
+
+
+def assert_and_throw_if_line_an_error(line: str):
```
czerwingithub, line 129: I think the better name is just `assert_not_an_error`.  The throwing is implied.

tests/package_tests/common.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    Check if the agent log line is error.
+    """
+    if re.match(rf"{AGENT_LOG_LINE_TIMESTAMP} ERROR .*", line):
+        raise TestFail(f'The next line message is an error:\n"{line}"')
```
czerwingithub, line 134: Why "next" line?  Just say something like "line message is an error"

tests/package_tests/docker_test.py
```diff
@@ -0,0 +1,119 @@
...
+import argparse
+import pathlib as pl
+import subprocess
+import sys
+import time
+import datetime
+import os
+from typing import Union
+
+sys.path.append(str(pl.Path(__file__).absolute().parent.parent.parent))
```
czerwingithub, line 25: Add a comment as to what this is doing.

tests/package_tests/k8s_test.py
```diff
@@ -0,0 +1,205 @@
...
+import os
+import re
+import subprocess
+import pathlib as pl
+import tempfile
+import time
+import sys
+from typing import Union
+
+__SOURCE_ROOT__ = pl.Path(__file__).absolute().parent.parent.parent
```
czerwingithub, line 27: comment as to why.

tests/package_tests/k8s_test.py
```diff
@@ -0,0 +1,205 @@
...
+    scalyr_agent_manifest = scalyr_agent_manifest_source_path.read_text()
+
+    # Change the production image name to the local one.
+    scalyr_agent_manifest = re.sub(
+        r"image: scalyr/scalyr-k8s-agent:\d+\.\d+\.\d+",
+        "image: scalyr/scalyr-k8s-agent:k8s_test",
+        scalyr_agent_manifest
+    )
+
+    # Change image pull policy to be able to bull the local image.
```
czerwingithub, line 119: s/bull/pull/

czerwingithub (changes requested):
Mostly minor stuff.

ArthurKamalov:
@czerwingithub @Kami 
I did some refactoring to reduce code duplication in different package tests. I also added abstractions that can verify the log file's content. We have something similar in our existing smoketests, but new ones should be less complex.

tests/package_tests/common.py
ArthurKamalov: After another refactoring, I removed this class. Now there is a new - LogVerifier class that reads new log data from the process pipe/file and does the verification of this file at the same time. We just have to implement how we read data.

czerwingithub (approved):

PR #772, opened 2021-09-01
Update Python version for the Windows builder to fix the UDP/TCP server memory leak,

ArthurKamalov:


Kami (approved):
LGTM, but let's please add a changelog entry.

Kami (approved):

.circleci/config.yml
```diff
@@ -2113,10 +2113,10 @@ jobs:
           command: |
             $ProgressPreference = "SilentlyContinue"
             mkdir python_installer -Force
-            if (!(Test-Path .\python_installer\python-3.8.7-amd64.exe -PathType Leaf)) {
-                wget -O .\python_installer\python-3.8.7-amd64.exe https://www.python.org/ftp/python/3.8.7/python-3.8.7-amd64.exe
+            if (!(Test-Path .\python_installer\python-3.8.10-amd64.exe -PathType Leaf)) {
+                wget -O .\python_installer\python-3.8.10-amd64.exe https://www.python.org/ftp/python/3.8.10/python-3.8.10-amd64.exe
```
Kami, line 2117: Looks like 3.8.12 is out so we may want to update to that during the next release - https://docs.python.org/release/3.8.12/whatsnew/changelog.html#python-3-8-12-final.

I don't think any of those issues affect us directly, but just to be up to date.

PR #773, opened 2021-09-14
Fix missing timestamp in file processors checkpoint.

ArthurKamalov:
This PR fixes a very nasty bug that was introduced with sharded copying manager. 

The copying manager does not put the 'time' field (it was also added with the copying manager) to the checkpoint object for the files which are immediately removed (by some monitor, as fact). Because of the missing 'time' field, another step - merging checkpoints from all workers, also works wrong and loses such checkpoints. This makes the copying manager reuse the older checkpoint when the log file was removed and added back, and old log lines may be re-uploaded.

The copying manager support tho ways to remove a file from being copied:

1.  To schedule it for removal and so it waits for EOF. 
2.  Remove immediately. 

The first case worked as it should but the second produced checkpoints without a 'time' field.

The fix is to add this field in the 'LogFileProcessor.get_checkpoint' method, to guarantee that every checkpoint has its `time` field.


czerwingithub (approved):
It looks like you have a lint violation.  Go ahead and fix that, then submit.  Thanks!

czerwingithub (approved):

PR #774, opened 2021-09-14
Implement more robust check if buildx is supported by docker

czerwingithub:
There was a bug in the existing checker (checked if a variable was
non-zero length instead of zero length).  I fixed that but also
made the check a little more robust.


yanscalyr (approved):
LGTM as far as I can tell

PR #775, opened 2021-09-22
CT-444 - k8s log events truncated at 16KB

yanscalyr:
It was found that Docker's logging mechanism will split lines at the 16KB point with no way to configure that. Some customers ran into this and assumed it was the Agent splitting the lines.

This PR implements a feature where JSON parsed lines will be merged together depending on if they end in a newline character or not, as not ending in a newline is a sign that the line was split by Docker.

TODO: This currently only implements the happy path, as handling cases of partial lines in the buffer and things like that resulted in issues with the `__fill_buffer` method. I will work on understanding the code better to handle that.

TODO: Tests.

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -2874,7 +2874,7 @@ class Configuration(object):
         self.__verify_or_set_optional_bool(
             config,
             "merge_json_parsed_lines",
-            True,
+            False,
             description,
             apply_defaults,
             env_aware=True,
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -1058,6 +1058,13 @@ class Configuration(object):
         return self.__get_config().get_bool("include_raw_timestamp_field")
 
     @property
+    def merge_json_parsed_lines(self):
+        """If True, and events are getting parsed with the parse_as_json feature, events with messages that do not
+        end in a newline character will be joined together until a newline character is found. This is to work around
+        Docker's logging message length limit of 16KB, which is usually well below our own max message size."""
+        return self.__get_config().get_bool("merge_json_parsed_lines")
+
+    @property
     def enable_profiling(self):
         return self.__get_config().get_bool("enable_profiling")
 
@@ -2858,6 +2865,15 @@ class Configuration(object):
         self.__verify_or_set_optional_bool(
             config,
             "include_raw_timestamp_field",
+            True,
+            description,
+            apply_defaults,
+            env_aware=True,
+        )
+
+        self.__verify_or_set_optional_bool(
+            config,
+            "merge_json_parsed_lines",
             True,
             description,
             apply_defaults,
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -1061,7 +1061,8 @@ class Configuration(object):
     def merge_json_parsed_lines(self):
         """If True, and events are getting parsed with the parse_as_json feature, events with messages that do not
         end in a newline character will be joined together until a newline character is found. This is to work around
-        Docker's logging message length limit of 16KB, which is usually well below our own max message size."""
+        Docker's logging message length limit of 16KB, which is usually well below our own max message size.
+        The final merged line will use the timestamp and other attributes of the first line."""
         return self.__get_config().get_bool("merge_json_parsed_lines")
 
     @property
--- a/scalyr_agent/log_processing.py
+++ b/scalyr_agent/log_processing.py
@@ -329,8 +329,8 @@ class LogFileIterator(object):
 
                 if "position" in checkpoint:
                     self.__position = checkpoint["position"]
-                    self.__extended_line_position = LogFileIterator.ExtendedLinePosition.from_checkpoint(
-                        checkpoint
+                    self.__extended_line_position = (
+                        LogFileIterator.ExtendedLinePosition.from_checkpoint(checkpoint)
                     )
                     for state in checkpoint["pending_files"]:
                         if not state["is_log_file"] or self.__file_system.trust_inodes:
@@ -645,8 +645,8 @@ class LogFileIterator(object):
         # Do we need more bytes to have at least max_line_bytes in available in the buffer.
         need_more_bytes_for_max_line = available_buffer_bytes < self.__max_line_length
         # If we are on an extended line, do we need more bytes in buffer to read the next fragment
-        need_more_bytes_for_fragment_position = not self.__have_buffer_for_fragment_position(
-            available_buffer_bytes
+        need_more_bytes_for_fragment_position = (
+            not self.__have_buffer_for_fragment_position(available_buffer_bytes)
         )
 
         if more_file_bytes_available and (
@@ -1353,9 +1353,12 @@ class LogFileIterator(object):
             tmp = self.__buffer.read()
             new_buffer.write(tmp)
             if expected_bytes != new_buffer.tell():
-                assert expected_bytes == new_buffer.tell(), (
-                    'Failed to get the right number of left over bytes %d %d "%s"'
-                    % (expected_bytes, new_buffer.tell(), tmp,)
+                assert (
+                    expected_bytes == new_buffer.tell()
+                ), 'Failed to get the right number of left over bytes %d %d "%s"' % (
+                    expected_bytes,
+                    new_buffer.tell(),
+                    tmp,
                 )
 
         self.__buffer = new_buffer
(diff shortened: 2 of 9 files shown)


czerwingithub:
Yan, ok, this is going to be larger change than I thought.  My apologies, I should have better anticipated some of the problems that you would have.  You have a good first start, but yes, it does not address some of the important functionality that you have to preserve.

The most important one is that you need to make sure the `position` of the `LogFileIterator` accurately reflects what has been returned from `readline`.  Since you haven't addressed the corner case when you have a partial line, I think you are doing the right thing now wrt to position.  However, when you try to address that corner case, you are going to have problems.

Basically, `position` should always reflect the exact place in the file where the agent should start reading from to get the next line after the one that was last returned by `readline`.  This is very important to make sure the checkpointing works correctly.

One simple approach to handle the partial line issue (that won't work) is:
* If you don't find the end of the Docker log line during a call to `readline`, just store the current line you've built up in an instance variable and return empty.
* On the next call, keep looking for that newline, adding to the instance variable as you go.

The problem with this approach is that `position` will point to wherever you left off building up that line, looking for the newline.  It will not point at the position just before the line you are building up.  That's necessary because if the agent is restarted, we need to pick up reading from that point again.

Instead, I think the simplest approach that we can take is:
* If you don't see a newline before you see a partial line, then reset the position to its original location from the start of the `readline` call.  And return empty to indicate the line is not yet ready.

This means that on the next call to `readline`, you are going to be reparsing all of the JSON that you just finished parsing... but, this is probably going to be a corner case that does not get exercised that often.

I would suggest you check to see if that approach is viable.  Make sure you can safely reset position.  Also, see how it interacts with the extended buffer.  I will also say that we should probably never be trying to use extended buffer at the same time we are doing this line joining.  (The extended buffer is meant to handle the case when the Docker logs generated very long log lines.  The stiching is needed because Docker is now making sure it doesn't emit long log lines).  So, you could make the simpilfying assumption that you never need both at the same time.

Kami:
On a somewhat related note - do we need / want some kind of timeout in this case (aka if we don't see ``\n`` / complete line for whatever reason we simply return partial one so we don't potentially get stuck while waiting for a complete line)?

So basically something similar as we do in the syslog monitor. Granted syslog monitor and protocol is quite different.

czerwingithub:
On Mon, Sep 27, 2021 at 1:52 AM Tomaz Muraus ***@***.***>
wrote:

> On a somewhat related note - do we need / want some kind of timeout in
> this case (aka if we don't see \n / complete line for whatever reason we
> simply return partial one so we don't potentially get stuck while waiting
> for a complete line)?
>

We have an existing timeout for partial lines to handle the case of a
newline taking too long to be observed from the log file.  Yes, we should
use that same timeout and depending on how Yan piggybacks this, he might be
able to leverage that same behavior.

yanscalyr:
My current (uncommitted) approach was to use a copy timing mechanism we use handling partial lines, in `log_processing.py` where I join the lines. I don't believe the existing mechanism will work since earlier chunks of the line will reset the timer.

scalyr_agent/log_processing.py
```diff
@@ -750,6 +738,39 @@ def readline(self, current_time=None):
...
+
+                            if (
+                                current_time - self.__merge_json_line_time
+                                < self.__line_completion_wait_time
+                            ):
+                                self.__buffer.seek(original_buffer_position)
+                                self.__position = self.__determine_mark_position(
+                                    self.__buffer.tell()
+                                )
+                                return LogLine(line=b"")
```
Kami, line 762: Should we also log something here (I guess at least at the DEBUG level although since it's an edge case which shouldn't happen often, it may also be OK to use INFO or similar).

scalyr_agent/log_processing.py
```diff
@@ -810,6 +831,28 @@ def readline(self, current_time=None):
...
+        next_line = self.__line_matcher.readline(self.__buffer, current_time)
+
+        # Check to see if we allow for extended lines, and if so, then read more pages so that
+        # we can parse an entire extended line.
+        if (
+            len(next_line) == 0
+            and self.__max_extended_line_length > self.__max_line_length
+        ):
+            while (
+                self.__available_buffer_bytes() < self.__max_extended_line_length
```
Kami, line 845: Just trying to understand if we need any additional (time / iteration count based) guard in this while loop to prevent infinite loop or being stuck here for a long time.

I did check all the methods which are called here and it appears it shouldn't be needed, but wanted to double check.

scalyr_agent/log_processing.py
```diff
@@ -750,6 +738,39 @@ def readline(self, current_time=None):
                     result.timestamp = timestamp
                     if attrs:
                         result.attrs = attrs
+                if self.__merge_json_parsed_lines:
+                    while result.line and not result.line.endswith(b"\n"):
```
Kami, line 742: (just thinking out loud) This should also work correctly if lines / messages contain carriage return since that's always presents between line feed character, right?

Kami:
Looking at the docs, code and other implementations for this 16K line length limitation  (https://github.com/elastic/beats/pull/6967/files#diff-e2f65dc2a99141c783ee47e52128190ee50abd157da0a37f34450904e4e2c971R99, https://stackoverflow.com/questions/61360759/logging-with-docker-and-kubernetes-logs-more-than-16k-split-up), it appears Docker will also include ``partial: true`` attribute in the log line object if the line is partial.

Would it make sense to be more proactive and user friendly and log a warning message in case we encounter entry where ``partial is True`` and ``merge_json_parsed_lines`` config option value is ``False``? Aka something along the lines of "Detected partial line in log file foo, you may want to set config option foo to join partial lines..."

And it seems at least filebeat enables this behavior by default. Something we may perhaps also want to consider in the future once this functionality goes through more battle testing.

scalyr_agent/log_processing.py
yanscalyr: I couldn't figure out any way to get stuck here myself, I thought there might be something with the interaction between this timer and the one used when reading from the file but since this timer never resets unless a line is emitted and we move on I don't think there is any real risk.

scalyr_agent/log_processing.py
yanscalyr: As far as I'm aware `\r` will be before `\n` if it is present at all yes.

yanscalyr:
From local testing I don't see `partial` as an attribute in the JSON we parse, it might just be something beats puts in for testing. In any case if it isn't present in my environment that means we shouldn't rely on it in general. I am a bit concerned about logging a warning for any instance of a line that doesn't end in a newline with json parsing because I don't fully understand if a line like that could be something a user actually expects.
Do we use our JSON parser in any case besides reading docker logs?

scalyr_agent/log_processing.py
```diff
@@ -750,6 +738,49 @@ def readline(self, current_time=None):
...
+                                    scalyr_logging.DEBUG_LEVEL_3,
+                                    "Reached max wait time for incomplete merged line in file %s, emitting line as-is.",
+                                    self.__path,
+                                )
+                                break
+                        next_attrs = scalyr_util.json_decode(
+                            next_line.decode("utf-8", "replace")
+                        )
+                        line = next_attrs.pop(self.__json_log_key, None)
+                        if line is None:
```
czerwingithub, line 780: Why would `line` be `None` here?  Can you add in documentation capturing what this case represents.

scalyr_agent/configuration.py
```diff
@@ -1057,6 +1057,13 @@ def include_raw_timestamp_field(self):
         message as recorded by those systems."""
         return self.__get_config().get_bool("include_raw_timestamp_field")
 
+    @property
+    def merge_json_parsed_lines(self):
+        """If True, and events are getting parsed with the parse_as_json feature, events with messages that do not
+        end in a newline character will be joined together until a newline character is found. This is to work around
```
czerwingithub, line 1063: You should also probably mention that the timestamp for the first line (among all of those joined) will be used for the joined line's timstamp.

scalyr_agent/log_processing.py
yanscalyr: Actually I was missing a warning log for this case that appears when handling this elsewhere in the code, I think it's message describes this well enough (we didn't find the expected message field in the parsed JSON).

As well, when not merging lines and we run into a json line with no message we output the unparsed line. This isn't really something we can do mid-merge unless we tack it on to the end of the line so far, which is how I'll handle this for now so we don't lose data.

czerwingithub (approved):

PR #776, opened 2021-09-28
New process level agent benchmark with 500k log line size

Kami:
This PR adds and hooks up new benchmark to CodeSpeed.

This benchmark measures agent resource utilization for an existing which ingests existing 50 MB log file with 500k log lines (generated with scalyr-genlog) log file on disk.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1852,7 +1852,7 @@ jobs:
       - codespeed_agent_process_benchmark:
           codespeed_executable: "Python 2.7.17 - loaded conf 5"
           codespeed_environment: "Circle CI Docker Executor Medium Size"
-          agent_config: "benchmarks/configs/agent_single_50mb_access_log_file.json"
+          agent_config: "benchmarks/configs/agent_single_50mb_genlog_500k_line_file.json"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1807,6 +1806,27 @@ jobs:
           capture_line_counts: true
           cache_key_name: "benchmarks-loaded-agent-growing-log-file-py-35"
 
+  benchmarks-loaded-agent-single-growing-log-file-180mb-py-27:
```
Kami, line 1809: I just moved this entry so they are arranged in order by the executable name.

benchmarks/configs/agent_single_50mb_genlog_500k_line_file.json
```diff
@@ -0,0 +1,18 @@
+{
+  // We want to consume whole file from the beginning so we set max offset to 100 MB
+  max_log_offset_size: 104857600,
+
+  // 600 KB
+  max_line_size: 614400,
```
Kami, line 6: I can also change this to something a bit closer to 500k (e.g. 510k) to see if the memory usage drops a bit (it probably should).

ArthurKamalov (approved):
LGTM, slack notification will be re-enabled later, right?

Kami:
Yep, I just temporarily disabled it to avoid spam to the channel while working on this branch.

PR #777, opened 2021-10-01
Update version and changelog to version 2.1.24

yanscalyr:
Updating the version to be able to make a preview release, and of course for the eventual real release.


CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
...
 =================================
 
+## 2.1.24 "Xuntian" - Oct 1, 2021
+
+<!---
+Packaged by Yan Shnayder <yans@sentinelone.com> on Oct 1, 2021 14:10 -0800
+--->
+
+Improvements:
+* Implemented optional line merging when running in Docker based Kubernetes to work around Docker's own 16KB line length limit. Use the ``merge_json_parsed_lines`` config option to enable this functionality.
```
czerwingithub, line 11: Can you also list the environment variable you need to set?  I believe most config options will have an env var by default.  It is better for folks on K8s / Docker.

czerwingithub (changes requested):

CHANGELOG.md
czerwingithub: Or, is this a per-logfile config option?

yanscalyr: It is not.

czerwingithub (approved):

PR #780, opened 2021-11-04
CT-463: Add the LetsEncrypt root cert to our included bundle

yanscalyr:
Connecting to dvus requires trusting LetsEncrypt, this can be worked around by pointing the agent at system certs but it would be preferable for this to work out of the box. This PR adds the LetsEncrypt cert to our packaged cert bundle.


czerwingithub (approved):

Kami:
I wanted to say it would be a good idea to add cert expiration check to the script we use for that, but it looks like it already checks that for all the certs in that directory and we are fine (and letsencrypt root cert expired just recently so it's unlikely new one will expire any time soon).

PR #781, opened 2021-11-08
Update Kubernetes monitor to log INFO message when containerd runtime environment is detected

Kami:
This pull request updates Kubernetes monitor to print a message which warns used that some metrics won't be reported / available when container level metrics reporting is enabled and containerd runtime is detected.

This message is logged under INFO and not WARN log level since it's an informational message and there isn't anything we or user can do at the moment since this data is not available through Kubelet metrics API when using containerd runtime.


yanscalyr (approved):
LGTM assuming the tests ever decide to run

PR #782, opened 2021-11-09
Fix graphite monitor

Kami:
This pull request fixes a bug in Graphite monitor where we didn't correctly convert incoming request data (metric name) from bytes to unicode type which is expected by the ``emit_value()`` method.

In addition to that, I added some test cases since we had none. There is a simple test case where we just exercise the parsing code path and there is a bit more involved one which starts actual TCP servers so we also exercise the TCP server request processing code.

Tests are still not exhaustive and don't cover all the possible scenarios and edge cases, but it's a start.


yanscalyr (approved):
LGTM

PR #784, opened 2021-11-12
New docker builds draft

ArthurKamalov:
This draft PR shows how we can produce Docker-based releases. The main idea is simple - the build has to be done from the static Dockerfile.  After that, the build and publishing process has to become much easier.  We also can drop building images on Jenkins and just start pushing docker images from GitHub actions, for example, on the new release tag, (there's also [docker automated builds](https://docs.docker.com/docker-hub/builds/) but it may require a subscription.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM python:2.7-slim as build
+
+MAINTAINER Scalyr Inc <support@scalyr.com>
+
+RUN apt-get update && apt-get install -y build-essential
+
+# install python dependencies
+ADD ./docker/requirements.txt /tmp/requirements.txt
+RUN pip --no-cache-dir install --root /tmp/dependencies -r /tmp/requirements.txt
+
+
+ADD . /scalyr-agent-2
+WORKDIR /tmp/build
+
+# build container tarball.
+RUN python /scalyr-agent-2/build_package.py k8s_builder --only-container-tarball
+
+RUN mkdir /tmp/build/source
+RUN tar -xf scalyr-k8s-agent.tar.gz -C /tmp/build/source
+
+FROM python:2.7-slim
+
+COPY --from=build /tmp/build/source /
+COPY --from=build  /tmp/dependencies/ /
+
+CMD ["/usr/sbin/scalyr-agent-2", "--no-fork", "--no-change-user", "start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ WORKDIR /tmp/build
 RUN python /scalyr-agent-2/build_package.py docker_json_builder --only-container-tarball
 
 RUN mkdir /tmp/build/source
-RUN tar -xf scalyr-k8s-agent.tar.gz -C /tmp/build/source
+RUN tar -xf scalyr-docker-agent.tar.gz -C /tmp/build/source
 
 FROM python:2.7-slim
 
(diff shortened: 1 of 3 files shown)


Kami:
So far, LGTM, but I have a couple of questions comments.

1. What does "static" Docker file mean in this context?
2. We will still need to build and publish multiple Docker images as we do today (k8s, docker, docker-syslog, docker-json - which really just means using a different set of configs inside the image)
3. It looks like that build using QEMU doesn't take long which is good. I have bad experience with ARM builds for projects which relied on grpc and other libraries without ARM wheels where build could easily take 30+ minutes...
4. Do you think we could already use new GHA workflow to build and publish images even before we fully switch to frozen binaries and get rid of Jenkins? I think it should be possible, we just need to set up rules for the release + master branch and add some if checks to the build workflow where we check the version and check if image with such tag already exists.
5. I think we should enable building and pushing all the images on every push to master. We can use image tag along the lines of ``dev-<date string>-<commit hash>`` or similar. This way we can get rid of manual "build preview release step" and make it easier to faster to share dev builds with users, when needed.

ArthurKamalov:
@Kami 
1. By static I meant that the Dockerfile itself has to be enough to produce the image. For now, we have to rely on the build script and the Dockerfiles are not self-sufficient and we can not use them just by doing `docker build -f <needed_dockerfile>`. It's may not be mandatory, but I think that is how Dockerfiles are meant to be used :)
2.  We still can use scripting or, as I've shown, even use the existing `build_packages.py`  script. That scripts can appropriately handle different configurations. The main idea is to do that as part of image build. We also can use multi-stage build + build-args to configure the build just by using docker. 
3. Yeah, it should be fine because we don't have to compile anything (maybe just some python library extensions, but it seems that it doesn't take long)
4. I think we can. Yeah, some conditional step that matches needed branches or tags and then pushes the build should do the trick. I have already. mentioned Docker automated builds. It can create a hook on needed branches or tags and build and publish images completely on the docker hub side. But, as I said, it may require a subscription. They say that they provide it for free for open-source projects, but it still may not be trivial.

5. That's a good idea.

Kami:
4. I'm fine with either way (either using Docker hub auto build feature or building the image + pushing  it to Dockerhub ourselves), but building it ourselves may give us more flexibility (aka we have more flexibility to determine if and when to build the image) + we can potentially also push to other registries and similar.

build_package.py
```diff
@@ -1855,6 +1862,15 @@ class BadChangeLogFormat(Exception):
         help="Enable coverage analysis. Can be used in smoketests. Only works with docker/k8s.",
     )
 
+    parser.add_option(
+        "",
+        "--only-container-tarball",
+        dest="only_container_tarball",
+        action="store_true",
+        default=False,
+        help="Only works with docker/k8s.",
```
Kami, line 1871: Please clarify this help string since it's not really clear from the current help string what's the purpose of this option and what purpose it serves.

Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM python:2.7-slim as build
```
Kami, line 1: While at it, we should also switch to Python 3 for the Docker image.

ArthurKamalov:
@Kami I tried to address all comments in the new PR. https://github.com/scalyr/scalyr-agent-2/pull/794.

PR #785, opened 2021-11-12
Linux system metrics monitor improvements (netstat)

Kami:
This pull request includes the following changes to the netstat part of the code of Linux system metrics monitor.

1. Sync tcollector code base with latest upstream version. It's worth noting that I didn't pull in upstream code which also parses metrics from ```/proc/net/snmp`` file. That would potentially include additional metrics and incur additional end user costs due to the additional metrics. If we wanted to also include metrics from that file (when available), we should likely make that an opt-in config option / feature flag.
2. Refactor the netstat collector code so it's more easier to test it. We already have local modifications to the tcollectors code base and I made similar testing related changes to dfstat tcollector code in the past.
3. Add very basic test case for it (we can expand on it in the future with fixture files for other popular kernel versions)
4. Update the code so the monitor doesn't print and error and we ignore ``MPTcpExt:`` line we don't support parsing metrics for.

The issue which 4) addresses was reported by John Tan. Initially I wasn't able to replicate it since it only affects some kernel versions.


ArthurKamalov (approved):
LGTM

PR #788, opened 2021-11-15
Move all command line parsing to agent main

ArthurKamalov:
Move every command line handling to the agent main in order to have a single executable.  It is crucial for us since each frozen binary executable is pretty heavy and affects package size. All config and windows service command-line parsing is moved to the agent_main.py as separate sub-parsers.


.circleci/config.yml
```diff
@@ -1823,26 +1843,37 @@ jobs:
           capture_line_counts: true
           cache_key_name: "agent_2-growing_logs-monitors-multiprocess-2-worker-py-35"
 
-  benchmarks-loaded-agent-single-growing-log-file-180mb-py-27:
+  benchmarks-loaded-agent-single-50mb-log-file-with-500k-lines-py-27:
```
Kami, line 1846: Just curious why this benchmark setup was changed?

Kami: EDIT: Looks like this may be due to merge conflict resolution or similar, right?

Kami:
Diff seems to be a bit messed up, any change you could sync up ``agent-frozen-binaries`` with latest master / similar to make review easier?

.circleci/config.yml
ArthurKamalov: Yeah, sorry. I haven't sync the target branch with mater.  Please see this PR https://github.com/scalyr/scalyr-agent-2/pull/788

PR #789, opened 2021-11-15
Move all command line handling to the agent main.

ArthurKamalov:
Move every command line handling to the agent main in order to have a single executable. It is crucial for us since each frozen binary executable is pretty heavy and affects package size. All config and windows service command-line parsing is moved to the agent_main.py as separate sub-parsers.

### DIFF
--- a/scalyr_agent/config_main.py
+++ b/scalyr_agent/config_main.py
@@ -38,6 +38,7 @@ import tempfile
 import traceback
 import errno
 import itertools
+import argparse
 from io import open
 
 from optparse import OptionParser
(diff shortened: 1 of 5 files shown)


scalyr_agent/config_main.py
```diff
@@ -1590,20 +1570,19 @@ def configure_agent_service_for_systemd_management():
...
             help="Configure the agent to be managed by systemd instead of init.d which is a default",
         )
 
-    (options, args) = parser.parse_args()
-    if len(args) > 1:
-        print("Could not parse commandline arguments.", file=sys.stderr)
-        parser.print_help(sys.stderr)
-        sys.exit(1)
+    options = parser.parse_args(args=argv)
+    # if len(args) > 1:
```
Kami, line 1582: Is this a testing change, or?

Kami:
It looks like that with some of that refactoring it should now also be a bit easier to write some unit tests for that CLI arguments parsing code. Is that indeed the case?

If that is indeed the case it would be good to add some very basic unit tests for that argument parsing code.

scalyr_agent/config_main.py
ArthurKamalov: Oops, sorry. It is a leftover from the previous command line handling logic, where there was `optparser` instead of `argparse`

Kami:
As discussed in the past - is there any way for us to keep this change backward compatible from the end user perspective?

Aka so users can still use the config binary, but it routes the request to the main agent binary? 

Perhaps via bash script on Linux and batch script on Windows (which would just be very small and light wrappers which call into the main agent binary via ``config`` + request of the user provided CLI arguments).

PR #790, opened 2021-11-17
Include pympler dependency for memory profiling as part of the Docker Image

Kami:
This pull request updates Docker requirements file so we also include ``pympler`` dependency which is used for memory profiling.

Profiling functionality is opt-in via scalyr agent config option. We already include dependency which is used for CPU profiler in the Docker Image, but we don't include one which is used for memory profiling.

Including it makes it easier to troubleshoot (profile) for the end user since they can just turn on the corresponding config option and there is no need to re-build and modify Docker images to get that dependency installed.

## TODO

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 Scalyr Agent 2 Changes By Release
 =================================
 
+## 2.1.26 "TBW" - Dec 17, 2021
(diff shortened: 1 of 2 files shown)


yanscalyr (approved):
LGTM

PR #791, opened 2021-11-18
Add response message to the Scalyr client connection error log.

ArthurKamalov:
A small addition to the bad request error log message, so it also contains a response message from the server.


scalyr_agent/scalyr_client.py
```diff
@@ -718,8 +719,9 @@ def __receive_response(self, body_str, send_time):
                 elif status == "error/client/badParam":
                     log.error(
                         "Request to '%s' failed due to a bad parameter value.  This may be caused by an "
-                        "invalid write logs api key in the configuration",
+                        "invalid write logs api key in the configuration. Response message: %s",
                         self.__full_address,
+                        json.loads(response).get("message"),
```
Kami, line 724: It seems like ``response_as_json`` variable should already be available here. We should probably use that (this way we can avoid parsing response twice and the need to wrap this block in try / catch in case parsing fails).

 Adding a test case would also be good.

Kami:
I added a test case for it. I will go ahead and merge it into master once the build passes.

I'm still battling with the build failure which seems to be related to a transitive dependency being updated and not supporting Python 2 anymore.

Kami:
snyk failure is a false positive in this case.

I just pinned an existing transitive test dependency on which we already dependent in the past for tests (it just wasn't explicitly pinned, but we need to pin it now to avoid issues with new version which doesn't support Python 2.7.).

PR #736, opened 2021-03-18
[WIP] Add initial basic JMX expoter metrics scraper monitor

Kami:
This pull request adds a monitor which can scrape metrics exposed by jmx exporter java agent via HTTP interface (https://github.com/prometheus/jmx_exporter).

Technically, JMX exporter uses "standard" Prometheus ``/metrics/`` endpoint, but this monitor hasn't been extensively tested with other ``/metrics`` endpoints and app so it likely contains many edge cases which are not handled correctly.

PR #792, opened 2021-11-21
[WIP] Add OpenMetrics / Prometheus metrics API scraper monitor

Kami:
This pull request adds a new monitor which supports scraping metrics from OpenMetrics / Prometheus metrics API endpoints.

It's loosely based on top #736 which was just handling JMX exporter metrics which are returned in OpenMetrics compatible format. 

It makes more sense and it's more useful to have a generic OpenMetrics monitor. This will also allow us to dog food Scalyr metrics for various services and deployments which expose metrics in Prometheus format.

Since OpenMetrics format supports more pre-aggregated formats such as histograms and summaries which are not directly supported by Scalyr metrics API we ignore metrics with those types.

## Note on Metrics Filtering

A lot of Prometheus metric exporters (think JMX exporter, etc.) expose many metrics by default (in some cases, many thousands).

In most cases, user will not be interested in all those metrics so we support filtering and only ingesting metrics user is interested in.

This is achieved using two whitelist / allowlist type of filters - with those two filters user can specify which metrics to include based on the metric name glob or based on the metric component value glob (in a lot of cases, many different metrics will have the same name, but a different component values).

Of course, where possible, users are encouraged to filters those metrics at the source (e.g. JMX exporter configuration) since filtering metrics inside the monitor includes some overhead.

## TODO

- [x] More tests (use actual response fixtures)
- [ ] Changelog entry


Kami:
Closing in favor of #806 which has been merged.

PR #793, opened 2021-11-30
Frozen binaries: initial refactor scalyr of scalyr_agent package

ArthurKamalov:
This is the first "frozen binary" PR that contains initial refactoring for the frozen binaries.  
The main changes are in the `__scalyr__.py` module. The main changes:
* It contains a new logic on how to determine the installation type. In order to determine the package type correctly, it relies on a special file that has to be placed in the "scalyr_agent" package directory.
* It does not import third party libraries anymore, except `tcollectors`

Changes in the majority of other files are mostly the result of the changes in `__scalyr__.py`.

There are also changes in unit tests since they are also affected by main changes. 

Unit tests added to the Github Actions. (Most of the CircleCI tests are commented for now, but they will be also moved to the Github action in further PR's)

### DIFF
--- a/.github/workflows/agent.yml
+++ b/.github/workflows/agent.yml
@@ -1,0 +1,32 @@
+name: Build agent packages.
+
+on: [push]
+
+env:
+  DOCKER_BUILDKIT: 1
+
+
+jobs:
+  unit-tests:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ "ubuntu-20.04", "windows-2019", "macos-11" ]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install python.
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8.10
+
+      - name: Install python dependencies
+        run: |
+          python3 -m pip install -r dev-requirements.txt
+
+      - name: Run unit tests.
+        run: |
+          python3 -m pytest tests/unit
--- a/benchmarks/micro/requirements-compression-algorithms.txt
+++ b/benchmarks/micro/requirements-compression-algorithms.txt
@@ -1,3 +1,1 @@
-zstandard
 brotli
-lz4
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -19,6 +19,7 @@ six==1.14.0
 ujson==1.35
 # Needed by MockHTTPServer class and related tests
 flask==1.1.1
+distro==1.6.0; platform_system != "Windows"
 
 # Used for performance optimized versions of rfc3339_to_* functions
 # NOTE: Not supported on windows
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -45,7 +45,7 @@ requests==2.26.0
 pg8000==1.10.6
 pymysql==0.9.3
 redis==2.10.5
-pysnmp=4.3.0
+pysnmp==4.3.0
 
 psutil==5.7.0; platform_system == "Windows"
 pywin32==300; platform_system == "Windows"
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -32,6 +32,10 @@ udatetime==0.0.16; platform_system != "Windows"
 git+https://github.com/apache/libcloud.git@trunk#egg=apache-libcloud; platform_system != "Windows"
 paramiko==2.7.2; platform_system != "Windows"
 
+# benchmarks dependencies
+numpy==1.16.6
+brotli
+
 
 #agent dependencies
 orjson==3.6.4
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,11 +2,11 @@
 # These dependencies are not needed to run the Scalyr Agent, but mainly used for
 # running test suites, and mocking Python Objects for the Scalyr Agent.
 
+
 # Testing tools and libraries
 mock==3.0.5
 psutil==5.7.0
-pytest==4.6.9; python_version < '3.0'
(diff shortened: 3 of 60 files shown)


benchmarks/micro/requirements-compression-algorithms.txt
```diff
@@ -1,3 +1 @@
-zstandard
```
Kami, line 1: Any reason this has been moved into dev-requirements?

.github/workflows/agent.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/checkout@v2
+
+      - name: Install python.
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8.10
+
+      - name: Install python dependencies
+        run: |
+          python3 -m pip install -r dev-requirements.txt
```
Kami, line 28: We probably also want to cache the dependencies to speed up the build, right?

scalyr_agent/build_info.py
```diff
@@ -25,49 +25,30 @@
...
-BUILD_INFO_PATH_LINUX = os.path.abspath(os.path.join(BASE_DIR, "../../build_info"))
-
-# On Windows, it's stored under bin/ directory and there we also use binary so we can't work with
-# relative paths
-BUILD_INFO_PATH_WINDOWS = "C:\\Program Files (x86)\\Scalyr\\bin\\build_info"
-
 GIT_GET_HEAD_REVISION_CMD = "git rev-parse HEAD"
 
+# The build_info file should be located in the install root directory.
+BUILD_INFO_FILE_PATH = str(pl.Path(__scalyr__.get_install_root(), "build_info"))
```
Kami, line 38: :+1:

scalyr_agent/platform_linux.py
```diff
@@ -60,38 +50,7 @@ def can_handle_current_platform(self):
         @return:  True if this platform instance can handle the current server.
         @rtype: bool
         """
-        return _platform.lower().startswith("linux")
-
-    @property
-    def default_paths(self):
```
Kami, line 66: Why is this method not needed anymore? Do we now simply rely on the parent (posix) class implementation?

scalyr_agent/util.py
```diff
@@ -2196,9 +2210,10 @@ def get_agent_start_up_message():
         used_locale,
     ) = get_language_code_coding_and_locale()
 
-    msg = "Starting scalyr agent... (version=%s) (revision=%s) %s (Python version: %s) " "(OpenSSL version: %s) (default fs encoding: %s) (locale: %s) (LANG env variable: %s)" % (
-        SCALYR_VERSION,
+    msg = "Starting scalyr agent... (version=%s) (revision=%s) (arch=%s) %s (Python version: %s)" "(OpenSSL version: %s) (default fs encoding: %s) (locale: %s) (LANG env variable: %s)" % (
+        __scalyr__.SCALYR_VERSION,
         build_revision,
+        platform.processor(),
```
Kami, line 2216: :+1:

dev-requirements.txt
```diff
@@ -2,37 +2,57 @@
...
+numpy==1.16.6
+brotli
+
+
+#agent dependencies
+orjson==3.6.4
+# TODO: remove six from the code base.
+six==1.14.0
+requests==2.26.0
+pg8000==1.10.6
```
Kami, line 45: Does this mean we can get rid of third party directory now since those are installed as regular pip dependencies?

scalyr_agent/platform_linux.py
ArthurKamalov: Right, they are identical.

dev-requirements.txt
ArthurKamalov: Correct, I just don't touch anything that can be removed to reduce the amount of noise in the PR.  Everything will be removed at the last step.

.github/workflows/agent.yml
ArthurKamalov: Right, There is already logic for that. I just didn't put it right now to reduce the number of new files, but it has already been added in the parallel PR with new docker images https://github.com/scalyr/scalyr-agent-2/pull/794.
In both PRs they have not been used fully yet, but in the next PR, they are used exactly for the reason that you've noticed here.

benchmarks/micro/requirements-compression-algorithms.txt
Kami: I would prefer to keep multiple requirements files unless there is a specific reason to only have one.

ArthurKamalov: Yeah, Sorry those changes didn't get there, they will be separated.

scalyr_agent/__scalyr__.py
```diff
@@ -19,239 +19,168 @@
...
 
-    file_names = ["VERSION"]
+    source_root = pl.Path(__file__).absolute().parent.parent
 
-    if __is_frozen__:
-        # also check for VERSION.txt file because there is a reserved filename - "VERSION" in Pyinstaller,
-        # and it expects that this file is a DLL.
-        file_names.append("VERSION.txt")
+    #
+    package_info_path = source_root / "scalyr_agent" / "package_info.json"
```
Kami, line 126: It almost seems like we should also store build information in this file so we don't have two files (existing build_info + new package_info.json).

benchmarks/micro/requirements-compression-algorithms.txt
ArthurKamalov: Added multiple requirements files. Now there are only test and benchmark-related requirements, but there will be new ones in further PRs.

PR #794, opened 2021-11-30
Add new docker image build

ArthurKamalov:
This PR adds new build logic for the Agent docker images.  This build logic is only introduced partially here and will be added in the future with other "frozen binary builders."  For now, there are only builders that create a filesystem for the docker images. Those builders are executed inside a single dockerfile, so there's no need to keep separate docker files for each image build.  NOTE: With the builder logic I had to also migrate another helper library - `environment_deployments`. It is not used for now, but it is a dependency for the package builders and it will be heavily used for other frozen binary builds.

Also was added new GitHub actions workflow that builds multi-platform images and pushes them to the registry.


.github/workflows/docker-and-ks8.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Build and push image from master branch.
+        uses: docker/build-push-action@v2
+        if: ${{ github.ref == 'refs/heads/master' }}
+        with:
+          context: .
+          file: docker/Dockerfile.unified
+          push: true
+          platforms: linux/amd64,linux/arm64
+          tags: |
+            ${{ matrix.image-name }}:${{ github.sha }}
```
Kami, line 70: :+1:

agent_build/package_builders.py
```diff
@@ -0,0 +1,802 @@
...
+
+    # Add agent source code as a bundled frozen binary if True, or
+    # add the source code as it is.
+    USE_FROZEN_BINARIES: bool = True
+
+    # Specify the name of the frozen binary, if it is used.
+    FROZEN_BINARY_FILE_NAME = "scalyr-agent-2"
+
+    # The type of the installation. For more info, see the 'InstallType' in the scalyr_agent/__scalyr__.py
+    INSTALL_TYPE: str
```
Kami, line 144: It actually seems that ``*_INSTALL`` constants in that module are ints and not strs?

In any case, once we drop support for Python < 3.6 we should probably just update the code utilize enum type for that.

agent_build/package_builders.py
```diff
@@ -0,0 +1,802 @@
...
+                str(dist_path),
+                "--workpath",
+                str(pyinstaller_output / "build"),
+                "-n",
+                type(self).FROZEN_BINARY_FILE_NAME,
+                "--paths",
+                ":".join(paths_to_include),
+                *add_data_options,
+                *hidden_import_options,
+                "--exclude-module",
```
Kami, line 497: Any reason why we exclude those modules? Would be good to add a comment and clarify why we need to exclude those.

agent_build/tools/build_in_docker/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# This frozen binary later can be used to run package tests on systems without python.
+FROM $BASE_IMAGE_NAME as stage-test
+RUN rm -rf /scalyr-agent-2
+ADD ./agent_tools /scalyr-agent-2/agent_tools
+ADD ./agent_build /scalyr-agent-2/agent_build
+ADD ./tests/package_tests/ /scalyr-agent-2/tests/package_tests/
+
+
+# Pick only specified stage.
+FROM stage-$BUILD_STAGE
```
Kami, line 39: So it looks we we simply base the final image on the build stage?

Would there be a benefit if we used a new stage for the final artifact and only copy subset of files from the build stage into this final stage (or we need everything from the build stage)?

Kami: EDIT: Never min - this docker file is only used for the build process itself and it's not something we distribute to the end users.

agent_build/tools/constants.py
```diff
@@ -0,0 +1,40 @@
+import enum
```
Kami, line 1: Please add ASF license header.

agent_build/package_builders.py
ArthurKamalov: Correct, that's all about the size, I'll add the commant.

agent_build/package_builders.py
ArthurKamalov: Yeah, I just don't want to pull the scalyr_agent package because it will depend on a bunch of third-party libraries and I just want to keep build libraries as independent as possible, just in case.

.github/workflows/docker-and-ks8.yml
```diff
@@ -0,0 +1,89 @@
...
+        with:
+          image: tonistiigi/binfmt:latest
+          platforms: all
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          driver-opts: network=host
+
+      - name: Build and push image to the local registry
```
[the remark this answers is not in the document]
ArthurKamalov: Testing added.

agent_build/package_builders.py
```diff
@@ -0,0 +1,802 @@
...
+        package_info_file = self._intermediate_results_path / "package_info.json"
+
+        package_info = {"install_type": type(self).INSTALL_TYPE}
+        package_info_file.write_text(json.dumps(package_info))
+
+        # Add this package_info file in the 'scalyr_agent' package directory, near the __scalyr__.py file.
+        add_data = {str(package_info_file): "scalyr_agent"}
+
+        # Add monitor modules as hidden imports, since they are not directly imported in the agent's code.
+        hidden_imports = [
```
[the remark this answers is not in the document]
ArthurKamalov: Changed. Now it adds all monitors from the 'monitors' directory and then excludes monitors that are specified to be excluded for each package type.

.github/workflows/docker-and-ks8.yml
```diff
@@ -0,0 +1,128 @@
...
+            --tag ${{ github.sha }}
+
+      - name: Push image using master branch commit hash.
+        if: ${{ github.ref_type == 'tag' }}
+        run: |
+
+          if [[ "${{ github.ref_name }}" == "v$(cat VERSION)" ]]; then
+            # if the this is a "release" tag with format "v*.*.*", then use version as tag
+            # also add the 'latest' tag.
+            tags="--tag $(cat VERSION) --tag latest"
```
Kami, line 115: Could adding latest tag in this scenario cause issues under some situations? I guess it could potentially cause issues in case a build for an old tag is re-triggered for some reason - in that case it would potentially overwrite ``latest`` tag with some older ref.

Is this something we should worry about and need some guards for it?

.github/workflows/docker-and-ks8.yml
```diff
@@ -0,0 +1,128 @@
...
+          else
+            # Just use GitHub tag as image tag.
+            tags="--tag ${{ github.ref_name }}"
+          fi
+
+          python3 build_package_new.py ${{ matrix.image-type }} \
```
Kami, line 121: At what point do we plan to switch to Dockerhub?

I think once the tests are running + passing it should likely be fine to switch to Docker hub for the master dev builds.

.github/workflows/docker-and-ks8.yml
```diff
@@ -0,0 +1,128 @@
...
+      - name: Perform the deployment of the test environment.
+        uses: ./.github/actions/perform-deployment
+        with:
+          deployment-name: "test_environment"
+
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+        with:
+          image: tonistiigi/binfmt:latest
```
Kami, line 35: Do we want to pin this image to a specific tag to avoid surprises and ensure better reproducibility?

ArthurKamalov:
@Kami I removed all image publishing except pushing from the master with the current commit SHA.  The release process itself I am planning to organize differently and I working on a. draft PR.

.github/workflows/docker-and-ks8.yml
ArthurKamalov: Yeah, there may be many various edge cases. I removed the publishing of "production" tags.

.circleci/smoketest_k8s.sh
```diff
@@ -161,7 +165,7 @@ echo ""
 echo "=================================================="
 echo "Starting verifier"
 echo "=================================================="
-kubectl run -it --restart=Never ${contname_verifier} --image=${smoketest_image} -- \
+kubectl run -i --restart=Never ${contname_verifier} --image=${smoketest_image} -- \
```
Kami, line 168: Any special reason why ``-t`` option was removed and we don't want / need tty anymore?

.dockerignore
```diff
@@ -0,0 +1,18 @@
+debug
```
Kami, line 1: Yeah, good call on ``.dockerignore``. 

We do want to exclude as much as possible to speed up the context transfer and as such, the build itself.

To be on the safe side we should also exclude ``.tox`` which can be very large as well (although not sure if it will be there when we run the docker image build, but still).

.github/actions/perform-deployment/action.yml
```diff
@@ -0,0 +1,17 @@
+name: 'Perform deployment.'
```
Kami, line 1: You meant want to explain (with examples) what a deployment really is / what it represents (in readme or here).

People who won't check the code itself may be confused on what deployment really is in this case.

.github/actions/perform-deployment/action.yml
```diff
@@ -0,0 +1,17 @@
...
+description: |
+  Prepares current machine by perfroming the set of the deployments, which are defined in the 'agent_build/environment_deployers.deployers.py.' folder.
+  For more info, please look at the docs in the same file.'
+inputs:
+  deployment-name:
+    description: "Name of the deployer"
```
Kami, line 7: deployer -> deployment

.github/actions/perform-deployment/index.js
```diff
@@ -0,0 +1,123 @@
+// Copyright 2014-2021 Scalyr Inc.
```
Kami, line 1: Some tests for this action would also be nice, but I know this may require quite a bit of mocking.

.github/actions/perform-deployment/index.js
```diff
@@ -0,0 +1,123 @@
...
+        [deployment_helper_script_path, "deployment", deploymentName, "deploy", "--cache-dir", cacheDir],
+        {stdio: 'inherit'}
+    );
+
+    if ( fs.existsSync(cacheDir)) {
+      console.log("Cache directory is found.")
+
+      // Run through the cache folder and save any cached directory within, that is not yet cached.
+      const filenames = fs.readdirSync(cacheDir);
+      for (const name of filenames) {
```
Kami, line 76: I wonder if we could refactor this code path a bit to reduce levels of indentation and make it a bit easier to read.

Perhaps return early in case cacheDir doesn't exist ?

.github/actions/perform-deployment/index.js
```diff
@@ -0,0 +1,123 @@
...
+    );
+
+    // Read and decode names from json.
+    const json_encoded_deployment_names = buffer.Buffer.from(code, 'utf8').toString()
+    const deployer_cache_names = JSON.parse(json_encoded_deployment_names)
+
+    const cache_hits = {}
+
+    // Run through deployment names and look if the is any existing cache for them.
+    for (let name of deployer_cache_names) {
```
Kami, line 46: Same here, may be easier to follow / test it if this block is in a separate utility function (e.g. ``getCacheInfoFromFile`` or similar).

.github/actions/perform-deployment/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "name": "perform-deployment",
+  "version": "1.0.0",
+  "description": "This action uses python module `agent_build/tools/environment_deployments.py` to perform some deployment which is defined in it. The main purpose of the action is to cache results of this deployment to the GitHub Actions cache. Since the deployments consist of  steps, and different deployments can use the same step, then it's more reasonable to cache each step separately. That means that the regular [@actions/cache](https://github.com/actions/cache)  is not enough, and we need to use more flexible JS scripting to call GitHub Actions caching dynamically.",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
Kami, line 11: All the code is written by us, right? If so, we should use Apache 2.0 license.

.github/workflows/agent-build.yml
```diff
@@ -0,0 +1,39 @@
+# The main "setup" workflow, that calls other workflows.
+name: Scalyr Agent Build.
+
+on: [push]
+
+# Agent image tests.
+jobs:
+  docker-json:
+    uses: scalyr/scalyr-agent-2/.github/workflows/docker-and-k8s.yml@add-new-docker-image-build
```
Kami, line 9: Cool, this re-usable workflows / workflows as actions are pretty new functionality, right?

.dockerignore
Kami: And probably also good idea to add ``node_modules`` to be on the safe side.

agent_build/tools/build_in_docker/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+            "BUILD_STAGE": stage_name,
+        },
+    )
+
+    # If there are output mapping specified, than we has to copy result files from the result build.
+    # To do that we have to create a container and copy files from it.
+    if output_path_mappings:
+        container_name = image_name
+        # Remove the container with the same name if exists.
+        subprocess.check_call(["docker", "rm", "-f", container_name])
```
Kami, line 115: To make debugging and troubleshooting a bit easier, do we also want to add some logging and print which commands are ran? Since, IIRC (please correct me if I'm wrong, ``check_call()`` doesn't print which command it runs).

Kami:
Some of the caching related code appears a bit complex (aka there is quite a lot of it), but if caching substantially speeds up the build it should indeed be worth the effort and the complexity.

Having said that, do we have any idea how much of storage space this cache will take? And can we control how long it's saved for? (aka purge cache after X days in LRU fashion). Since per-repo cache storage space is limited (IIRC, cache size was just increased recently to 10 GB, not sure though if paid plans have any higher limit though).

Since if we build new cache for every PR / push, I would imagine we could hit that limit pretty fast with Docker images.

Kami:
It appears that "Run image test" step takes quite a while (~7 minutes).

Will this get any faster when cache is populated or similar? Or there is no easy way to speed it up since we need to build image from scratch each time?

ArthurKamalov:
> Some of the caching related code appears a bit complex (aka there is quite a lot of it), but if caching substantially speeds up the build it should indeed be worth the effort and the complexity.
> 
> Having said that, do we have any idea how much of storage space this cache will take? And can we control how long it's saved for? (aka purge cache after X days in LRU fashion). Since per-repo cache storage space is limited (IIRC, cache size was just increased recently to 10 GB, not sure though if paid plans have any higher limit though).
> 
> Since if we build a new cache for every PR / push, I would imagine we could hit that limit pretty fast with Docker images.

There have to be just 2-3 docker images. That's 700-800MB max. New deployments logic allows us to reuse those images among multiple deployments, so that has to save cache space significantly.

ArthurKamalov:
> It appears that "Run image test" step takes quite a while (~7 minutes).
> 
> Will this get any faster when cache is populated or similar? Or there is no easy way to speed it up since we need to build image from scratch each time?

Yeah, since the image contains the code base of the agent it has to be rebuilt on each change. Such changes have to happen almost on each new commit. It's possible to cache some progress before the agent code is added, but the build process and Dockerfile will be less straightforward.

agent_build/tools/build_in_docker/__init__.py
ArthurKamalov: Added wrapper for the `subprocess.check_call` which logs its command.

.github/actions/perform-deployment/package.json
ArthurKamalov: All, except dependencies :)

.github/workflows/agent-build.yml
ArthurKamalov: Yes, it's pretty new and it also has one unresolved issue. For now, it can not properly run the re-usable workflow locally, from the same repo, so we have to specify its full path with `@<branch_name>`. The problem is that this `@<branch_name>` can not be dynamically evaluated with parameter expression and has to be hardcoded.
https://github.community/t/ref-head-in-reusable-workflows/203690. I added the verifier job that checks branch names and fails if they are don't match. Will hope that they fix that issue soon.

.github/actions/perform-deployment/action.yml
ArthurKamalov: Added the DEPLOYMENT.md readme file.

.circleci/smoketest_k8s.sh
ArthurKamalov: For some reason, it couldn't run the container on GitHub actions when I tried to run old smoke tests there. For now, I just left them on circleci without moving them here, because there are many things to change in order to make them work on GH actions.

.github/actions/perform-deployment/index.js
ArthurKamalov: Yeah, that js code is really hard to test.

agent_build/tools/environment_deployments/steps/cache_lib.sh
```diff
@@ -1,99 +0,0 @@
-#!/bin/bash
```
ArthurKamalov: @Kami This file just moved to `agent_build/tools/environment_deployments/steps/step_runner.sh` I am not really sure why GitHub treats it as deleted and new files instead of renamed.

.circleci/config.yml
```diff
@@ -2394,14 +2394,14 @@ workflows:
...
          <<: *master_and_release_only
-     - ami-tests-development:
-          context: scalyr-agent
-          requires:
-            - build-windows-package
-            - build-linux-packages-and-installer-script
-     - ami-tests-stable:
-          context: scalyr-agent
-          <<: *master_and_release_only
+#     - ami-tests-development:
```
Kami, line 2397: When will those be uncommented? Before merging the PR, or?

.circleci/modernize/modernize.py
```diff
@@ -430,6 +430,9 @@ def _run_tests():
         os.path.join(
             root, ".circleci/docker_unified_smoke_unit/smoketest/smoketest.py"
```
Kami, line 431: Once we drop support for Python 2.7, we can also remove modernize. I hope this will be soon, since I have tons of issues with modernize hanging stuck locally.

.github/workflows/agent-build.yml
```diff
@@ -0,0 +1,56 @@
...
+
+on: [push]
+
+# Agent image tests.
+jobs:
+  match-branches:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Check branches.
```
Kami, line 12: Will this step bill removed once this PR is merged into master?

.github/workflows/docker-and-k8s.yml
```diff
@@ -0,0 +1,110 @@
...
+        type: string
+        default: 3.8.10
+      publish:
+        description: "If true, then the result image will be published on master branch or any tag."
+        type: boolean
+        required: false
+        default: true
+
+    secrets:
+      SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN:
```
Kami, line 21: May not be a bad idea to add description for those two secrets. 

At some point I may also rename them to reduce the confusing and use consistent naming format since both of them refer to the same Scalyr environment and account.

.github/workflows/docker-and-k8s.yml
```diff
@@ -0,0 +1,110 @@
...
+        env:
+            SCALYR_API_KEY: ${{ secrets.SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN }}
+        run: |
+          python3 tests/package_tests/run_package_test.py run-package-test \
+            ${{ inputs.image-type}}_test \
+            --name-suffix "_${{ github.run_number }}${{ github.run_attempt }}"
+
+
+  publish:
+    needs: [test]
```
Kami, line 80: Is there any way for us to run this workflow only after Circle CI tests pass? Or that would be too hard to do?

agent_build/DEPLOYMENTS.md
```diff
@@ -0,0 +1,101 @@
...
+to reuse that outside of the GitHub actions.
+
+
+#### Deployments usage example
+Let's have a look at more complex example and how we can achieve the same results with `Deployments`.
+
+We need to install Python dependencies from two requirement files and we also need to download browser driver for 
+Selenuim.
+
+To achieve that we have a Deployment object in the ``agent_build/tools/environment_deployments/deplyments.py`` file.
```
Kami, line 36: Small typo in the filename

agent_build/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+
+    logging.info(f"RUN COMMAND: '{cmd_str}':\n")
+    return _ORIGINAL_POPEN_CALL(
+        *args,
+        **kwargs
+    )
+
+
+# Replace original subprocess with its wrapper that logs additional information about running command.
+subprocess.Popen = subprocess_popen_with_log
```
Kami, line 48: Not the most ideal thing, but I guess it's fine for now :D

Kami: That is, as long as that module is only imported inside those deployments modules.

It could cause issues with pytest and other tools which search and load all the Python files in a project.

Kami:
If not already (sorry if I messed it), can you please also add short information to the README somewhere on how to build those various Docker images locally (for testing and other purposes).

Kami (approved):
LGTM

.github/workflows/agent-build.yml
ArthurKamalov: It shouldn't be removed, but we'll have to change the branch name to master, so it will call a re-usable workflow file from an appropriate branch.

.github/workflows/docker-and-k8s.yml
ArthurKamalov: Not so easy, I'm afrad.

agent_build/__init__.py
ArthurKamalov: I guess that's was overkill, so I tried to remove unneeded stuff.

agent_build/tools/files_checksum_tracker.py
```diff
@@ -0,0 +1,120 @@
+# Copyright 2014-2021 Scalyr Inc.
```
ArthurKamalov, line 1: @Kami This new file mostly contains logic that has been moved from the `deployments.py` module.

.github/workflows/docker-and-k8s.yml
```diff
@@ -0,0 +1,132 @@
...
+        with:
+          image: tonistiigi/binfmt:qemu-v6.1.0
+          platforms: all
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          driver-opts: network=host
+
+      - name: Push image useing the master branch commit SHA.
```
Kami, line 123: So version push will happen as part of a manual workflow and not automatically on push to a tag branch?

ArthurKamalov: Right, for now, it just automatically pushes from master on a new commit. Do you think I should back the pushing on tag?

Kami: If we can address the edge case I mentioned the other case then I think we should.

This should also allow us to simplify the release pipeline since we don't need to call manual workflow from Jenkins pipeline, right?

We would just need to make sure tests run before the image is pushed.

Kami (approved):

build_package_new.py
```diff
@@ -0,0 +1,158 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This is a new package build script which uses new package  build logic.
+# usage:
+#       build_package_new.py <name of the package> --output-dir <output directory>
```
Kami, line 17: Would be good to add a quick check to this script which fails with more useful error message when running under Python version older than 3.8 - otherwise bit confusing error is printed and it's not immediately clear why it's failing to run.

PR #795, opened 2021-11-30
Update MySQL monitor docs, clarify which permissions are needed

Kami:
This pull request updates MySQL monitor docs and clarifies which permissions are needed for the user which is used for monitoring purposes.

Sadly there are some differences between MySQL versions and implementations so we can't offer a single solution which would work in all the scenarios and grant correct permission for ``SHOW SLAVE STATUS`` command.


ArthurKamalov (approved):
LGTM

PR #796, opened 2021-12-01
Add new "collect_replica_metrics" config option to the MySQL monitor

Kami:
This pull request adds new ``collect_replica_metrics`` config option to the MySQL monitor (which defaults to ``True`` for backward compatibility reasons).

When this option is set to False (e.g. when user is monitoring a master and not a replica), monitor won't try to collect replica metrics and as such, user which is used to connect to the database only requires ``PROCESS`` permission grant and nothing else.

Technically, we could also change the behavior to make it even more permissive - if we encounter permission related error when querying for replica metrics we could only print an error that monitor user needs additional permissions and continue to operate normally (right now if we encounter this error it's a fatal one and monitor won't start and collect other metrics). This would require a bit more code changes though.

In the future, we may also want to make this new config option opt-in and not opt-out by default.

I also removed some Python < 2.7 support code which is not needed anymore.


ArthurKamalov (approved):
LGTM

PR #797, opened 2021-12-09
Upgrade requests to 2.20

ArthurKamalov:
This pull request upgrades the `requests` library to 2.20. 

The main change besides the upgrade itself is that the request library now does not store its own dependencies "baked" in its own source code and just moved them to the normal package dependencies. For us, that means that we have to move those dependencies to other dependencies in the `third_party` directory.  

The `requests` 2.20 also has an additional dependency - 'certifi'. It has been also added to third-party libraries.


Kami:
LTTM.

As far as certifi goes - it contains a list of valid and revoked CAs so we should make sure it's updated frequently / as part of every release (especially once we move to frozen binaries).

Kami:
And once we move to frozen binaries and drop support for old Python versions we can just pin requests to the latest stable version.

Kami (approved):

ArthurKamalov:
> And once we move to frozen binaries and drop support for old Python versions we can just pin requests to the latest stable version.

Good idea.

PR #798, opened 2021-12-18
Docker imate support for linux/armv7 platform

Kami:
This pull request adds support for ``linux/arm`` architecture.

This architecture includes ``linux/armv7`` which is an architecture used by Raspberry Pi and some other lower powered embedded ARM and IoT devices.

Older versions of Raspberry Pi used ``armv6``, but not many places support this architecture anymore so I didn't add support for it.

Overall, this PR includes the following changes:

- Support for building and pushing ``linux/armv7`` images
- Add new ``--reuse-local-cache`` argument to the build script. This comes handy when building the image locally (outside CI) to speed up the builds.
- Fix tag name when using custom registry. Previous code wouldn't not work correctly when using custom registry when the user / org name is not ``scalyr``. Constructed URL would end up being ``<registry url>/<user>scalyr/scalyr-agent-2:tag``, but we want ``<registry url>/<user>/scalyr-agent-2:tag``. @ArthurKamalov can you please review this change? I just did a quick solution which works, but there are likely other possible solutions (e.g. remove user / org name from the "image_name", etc.)
- Update docker image requirements file so we don't bundle ``orjson`` and ``ujson`` dependency. Those dependencies don't offer armv7 compatible pre-built wheels at this point. This means we would need to build them inside the Docker image. To do that, we would need to install rust, etc. Since the building is slow, I decided to skip it for now and simply exclude those dependencies on armv7 (we still use those dependencies on arm64 for which wheels are available). It means that we will simply fall back to stdlib json library.

Example I used to build the image locally:

```bash
python3 build_package_new.py k8s --tag "$(date +%s)-dev" --registry kami007 --reuse-local-cache --remove-image-name-prefix --push
```

### DIFF
--- a/.github/workflows/agent-build.yml
+++ b/.github/workflows/agent-build.yml
@@ -3,28 +3,10 @@ name: Scalyr Agent Build.
 
 on: [push]
 
-# Agent image tests.
+# Agent Docker image tests
 jobs:
-  match-branches:
-    runs-on: ubuntu-20.04
-
-    steps:
-      - name: Check branches.
-        env:
-          TARGET_BRANCH: master
-
-        # Because of the issue https://github.community/t/ref-head-in-reusable-workflows/203690
-        # there's no way to specify local re-usable workflow. The branch of the workflow has to be specified manually and statically
-        # To avoid confusion, run this check that will fail if the current branch does not match the branch of the workflow.
-        # TODO: Create pre-commit hook that verifies branch name and changes it if needed. That may be a good workaround
-        # until Github add needed feature.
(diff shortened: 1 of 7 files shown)


Kami:
@ArthurKamalov we also need to fix / change the build setup a bit so it doesn't cause the build to be marked as failed in case the branch doesn't match master (https://github.com/scalyr/scalyr-agent-2/runs/4570525718).

Kami:
Re match-branches issue I mentioned on Slack. Can't we just use something like this - ca4f4b5b? Or am I missing something?

Kami:
Once this PR is fully finished and merged (+ permissions issue fixed), we should also enable publishing of dev images to Docker Hub on merge / push to master.

ArthurKamalov (approved):
LGTM

PR #800, opened 2021-12-20
Add docker images push by tag

ArthurKamalov:
This PR adds a new step, in the docker publishing job in the GitHub actions, that verifies the given tag of the repository and publishes appropriate image tags:
- custom image tag.
- or production tag (e.g. 2.1.25) with the `latest` tag. That's only the case when the repo tag matches a  pattern of the repo's production tag, e.g - `v2.1.25`


VERSION
```diff
@@ -1 +1 @@
-2.1.26
+2.1.25
```
Kami, line 1: Was this accidental change?

.github/workflows/agent-build.yml
```diff
@@ -6,7 +6,7 @@ on: [push]
 # Agent Docker image tests
 jobs:
   docker-json:
-    uses: scalyr/scalyr-agent-2/.github/workflows/docker-and-k8s.yml@master
+    uses: scalyr/scalyr-agent-2/.github/workflows/docker-and-k8s.yml@add-images-prod-push
```
Kami, line 9: I assume this testing branch name change will be removed before merging, right? :)

VERSION
ArthurKamalov: I accidentally bumped the version in previous PR's and forgot to return it back :)

.github/workflows/agent-build.yml
ArthurKamalov: Right, it seems that's pointless and we need to use it from master)

VERSION
Kami, line 1: Why did you revert this change?

IIRC, we can't do internal build via Jenkins if the VERSION file is not set to this value.

Kami (approved):
Approving, but please make sure VERSION file stuff is sorted out / understood as discussed on Slack.

Kami:
OK, all the builds are green :)

PR #801, opened 2021-12-21
Add support for Alpine Linux based Docker images

Kami:
This pull request includes Dockerfile and other changes which are needed to produce Alpine Linux based Docker images.

Alpine Linux is a popular base image choice since it's smaller than the alternatives. 

One of the disadvantages with alpine is that it uses musl and not libc and pre-built musl specific binaries (in our case wheels) are not available and that increases the build times.

In our case, we only have one dependency like that (orjson) and IIRC, recent manywheel wheel specification already supports musl specific wheels (they just need to be built and provided by the library developer).

For now, we avoid very long build times by excluding orjson and zstandard dependency on ARMv7 - both of those are optional dependencies at this point and don't have any negative consequences for the end user.

Image sizes:

* standard one (based on buster-slim) - 146 MB uncompressed, 47 MB compressed
* alpine one - 63 MB uncompressed, 20 MB compressed

```bash
(py3.8) $ docker image ls | grep kami007 | grep dev
kami007/scalyr-k8s-agent   buster-dev        eae692e628cc   2 hours ago    146MB
kami007/scalyr-k8s-agent   alpine-dev        af5027905701   21 hours ago   63.9MB
```

Screenshot from 2021-12-21 15-03-27

Example usage:

```bash
# Regular buster slim based image
python3 build_package_new.py k8s --tag "buster-dev" --remove-image-name-prefix  --push

# Alpine based image
python3 build_package_new.py k8s --dockerfile-name "Dockerfile.alpine" --tag "buster-dev" --remove-image-name-prefix --push
```

As far as tagging goes, usually the convention is to add ``-alpine``` suffix (e.g. ``v2.11.25-alpine``, ``<sha>-alpine``).

Image has been tested and everything which was testing is working fine.

It also includes some other changes:

* ``ujson`` dependency has been removed from the Docker image. It's not really maintained anymore and we have a better alternative in place (``orjson``) which is already used by default under Python 3.

## Notes, Limitations

Since pre-built wheels for ARM + musl are not available and building and compiling Rust is very slow in emulated ARM environments, for the time being we don't include optional ``orjson`` and ``zstandard`` dependency in Alpine based ARM images.

If we really wanted to, we could include it, but this would require building those wheels which would substantially slow the build time for those ARM images(+20-30 minutes or more).

## TODO

- [x] Hook up builds and publish, ensure correct cache is used for buster and alpine based images (need to update checksum calculation code to support using different path to Dockfile)


Kami:
@ArthurKamalov Alright, I pushed the changes (2dd8253d, e2e151f6) which refactor the image builder classes to support multiple builders per package - e.g. one for Buster and another for Alpine based image.

With this change, we now have type sets of Docker image builder - ``<package type>-<image type>``, e.g. ``k8s-buster`` and ``k8s-alpine``.

Let me know what you think about those changes / abstractions. I'm not too happy about it, but I can't come up with a better alternative. I would be more than happy if we could come up with a simpler approach which still works.

.github/workflows/docker-and-k8s.yml
```diff
@@ -72,13 +73,13 @@ jobs:
 
       - name: Get checksum of the base image files
         run: |
-          python3 build_package_new.py ${{ inputs.image-type }} --files-checksum > image_files_checksum.txt
```
ArthurKamalov, line 75: shouldn't this also be `${{ matrix.image_distro_name }}"`

.github/workflows/docker-and-k8s.yml
```diff
@@ -72,13 +73,13 @@ jobs:
...
+          python3 build_package_new.py "${{ inputs.image-type }}-buster" --files-checksum > image_files_checksum.txt
```
ArthurKamalov, line 76: shouldn't this also be ${{ matrix.image_distro_name }}"

Kami: To reduce the build time, we only run test jobs for buster. But if you think it's a good idea, I can update the workflow to also run tests for alpine.

Kami: Pushed a change to also run tests under Alpine. Will see how long it takes.

ArthurKamalov (approved):

PR #804, opened 2021-12-22
Use self hosted runners for Docker build jobs

Kami:
Change to use self hosted runners for new Docker image build jobs.

### DIFF
--- a/.github/workflows/agent-build.yml
+++ b/.github/workflows/agent-build.yml
@@ -11,12 +11,16 @@ on: [push]
 
 # 2. In this workflow, change docker hub secrets to utilize testing and not prod account so images
 #    get pushed to testing account. Change ``_PROD_` in the secret name to ``_TEST``,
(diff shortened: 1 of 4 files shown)


Kami:
I also added some info to the workflow file on how to test those workflow changes with a non-master branch - 
3beff77cd5f3faf131467164fe5199e05ea966f1 and pushed a change to verify that those changes indeed work correctly (b51f55bb5dd12d60c7d9516c77ad13c39ee902e7)

I also added ``pre-job`` which automatically cancels redundant old builds for non-master branches - 3e9d413df83e52c887651fb2a832de6e3ef1fca8.

Kami:
I switched back to cloud runners since there appear to be some issues with self hosted ones and there's is actually no need to use self hosted runners at this point since everything is public and we have enough concurrency still available.

Kami:
Once the tests pass, I will still go ahead and merge this PR since it includes two other useful changes.

PR #805, opened 2021-12-24
Improve caching

ArthurKamalov:
This PR: 
* Splits agent docker image Dockerfile into the base and final dockerfiles. Base image build does all of the hard work and then can be cached to speed up the process. All build instructions for the `buster` and `alpine` version now are in the same dockerfile - `agent_build/docker/Dockerfile.base`. The final Dockerfile does nothing except the creating of the agent files, which is a completely non-time-consuming process, so the final Dockerfile does not require caching.
* I also reorganized requirement files.  <DEBATABLE TOPIC ALERT>:  I Enabled `orjson` and `zstandard` for alpine to keep fewer requirement files.  Yeah, the build time under QEMU is terrible ( at least 40 min), but, caching has to reduce that to 1-2 min. We only have to wait that long when we change a very small set of files that can change the checksum of the deployment. 
* Cached docker images now take only a few hundred megabytes!.
* Docker images are now all built-in one workflow but in a different job by using job matrix. The next step is to add the final "publish" job that aggregates all results and pushes all images at once, from one job. That has to make the publishing process more "atomic", as we discussed earlier.

### DIFF
--- a/.github/workflows/agent-build.yml
+++ b/.github/workflows/agent-build.yml
@@ -5,46 +5,10 @@ on: [push]
 
 # Agent Docker image tests
 jobs:
-  docker-json:
-    uses: scalyr/scalyr-agent-2/.github/workflows/docker-and-k8s.yml@master
-    with:
-      image-type: docker-json
-      publish: true
+  build-images:
+    uses: scalyr/scalyr-agent-2/.github/workflows/docker-and-k8s.yml@refactor-deployment-caching
     secrets:
       SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN: ${{ secrets.SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN }}
       SCALYR_CLOUDTECH_TESTING_DEV_SCALYR_READ_API_KEY: ${{ secrets.SCALYR_CLOUDTECH_TESTING_DEV_SCALYR_READ_API_KEY }}
-      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME_PROD_ACCOUNT }}
-      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD_PROD_ACCOUNT }}
-
-  docker-syslog:
-    uses: scalyr/scalyr-agent-2/.github/workflows/docker-and-k8s.yml@master
-    with:
-      image-type: docker-syslog
-      publish: true
-    secrets:
-      SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN: ${{ secrets.SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN }}
-      SCALYR_CLOUDTECH_TESTING_DEV_SCALYR_READ_API_KEY: ${{ secrets.SCALYR_CLOUDTECH_TESTING_DEV_SCALYR_READ_API_KEY }}
-      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME_PROD_ACCOUNT }}
-      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD_PROD_ACCOUNT }}
-
-  docker-api:
-    uses: scalyr/scalyr-agent-2/.github/workflows/docker-and-k8s.yml@master
-    with:
-      image-type: docker-api
-      publish: false
-    secrets:
-      SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN: ${{ secrets.SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN }}
-      SCALYR_CLOUDTECH_TESTING_DEV_SCALYR_READ_API_KEY: ${{ secrets.SCALYR_CLOUDTECH_TESTING_DEV_SCALYR_READ_API_KEY }}
-      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME_PROD_ACCOUNT }}
-      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD_PROD_ACCOUNT }}
-
-  k8s:
-    uses: scalyr/scalyr-agent-2/.github/workflows/docker-and-k8s.yml@master
-    with:
-      image-type: k8s
-      publish: true
-    secrets:
-      SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN: ${{ secrets.SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN }}
-      SCALYR_CLOUDTECH_TESTING_DEV_SCALYR_READ_API_KEY: ${{ secrets.SCALYR_CLOUDTECH_TESTING_DEV_SCALYR_READ_API_KEY }}
-      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME_PROD_ACCOUNT }}
-      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD_PROD_ACCOUNT }}
+      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME_TEST_ACCOUNT }}
+      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD_TEST_ACCOUNT }}
--- a/.github/workflows/docker-and-k8s.yml
+++ b/.github/workflows/docker-and-k8s.yml
@@ -3,19 +3,10 @@ name: Build Agent Images
 on:
   workflow_call:
     inputs:
-      image-type:
-        description: "Name of the image build type (docker-json, docker-syslog, docker-api, k8s)."
-        required: true
-        type: string
       python-version:
         required: false
         type: string
         default: 3.8.10
-      publish:
-        description: "If true, then the result image will be published on master branch or any tag."
-        type: boolean
-        required: false
-        default: false
 
     secrets:
       SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN:
@@ -39,9 +30,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - image_distro_name: "buster"
-          - image_distro_name: "alpine"
+        image-type: ["docker-json", "docker-syslog", "docker-api", "k8s"]
+        image-distro-name: ["buster", "alpine"]
+
+    env:
+      # Set this variable to tell the agent build code that it runs in CI/CD and it needs to use caching.
+      AGENT_BUILD_IN_CICD: "1"
 
     steps:
       - name: Checkout repository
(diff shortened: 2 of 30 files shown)


Kami:
> I also reorganized requirement files. "DEBATABLE TOPIC ALERT": I Enabled orjson and zstandard for alpine to keep fewer requirement files. Yeah, the build time under QEMU is terrible ( at least 40 min), but, caching has to reduce that to 1-2 min. We only have to wait that long when we change a very small set of files that can change the checksum of the deployment.

If we only need to run full rebuild infrequently (e.g. when updating one of those two dependencies or similar), then I'm fine with it.

It likely allows us to simplify and unify things and have less conditionals.

agent_build/docker/Dockerfile
```diff
@@ -0,0 +1,87 @@
+# Final Scalyr Agent docker image that uses the previously built base image.
```
Kami, line 1: If this file is auto generated from the base file, please add a comment which say auto-generated with instructions on how to generate it.

agent_build/docker/Dockerfile.base
```diff
@@ -0,0 +1,81 @@
...
+    python3-dev \
+    patchelf \
+    git
+
+# Install rust and cargo. It is needed to build some of the Python dependencies of the agent.
+FROM scalyr-dependencies-$BASE_IMAGE_SUFFIX as scalyr-dependencies-rust-
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN rustup toolchain install nightly
+RUN rustup default nightly
```
Kami, line 36: I see you removed manylinux2014 wheel workaround, do you think we don't need it anymore?

With that workaround, we can re-use existing zstandard + orjson wheels for arm64 (yes, sadly just arm64 and we still need to build it from scratch for alpine + armv7).

agent_build/tools/environment_deployments/deployments.py
```diff
@@ -610,11 +617,193 @@ def get_deployment_by_name(name: str) -> Deployment:
...
+                result_image_name=f"{type(self).BASE_IMAGE_NAME}_testing",
+                with_coverage=True
+            )
+            # build prod version of the base image.
+            _build_base_image(
+                result_image_name=type(self).BASE_IMAGE_NAME,
+                with_coverage=False
+            )
+
+        if common.IN_CICD and not hit:
```
Kami, line 749: Does caching still works if you run build package script locally?

agent_build/docker/Dockerfile
ArthurKamalov: No, that file is static and does not change. It just waits. that we specify the `BASE_IMAGE ` build arg to use it as a base image.

agent_build/docker/Dockerfile.base
ArthurKamalov: If we agree on the first slow build and then we rely on caching, then, I guess, this workaround won't win more time for us anymore.

agent_build/tools/environment_deployments/deployments.py
ArthurKamalov: Locally it is handled by docker's own caching. You just have to build it once, and docker will reuse cached layers, so there's no need to enable the `IN_CICD ` flag locally.

ArthurKamalov:
TODO: 

- [ ] publish all images from one job.

agent_build/docker/Dockerfile.base
Kami: I guess it may still save us some time during the build of the first base image :)

agent_build/docker/Dockerfile.base
```diff
@@ -0,0 +1,81 @@
...
+    g++ \
+    make \
+    curl \
+    python3-dev \
+    patchelf \
+    git
+
+# Install rust and cargo. It is needed to build some of the Python dependencies of the agent.
+FROM scalyr-dependencies-$BASE_IMAGE_SUFFIX as scalyr-dependencies-rust-
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
Kami, line 33: I just did curl + bash as a quick solution to test it out a while ago, but to be on the safe side we should probably either vendor the installer script file inside this repo or download the installer file and check the checksum against the one statically specified in this Dockerfile.

Kami: Btw - does building wheel works on armv7 under alpine? I had issues and couldn't get it to work since rust toolchain for cross compiling is not available for armv7 + musl.

Kami: EDIT: Never mind, I see this comment below.

>  # if TARGETVARIANT is 'v7', then we use stage without rust.

IIRC, that is only the case for armv7 + alpine (musl) and not armv7 + buster (glibc).

Kami:
@ArthurKamalov What command should now be used to build the images (locally)? It seems like the same command which worked before, doesn't work anymore.

Kami:
Once this PR is finished, we can remove ``Dockerfile`` and ``Dockerfile.alpine`` from the repo root since those files are not used anymore, right?

agent_build/docker/Dockerfile.base
```diff
@@ -0,0 +1,90 @@
...
+# Use one of the two previous stages according to the TARGETVARIANT suffix.
+# if TARGETVARIANT is 'v7', then we use stage without rust.
+# IMPORTANT: that may be needed to revise if new architectres are added.
+FROM scalyr-dependencies-rust-$TARGETVARIANT as scalyr-dependencies
+
+ARG TARGETVARIANT
+ARG COVERAGE_VERSION
+
+ADD agent_build/requirement-files/docker-image-requirements.txt agent_build/requirement-files/docker-image-requirements.txt
+ADD agent_build/requirement-files/compression-requirements.txt agent_build/requirement-files/compression-requirements.txt
```
Kami, line 51: I think there is a small change from before - in current master we don't include ``lz4`` inside the final built Docker image.

It's probably fine to leave it in place though in case we decide to expose lz4 support on the server side to the users in the future.

Kami: Actually, since that's not included in master Docker image and we are not officially supporting it server side yet, we may want to remove this dependency from docker image from the time being - this should speed base image build a bit, since it appears lz4 doesn't offer pre-built wheels (even for amd64).

Kami:
Per my comment on Slack - I finally got it working and building locally (and it's also now faster due to the removed lz4 dependency), but for some reason built buster and alpine uncompressed and compressed image size is almost the same which means that likely something is not correct.

I built the images like this:

```bash
python3 build_package_new.py k8s --tag "2.1.26dev-test" --remove-image-name-prefix  --push --registry kami007

python3 build_package_new.py k8s-alpine --tag "2.1.26dev-alpine-test" --remove-image-name-prefix  --push --registry kami007
```

If you check compressed image size on Docker Hub you will see it's almost the same in both cases:

Screenshot from 2022-01-03 17-09-09

And here is Alpine image built on top of master without changes from this branch:

Screenshot from 2022-01-03 17-09-24

I didn't have a chance yet to dig in what is going on, but this is a blocker and needs to be fixed.

Kami:
While working on #810, I noticed that ``smoke-k8s`` Circle CI job still utilizes old Docker image.

We should update that job to utilize new Docker image. Or if that job is now not needed anymore with the new job and images, it should be removed.

Kami:
In #810 I fixed and updated ``smoke-k8s`` tests to utilize new Docker image.

We should cherry pick that changes into this branch since we likely won't be able to merge that PR until frozen binary stuff is ready and merged.

On that note - we need to make sure base buster image builds fast (aka we use wheels for most things) since we don't utilize this new caching on Circle CI otherwise that job will take forever ever...

tests/package_tests/all_package_tests.py
```diff
@@ -201,6 +202,27 @@ def _test_pushed_image():
...
+                        "docker",
+                        "run",
+                        "-i",
+                        "--rm",
+                        str(full_image_name),
+                        "/bin/cat",
+                        "/etc/os-release"
+                    ]).decode()
+
+                    assert expected_os_name in os_release_content.lower()
```
Kami, line 224: :+1:

Kami:
@ArthurKamalov I tested new changes and confirmed that the alpine image is being built correctly now.

Per my comment above, I also cherry picked some changes and fixes from #810:

* 5340e2d3fda1b8624c4f8fc87e7c59a4a50de91a
* 4d24a18be580455742bad50eb47ab0d5a5193f67
* 398d17a18f5a93eac30eae509288270d48853c79

What's left before I can perform a final review of this PR? Can you please make it ready for a final review (re-enable tests, verify caching is working, etc.)? Thanks.

And as discussed before let's please clean up the repo and remove all old Dockerfiles and requirements files we don't need anymore.

And while at it, let's please add some short section to the readme on how to build docker image - in the past it was simpler due to Dockerfile being directly in the repo root and now the user needs to use the build script, etc.

Kami:
It looks like smoke-k8s is now failing because of some upstream apt repo issue, but other smoke jobs appear to be failing because something has changed in this PR so will likely need to be updated.

Kami:
OK, ``smoke-k8`` Circle CI job is now finally passing with new Docker image (https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/7773/workflows/407008be-d187-4dc6-9790-1c66a387792e/jobs/96195).

Other ``smoke-docker-*`` jobs are still failing. I assume similar changes will need to be made as I made to smoke-k8s to utilize new Docker image and to make it pass. @ArthurKamalov please look into this when you get a chance.

ArthurKamalov:
@Kami I pushed new changes with a new publishing logic. In short, all publishing jobs initially push their images to the docker hub with a special tag (I picked `_temp_release`). If everything is ok, then the final job runs a simple script and “re-tags” those images to the final tags. That approach has to provide some kind of atomicity, I think.

.github/workflows/docker-and-k8s.yml
```diff
@@ -87,93 +87,67 @@ jobs:
...
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_PASSWORD }}
+
+      - name: Push image using the master branch commit SHA.
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: |
+
+          # Push result docker image with a special tag "_temp_release".
```
Kami, line 133: Would be good to add a note here that this is needed because this step doesn't just push the image, but it also builds it (which may not be immediately obvious since the step name just stay push, maybe we should update it to say Build and push docker image using master branch commit sha).

agent_build/package_builders.py
```diff
@@ -930,20 +930,21 @@ def build(
...
-                    # prefix, we remove it in case remove_image_prefix argument is provided (
-                    # this is desired in a lot of cases when using custom registry)
-                    # e.g. scalyr/scalyr-agent-2 -> scalyr-agent-2
-                    if remove_image_name_prefix and "/" in image_name:
-                        image_name = image_name.split("/")[1]
-                    tag_options.append(f"{registry_prefix}{image_name}:{tag}")
+
+            full_name = image_name
+
+            if user:
```
Kami, line 935: I see you refactored that, good (previous approach with --remove-image-tag-name-prefix I did was a quick hack / workaround, this one should be better).

.github/workflows/docker-and-k8s.yml
ArthurKamalov: Yeah, it also builds it, I'll add that. Another notion is that it does not actually push using master branch commit sha. The main purpose of this job is just to build and push to the docker hub with the `_temp_release` tag if we are in the master branch or some tag. The following job has to run a small script file `scripts/cicd/publish-docker-images-tags.py` that has to re-tag previously pushed `_temp_release` images to normal tags.

Another note: I have also added a concurrency option to these jobs, to avoid race conditions on multiple parallel pushes.

Kami: Ack, thanks for clarifying.

Kami (approved):
LGTM

ArthurKamalov:
@Kami I moved the docker image test into the outer workflow, so the reusable workflow can contain only publishing-related jobs. In this case, we can use the concurrency option for the whole workflow, which has to provide much safer protection from race conditions.

.github/workflows/agent-build.yml
```diff
@@ -14,54 +14,84 @@ on: [push]
...
+          SCALYR_API_KEY: ${{ secrets.SCALYR_PROD_CLOUDTECH_TESTING_WRITE_TOKEN }}
+        run: |
+
+          python3 tests/package_tests/run_package_test.py package-test \
+           "${{ matrix.image-type }}-${{ matrix.image-distro-name }}_test" \
+            --name-suffix "_${{ github.run_number }}${{ github.run_attempt }}" \
+
+
+  publish-images:
+    if: ${{ github.ref == 'refs/heads/improve-caching' || github.ref_type == 'tag' }}
```
Kami, line 88: I assume you will change this to master branch name after it's confirmed working, right?

.github/workflows/publish-docker-images.yml
```diff
@@ -0,0 +1,96 @@
+name: Build Agent Images
```
Kami, line 1: Please fix the name to publish, not build

.github/workflows/publish-docker-images.yml
```diff
@@ -0,0 +1,96 @@
...
+
+      - name: Push image using the a temporary tag.
+        run: |
+          # Push result docker image with a special tag "_temp_release".
+          # The next job will "re-tag" that image to an appropriate one.
+          # That is done to guarantee that the real, production tag only pushed when all job are succeeded.
+          python3 build_package_new.py "${{ matrix.image-type }}-${{ matrix.image-distro-name }}" \
+            --push \
+            --user "${{ secrets.DOCKER_HUB_USERNAME }}" \
+            --tag "_temp_release-${{ matrix.image-distro-name }}"
```
Kami, line 73: Once we re-tag the image with the final tag, will this temporary tag be removed or?

.github/workflows/agent-build.yml
ArthurKamalov: right

Kami (approved):
LGTM, nice work 👍

.github/workflows/publish-docker-images.yml
ArthurKamalov: Should it be removed? I just thought it's ok to leave it until the next time.

Kami: I'm fine either way, I also don't see a problem if we leave it there.

PR #806, opened 2021-12-26
Add new Kubernetes Open Metrics monitor

Kami:
(the code is still WIP and still needs a bunch of refactoring, please only leave comments about the architecture and open questions and not about the actual code yet)

## Description

This pull request builds on top of #792 and adds new ``KubernetesOpenMetricsMonitor`` monitor.

This monitor is designed to run on scaly agents which are deployed in a Kubernetes cluster as a ``DaemonSet`` (it doesn't work if agent is deployed as a ``Deployment`` or a side car container).

It supports automatically discovering Prometheus metrics exporter endpoints and scraping them (think node exporter, jmx exporter, app specific exporters, etc.).

In addition to those dynamic exporters, it also supports scrapping two fixed metrics endpoints exposed by the Kubernetes API in the OpenMetrics format.

Those are ``https://kubernetes.default.svc:443/api/v1/nodes/<node name>/proxy/metrics`` and ``https://kubernetes.default.svc:443/api/v1/nodes/<node name>/proxy/metrics/cadvisor``.

First endpoint returns various Kubernetes API related metrics and the second one returns various cAdvisor metrics (things like pod resource utalization, etc.).

Both of those endpoints operate on the node level and return node specific metrics - so that plays along well with our DaemonSet mode and work distribution (each agent is only responsible for exporters running on the same node the agent is running on).

It's worth noting that there are some similarities and overlap between this endpoint and the "/stats/summary" Kubelet API endpoint Kubernetes monitor relies on, but the latter returns JSON response and different set of metrics.

As discussed before, ``/stats/summary`` Kubelet API has multiple issues such as no official schema and documentation and plans that it will get deprecated in the future (there is no fixed timeline for that yet though). So it could very well be, that scrapping this cAdvisor metrics endpoint could also serve as a replacement for the Kubelet API metrics (but that will of course require more digging in into the available metrics and that was not the main purpose of the changes in this PR).

## How it works

This monitors auto discovers node exporters by querying Kubelet API for pods running on the local node and then finding exporters based on the pod annotations.

We rely on the following annotations:

* ``prometheus.io/port``
* ``prometheus.io/scheme``
* ``prometheus.io/path``
* ``monitor.config.scalyr.com/scrape``

First 3 are the de-facto Prometheus auto-discovery annotations the last one is a new one I introduced as part of this PR.

Prometheus also utilizes ``prometheus.io/scrape`` annotation to decide if a specific exporter should be scraped, but we don't want to utilize that since we want user to explicitly opt-into this metrics gathering by the agent on the pod level - a lot of those metric exporters expose tons of metrics so ingesting them could be expensive.

Here is an example YAML definition for node exporter ``DaemonSet`` which will be auto-discovered by this monitor.

```yaml
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  labels:
    app.kubernetes.io/component: exporter
    app.kubernetes.io/name: node-exporter
  name: node-exporter
  namespace: monitoring
spec:
  selector:
    matchLabels:
      app.kubernetes.io/component: exporter
      app.kubernetes.io/name: node-exporter
  template:
    metadata:
      labels:
        app.kubernetes.io/component: exporter
        app.kubernetes.io/name: node-exporter
      annotations:
        prometheus.io/scrape:             'true'
        prometheus.io/port:               '9100'
        monitor.config.scalyr.com/scrape: 'true'
    spec:
      containers:
      - args:
        - --path.sysfs=/host/sys
        - --path.rootfs=/host/root
        - --no-collector.wifi
        - --no-collector.hwmon
        - --collector.filesystem.ignored-mount-points=^/(dev|proc|sys|var/lib/docker/.+|var/lib/kubelet/pods/.+)($|/)
        - --collector.netclass.ignored-devices=^(veth.*)$
        name: node-exporter
        image: prom/node-exporter
        ports:
          - containerPort: 9100
            protocol: TCP
        resources:
          limits:
            cpu: 250m
            memory: 180Mi
          requests:
            cpu: 102m
            memory: 180Mi
        volumeMounts:
        - mountPath: /host/proc
          name: proc
          readOnly: false
        - mountPath: /host/sys
          name: sys
          readOnly: false
        - mountPath: /host/root
          mountPropagation: HostToContainer
          name: root
          readOnly: true
      nodeSelector:
        kubernetes.io/os: linux
      securityContext:
        runAsNonRoot: true
        runAsUser: 65534
      volumes:
      - hostPath:
          path: /proc
        name: proc
      - hostPath:
          path: /sys
        name: sys
      - hostPath:
          path: /
        name: root
```

## Implementation

Internally this monitor utilizes regular ``OpenMetrics`` monitor I added in #792 (which many improvements in this PR).

It works by automatically scheduling and starting + stopping ``OpenMetrics`` monitors as needed. This scheduling work happens on every ``gather_sample()`` loop (so by default, that happens every 60 seconds).

Since our ``MonitorsManager`` abstraction didn't expose APIs for dynamically managing monitors, I needed to implement that (it works, but it still needs some refactor and improvements).

To avoid cyclic reference between ``MonitorsManager`` and actual ``Monitor`` class I introduced new method for retrieving reference to the currently active ``MonitorsManager`` instance (which is a singleton).

Nice thing about re-using MonitorsManager framework is that we get the related functionality for "free" (aka thread per monitor and management of those threads).

Technically using some kind of async framework such as eventlet or gevent may be more resource efficient since the code is mostly IO bound, but confirming that theory would require actual benchmarking and it's not worth the effort at this point since having eventlet / gevent monkey patched event loop running inside the same agent process would be challenging (and the best way to approach that would be to actually run a separate process for MonitorsManager which utilizes eventlet loop, but that would require a lot of work and not something which is feasible or worth it in the mid term).

## Testing

I tested this on a local minikube k8s and Raspberry Pi k3s cluster. It works correctly and auto-discovers and scrapes metrics for the following exporters I have set up:

* node exporter
* rpi / arm exporter
* kube state metrics exporter
* jmx exporter which runs as an agent along java app

It's very likely though that it doesn't cover all the various scenarios and configurations, but only time and testing will tell (e.g. right now we perform auto discovery on pod level, but I imagine in some special setups someone may want to perform it on Service / Endpoint level, but this would make it a lot more challenging especially with the Daemonset deployment model).

## Open Questions

1. Some of those exporters produce a lot of data. We will need a manageable way to specify metrics name include and exclude lists per exporter (e.g. as pod annotations). A lot of exporters, especially the Kubernetes API + cAdvisor ones return many metrics per scrape (2500+). In the former, a lot of those metrics are different histogram quantile values for each API path. Think response timing and counters and 10 or so quantiles per API endpoint which adds up really fast and most people are probably not interested in ingesting those. 

2. For better integration of this functionality, we will likely need some additional changes / support on the server side (think dashboards, billing, etc.)

3. Decide on which additional attributes we should include with each metric log line (think kubernetes cluster name, node name, ip, etc.). This is also somewhat tied to billing discussion since it's a lot of duplicated data (although data which compresses well).

## TODO

- [x] Refactor the code
- [x] Finish the implementation
- [x] Unit tests (with fixtures)
- [ ] End to end tests (spin k8s minikube cluster, deploy various exporters, send data to Scalyr and verify it works correctly)


tests/package_tests/internals/common.py
```diff
@@ -252,17 +252,20 @@ def perform(
                 if bytes_sent <= 0:
                     return (
                         LogVerifierCheckResult.FAIL,
-                        "Agent log says that during the run the agent has sent zero bytes.",
+                        f"Agent log says that during the run the agent has sent zero bytes.\n"
+                        f"Whole log content: {whole_log_text}",
```
ArthurKamalov, line 256: It seems that some changes have not been merged yet from the master, but there has to be the similar output of the whole test on the outer level https://github.com/scalyr/scalyr-agent-2/blob/79e7066a73b868121677068e276cbe6a8a597342/tests/package_tests/internals/common.py#L189-L200

There may be a double output, I guess

ArthurKamalov (approved):
LGTM

tests/package_tests/internals/common.py
Kami: I will test the change on failure and see how it behaves and remove this change if it's not needed anymore (I added it since in the pat it was hard to troubleshoot failures since they were missing the context).

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -2336,6 +2341,10 @@ def __init__(
     def _build_kubelet_url(kubelet_url, host_ip, node_name):
         if node_name and host_ip:
             return kubelet_url.substitute(node_name=node_name, host_ip=host_ip)
+
+        global_log.warn(
```
Kami, line 2345: Technically, we could also throw here since in case this method returns ``None`` everything else will fail (and IIRC that method is called only once in the constructor so there is no chance for it evaluating to a different value later on).

scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
```diff
@@ -0,0 +1,983 @@
...
+
+        # 1. Query Kubelet API for pods running on this node and find exporters we want to scrape
+        # (based on the annotations)
+        k8s_pods = self.__get_k8s_pods()
+        node_name = self.__get_node_name()
+
+        self._logger.info(f"Found {len(k8s_pods)} pods on node {node_name}")
+
+        # Maps scrape URL to the corresponding monitor config and K8sPod
+        scrape_configs: Dict[str, Tuple(OpenMetricsMonitorConfig, K8sPod)] = {}
```
ArthurKamalov, line 700: I guess there have to be square brackets.

scalyr_agent/builtin_monitors/openmetrics_monitor.py
```diff
@@ -0,0 +1,599 @@
...
+        lines = resp.text.splitlines()
+
+        result = []
+
+        metric_name_to_type_map = {}
+
+        metrics_count = 0
+
+        for line in lines:
+            comment_metric_name = None
```
ArthurKamalov, line 373: it seems like those 2 variables are redeclared inside of the if statement below, so the current ones can be removed.

ArthurKamalov (approved):
Looks good to me, as far as I could fill my knowledge gaps. I have nothing but some small lint issues.  

I see that end-to-end tests are also in the plan. I guess they will be introduced in the next PRs?

scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
Kami: Yeah, that's type annotations syntax.

scalyr_agent/builtin_monitors/openmetrics_monitor.py
Kami: Good catch, will remove it.

PR #809, opened 2022-01-04
Update numpy to the latest version

Kami:
This PR updates numpy to the latest version which includes a fix for a security issue (we only use numpy inside the tests / benchmarks).


Kami:
Looks like we can't fully upgrade until we drop support for Python 2.7 and Python 3 < 3.8.

PR #810, opened 2022-01-04
[WIP] Drop support for Python < 3.8

Kami:
This pull request updates all the tox lint + unit tests targeted to utilize Python 3.8.

Python 3.5 and Python 3.6 have both been EOL now (3.5 has been EOL for more than a year). New Docker images utilize Python 3.8 and the same will be the case for new frozen binaries.

As soon as frozen binaries stuff is ready and merged we will also drop support for Python 2.7 which has been EOL for a long time now and it's making it very hard to keep things up to date since most libraries don't support that version anymore.

As you can see, this change also allows us to simplify CI a bit and in the future we will also be able to get rid tons of legacy compatibility related code.

## TODO

- [x] Remove Python 2 targets
- [ ] Update CodeSpeed benchmarks fixtures
- [x] Update requirements files and simplify them
- [x] Update smoke-k8s tests to utilize new Docker images


Kami:
Finally, all green :tada: 

But as discussed, we likely can't merge this PR until frozen binaries stuff is merged, ready and released.

PR #816, opened 2022-01-19
Allow debug log level to be set for subset of loggers

Kami:
## Background, Context

Today, user can enable debug log level for all the loggers using ``debug_level: [0,5]`` config option.

This sets debug log level for all the loggers. 

Problem is that a lot of debug loggers are very noisy and in a lot of scenarios, you only really want to debug subset of the agent (e.g. a specific monitor) so other debug logs will just result in noise.

## Proposed Implementation

This pull request allows user to only enable debug log level for a subset of loggers by utilizing new ``debug_level_logger_names`` config option. 

With this config option, user specifies for which loggers the debug level should be enabled.

For example:

```javascript
...
"debug_level_logger_names": ["scalyr_agent.copying_manager.worker", "scalyr_agent.builtin_monitors.linux_process_metrics(agent)"],
...
```

## Notes, Limitations

Right now, there is no easy / straight forward way to discover all the available logger names (beyond digging into the code).

Problem is that a lot of loggers are initialized lazy / when initially used which means we can't add ``scalyr-agent-2 --discover-logger-names`` or similar CLI flag since that will only discover loggers which are before that code is called.

One option is to enable debug log level for all the loggers (``debug_level``), find messages and modules we are interested in and then only enable it for those specific loggers.

## TODO

- [ ] Documentation (after the PR has been merged)


tests/unit/scalyr_logging_test.py
```diff
@@ -72,6 +73,63 @@ def test_formatter(self):
...
+            expression="XX Test debug 3", file_path=self.agent_debug_log_path
+        )
+
+        # 2. Now we enable debug level for 2 loggers only
+        scalyr_logging.set_log_level(
+            level=logging.INFO, debug_level_logger_names=["logger2", "logger3"]
+        )
+
+        logger1.info("YY Test info 1")
+        logger2.debug("YY Test debug 1")
```
ArthurKamalov, line 111: Should it be the `logger1` instead of 2?

tests/unit/scalyr_logging_test.py
```diff
@@ -72,6 +73,63 @@ def test_formatter(self):
...
+        logger2.debug("YY Test debug 2")
+
+        logger3.info("YY Test info 3")
+        logger3.debug("YY Test debug 3")
+
+        self.assertLogFileContainsLineRegex(expression="YY Test info 1")
+        self.assertLogFileContainsLineRegex(expression="YY Test info 2")
+        self.assertLogFileContainsLineRegex(expression="YY Test info 3")
+
+        self.assertLogFileContainsLineRegex(
```
ArthurKamalov, line 123: If logger1 is excluded, should we expect that the log file won't contain its debug message now?

ArthurKamalov: UPD: this one passes only because the `logger2` is used instead of logger1, so it seems that it might be done intentionally, but still not sure:)

ArthurKamalov (approved):
Just added small comments.

tests/unit/scalyr_logging_test.py
Kami: Good catch - thanks, will fix it.

tests/unit/scalyr_logging_test.py
Kami: Good catch - thanks!

Pushed a fix in a62504fa.

PR #818, opened 2022-01-20
Fix copy manager functionality so it works under Python >= 3.8 on OS X

Kami:
This pull request fixes copy manager functionality and related tests so they work on OS X when using Python >= 3.8.

In Python 3.8, local scoped objects can't be pickled anymore so the code fails. The solution is to use ``_SharedObjectManager`` class definition which is used for communication and is pickled by the multiprocessing framework to a top level scope.

Related links:

* https://gist.github.com/Kami/8386c79d2db7c95329e6c182ec639f49
* https://github.com/spack/spack/issues/14102

### DIFF
--- a/scalyr_agent/copying_manager/worker.py
+++ b/scalyr_agent/copying_manager/worker.py
@@ -1368,6 +1368,81 @@ _CopyingManagerWorkerSessionProxy = multiprocessing.managers.MakeProxyType(  # t
 class CopyingManagerWorkerSessionProxy(_CopyingManagerWorkerSessionProxy):  # type: ignore
     pass
 
+# NOTE: This class needs to be top-level in the module scope since Python >= 3.8 doesn't support
+# pickling local objects anymore and this object is indeed used for communication and is pickled.
+# - https://gist.github.com/Kami/8386c79d2db7c95329e6c182ec639f49
+# - https://github.com/spack/spack/issues/14102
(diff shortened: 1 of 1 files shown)


ArthurKamalov:
@Kami The copying manager-related test still fails. CPython switched to the spawn method from the fork in 3.8, I guess that's why `multiprocessing.Manager` can not work properly.  As a quick workaround, we can disable multiprocess workers to threaded ones as we did for Windows.

Kami:
@ArthurKamalov Thanks. Tests were passing locally after I made this change, but it looks like I missed the spawn change which is also needed.

For now, I will indeed likely looking into skipping those tests on OS X.

Kami:
@ArthurKamalov Pushed a change to also skip multiprocess copying manager worker tests under OS X.

On a related note, does the same issue also exists on Linux? If so, we will want to fix this so we can still use multiprocess functionality on Linux under Python >= 3.8.

ArthurKamalov:
> @ArthurKamalov Pushed a change to also skip multiprocess copying manager worker tests under OS X.
> 
> On a related note, does the same issue also exists on Linux? If so, we will want to fix this so we can still use multiprocess functionality on Linux under Python >= 3.8.

I pushed the commit which enables smoke and unit tests for python 3.8. The smoke test seems like working, but the unit test hasn't appeared in CircleCI for some reason.

ArthurKamalov (approved):
LGTM

PR #820, opened 2022-01-24
Add "logger_include_node_name" config option to K8s Open Metrics Monitor

Kami:
This config option allows us to enable debug level using new ``debug_level_logger_names`` global config option for all the monitor instances without needing to know the node name up front.


ArthurKamalov (approved):
LGTM

PR #821, opened 2022-01-24
Remove support for ujson in favor of orjson

Kami:
This pull request drops support for ``ujson`` in favor of more correct and performant ``orjson`` library.

We already utilize orjson in Docker images and we should also utilize it and bundle it by default with new frozen binaries (/cc @ArthurKamalov).


ArthurKamalov (approved):
LGTM

Kami:
I'll get that updated and in sync with master for inclusion with the next release (v2.1.32).

PR #822, opened 2022-01-25
Ingest memory profiling data by default when profiling is enabled

Kami:
This pull request updates the code so we ingest memory profiling dump / data by default when profiling is enabled (``enable_profiling`` configuration option).

We already do that by default for CPU profiling data, but we haven't do it for memory profiling data. This way the behavior should be consistent and it should also help with the end user experience.

NOTE: I also changed the parser name we attach to the profiler data - I checked the (server side code) and it doesn't seem we define any parser for this name on the server side, but in case I missed something, please let me know.

On a somewhat related note - in the future it may make sense for us to define server side parser / line join rule for those two parsers to provide a better user experience.


ArthurKamalov (approved):
LGTM

PR #823, opened 2022-01-25
Fix URL monitor handling of request data for POST requests under Python 3

Kami:
This pull request fixes URL monitor so it works correctly under Python 3 when the POST method is used and the request data is specified.

Previously under Python 3, user would see errors like this when specifying POST data:

```bash
'failed', extra_fields={'url': 'http://127.0.0.1:11505/200_post', 'status': 0, 'length': 0, 'unknown_error': 'POST data should be '}
```

In addition to that, I also included two other improvements:

* If the error message is truncated, indicate that with ``...`` and include 30 characters instead of 20 by default. Previously we didn't indicate that a message has been truncated which would result in "weird" errors such as``'POST data should be '``.
* On unknown errors, also log the whole traceback. This should help with troubleshooting (and unknown errors would indicate some kind of fatal error or a bug like in this case so it's not something which would get logged during normal use so it won't result in a log volume increase)

NOTE: Long term (after we drop support for Python 2) we should migrate all that home grown glue code for handling HTTP requests to the requests library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Docker images:
 * Upgrade Python used by Docker images from 3.8.10 to 3.8.12.
 * Base distribution version for non slim Alpine Linux based images has been upgraded from Debian Buster (10) to Debian Bullseye.
 
+Bug fixes:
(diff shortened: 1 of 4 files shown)


ArthurKamalov (approved):
LGTM

PR #828, opened 2022-02-02
Fix centos 8 ami

ArthurKamalov:
Switch centos8 ami to the vault repo source, due the distro's EOL


tests/ami/scripts/test_rpm.sh.j2
```diff
@@ -57,6 +57,16 @@ id
 echo_with_date ""
 echo_with_date "==========================="
 
+# If it is a centos 8, then change repo source to centos vault.
```
Kami, line 60: Please just add a comment on why we need to that - CentOS 8 now being EOL and repo not being active anymore.

Kami (approved):
Thanks for fixing this!

PR #829, opened 2022-02-04
Refactor package type logic

ArthurKamalov:
This PR mostly prepares already existing packages to have the same codebase as in the upcoming frozen binaries. All of that was part of the abandoned frozen binaries PR, but now it was decided it bring changes iteratively without breaking existing packages.
The main change here is to get rid of the `MSI_INSTALL` install type and make the Windows package work as a regular `PACKAGE_INSTALL` type. To determine the package type we use the `build_info` file now, but it is renamed to `install_info` and now it's a JSON where the previous actual build_info string is stored as one of the fields. The install type is stored in the `install_type` field. 
The existing build process of the widows installer also has been changed in order to work with this `install_info` file. The info file now is embedded into the frozen binary.

TODO: 

- [ ] Make windows event monitor work.  For some reason the windows event monitor has stopped working and I can not find any connection with new changes.


build_package.py
```diff
@@ -314,41 +299,117 @@ def build_win32_installer_package(variant, version):
...
+        "scalyr_agent.builtin_monitors.mysql_monitor",
+        "scalyr_agent.builtin_monitors.nginx_monitor",
+        "scalyr_agent.builtin_monitors.shell_monitor",
+        "scalyr_agent.builtin_monitors.syslog_monitor",
+        "scalyr_agent.builtin_monitors.test_monitor",
+        "scalyr_agent.builtin_monitors.url_monitor",
+        "scalyr_agent.builtin_monitors.windows_event_log_monitor",
+        "scalyr_agent.builtin_monitors.windows_system_metrics",
+        "scalyr_agent.builtin_monitors.windows_process_metrics",
+        "scalyr_agent.builtin_monitors.openmetrics_monitor",
```
Kami, line 326: May also need to add kubernetes_openmetrics_monitor module which was added recently. But that monitor is kinda specific to Kubernetes version of the agent so one option is to exclude it from the Windows bundle.

scalyr_agent/__scalyr__.py
```diff
@@ -155,13 +195,35 @@ def get_install_root():
...
+    else:
+        executable_path = __file__
+
+    # Resolve possible symlinks until we reach the real file.
+    while os.path.islink(executable_path):
+        executable_path = os.path.realpath(executable_path)
+
+    if __is_frozen__:
+        # In case if frozen binaries the executable has to be located in the '<install root>/bin', so we just
+        # go down until we reach the the install root.
```
Kami, line 216: I guess here it should be "go up" instead of "go down", since if I understand it correctly, we go to the parent directory of bin/, right?

scalyr_agent/agent_main.py
```diff
@@ -2183,34 +2208,33 @@ def stop(self):
...
+        # Create fully valid command line args so the Windows service could handle it properly.
+        argv = [sys.argv[0]]
+        argv.extend(other_argv)
+        platform_windows.parse_options(argv)
+        exit(0)
+
+    if command_args.command == "config":
+        # Add the config option. So we can configure the agent from the same executable.
+        # It has to save us from building multiple frozen binaries, when packaging.
+        config_main.parse_config_options(other_argv)
```
Kami, line 2232: I assume ``parse_config_options`` throws in case of errors / similar and we exit with non-zero and the line below will never be reached (`exit(1)`)?

tests/unit/fixtures/build_info/build_info
```diff
@@ -1,4 +1,4 @@
-Packaged by: jenkins@scalyr.com
-Latest commit: 7d4c4e2e94242ee25320a75c510d52967cfe50eb
-From branch: release
-Build time: 2020-05-06 17:59:21 UTC
+{
+    "build_info": "Packaged by: jenkins@scalyr.com\nLatest commit: 7d4c4e2e94242ee25320a75c510d52967cfe50eb\nFrom branch: release\nBuild time: 2020-05-06 17:59:21 UTC\n",
```
Kami, line 2: Since this file now contains a JSON object, it may make sense to split this built info into multiple fields instead of using e raw string - e.g. packaged_by, revision_sha, branch_name, build_time or similar.

This may give us more flexibility and make it a bit easier to parse out the commit sha and similar (and we can still re-constructor single raw string from those values, if needed).

Kami (approved):
A couple of small questions and one comment on the ``build_info`` field value.

Besides that, LGTM.

scalyr_agent/agent_main.py
ArthurKamalov: Yes, this function has its own argument parser that has to exit with non-zero. The line below is just to exit for sure if the parsing of the `scalyr-agent-2-config` arguments is successful.

scalyr_agent/__scalyr__.py
ArthurKamalov: Yeah, I guess that's more correct to have root at the top)

build_package.py
```diff
@@ -1035,15 +1099,15 @@ def build_base_files(base_configs="config"):
...
-    # create copies of the config_main.py with python2 and python3 shebang.
-    config_main_path = os.path.join(agent_source_root, "scalyr_agent", "config_main.py")
-    config_main_py2_path = os.path.join("scalyr_agent", "config_main_py2.py")
-    config_main_py3_path = os.path.join("scalyr_agent", "config_main_py3.py")
-    replace_shebang(config_main_path, config_main_py2_path, "#!/usr/bin/env python2")
-    replace_shebang(config_main_path, config_main_py3_path, "#!/usr/bin/env python3")
-    config_permissions = os.stat(config_main_path).st_mode
-    os.chmod(config_main_py2_path, config_permissions)
-    os.chmod(config_main_py3_path, config_permissions)
+    # # create copies of the config_main.py with python2 and python3 shebang.
```
Kami, line 1102: Was this commented out for testing purposes or it can be fully removed?

scalyr_agent/__scalyr__.py
```diff
@@ -90,6 +91,55 @@
...
+PACKAGE_INSTALL = 1  # Indicates source code was installed via a package manager such as RPM or Windows executable.
+TARBALL_INSTALL = 2  # Indicates source code was installed via a tarball created by the build_package.py script.
+DEV_INSTALL = 3  # Indicates source code is running out of the original source tree, usually during dev testing.
+
+
+def read_install_info():
+    """
+    Read the 'install_info' file that has to be located near this file. In opposite return empty dict.
+    """
+    install_info_path = os.path.join(os.path.dirname(__file__), "install_info")
```
Kami, line 104: Should we add .json suffix to the file? I guess it's just a matter of personal personal and doesn't matter much :)

scalyr_agent/__scalyr__.py
```diff
@@ -90,6 +91,55 @@
...
+    if not install_type:
+        return DEV_INSTALL
+
+    if install_type in ["package"]:
+        return PACKAGE_INSTALL
+    elif install_type == "tar":
+        return TARBALL_INSTALL
+
+    raise ValueError(
+        "Unrecognized package type {} in the 'build_info' file".format(install_type)
```
Kami, line 135: Perhaps also include actual value in the error message to make it easier / faster to troubleshoot?

agent_build/linux/scalyr-agent-2-config
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
Kami, line 1: That's here for backward compatibility reasons, right?

agent_build/linux/scalyr-agent-2-config
```diff
@@ -0,0 +1,5 @@
...
+
+set -e
+
+scalyr-agent-2 config "$@"
```
Kami, line 5: Depending on what we want to achieve, using ``exec`` may also be desired here to replace the script process.

agent_build/package_builders.py
```diff
@@ -405,13 +416,12 @@ def _build_frozen_binary(
 
         # Create package info file. It will be read by agent in order to determine the package type and install root.
         # See '__determine_install_root_and_type' function in scalyr_agent/__scalyr__.py file.
-        package_info_file = self._intermediate_results_path / "package_info.json"
+        install_info_file = self._intermediate_results_path / "install_info.json"
```
Kami, line 419: Here I see the file install_info has .json extension, but in other places is without extension. Or am I missing something?

Kami: EDIT: Sorry, I needed to refresh to see the latest changes.

scalyr_agent/__scalyr__.py
ArthurKamalov: Yes, I think adding suffixes should make it more clear and there's no significant reason to keep it without extension.

tests/ami/scripts/test_windows.ps1.j2
```diff
@@ -176,19 +176,6 @@ Log-Message ""
 Log-Message "native agent status output verified"
 Log-Message ""
 
-#
-# Verify build_info file exists
-#
-
-Log-Message ""
-Log-Message "Verifying build_info file exists"
```
Kami, line 184: I guess we don't need this check anymore since agent will fail to start if the install info file doesn't exist, right?

agent_build/linux/scalyr-agent-2-config
ArthurKamalov: Yes.  I think there's a high chance that some of the users use it during their upgrades or something else.

tests/ami/scripts/test_windows.ps1.j2
ArthurKamalov: Yes, And in the case of frozen binaries (and Windows MSI has frozen binaries), the `install_info.json` file is "baked" to the binary itself.

agent_build/package_builders.py
ArthurKamalov: Sorry, I pushed recently another commit where I addressed this comment https://github.com/scalyr/scalyr-agent-2/pull/829#discussion_r808848373

agent_build/linux/scalyr-agent-2-config
ArthurKamalov: Ah sure, that should be a better approach. 

I guess it's also worth changing it to `sh`. I don't think that we'll add something complex here, that is only supported by bash. I am also not sure that the config command may be somehow required in docker images, but if so, our alpine images also would require a POSIX shell.

Kami: Yeah, sh should be fine here since we are not doing anything complicated.

Kami (approved):
LGTM 👍

PR #830, opened 2022-02-07
Fix small config parse log regression

Kami:
This pull request fixes a small regression and makes sure we don't log any messages when ``status -v`` and other commands are used - we only want to log that message when using a regular entry point and starting the agent.


Kami:
Since we had similar small regressions in the past, I also added a small unit tests which will hopefully help catch similar issues in the future - 7f054878 (granted, it likely won't catch all of them since some of the messages are only produced based on specific config values).

ArthurKamalov (approved):
LGTM. Good check.

PR #831, opened 2022-02-15
Add support for non-UTC timestamps in the container log lines

Kami:
## Description

This pull request updates our internal date and time parsing library which is used to parse timestamps from the container logs (used by the Docker and Kubernetes monitor) to support non-UTC timestamps with custom time zones.

## Background, Context

This issue was reported by our user.

Before, we simply assumed everyone is using UTC on their servers / containers (which is indeed very common and best practice for various reasons - always have servers set to UTC timezone, always store date time values in UTC and then only perform UTC -> selected / local timezone conversion and formatting when data is being displayed).

## Implementation Details

In the past we performed multiple benchmarks (IIRC, we also have micro benchmarks for it) and it turned out that time parsing takes a non-trivial amount of time in high load situations (which is to be somewhat expected since that function is called on every single log line produced by the containers).

Because of that, we have multiple versions of performance optimized functions for that functionality. One version utilizes Python built-in functionality and the other one utilizes ``udatetime`` library.

udatetime is a 3rd party library with C extension and it's an optional dependency which doesn't come bundled with the agent.

Right now if the user doesn't have ``udatetime`` library installed and log lines contains a non UTC timezone information (udatetime supports time values with timezone information), this will likely add some overhead to the date time parsing. 

It likely won't be substantial, but it won't be non-trivial either (I will add micro benchmark when I get a chance, to quantify it).

Working and testing this change also uncovered some edge cases with some implementations where a fractional part (microseconds) hasn't been correctly preserved or handled. This has also been fixed.

Keep in mind that right now there is still an edge case udatetime implementations doesn't support - our custom functions support timestamps which are not valid and don't follow the spec and contain more than 6 fractional components. udatetime implements the spec and doesn't support that.

Depending if we think that's an issue or not, we can update our udatetime based implementation to also support that.

## TODO (longer term)

- [x] Bundle udatetime with Docker images
- [ ] Bundle udatetime with frozen binaries
- [ ] Once we use udatetime everywhere, remove bundled dateutil library which won't be needed anymore
- [ ] Decide if we should utilize udatetime or smth else (udatetime works, but it doesn't support Python 3.9 at this point and there may be better alternatives out there)

### DIFF
--- a/scalyr_agent/date_parsing_utils.py
+++ b/scalyr_agent/date_parsing_utils.py
@@ -22,7 +22,6 @@ if False:  # NOSONAR
     from typing import Optional
 
 import re
-import time
 import calendar
 import datetime
 
@@ -51,8 +50,6 @@ except ImportError:
     udatetime = None
 
 from dateutil.parser import isoparse
-
-from scalyr_agent.compat import custom_any as any
 
 
 if six.PY3:
--- a/scalyr_agent/date_parsing_utils.py
+++ b/scalyr_agent/date_parsing_utils.py
@@ -114,26 +114,7 @@ def _rfc3339_to_nanoseconds_since_epoch_strptime(string):
         return None
 
     nano_seconds = int(calendar.timegm(tm[0:6])) * 1000000000
-
-    nanos = 0
-
-    # now add the fractional part
-    if len(parts) > 1:
-        fractions = parts[1]
-        # if the fractional part doesn't end in Z we likely have a
-        # malformed time, so just return the current value
-        if not fractions.endswith("Z"):
-            # we don't handle non UTC timezones yet
-            if any(c in fractions for c in "+-"):
-                return None
-
-            return nano_seconds
-
-        # strip the final 'Z' and use the final number for processing
-        fractions = fractions[:-1]
-        to_nanos = 9 - len(fractions)
-        nanos = int(int(fractions) * 10**to_nanos)
-
+    nanos = _get_fractional_nanos(value=string)
     return nano_seconds + nanos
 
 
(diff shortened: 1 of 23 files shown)


Kami:
Per my comment above, I added a micro benchmark which shows the impact of parsing timestamps with timezones using dateutil (pay attention to the first table and ``*_udatetime``and ``*_string_split`` version).

Screenshot 2022-02-15 at 16 13 27


With tz parsing, string split version of the function which falls back to dateutil library is around 5x slower. 

Granted actual absolute timings are still low, but as mentioned above (and measured in the past), we could be calling this function many times per second.

Good news is that it doesn't have impact on ``udatetime`` which supports timezones out of the box. This means that once we move to the frozen binaries, we can get rid of the dateutil and other implementation in favor of udatetime everywhere.

Kami:
I decided to only keep string_split and udatetime versions of those functions. 

Those are only two functions which are used right now (and in the future we will have udatetime based one anyway). Other functions were used for benchmark purposes in the past.

This will allow us to simplify things, especially now that we are handling more formats which make it harder to implement in all the various implementations.

Kami:
Since this PR also includes "bundle udatetime with docker image by default" change, I will bake it on our environments for a while before release.

I don't expect surprises though since now udatetime supports all the scenarios which are supported by our native Python implementation (+timezones).

scalyr_agent/date_parsing_utils.py
```diff
@@ -14,15 +14,30 @@
 
 """
 Module containing various date parsing related utility functions.
+
+All the functions for parsing dates in RFC3339 format have been optimized for performance since
+they are parth of a critical code path when ingesting Docker / Kubernetes logs.
```
ArthurKamalov, line 19: typo?

ArthurKamalov (approved):
Nothing that catches the eye, LGTM.

tests/unit/util/date_utils_test.py
```diff
@@ -84,29 +240,33 @@ def test_rfc3339_to_datetime(self):
 
         s = "2015-08-06T14:40:56.123456"
         actual = scalyr_util.rfc3339_to_datetime(s)
-        self.assertEquals(datetime.datetime(2015, 8, 6, 14, 40, 56), actual)
+        self.assertEquals(datetime.datetime(2015, 8, 6, 14, 40, 56, 123456), actual)
 
-    def test_rfc3339_to_datetime_with_and_without_strptime(self):
+    @skipIf(not udatetime, "udatetime not available, skipping test")
+    def test_rfc3339_to_datetime_with_and_without_udatetime_compatibility_check(self):
         # Verify corectness between two different implementations of this function
```
ArthurKamalov, line 247: typo

tests/unit/util/date_utils_test.py
```diff
@@ -143,20 +295,25 @@ def test_rfc3339_to_nanoseconds_since_epoch(self):
...
         )
         actual = scalyr_util.rfc3339_to_nanoseconds_since_epoch(s)
         self.assertEquals(expected, actual)
 
-    def test_rfc3339_to_nanoseconds_since_epoch_with_and_without_strptime(self):
+    @skipIf(not udatetime, "udatetime not available, skipping test")
+    def test_rfc3339_to_nanoseconds_since_epoch_with_and_without_udatetime_compatibility_check(
+        self,
+    ):
         # Verify corectness between two different implementations of this function
```
ArthurKamalov, line 309: typo

ArthurKamalov (approved):

PR #833, opened 2022-02-21
Date parsing logging fix and related changes

Kami:

- Fix logging inside date_parsing_utils module so we use scalyr logging module so messages get correctly routed to the agent log file.

- Remove old alias import in scalyr_agent.util module which is not needed anymore to work around cyclic import issue (aka so we can import scalyr_agent.utils inside scalyr_agent.date_parsing_utils module).

- Update orjson since latest version includes some performance improvements and pre-built wheels for armv7 (https://github.com/ijl/orjson/blob/master/CHANGELOG.md#367---2022-02-14)


ArthurKamalov (approved):
LGTM, so, that means that we can enable orjson for our docker images right?:)

Kami:
@ArthurKamalov orjson dependency has already been added to Docker images in the past and should be used by default there since v2.1.26.

ArthurKamalov:
> @ArthurKamalov orjson dependency has already been added to Docker images in the past and should be used by default there since v2.1.26.

sorry that I haven't clarified, I meant for armv7.

Kami:
@ArthurKamalov 

> sorry that I haven't clarified, I meant for armv7.

Yeah, I believe so, but only on Debian based image. I will test / confirm it locally.

Kami:
I confirmed locally that now we can indeed use orjson on ARMv7 on Debian. Problem is that it still doesn't work on Alpine Linux and there isn't an environment marker (https://www.python.org/dev/peps/pep-0496/) which is specified enough which would allow us to include it on Debian, but not on Alpine when running on armv7.

Kami:
@ArthurKamalov I came up with this "workaround" - ee046ed6d327b50bc5b2f0508298e4009caaeb0f.

Hopefully in the future we will be able to get rid of it and simplify the build process once armv7 + musl wheels are available.

PR #836, opened 2022-02-22
Fix package tests

ArthurKamalov:
This PR adds other small fixes to the package tests to make them work after the `scalyr-agent-2-config` file was replaced with a shell script.

PS. Also disabled AMI test for centos 8 because of the bad vault repo availability.


ArthurKamalov:
PS: The "only master" will be returned to the CircleCi job before the merge.

Kami (approved):

PR #839, opened 2022-02-24
Fix package build info

ArthurKamalov:
This PR fixes the `set_build_info` function in the `build_package.py` script, so now it handles input as JSON. Before that, it accepted build info as a string. Our release pipeline relies on this function and keeps writing build info as a string, which leads to type error.

I Also reverted some last changes that add back build info file, because it seems that wasn't a root cause.

Another change is that now we use a stable install script in AMI tests even in internal Jenkins build. Before that, we used install script that downloaded packages from an internal repository.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2345,11 +2345,11 @@ workflows:
          requires:
            - package-test-deb
          <<: *master_and_release_only
-     - ami-tests-development:
-          context: scalyr-agent
-          requires:
-            - build-windows-package
-            - build-linux-packages-and-installer-script
+#     - ami-tests-development:
+#          context: scalyr-agent
+#          requires:
+#            - build-windows-package
+#            - build-linux-packages-and-installer-script
      - ami-tests-stable:
           context: scalyr-agent
           <<: *master_and_release_only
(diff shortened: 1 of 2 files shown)


Kami:
> Another change is that now we use a stable install script in AMI tests even in internal Jenkins build. Before that, we used install script that downloaded packages from an internal repository.

I think I don't 100% understand this part. Don't we want to intentionally use internal / development version of the installer script so we are using a correct bundled package? Or am I missing something?

ArthurKamalov:
> > Another change is that now we use a stable install script in AMI tests even in internal Jenkins build. Before that, we used install script that downloaded packages from an internal repository.
> 
> I think I don't 100% understand this part. Don't we want to intentionally use internal / development version of the installer script so we are using a correct bundled package? Or am I missing something?

Using an install script from a current revision (development) instead of a stable one is good because we also test changes in the script itself. But, in our current case, when we start internal release, we also use an "internal" version of the install script, which downloads packages from the internal repo.  I think the optimal way is still to use the install script from local revision but make it refer to the stable repo.

Kami (approved):

PR #842, opened 2022-03-02
Fix dynamic pod metadata updates in the Kubernetes monitor, additional debug level logging, new GHA based Kubernetes end to end tests

Kami:
## Description

This pull request fixes Kubernetes monitor to correctly handle dynamic pod metadata updates when Kubernetes cluster is using containerd container run time.

## Background, Context

When using containerd runtime, scalyr agent would not automatically pick up any pod metadata changes (e.g. adding or changing Scalyr related annotation) without the agent pod restart.

This issue only affected containerd runtime. When using Docker runtime, `ControlledCacheWarmed` abstraction which dynamically populates the cache in background is used.

## Proposed Implementation / Fix

This fix involves making sure we don't always use locally cached value - now when retrieving pod information from the ``KubernetesCache`` object, we will query Kubernetes API in case the cached entry is stale / expired.

An alternative would be utilize ``ControlledCacheWarmer`` abstraction to continuously update (warm) cache in the background for active pods, but that approach would be more involved so I decided to go with a simpler approach first.

In case it turns out that this approach has too big of a performance overhead (on the Kubernetes API or similar), we may need to consider ``ControlledCacheWarmer`` approach.

It's worth noting though that in aggregate, CPU used by both approaches should be the same - only difference is, that in theory, CCW approach should result in those Kubernetes API requests potentially being spread over longer time frame instead of happening "on demand" when inspecting pods inside the monitor main loop.

## Other (related) Changes

In addition to that, it includes the following changes:

1. Add additional debug level logging to the Kubernetes monitor. 

While troubleshooting the issue with original set of debug logs it was impossible to pin point down the root cause because the logs didn't contain enough information. This change should improve that.

2. Fix docker image build script on OS X.

On Monterey, port 5000 is already used by a system service meaning script would fail. This pull request fixes it by using a different port. If this becomes a bigger issue in the future, we may want to update the code to pick a random unused port.

3. Some docker image build docker fixes and info on how to use locally built images with local minikube cluster

4. Fixed usage of mutable default values for method argument in one places. I don't think there was an issue related to that because of they way that code has been used, but it's a bad practice and can cause issues.

5. New initial set of Kubernetes end to end tests running on Github actions (including regression test for bug fix in this PR)

### DIFF
--- a/.github/actions/install-k8s-agent-daemonset/action.yml
+++ b/.github/actions/install-k8s-agent-daemonset/action.yml
@@ -69,7 +69,7 @@ runs:
         kubectl apply -f "${{ inputs.daemonset_yaml_path }}"
 
         # Give it some time to start up
-        sleep 40
+        sleep 60
         kubectl -n scalyr get event
         kubectl -n scalyr get serviceaccount
         kubectl -n scalyr get pods
--- a/.github/actions/install-k8s-agent-daemonset/action.yml
+++ b/.github/actions/install-k8s-agent-daemonset/action.yml
@@ -50,7 +50,8 @@ runs:
           --from-literal=SCALYR_K8S_CLUSTER_NAME="${{ inputs.scalyr_cluster_name }}" \
           --from-literal=SCALYR_K8S_EVENTS_DISABLE="true" \
           --from-literal=SCALYR_K8S_VERIFY_KUBELET_QUERIES="false" \
-          --from-literal=SCALYR_SERVER="${{ inputs.scalyr_server }}"
+          --from-literal=SCALYR_SERVER="${{ inputs.scalyr_server }}" \
+          --from-literal=SCALYR_K8S_CACHE_EXPIRY_SECS="5"
 
         # Apply any additional custom resources
         if [ ! -z "${{ inputs.extra_yaml_paths }}" ]; then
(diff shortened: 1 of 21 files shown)


Kami:
To make testing easier without triggering the whole preview build, I will likely merge this into master to have images built and pushed to Docker Hub so those can be tested.

As discussed earlier, I think it would be good if we modified our "publish images" GHA workflow to also push images on each upstream push to non-master branch, but we should publish those images to a separate, development Docker Hub account and likely utilize the same tag (aka branch name) for each branch push to avoid new tag being pushed in each commit.

Kami:
For now, I decided to use out testing docker hub account and push images build for this branch there - https://hub.docker.com/r/test4scalyr/scalyr-k8s-agent/tags?page=1&ordering=last_updated.

Sadly this was quite painful - I encounter a bunch of "random" 401 docker hub errors and other issues. I have no idea why, since some image pushes succeeded and some didn't and we use exactly the same credentials for everything. It could be rate limiting or similar, but then I would expect something else than 401 🤷 

So to get that working I needed to get a bit creative and add some workarounds. I also needed to make update to the publish image script so it supports publishing subset of all the images (I went with a very simple change, if we wanted to use that functionality more frequently, we likely want to refactor and improve that change).

ArthurKamalov (approved):
LGTM. I just realized that I missed that the e2e test also could be implemented by extending the tests that I introduced with new docker images. Do you think that its possible to somehow combine them together to reduce redundancy?

Kami:
@ArthurKamalov Yeah, if we already have a good pattern in place for such tests, I think re-using it would make sense.

PR #843, opened 2022-03-08
Fix startup checkpointing bug related to ephemeral log files (/var/log/scalyr-agent-2/*.log) on Kubernetes deployments

Kami:
## Description

This pull request fixes a bug where we incorrectly preserved checkpoints related state for ephemeral logs in ``/var/log/scalyr-agent-2/`` across agent pod restarts (aka new pods being provisioned).

This issue meant that some of the internal agent log messages which are emitted during the early agent startup phase would be skipped (aka not ingested into Scalyr), because the agent would try to use old and stale checkpoint data which refers to the previous container state. 

This would cause the agent to detect invalid checkpoint state and ingest those files from the end (and not from the beginning as would be desired in this specific scenario).

## Background, Context

A while back, a change was made to the Scalyr Agent Kubernetes DaemonSet deployment to use a persistent mount for ``/var/lib/scalyr-agent-2`` directory inside the container to preserve checkpoints state across agent pod runs.

This was done intentionally to avoid ingesting duplicate container logs when agent pod is re-created.

An unintentional side affect of that change was to also re-use checkpoints for any other log files ingested by the agent and not just the container pod logs (which are persistent since they are stored on the host on which the agent DaemonSet deployment runs).

This includes all the logs inside ``/var/log/scalyr-agent/`` directory. This directory is ephemeral to the container which means it's gone after the container is stopped / deleted / similar.

Actual symptoms of that bug are that some of the internal logs produced by the agent during the start up phase are not ingested into Scalyr - luckily we already have a safe guard in place to start ingesting a file from the end in case invalid checkpoint state is found, but in this case the desired behavior which would ensure that no logs are skipped would be to ingest from the beginning in such scenario.

## Proposed Solution

This pull request fixes this bug by introducing a new ``ignore_checkpoints_on_startup_path_globs`` config option and setting the default value for that config option on Kubernetes deployments to ``["/var/log/scalyr-agent-2/*.log"]``.

This config option tells the agent checkpoints for which file paths (globs) should be ignored when initially reading and consolidating checkpoints on agent start up.

Technically, we could use a similar approach and not persist checkpoints data for specific globs on graceful shutdown, but graceful shutdown is not guaranteed which means we would still need to have the same logic in place on start up to ensure we actually skip checkpoints for those files.

## Notes

To get test Docker images built and published to DockerHub account, I needed to pull in some changes / fixes from #842.

### DIFF
--- a/docker/k8s-config/agent.json
+++ b/docker/k8s-config/agent.json
@@ -10,4 +10,11 @@
   // metrics on the container running the agent.
   implicit_metric_monitor: false,
   implicit_agent_process_metrics_monitor: false,
+
+  // Checkpoints file is persisted across pod restarts to ensure we don't
+  // emit duplicate lines for container logs on on pod restart, but we don't
+  // want to persist checkpoint data for ephemeral log files (aka log files
+  // which are not persisted across restarts and are ephemeral to the
+  // container - think anything in /var/log/scalyr-agent-2/*.log)
+  ignore_checkpoints_on_startup_path_globs: ["/var/log/scalyr-agent-2/*.log"],
 }
--- a/scalyr_agent/copying_manager/copying_manager.py
+++ b/scalyr_agent/copying_manager/copying_manager.py
@@ -25,6 +25,7 @@ import time
 import operator
 import signal
 import errno
+import fnmatch
 
 WORKER_SESSION_PROCESS_MONITOR_ID_PREFIX = "agent_worker_"
 
@@ -781,6 +782,8 @@ class CopyingManager(StoppableThread, LogWatcher):
                 # prepare and read checkpoints.
                 # read all checkpoints from the manager's previous run and combine them into one mater file.
                 checkpoints = self.__find_and_read_checkpoints(warn_on_stale=True)
+                checkpoints = self.__process_checkpoints_on_startup(checkpoints)
+
                 if checkpoints:
                     self.__consolidate_checkpoints(checkpoints)
                 else:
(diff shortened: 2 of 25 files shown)


Kami:
I will get test imaged build, pushed to test docker account and test them on staging.

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1401,6 +1404,40 @@ def _consolidated_checkpoint_file_path(self):
...
+                    return True
+
+            return False
+
+        checkpoint_file_paths = list(checkpoints.keys())
+
+        for file_path in checkpoint_file_paths:
+            if path_matches_any_glob(file_path=file_path):
+                log.log(
+                    scalyr_logging.DEBUG_LEVEL_0,
```
Kami, line 1433: Right now this is being logged under INFO, but I'm kinda torn between DEBUG and INFO here.

INFO may be a bit verbose and some users may get concerned by that log message, but in case we change it to DEBUG, it may make debugging any issues related to this functionality harder.

ArthurKamalov: If I'm not mistaken, that has to be only on startup, so it shouldn't generate a constant noise if we keep it as INFO

Kami: Correct, only once on startup (message per ignored log file), but it may still alert some users so dunno.

dariocalia: Recommend keeping it at INFO.  The changelog .md is pretty clear.  Should we update the readme.md?  I could not locate any content on this.

Kami: @dariocalia Thanks for the input. 

Not sure what exactly you mean with updating readme though - readme in this repo is mostly used for people interested in contributing to the project and / or using the source code of the agent directly.

So not sure what we should put in the readme with regards to the change / bug fix :) Or did you perhaps mean RELEASE_NOTES.md?

docker/k8s-config/agent.json
```diff
@@ -10,4 +10,11 @@
   // metrics on the container running the agent.
   implicit_metric_monitor: false,
   implicit_agent_process_metrics_monitor: false,
+
+  // Checkpoints file is persisted across pod restarts to ensure we don't
+  // emit duplicate lines for container logs on on pod restart, but we don't
+  // want to persist checkpoint data for ephemeral log files (aka log files
+  // which are not persisted across restarts and are ephemeral to the
+  // container - think anything in /var/log/scalyr-agent-2/*.log)
+  ignore_checkpoints_on_startup_path_globs: ["/var/log/scalyr-agent-2/*.log"],
```
ArthurKamalov, line 19: I guess there shouldn't be so many reasons for users to deal with k8s agent config file directly, right? I just thought about putting that addition on the code level for the k8s monitor, so we can keep the config as "blank" as possible as we have for the "regular" agent config.

ArthurKamalov (approved):
LGTM.
Just want to note the need to change the test docker hub account to prod when it's time to release, because it's really easy to miss/forget about it later (as I did, and had to push everything manually :sweat_smile:)

docker/k8s-config/agent.json
Kami: Correct - on K8s (docker images), user shouldn't be editing base config files directly - they should either utilize environment variables or a ConfigMap to configure the agent.

And the problem is that this change and config option is global and not specific to k8s monitor (so putting config option into k8s monitor may make it a bit confusing).

But yeah, in short, I also don't think there should ever be a need to utilize this config option outside of the k8s environment unless a user has some specific use case.

PR #847, opened 2022-03-29
Optionally include cluster name with all the Kubernetes Open Metrics monitor log lines

Kami:
This pull request includes a small changes which allows user to configure the new Kubernetes Open Metrics monitor to include Kubernetes cluster name with each of the emitted metrics log lines.

It follows the same approach as we already follow for the node name inclusion.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2142,6 +2142,8 @@ jobs:
       - run:
           name: Build MSI package
           command: |
+              # Make sure step fails on any command failure
+              $ErrorActionPreference = "Stop"
               $Env:WIX = ";C:\wix311;"
               C:\Users\circleci\AppData\Local\Programs\Python\Python38\python.exe -m pip install -r win32\windows-build-requirements.txt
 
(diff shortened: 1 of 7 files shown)


Kami:
@ArthurKamalov See my comments on Slack - I pushed a change which fixes the build and a change which fails that Circle CI job step in case creating an executable fails.

"Fix" was upgrading to the latest version of pyinstaller (4.10) since for some reason pre-built wheels for 4.2 we used are not available anymore. Not sure if they were ever available or we pre-built and cached them on CI. 

Would be good to double check that everything still works as expected with the latest pyinstaller version (and hopefully tests should catch most obvious issues).

I did check changelog - https://pyinstaller.readthedocs.io/en/stable/CHANGES.html.

ArthurKamalov (approved):
@Kami Thanks for handling that. I also looked at the changelog and it seems that the upgrade of the PyInstaller should be ok. I also couldn't trace how. the pyinstaller==4.2 had been installed but I think that the pre-compiled wheel had been used.

PR #855, opened 2022-04-14
Add a new internal testing monitor which allows us to ingest single file into multiple accounts

Kami:
This pull request adds a new "symlink_file_monitor".

This monitor is meant to be used a short term internal hacky workaround which allows us to ingest the same file into multiple DataSet accounts (by multiple workers).

It works by creating a symlink for a matching file path and then each worker process can use a symlink which is unique to them.

Example agent and monitor config which shows an example of how to ingest same file into DataSet:

```javascript
 logs: [
   {
      "path": "/tmp/file1*.txt",
      "copy_from_start": true,
   },
   {
      "path": "/tmp/worker-0-file1*.txt",
      "worker_id": "test_team_one",
      "copy_from_start": true,
   },
   {
      "path": "/tmp/worker-1-file1*.txt",
      "worker_id": "test_team_two",
      "copy_from_start": true,
   },
  ],

"monitors": [
        {
            module: "scalyr_agent.builtin_monitors.symlink_file_monitor",
            "input_path_glob": "/tmp/file1_*.txt",
            "symlink_file_name_format": "worker-{index}-{file_name}",
            "symlinks_count": 2,
            "sample_interval": 10,
        }
    ],

```

Long term, we want to update the agent to directly support this use case, but this will require more work. Right now the key in the checkpoints file is just the file path. If we wanted to support this functionality, we would need to change it to be (file_path, worker_id) or similar (and we would of course need to keep it backward compatible which makes it a bit more complicated).

It's worth noting that the file path won't be exactly the same, but that's not a problem for our use case.


ArthurKamalov (approved):
LGTM

PR #857, opened 2022-04-20
[WIP] Periodically re-read authentication token from token volume mount file on disk

Kami:
This pull request updates Kubernetes code (KubernetesApi, KubeletApi) to periodically re-read token which is used to authenticate from `/var/run/secrets/kubernetes.io/serviceaccount/token` volume mount file on disk.

## Background, Context

In recent versions, Kubernetes supports authentication tokens which are periodically refreshed (https://thenewstack.io/no-more-forever-tokens-changes-in-identity-management-for-kubernetes/).

In our agent code base, we used the same token value for the whole duration of the life time of the agent process (Kubernetes monitor, https://docs.kubermatic.com/kubermatic/v2.20/architecture/concept/kkp-concepts/service_account/service_account_token_projection/) which doesn't play along nicely with the refresh functionality.

## Proposed Implementation

There are two ways to support tokens which are periodically refreshed / rotated.

1. Rely on "Service Account Token Volume Projection" functionality and periodically re-read token value from the file on disk (service account token volume projection will take care of updating that volume mount file content)
2. Directly talk to the Kubernetes API and handle refreshing the token ourselves

In this PR, I went with the approach 1) since it simpler and it should be drop in for the current approach - new code will now periodically re-read token file from disk and in case the value has changed, new value will be used for subsequent requests.

Per recommendation in the Kubernetes docs, we re-read the value from the file every 5 minutes - https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/.

> The application is responsible for reloading the token when it rotates. Periodic reloading (e.g. once every 5 minutes) is sufficient for most use cases.

If this ever becomes a problem, we can lower that value (the interval can also be changed using ``k8s_re_read_token_interval`` config option). An option also is to try to re-read the token in case we get 401 back, but I decided to go with the simpler approach, especially since official documentation says that re-reading the token every X minutes should be sufficient for most cases.

It's also worth noting that with the this approach, there doesn't appear to be any easy way to retrieve token TTL directly so we could use a smarter logic to determine when to re-read the token.

### DIFF
--- a/tests/unit/builtin_monitors/kubernetes_monitor_test.py
+++ b/tests/unit/builtin_monitors/kubernetes_monitor_test.py
@@ -22,6 +22,7 @@ import re
 import time
 import warnings
 import platform
+import tempfile
 from string import Template
 from collections import OrderedDict
 
@@ -57,6 +58,7 @@ from scalyr_agent.test_base import ScalyrTestCase, BaseScalyrLogCaptureTestCase
 from scalyr_agent.test_util import ScalyrTestUtils
 
 from tests.unit.monitor_utils.k8s_test import FakeCache
+from tests.unit.monitor_utils.k8s_test import KubernetesApi
 from tests.unit.configuration_test import TestConfigurationBase
 from tests.unit.copying_manager_tests.copying_manager_test import FakeMonitor
 from scalyr_agent.test_base import skipIf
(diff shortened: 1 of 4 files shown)


tests/unit/builtin_monitors/kubernetes_monitor_test.py
```diff
@@ -1267,3 +1271,195 @@ def mock_k8s_cache_pod(
...
+        self.assertEqual(k8s.token, "token1")
+
+        # Update the token
+        with open(token_file_path, "w") as fp:
+            fp.write("token2")
+
+        self.assertEqual(k8s.token, "token1")
+
+        # 2. After the token has been updated, the code should eventually re-read a new token
+        time.sleep(1.2)
```
ArthurKamalov, line 1310: I guess there're no easy, non-invasive ways to make API instances read tokens with a smaller interval. Just in case we can reduce the overall time for testing :)

tests/unit/builtin_monitors/kubernetes_monitor_test.py
```diff
@@ -1267,3 +1271,195 @@ def mock_k8s_cache_pod(
...
+            k8s_token_re_read_interval: 1,
+          }
+        """
+            % (token_file_path)
+        )
+        global_config = self._create_test_configuration_instance()
+        global_config.parse()
+
+        session = requests.Session()
+        adapter = requests_mock.Adapter()
```
ArthurKamalov, line 1334: Great lib, haven't used it before :)

ArthurKamalov (approved):
LGMT

tests/unit/builtin_monitors/kubernetes_monitor_test.py
Kami: Yeah, I was actually thinking about the same think, but then decided to not do it since it's only couple of seconds.

But since you also mentioned it, I decided I will make the change - I will change the option type from int to float and use a shorted interval (e.g. 0.1s).

Kami: EDIT: I just tested it locally and we don't save more than 2 seconds - I still need to keep the sleep to be at least 0.5-0.6 seconds to avoid false positives.

One option would be to replace sleep with counter based mock approach, but probably an overkill for now since it only adds 4 seconds to the test run.

ArthurKamalov: I don't see any problem with 4 seconds, especially with the fact that we don't add tests with such sleep so often.

PR #863, opened 2022-04-28
CT-523

joseph-makar:
From [CT-523](https://scalyr.myjetbrains.com/youtrack/issue/CT-523):

This change arranges for server responses that cannot be parsed as json (ie parseResponseFailed from ScalyrClientSession.__receive_response):
- To be logged as info instead of error (from the ticket, perhaps warning is more appropriate?)
- And to be retried (the same way as done for requestTooLarge)
  - If the retries continue to fail until its time limit (CopyingManagerWorkerSession.__config.max_retry_time) it is logged as an error and the events are dropped

This feels like a bit of a hack to me (both the implementation (is there a cleaner, alternative approach here?) and the intention) but I totally understand the desire to avoid alarming the customer with error logs.
Ideally I think it's better to focus the effort on making the nginx config more amenable to deployments or adapting the deployments to be better staggered, but I defer to your opinions @Kami & @oliverhsu77 

Also I don't see why this smoke test failed on the second commit: https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/8678/workflows/292fd70f-26e3-424c-9cf1-0893ccb8befd/jobs/111937
I would just rerun it but I don't have the permissions to do so.


scalyr_agent/scalyr_client.py
```diff
@@ -1187,6 +1187,10 @@ def __init__(
         # Used to record some performance timing data for debugging/analysis
         self.__timing_data = dict()
 
+        # Used to cache response status (from ScalyrClientSession.__receive_response)
+        # which is then used for special handling of specific timed-out requests
+        self.__receive_response_status = None
```
Kami, line 1192: Hm, I don't see this variable being used anywhere. Or did I miss something?

scalyr_agent/copying_manager/worker.py
```diff
@@ -489,6 +489,14 @@ def run(self):
                         # on the ground and advance.
                         if current_time - last_success > self.__config.max_retry_time:
                             if self.__pending_add_events_task is not None:
+                                if (
+                                    "parseResponseFailed"
+                                    in self.__pending_add_events_task.__receive_response_status
+                                ):
+                                    log.error(
+                                        "Repeatedly failed to parse response due to exception.  Dropping events",
```
Kami, line 497: I thought it may not be a bad idea to also log partial response here, but I believe we will already log this here (https://github.com/scalyr/scalyr-agent-2/pull/863/files#diff-4cb9befeecb4697d270768d8487b47aa20810c540fa7a9b6c86800735e7927a2R704) so it's not needed to do it here again?

Kami:
> Also I don't see why this smoke test failed on the second commit: https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/8678/workflows/292fd70f-26e3-424c-9cf1-0893ccb8befd/jobs/111937
I would just rerun it but I don't have the permissions to do so.

Yeah, looks like an intermediate / upstream error and re-run should fix it.

For Circle CI access, can you please open internal IT ticket to get access.

Kami:
> Ideally I think it's better to focus the effort on making the nginx config more amenable to deployments or adapting the deployments to be better staggered, but I defer to your opinions @Kami & @oliverhsu77

Yeah, we have an open issue for that, but sadly no ETA.

scalyr_agent/scalyr_client.py
```diff
@@ -700,7 +700,7 @@ def __receive_response(self, body_str, send_time):
                 response_as_json = scalyr_util.json_decode(response)
             except Exception:
                 # TODO: Do not just catch Exception.  Do narrower scope.  Also, log error here.
-                log.error(
+                log.info(
```
Kami, line 703: Yeah, for now it's probably better to start with ``warning`` (makes it easier for us to find those log messages).

In case people still find it confusing, we can then change it to info.

Kami (approved):
LGTM.

I will go ahead and merge this into master so the change can be included in the upcoming release.

Kami:
I noticed there are other places where we use ``ERROR`` log level for non-fatal errors which are retried - https://github.com/scalyr/scalyr-agent-2/blob/23934609b65d0af3bc1d78fd9a15e6f433b17702/scalyr_agent/scalyr_client.py#L581.

In the future, we should probably also update those to use warning level or similar.

scalyr_agent/copying_manager/worker.py
joseph-makar: Yes, I wanted to be explicit that the repeated failures were causing events to be dropped.  But I don't have a strong opinion & I'm ok with removing it if that seems better.

Kami: Sorry, my comment was a bit confusing.

I do think that having an error log message here is a good idea. 

I just thought it may also be a good idea to include actual server response here as well (e.g. ``scalyr_util.remove_newlines_and_truncate(response, 1000)``), but I actually think it's not needed since it will already be logged in the client code under warning / info.

PR #864, opened 2022-04-28
Add extra logging for troubleshooting duplicate log ingestion.

ArthurKamalov:


scalyr_agent/util.py
```diff
@@ -2203,13 +2204,14 @@ def get_agent_start_up_message():
         date_parsing_library = "native python"
 
     msg = (
-        "Starting scalyr agent... (version=%s) (revision=%s) %s (Python version: %s) "
+        "Starting scalyr agent... (version=%s) (revision=%s) %s (hostname: %s) (Python version: %s) "
```
Kami, line 2207: Since we are already changing that, I will also go ahead and change the format to foo=bar instead of foo: bar since the latter is not handled by our agent parser (that was an oversight a while back).

scalyr_agent/configuration.py
```diff
@@ -701,6 +701,7 @@ def print_useful_settings(self, other_config=None):
             "default_sessions_per_worker",
             "default_worker_session_status_message_interval",
             "enable_worker_session_process_metrics_gather",
+            "ignore_checkpoints_on_startup_path_globs",
```
Kami, line 704: I'll remove this for now under Python 2.7 since there is a bug with JsonArray under Python 2.7 which is pain to fix and not worth doing it since we will drop support for 2.7 in the future anyway.

Kami (approved):

PR #867, opened 2022-05-06
fix: handling python > 3.9

askalski85:
Set minimum Python2 to 2.7.

This replaces https://github.com/scalyr/scalyr-agent-2/pull/862, the common code is not extracted to the separate file, because preinstall script can't use other scripts. I don't want to allow postinstall script to run pre install script as its not natural.

Giving a check-python name for the pre install script is also wrong, as this script is meant to do more than just check python (now cleans up scalyr from /usr/share)

Lets unblock Ubuntu 22.04 users with this quick fix.

### DIFF
--- a/installer/scripts/postinstall.sh
+++ b/installer/scripts/postinstall.sh
@@ -13,24 +13,43 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Requirements
+MIN_REQ_PYTHON2_VERSION="2.7"
+MIN_REQ_PYTHON3_VERSION="3.5"
+
 # Used below to execute a command to retrieve the Python interpreter version.
 run_and_check_persion_version() {
-  command=$1
-  version=$(${command} 2>&1 | grep -o "[0-9].[0-9]")
+  local command=$1
+  local version=$(${command} 2>&1 | grep -Eo "[0-9](.[0-9]+)+")
   exit_code=$?
 
   # shellcheck disable=SC2072,SC2071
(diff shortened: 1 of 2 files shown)


Kami (approved):
LGTM, although I would feel more comfortable with tests.

installer/scripts/postinstall.sh
```diff
@@ -56,26 +75,26 @@ check_python_version() {
   # Verify that a suitable Python version is available and set up the agent symlinks
   if is_python_valid python; then
     # 'python' command exists
-    echo "The Scalyr Agent will use the default system python binary (/usr/bin/env python)."
-    /usr/share/scalyr-agent-2/bin/scalyr-switch-python default
+    echo "+ The Scalyr Agent will use the default system python binary (/usr/bin/env python)."
+  #  /usr/share/scalyr-agent-2/bin/scalyr-switch-python default
```
Kami, line 79: Is this change (comment) intentional?

Kami:
I've done basic manual testing locally and things appear to be working fine.

For now, I will merge it, but I will work with @ArthurKamalov to make sure we also add some tests for this functionality when we refactor the duplication.

Kami:
Merged into master via #868. Thanks for the contribution.

Kami:
Just an update from other PR - https://github.com/scalyr/scalyr-agent-2/pull/868#discussion_r866685549.

Sadly I needed to revert that PR until we fix all the test case and make sure all the edge cases are handled correctly to ensure that everything works correctly and there are no inadvertent regressions.

PR #868, opened 2022-05-06
Fix handling of Python >= 3.10 in rpm/deb package post install script

Kami:
This PR includes changes from #867.

I opened a new PR so we can make sure all the tests run and pass.

### DIFF
--- a/installer/scripts/preinstall.sh
+++ b/installer/scripts/preinstall.sh
@@ -18,29 +18,46 @@
 # In this cases script have to exit with an error.
 # This is important because all agent scripts rely on '/usr/bin/env python' command.
 
+# Requirements
+MIN_REQ_PYTHON2_VERSION="2.7"
+MIN_REQ_PYTHON3_VERSION="3.5"
 
 echo "Checking Python version."
 
 is_python_valid() {
-  command=$1
-  version=$(/usr/bin/env "${command}" --version 2>&1 | grep -o "[0-9].[0-9]")
+  local command=$1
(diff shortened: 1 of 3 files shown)


installer/scripts/preinstall.sh
```diff
@@ -18,29 +18,47 @@
...
+      return 1
+      ;;
+  esac
+
+  echo "+ Found Python $version [$command]"
+  return 0
 }
 
-if ! is_python_valid python && ! is_python_valid python2 && ! is_python_valid python3; then
+if ! is_python_valid python && ! is_python_valid python3 && ! is_python_valid python2; then
```
Kami, line 61: This changed the order in which we check the versions which caused some tests to fail on master.

Kami: @askalski85 Sadly merging this PR into master caused all kinds of issues and end to end tests to fail. 

I'm still digging in, but it looks like this change is problematic - it changes the order of Python versions we prefer. With this change we now prefer Python 3 over Python 2.

Is this change needed to get the agent to work on Ubuntu 22.04? If it's not, I would prefer to revert it to avoid the surface area of this change.

/cc @ArthurKamalov

askalski85: the preference is not needed but natural. The preference can be easly reverted though.

askalski85: updated my commit with the changelog and original order python2 python3

askalski85: https://github.com/scalyr/scalyr-agent-2/pull/871

PR #869, opened 2022-05-06
Upgrade paramiko, libcloud dependency used by the AMI tests

Kami:
This pull request upgrades paramiko and libcloud dependency used by the AMI tests.

New version of paramiko supports newer public key algorithms (e.g. Ubuntu 22.04) and the WIP version of Libcloud includes a workaround / fix so the deployment code still works correctly with older OpenSSH versions which don't support SHA2 based algorithms (e.g. Ubuntu 14.04).

This is needed due to the breaking change in paramiko >= 2.9.0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -695,7 +695,7 @@ commands:
           at: /tmp/workspace
       - restore_cache:
           name: "Restore Python Dependencies cache"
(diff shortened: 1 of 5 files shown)


Kami:
@ArthurKamalov Still work in progress, but I hope this will allow us to run AMI tests against 14.04 and 22.04.

Kami:
Confirmed that with those changes AMI tests now work under Ubuntu 14.04 and Ubuntu 22.04.

Tests will of course fail on 22.04 until the Python version issue post install bug is fixed.

ArthurKamalov (approved):
LGTM

PR #870, opened 2022-05-06
Fix pre / post install scripts for Python >= 3.10

ArthurKamalov:
Fixes and changes on top of #868.


Kami:
Comment from the other PR - this change also changes the default behavior - new version of the script now prefers Python 3 over Python 2.

I think that change should be fairly safe (and long term we only want to support Python 3), but I just wanted to make sure there are no surprised / inadvertent regressions hiding in there.

scripts/circleci/run-ami-tests-in-bg.sh
```diff
@@ -51,6 +51,7 @@ if [ "${TEST_TYPE}" == "stable" ]; then
   #python tests/ami/packages_sanity_tests.py --distro=WindowsServer2019 --type=install --to-version=/tmp/workspace/ScalyrAgentInstaller.msi &> outputs/WindowsServer2019-install.log &
 
   # Tests below utilize installer script to test installing latest stable version of the package
+  python tests/ami/packages_sanity_tests.py --distro=ubuntu2204 --type=install --to-version=current --installer-script-url="${INSTALLER_SCRIPT_URL}" &> outputs/ubuntu2204-install.log &
```
Kami, line 54: We will only be able to enable Ubuntu 22.04 tests for stable version once next agent release with this bug fix is out :)

So for now, we can only run "fresh install from deb package" tests - sadly I don't think "upgrade" test will work either since it will try to upgrade from current stable version which is broken.

Kami: Looks like this change worked - [c027458](https://github.com/scalyr/scalyr-agent-2/pull/870/commits/c0274584327c3760fa9cb51ebb965d0a4de32ed4) (https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/8754/workflows/f6bdcc39-68b0-42b9-bf5a-8434279a2b49/jobs/113205).

> Job finished successfully: "bash /home/circleci/.pyenv/libexec/pyenv-exec python tests/ami/packages_sanity_tests.py --distro=ubuntu2204 --type=install --installer-script-url=/tmp/workspace/installScalyrAgentV2.sh --to-version=/tmp/workspace/scalyr-agent-2.deb ".

We can re-enable stable + upgrade tests once new release with bug fix is out.

.circleci/config.yml
```diff
@@ -1968,9 +1969,9 @@ jobs:
       - image: cimg/python:3.6
     steps:
       - package-test:
-          distribution: "ubuntu1804"
```
Kami, line 1971: Hm, do we still want to keep 18.04 tests or not?

tox.ini
```diff
@@ -245,13 +245,13 @@ commands =
     # NOTE: We use short traceback formatting since we run py.test inside py.test which results in hard to read output
     py.test tests/distribution/python_version_change_tests/ubuntu1604 -s -vv --tb=short --durations=5 {posargs}
 
-[testenv:agent_distributions_tests_ubuntu1804]
+[testenv:agent_distributions_tests_ubuntu2204]
```
Kami, line 248: Same here - were those tests intentionally removed?

Kami (approved):
LGTM, thanks for adding additional tests 👍

PR #871, opened 2022-05-06
fix: handling python > 3.9

askalski85:
Set minimum Python2 to 2.7.

Follow up on https://github.com/scalyr/scalyr-agent-2/pull/868#discussion_r866809725


Kami:
@askalski85 Just wanted to give you quick update - we added new tests and managed to get all the existing tests to work in #870.

We hope to have a new release with this fix out this week (but it will take a couple of days since we need to follow normal release process which includes running new version on our infrastructure for a while to ensure there are no regressions).

Thanks for the patience.

PR #880, opened 2022-05-11
Fix no check remote regression

Kami:
This pull request fixes a small regression which caused the agent to never perform connectivity check on startup.

I believe this regression was inadvertently introduced in v2.1.29 via 1d1ce997e6ed23bb30e13cc7a2eacc2ca3bef60c.

With the change in the command parser which is used, the default value for that option is now ``False``. This means that the code would always end up in that if branch and set ``no_check_remote`` to ``True``.

## TODO

- [ ] Tests


Kami:
Note, this is more involved (looks like multiple issues related to this bug and multiple changes relied on this broken behavior) so we should work on it post v2.1.30 release.

ArthurKamalov (approved):
LGTM

PR #883, opened 2022-05-13
Update re-attempt log entries to be warnings

joseph-makar:
From [CT-540](https://scalyr.myjetbrains.com/youtrack/issue/CT-540):
- Change re-attempt log entries to be warnings
- Note there are other re-attempt entries in both scalyr_agent/scalyr_client.py and scalyr_agent/connection.py but they are currently log.exceptions() (untouched)


Kami (approved):
LGTM 👍

PR #886, opened 2022-05-20
Refactor Docker image builders

ArthurKamalov:
Refactor existing build steps that build base images from the agent docker image.


.github/actions/execute-builder/index.js
```diff
@@ -0,0 +1,174 @@
+// Copyright 2014-2021 Scalyr Inc.
```
ArthurKamalov, line 1: This file seems like new, but it was renamed and haven't changed much.

ArthurKamalov:
@Kami 

it seems like a massive amount of changes, but i just tried logically rearrange the existing code.

Now all package builders and its steps are located in the [agent_builders.py](https://github.com/scalyr/scalyr-agent-2/blob/arthur-rs/agent_build/agent_builders.py)
The implementation of the builders and steps now in the [tools/builder.py](https://github.com/scalyr/scalyr-agent-2/blob/arthur-rs/agent_build/tools/builder.py).

Dockerfiles were also refactored and all major commands such as rust installation and etc were moved to a separate scripts.

PR #887, opened 2022-05-25
Add support for defining arbitrary attributes which are included with every metric in the Open Metrics monitor

Kami:
This pull request updates Kubernetes Open Metrics monitor and allows user to define arbitrary attributes (key value pairs) which are included with every metric log line as part of labels / extra fields.

Attributes are defined using pod annotations. The actual value needs to be a serialized JSON string where a key is a string and a value is a string (complex / nested objects are not supported).

This comes handy in scenarios where user wants to add additional context (e.g. app name) to the metric to make querying easier.

Example configuration:

```yaml
...
    template:
        metadata:
        labels:
            app.kubernetes.io/component: exporter
            app.kubernetes.io/name: node-exporter
        annotations:
            prometheus.io/scrape:                          'true'
            prometheus.io/port:                            '9100'
            k8s.monitor.config.scalyr.com/scrape:          'true'
            k8s.monitor.config.scalyr.com/scrape_interval: '120'
            k8s.monitor.config.scalyr.com/scrape_timeout:  '5'
            k8s.monitor.config.scalyr.com/attributes:      '{"app": "${pod_labels_app}", "env": "eu"}'
...
```

## TODO (future)

- [x] Ability to use template syntax and reference other values from the pod metadata (e.g. pod name, metadata, labels, etc.).

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
@@ -81,7 +81,7 @@ for the exporter pod:
             k8s.monitor.config.scalyr.com/scrape:          'true'
             k8s.monitor.config.scalyr.com/scrape_interval: '120'
             k8s.monitor.config.scalyr.com/scrape_timeout:  '5'
-            k8s.monitor.config.scalyr.com/attributes:      '{"app": "${pod_labels_app}", "region": "eu"}'
+            k8s.monitor.config.scalyr.com/attributes:      '{"app": "${pod_labels_app}", "instance": "{pod_labels_app.kubernetes.io/instance}", "region": "eu"}'
         spec:
         containers:
         - args:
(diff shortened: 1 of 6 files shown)


scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
```diff
@@ -384,9 +387,9 @@
 )
 
 # Annotation related constants
-PROMETHEUS_ANNOTATION_SCAPE_PORT = "prometheus.io/port"
```
Kami, line 387: not every annotation wears a (s)cape :P

scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
```diff
@@ -151,21 +157,18 @@
 This monitor will only work correctly if the agent is deployed as a DaemonSet. That's because each
 monitor instance will only discover metrics endpoints which are local to the node the agent is
 running on.
-
-## TODO
-
-- [ ] Should we include node name (+cluster name - scalyr_k8s_cluster_name ?) with each metric? We
```
Kami, line 157: This TODO is not relevant anymore since it has been implemented.

ArthurKamalov (approved):
LGTM

Kami:
@ArthurKamalov 

> Shouldn't we also add upper case (A_Z)?

It's using re.IGNORECASE flag so A-Z is also supported - https://github.com/python/cpython/blob/main/Lib/string.py#L67

> And why there's an extra ]*?

I simplified the regex further (f16361fabb039bcd99dbb5c9ffe34810b0d1b122), so it's almost the same as upstream now (+ the additional characters we support) - https://github.com/python/cpython/blob/main/Lib/string.py#L77.

ArthurKamalov:
> @ArthurKamalov
> 
> > Shouldn't we also add upper case (A_Z)?
> 
> It's using re.IGNORECASE flag so A-Z is also supported - https://github.com/python/cpython/blob/main/Lib/string.py#L67

Ah, got it.

> > And why there's an extra ]*?
> 
> I simplified the regex further ([f16361f](https://github.com/scalyr/scalyr-agent-2/commit/f16361fabb039bcd99dbb5c9ffe34810b0d1b122)), so it's almost the same as upstream now (+ the additional characters we support) - https://github.com/python/cpython/blob/main/Lib/string.py#L77.

Great. But for final clarification, did `]*` served some purpose and it was also simplified, or it was just a typo? :)

Kami:
@ArthurKamalov 

> Great. But for final clarification, did ]* served some purpose and it was also simplified, or it was just a typo? :)

It served no purpose (afaik) :)

PR #890, opened 2022-05-26
 Escape special metric field names which are considered special / reserved

Kami:
This pull request updates ``emit_value()`` agent logger code which is used by all the monitors to emit metric values to escape / sanitize metric field names which are considered special / reserved (logfile, monitor, metric, value, instance, severity).

This is needed to avoid collisions with special / reserved event attribute names which are set by the server side agent metrics parser code and are needed for various tsdb queries to work correctly.

Those reserved field names are sanitized by suffixing them with `_` (e.g. ``metric`` becomes ``metric``). This suffix was chosen for consistency with the existing server-side parsing behavior (in some situations server side parsing code adds `_` to original field values, e.g. `timestamp_`, ``message_``, etc.).


ArthurKamalov (approved):
LGTM

PR #891, opened 2022-05-30
Client side rate calculation for counter metrics

Kami:
## Description

This pull request implements client (agent) side per second rate calculation for "counter" metric types (metric types which values increase over time).

Having this functionality done on the client and not on the server will give us more flexibility and also works better with the new future "stateless" ingestion API.

## Proposed Implementation

I went with an implementation where per second rate is calculated inside the ``emit_value()`` method where metric values are emitted. This implementation is the simplest and doesn't require any large refactoring.

---

If in the future we will want to perform more client side aggregations and transformations we may want to change and refactor the code then - perhaps also do that in another background thread so we don't "block" writing metrics to monitor log file on disk in case a lot of calculations need to be done (but this could also be problematic and we would need to decide how and when to handle writing rate metrics to a file on disk - ideally we want them written to the disk in the same monitor log file and ingested as close as possible to the original metric value from which the derivatives are calculated).

### Notes on Edge Cases

Here is a list of the edge cases which are currently handled.

#### 1. We don't have previous value yet

This happens on initial monitor start up and / or restart where we don't have a previous value yet. In such case, we can't perform a rate calculation yet so we simply skip calculating rate.

#### 2. Current timestamp <= previous timestamp

If current collection timestamp is smaller or equal than previous one, we don't perform any calculation since this would result in an invalid value and division by zero.

Keep in mind that such scenario likely indicates bug / developer error (invalid timestamp being passed to the ``emit_value()`` method or similar).

#### 3. New metric value is < previous value

Rate is calculated for counter metrics which values should be increasing over time. If a new value is smaller than the previous one, we don't calculate the rate for that value. We only calculate the rate once a new value which is larger than the previous one is available.

#### 4. Time delta between the current and previous collection timestamp is too large

If time delta between previous metric collection value timestamp and current collection value timestamp is too large, we ignore the previous collection value and don't calculate the rate for it.

I used the same threshold value which is currently used on the server side - 11 minutes.

---

Most of those edge cases are handled in the similar manner as they are currently handled in the backend / server side code, but now we also have the opportunity to change and improve any of that behavior if we think that is justified and it makes sense.

## Configuration

Right now we only calculate per second rates for metrics which are specified in the ``calculate_rate_metric_names`` config option.

To begin with, we will likely define default value for this config option ourselves (which may be deployment dependent - thing Docker / Kubernetes vs non Docker deployment).

## Open Questions / To Decide

- [x] Should we persist "metadata" which is needed to calculate the rates (previous metric value, previous timestamp) on disk? This would allow us to pick up where we left up on agent / monitor restart or similar, but it does increase the complexity a bit and may not be needed since we already have an upper bound on previous metric value (11 minutes) before we consider it stale and don't use it for future metric calculations. In most cases, sample collection interval is 60 seconds so losing a rate value for one or two samples is probably not a problem (?).

## TODO

- [x] Allow user to define for which Kubernetes Open Metrics monitor metrics rates should be calculated using pod annotations.
- [x] Include monitor name in the internal value cache to avoid conflicts where the are the same metric names used by multiple monitors.

## TODO (future)

- [ ] Detect "counter" metric types and automatically "enroll" those metrics into client / agent side rate calculation. This can be done for monitors such as Open Metrics one where we know metric schema / type. For now, we probably don't want to do that for every single Open Metrics monitor counter metric type since there are many of those and this could include non-trivial amount of overhead. For the time being, it will be an opt-in functionality (we will likely define default values as part of the standard agent config).

### DIFF
--- a/scalyr_agent/scalyr_logging.py
+++ b/scalyr_agent/scalyr_logging.py
@@ -75,11 +75,14 @@ DEBUG_LEVEL_5 = 4
 
 # Stores metric values and timestamp for the metrics which we calculate per second rate on the
 # agent.
-# Maps <monitor name short hash>.<metric name> to a tuple (<timestamp_of_previous_colection>,
-# <previously_collected_value>).
+# Maps <monitor name + monitor instance id short hash>.<metric name> to a tuple
+# (<timestamp_of_previous_colection>, <previously_collected_value>).
 # To avoid collisions across monitors (same metric name can be used by multiple monitors), we prefix
 # metric name with a short hash of the monitor FQDN (<monitor module>.<monitor class> name). We
 # use a short hash and not fully qualified monitor name to reduce memory usage a bit.
+# NOTE: Those values are not large but we should probably still implement some kind of watch dog
+# timer where we periodically purge out entries for values which are older than MAX_RATE_TIMESTAMP_DELTA_SECONDS
+# (since those won't be used for calculation anyway).
 RATE_CALCULATION_METRIC_VALUES = defaultdict(lambda: (None, None))
 
 # If the time delta between previous metric value and current metric value is longer than this
--- a/scalyr_agent/scalyr_logging.py
+++ b/scalyr_agent/scalyr_logging.py
@@ -73,21 +73,22 @@ DEBUG_LEVEL_3 = 6
 DEBUG_LEVEL_4 = 5
 DEBUG_LEVEL_5 = 4
 
-# Stores metric values and timestamp for the metrics which we calculate per second rate on the
+# Stores metric values and timestamp for the metrics which we calculate per second rate in the
 # agent.
 # Maps <monitor name + monitor instance id short hash>.<metric name> to a tuple
 # (<timestamp_of_previous_colection>, <previously_collected_value>).
 # To avoid collisions across monitors (same metric name can be used by multiple monitors), we prefix
-# metric name with a short hash of the monitor FQDN (<monitor module>.<monitor class> name). We
-# use a short hash and not fully qualified monitor name to reduce memory usage a bit.
-# NOTE: Those values are not large but we should probably still implement some kind of watch dog
-# timer where we periodically purge out entries for values which are older than MAX_RATE_TIMESTAMP_DELTA_SECONDS
+# metric name with a short hash of the monitor FQDN (<monitor module>.<monitor class name>) +
+# monitor instance id. We use a short hash and not fully qualified monitor name to reduce memory
+# usage a bit.
+# NOTE: Those values are not large, but we should probably still implement some kind of watch dog
+# job where we periodically purge out entries for values which are older than MAX_RATE_TIMESTAMP_DELTA_SECONDS
 # (since those won't be used for calculation anyway).
 RATE_CALCULATION_METRIC_VALUES = defaultdict(lambda: (None, None))
 
-# If the time delta between previous metric value and current metric value is longer than this
-# amount of seconds (11 minutes by default), we will ignore previous value and not calculate the
-# rate with old metric value.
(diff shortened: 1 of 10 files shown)


scalyr_agent/scalyr_logging.py
```diff
@@ -460,17 +623,39 @@ def emit_value(
...
         self.info(
-            string_buffer.getvalue(),
+            "%s %s" % (metric_name, util.json_encode(metric_value))
+            + extra_fields_string_buffer.getvalue(),
             metric_log_for_monitor=monitor,
             monitor_id_override=monitor_id_override,
             timestamp=timestamp,
         )
-        string_buffer.close()
+        # string_buffer.close()
```
ArthurKamalov, line 637: it seems like a leftover.

scalyr_agent/scalyr_monitor.py
```diff
@@ -120,6 +121,9 @@ def __init__(
         # The logger instance that this monitor should use to report all information and metric values.
         self._logger = logger
         self.monitor_name = monitor_config["module"]
+        self.short_hash = hashlib.sha256(self.monitor_name.encode("utf-8")).hexdigest()[
```
ArthurKamalov, line 124: IIRC, the `monitor_name` is guaranteed to be unique because we get the instance-specific name suffix from its logger, right?

ArthurKamalov (approved):
Looks good to me so far :+1:

scalyr_agent/scalyr_monitor.py
Kami: IIRC, ``monitor_name`` here contains just the fully qualified name to the monitor class and the value which is used for logging and contains the id as well is different, but I'll dig in.

Kami: Yep, it appears we only pass ``id`` to the logger instance (https://github.com/scalyr/scalyr-agent-2/blob/master/scalyr_agent/monitors_manager.py#L508) and we don't have access to it in the monitor class.

Having said that, I could also refactor the code to pass the ``id`` to the monitor class and use that instead of the hash.

EDIT: This actually brings up good point - we want to use ``<monitor module>.<monitor instance id>.<metric name>`` for the internal dict key and not just ``<monitor module>.<metric name>``.

So it turns out we actually need both :)

scalyr_agent/scalyr_logging.py
```diff
@@ -368,6 +395,143 @@ def __init__(self, name):
...
+        """
+        if not self._should_calculate_metric_rate(
+            monitor=monitor, metric_name=metric_name
+        ):
+            return None
+
+        if timestamp:
+            timestamp_s = timestamp / 1000
+        else:
+            timestamp_s = timestamp or int(time.time())
```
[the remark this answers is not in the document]
Kami: Indeed, it's redundant, will fix it :)

Kami:
I pushed some changes to make code a bit more generic and future-proof (so in the future we can add more functions aka client side pre-calculation of various derived values).

Having said that, I'm still thinking on how to approach configuring this. One option would be to have an object which maps metric function to a list of metrics for which those should be pre-calculated. Problem is that we don't have a great way for specifying "complex" (object) values in the k8s config via environment variables.

Maybe that's OK for now since that functionality is still in internal preview and won't be directly exposed to the end user just yet. And even when it will be exposed, we will likely control default values for this config option ourselves (e.g. via config fragment).

Kami:
For now, I decided to go with current simple approach of configuring this aka existing ``calculate_rate_metric_names`` config option.

Configuration notation where the config value is ``JsonObject`` and the object value itself is ``JsonArray`` is not supported today and this would require more work (we would also need to implement merging support so those values can be specified across multiple config fragments and merged together, etc.).

PR #895, opened 2022-06-08
Fix copying manager checkpoints selecting logic 

ArthurKamalov:
This PR has a potential fix for issues that potentially happened to some of our customers and on our staging servers as well.

The general problem is that in some circumstances, on start, the agent reuploads old logs causing a spike in log volume. I wasn't able to reproduce the customer's issue, but I found several places in our code, which may lead to similar behavior. 

The proposed changes fix:
*  Bug with not picking `active` checkpoint when `full` checkpoints are not presented.
*  Now, the "freshest" checkpoint for a log file that is presented in multiple checkpoint files is selected according to its own timestamp, not by the write timestamp of the whole checkpoint file. That should avoid the situation where a checkpoint file with the most recent checkpoint has been written earlier than another checkpoint file, so the second file is selected instead of the first.

### DIFF
--- a/scalyr_agent/copying_manager/copying_manager.py
+++ b/scalyr_agent/copying_manager/copying_manager.py
@@ -1497,9 +1497,9 @@ class CopyingManager(StoppableThread, LogWatcher):
         :return: Checkpoint state stored in dict.
         """
 
-        found_checkpoints = []
-
         current_time = time.time()
+
+        checkpoints_files_paths = []
 
         # search for all worker session checkpoint files.
         worker_checkpoints_glob = os.path.join(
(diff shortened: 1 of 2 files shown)


dlobue (approved):
LGTM

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -1497,26 +1497,54 @@ def __find_and_read_checkpoints(self, warn_on_stale=False):
...
+            active_worker_checkpoints_glob
+        )
+
+        # Fetch workers main checkpoint file paths from active checkpoint path
+        # in order to handle exotic ede case where an active checkpoint exists but main checkpoint doesn't
+        for active_workers_checkpoints_path in all_active_workers_checkpoints_paths:
+            active_workers_checkpoints_file_name = os.path.basename(
+                active_workers_checkpoints_path
+            )
+            worker_checkpoint_path = os.path.join(
```
Kami, line 1530: Eventually it would probably be better to refactor this in a utility function.

PR #898, opened 2022-06-13
Update requests lib to 2.25

ArthurKamalov:


Kami:
Per Slack discussion - requests version we used before (2.20.0) already didn't support Python 2.6 (https://requests.readthedocs.io/en/latest/community/updates/#id12) so upgrading to requests v2.25.1 should be fine.

Kami:
If possible, we should also upgrade version used by Docker Images. Ideally we would use the same version everywhere (or perhaps newer in Docker images since we use Python 3.8 there).

IIRC, another reason why we haven't upgraded requests in the past was something related to ``docker`` dependency. Not sure if that's still a problem or not - we will likely see if tests fail or similar after upgrading the version used by Docker images.

Kami:

Kami:
As discussed, let's include this change in the next release after this one (v2.1.32 instead of v2.1.31).

Kami:
It looks we can actually upgrade since upgrading to 2.25.1 shouldn't affect Python 2.6 compatibility (2.22.0 we use right now already doesn't support Python 2.6).

Kami (approved):

PR #899, opened 2022-06-13
Downgrade urllib3 to 1.23

ArthurKamalov:
The previously upgraded `urllib3` version was to high and didn't comply the requirements of the `requests` library.


Kami:
urllib3 changelog for reference - https://github.com/urllib3/urllib3/blob/main/CHANGES.rst

Checking the changelog, it sounds like we may have similar issues under Python 3.10 - https://github.com/urllib3/urllib3/blob/main/CHANGES.rst#1265-2021-05-26

> Fixed deprecation warnings emitted in Python 3.10.

But it sounds like that would only manifest itself under Python 3.10.

Kami:
Superseded by #898.

PR #900, opened 2022-06-14
Fix empty command parsing in agent_main.py

ArthurKamalov:
This PR fixes bug with empty command for the agent_main.py script.


Kami:
Was this issue there for a long time or we inadvertently introduced a regression recently when we refactored some of that argument parsing code?

In addition to that, does it affect all the platforms or only Windows?

Kami (approved):
LGTM 👍

(Please also add a changelog entry)

PR #902, opened 2022-06-15
Reduce permissions for GHA token used by GHA workflows to the minimum

Kami:
This pull request is an attempt at reducing the permissions for Github token used by the GHA workflows to the minimum.

By default, token gets assigned broad set of permissions and for security reasons, we want to reduce those to the bare minimum which is needed for the workflows to run.

Default permissions (from https://github.com/scalyr/scalyr-agent-2/runs/6904005239?check_suite_focus=true job)

```bash
GITHUB_TOKEN Permissions
  Actions: write
  Checks: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
```

NOTE: The permissions YAML syntax is a bit weird and in case only a single permission is specified, rest of the permissions will default to ``none``. Perhaps we may still want to be explicit and explicitly set rest of the permissions to ``none`` although most of the other repos I checked don't use this notation.

> You can use the permissions key in your workflow file to modify permissions for the GITHUB_TOKEN for an entire workflow or for individual jobs. This allows you to configure the minimum required permissions for a workflow or job. When the permissions key is used, all unspecified permissions are set to no access, with the exception of the metadata scope, which always gets read access.

References:

* https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions
* https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/
* https://docs.github.com/en/actions/security-guides/automatic-token-authentication

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1016,6 +1016,7 @@ jobs:
             python.exe -m pip install --user --upgrade "pip==<< pipeline.parameters.default_pip_version >>"
       - run:
           name: Run Unit Tests under Python 3.10
+          no_output_timeout: 8m
           command: |
             # NOTE: If we don't redirect the output, Windows this step doesn't correctly exit even
             # after the tox commands finishes. That's why we use the redirect workaround. Sadly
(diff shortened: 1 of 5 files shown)


ArthurKamalov (approved):
LGTM

PR #906, opened 2022-06-22
Fix invalid end of the json parsing.

ArthurKamalov:
### Issue:
Our custom JSON parser does check for the end of the document after its parsing. For example ,if we put extra closing curly brackets in the middle of the object, it doesn't raise any error and just returns "incomplete" object.

Here the result object will miss the `filed2` filed and won't raise anything.
```
{
    api_key: ",key."
    field1: 1,
}
    filed2: 2
}
```

There's also the case where the customer accidentally commented out the beginning of the nested object:
```
{
  api_key: ",key."
  // server_attributes: {       // the beginning of the nested object is commented.
     serverHost: "HOST",        // the 'serverHost' field is wrongly parsed as a filed of outer object.
  }                             // this brackets are treated as end of the whole JSON document.

  // Everything that remains is just ignored.
  monitors: [
    {
       module:  "my_module",
    },
  ]
}
```
and the `monitors` list was lost.

### Solution:
To do additional check for the end of the document after the "root" JSON value is parsed.

### DIFF
--- a/tests/unit/configuration_test.py
+++ b/tests/unit/configuration_test.py
@@ -1878,9 +1878,9 @@ class TestConfiguration(TestConfigurationBase):
             """{
             api_key: "hi there",
             max_line_size: 5,
-            workers: [{"sessions":2, "api_key": "foo1234", "id": "one"}]}
+            workers: [{"sessions":2, "api_key": "foo1234", "id": "one"}]
             }
-        """
+"""
         )
         mock_logger = Mock()
         config = self._create_test_configuration_instance(logger=mock_logger)
@@ -1899,7 +1899,7 @@ class TestConfiguration(TestConfigurationBase):
             """{
             api_key: "hi there",
             max_line_size: 9900,
-            workers: [{"sessions":10, "api_key": "foo1234", "id": "one"}]},
+            workers: [{"sessions":10, "api_key": "foo1234", "id": "one"}],
             }
         """
         )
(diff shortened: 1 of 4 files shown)


tests/unit/configuration_test.py
```diff
@@ -1878,9 +1878,9 @@ def test_print_config_when_changed(self):
             """{
             api_key: "hi there",
             max_line_size: 5,
-            workers: [{"sessions":2, "api_key": "foo1234", "id": "one"}]}
```
ArthurKamalov, line 1881: Caught the same issue, in existing tests :grin:

tests/unit/json_lib/parser_test.py
```diff
@@ -273,6 +279,98 @@ def test_parse_from_bytes(self):
...
+                {
+                   module:  "my_module",
+                },
+              ]
+            }
+"""
+        with pytest.raises(JsonParseException) as err_info:
+            JsonParser.parse(json_data)
+
+        assert "Expecting EOF, got 'm'" in str(err_info.value)
```
Kami, line 333: The error message is not the most user-friendly, but I imagine if we wanted to provide more context, we would need to substantially change the parsing code so we have access to more context and not just the next / current character.

On that note, does the error class / message already include the line number where the error occured?

ArthurKamalov: Changed error message to `"Expecting end of the parsed document at the character '105', got character 'f' instead"`, WDYT?

ArthurKamalov: @Kami Line number is also shown now. I've also tried to add the column number, but I guess it's better to do that separately since it requires more refactoring.

Kami (approved):

PR #908, opened 2022-06-22
Fix kubernetes_events.log duplication

ArthurKamalov:
### Issue: 
The metric log file of the kubernetes_events_monitor is processed by the copying manager multiple times.

```
Path /var/log/scalyr-agent-2/agent.log: copied 18778 bytes (114 lines), 0 bytes pending, last checked Thu May 12 16:07:15 2022 UTC
Path /var/log/scalyr-agent-2/kubernetes_monitor.log: copied 17488 bytes (80 lines), 0 bytes pending, last checked Thu May 12 16:07:15 2022 UTC
Path /var/log/scalyr-agent-2/kubernetes_events.log: no matching readable file, last checked Thu May 12 16:03:09 2022 UTC
Path /var/log/scalyr-agent-2/kubernetes_events.log: copied 0 bytes (0 lines), 0 bytes pending, last checked Thu May 12 16:07:15 2022 UTC

Glob: /var/log/scalyr-agent-2/agent-worker-session-*.log:: last scanned for glob matches at Thu May 12 16:07:09 2022 UTC
```

### Possible solution: 

It seems that the this file is added to the copying manager from two different places in code:
1. log file matcher being added to the copying manager dynamically from the monitor itself https://github.com/scalyr/scalyr-agent-2/blob/dc507bef5b5027ffe35ba07f95e1b0b7572f3728/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py#L667

2. copying manager goes through all monitors and adds their log files. https://github.com/scalyr/scalyr-agent-2/blob/dc507bef5b5027ffe35ba07f95e1b0b7572f3728/scalyr_agent/copying_manager/copying_manager.py#L703

so we have to keep only one.

IMO, it's better to remove the first one, since it has to only affect this monitor.


Kami:
Per Slack - it's also important to note this distinction.

Multiple monitors schedule log files itself (kubernetes, kubernetes events, kubernetes open metrics), but kubernetes and kubernetes open metrics schedule other log files which file name are not the same as the monitor file name (kubernetes monitor schedules log files for container logs and kubernetes open metrics monitor schedules log files for other open metrics monitors).

In this case we still want copy manager to schedule log file for the actual monitor log file which matches the monitor name).

On the other hand, Kubernetes events monitor is special since it dynamically schedules log file which is named the same as the monitor log file - that's likely the reason why the clash / this issue occurs.

Taking this into account and primary reason why k8s events monitor schedules log file itself (custom parser), it sounds like another approach is possible - ability for the monitor to define parser and any other attributes for the log file itself - but it appears there may be another edge case hiding there since log file path can be changed / overridden via config option - https://github.com/scalyr/scalyr-agent-2/blob/bf39c279b54e056f6fcc360c098474b46bc48afd/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py#L109.

ArthurKamalov:
It seems that the copying manager [checks](https://github.com/scalyr/scalyr-agent-2/blob/a921c24d7b038f04b26721fc9a7cb76ff14caa6b/scalyr_agent/configuration.py#L889-L893) for a parser before [adding the monitor's log file](https://github.com/scalyr/scalyr-agent-2/blob/a921c24d7b038f04b26721fc9a7cb76ff14caa6b/scalyr_agent/copying_manager/copying_manager.py#L704). That's why parser type is should be assigned correctly to the k8s_events monitor log.

ArthurKamalov:
PS: I verified that it is working by changing branch for e2e test.

Kami (approved):
LGTM 👍

.github/workflows/end_to_end_tests.yml
```diff
@@ -3,10 +3,10 @@ name: "Kubernetes End to End Tests"
 on:
   push:
     branches:
-      - master
+      - arthur/new-checks
```
Kami, line 6: I assume you will change this back to master before / after merging this branch?

PR #909, opened 2022-06-23
k8s agent support for windows nodes

half2me:
Since we run a mix OS cluster on AKS (linux and windows nodes) We need the Scalyr agent to collect logs from all nodes.
Installing the scalyr agent for k8s via helm, only provides support for linux nodes, and fails trying to run on windows nodes. Adding the `nodeSelector.kubernetes\\.io/os: linux` option restricts it to linux nodes only.

For windows nodes, I've essentially just made another DaemonSet restricted to windows nodes only. I've built a docker image, using `servercore:ltsc2019`, installed the latest python for windows, and all other dependencies. I've had to make a small change to the python code, because it is trying to check for a running service on windows, which is not the case here.

This is not meant to be merged, it is simply an example of how to get a windows node with the pod/container logs mounted in, so scalyr agent can upload them.

This is as far as I've got. The agent runs, it can communicate with the kubernetes API, but unfortunately atm it only uploads its own logs. If anyone has any ideas as to what is missing, please do let me know.

If you want to try it out, make sure you first install the scalyr helm chart, as I reuse the serviceaccount. Also, I've named the image "winscaler", make sure to change this to whatever you have built.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM mcr.microsoft.com/windows/servercore:ltsc2019
+
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
+RUN Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
+RUN choco install python git -y
+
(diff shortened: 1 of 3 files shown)


Kami:
@half2me Thanks for the contribution.

@ArthurKamalov will have a look at it as soon as he get a chance and we will see if we can use this as a starting point for Windows node support.

PR #912, opened 2022-06-28
Remove Rust tooling used for build step of Docker images

Kami:
This pull request removes Rust tooling which were installed as part of the Docker image build step.

We can not use existing pre-built musl + arm64 orjson wheel so there is no need for us anymore to install Rust + build dependency from scratch (which also speeds up the build quite a bit).

This change should have no end user impact since it only updates the Docker image build related code.


ArthurKamalov (approved):
LGTM

PR #926, opened 2022-07-04
Enable additional Kubernetes metrics collection via global config option

Kami:
This pull request adds new ``k8s_explorer_enable`` config option which enables additional metrics collection which is needed for Kubernetes Explorer functionality.

Right now Kubernetes Explorer also depends on some metrics from Kubernetes Events Monitor so when ``k8s_exlorer_enable`` is True, kubernetes events monitor is enabled as well.

For now, this PR also disables building ``k8s-with-openmetrics-monitor`` images (with those changes, those are not needed anymore, but I left code in place in case we may need them in the future - e.g. if we want to release new explorer agent version without main agent release).

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
@@ -38,9 +38,9 @@ monitor:
       building scrapper URL.
     * ``k8s.monitor.config.scalyr.com/scrape_interval`` (optional) - How often to scrape this endpoint.
       Defaults to 60 seconds.
-    * ``k8s.monitor.config.scalyr.com/scrape_timeout`` (optional) - How long to wait before timing out.
+    * ``k8s.monitor.config.scalyr.com/scrape_timeout`` (optional) - How long to wait before timing out. This should be at least 5-10 seconds shorter than scrape interval.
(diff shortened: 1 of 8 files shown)


scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
```diff
@@ -539,6 +541,18 @@ def _initialize(self):
 
         self.__previous_running_monitors_count = 0
 
+        self.__enable_monitor = self._global_config.k8s_explorer_enable
+
+    def run(self):
+        if not self.__enable_monitor:
```
ArthurKamalov, line 547: I guess that shouldn't be a big deal, unless there are some periodic "health checks" for running monitors, but I don't think that we have such checks now.

Kami: Yeah, correct - and we already use the same pattern in the kubernetes_events_monitor (which I copied).

Kami: Technically, we could also utilize "dynamic monitor scheduling functionality" which I implemented recently (monitor itself would ask MonitorsManagers to remove itself from running monitors), but this would be introducing a new pattern (and possible edge cases), so I went with the existing approach.

PR #933, opened 2022-07-11
Log a warning in case a mandatory metrics exporter is not being scrapped

Kami:
This pull request updates Kubernetes OpenMetrics monitor to log a warning in case OpenMetrics monitor for node exporter pod is not found.

Scrapping node-exporter pod is mandatory for a complete Kubernetes Explorer experience.

Same thing goes for kube state metrics exporter pod, but logic for detecting that has not been implemented yet. kube state metrics is a `Deployment` which means only a single pod will run in a cluster at a time.

Agent deployment and work distribution model is heavily based around the DaemonSet model where each agent pod instance is only responsible for work on a specific node the agent is running on.

That means that if we wanted to implement kube state metrics scrapper running detection logic we would need to check across all the running agent pod instances (aka across all the nodes) which would require us to implement basic coordination primitives.

Implementing something like that is definitely possible, but it would result in unnecessary complexity and overhead, especially since we can solve that much easier on the server (dashboard) side - we can have a simple widget which prints helpful error message in case a specific data is missing and points user to the docs on how to configure their Kubernetes cluster for Kubernetes Explorer.

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_openmetrics_monitor.py
@@ -223,6 +223,8 @@ DEFAULT_SCRAPE_SCHEME = "http"
 DEFAULT_SCRAPE_PORT = None
 DEFAULT_SCRAPE_PATH = "/metrics"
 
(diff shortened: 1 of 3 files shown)


ArthurKamalov (approved):
LGTM, as far as I am familiar with this )

PR #934, opened 2022-07-12
Fix sample interval overriding in the Windows System and Windows Process monitor

Kami:
This pull request fixes a bug in Windows System and Windows Process monitor which didn't allow user to override / change the default sample interval of 30 seconds using ``sample_interval`` monitor config option.


ArthurKamalov (approved):
LGTM

ArthurKamalov (approved):

PR #935, opened 2022-07-13
Fix a bug with "import_vars" functionality which would not work correctly when multiple variables used the same common prefix

Kami:
This pull request fixes bug in the ``import_vars`` functionality which would not work correctly when multiple variables used the same common prefix (e.g. ``SCALYR_FOO, SCALYR_FOO_BAR, SCALYR_FOO_BAZ``).

## Proposed Fix

Proposed fix in this PR includes sorting variable names by length in the descending order (from the longest to the shortest) when applying variable substitutions. This way we make sure that the correct and most "specific" variable is used when common prefix is shared by multiple variables.

It's worth noting that we can't use regular expressions directly since we support partial / inline substitutions, e.g. (``foo$VARIABLE_BARbaz``).


ArthurKamalov (approved):
LGTM

PR #939, opened 2022-07-18
CT-545 Migrate CircleCI docker smoke tests to GHA

dlobue:
Docker & k8s smoketests are now working in GHA.
I have not disabled the docker smoketests from running circleci. I wasn't sure if we wanted to run both side-by-side for a time first.

CT-545


.circleci/smoketest_docker.sh
```diff
@@ -80,7 +80,7 @@ contname_verifier="ci-agent-${log_mode}-${CIRCLE_BUILD_NUM}-verifier"
 # Kill leftover containers
 function kill_and_delete_docker_test_containers() {
     echo ""
-    echo "Killing and deleting all test containers..."
+    echo "::group::Killing and deleting all test containers..."
```
ArthurKamalov, line 83: Haven't known about that feature, cool!

ArthurKamalov (approved):
Everything looks pretty good to me. :+1:

Kami:
How does the tests run time compare to Circle CI? Do we need to implement some kind of Docker Layer Caching? (I've done that for another repo - logstash one, but it turns out the action I used is not great, too many edge cases and eventually the cache grows so large it's slower to download and upload the cache than building the image from scratch).

dlobue:
> How does the tests run time compare to Circle CI? Do we need to implement some kind of Docker Layer Caching? (I've done that for another repo - logstash one, but it turns out the action I used is not great, too many edge cases and eventually the cache grows so large it's slower to download and upload the cache than building the image from scratch).

The difference is shockingly low. The CircleCI version appears to be ~45s faster.  Here are the runs I'm comparing:
https://app.circleci.com/pipelines/github/scalyr/scalyr-agent-2/9746/workflows/e62f72bf-f7a6-4fb8-b019-771d84e006b1/jobs/130494
https://github.com/scalyr/scalyr-agent-2/runs/7588013576?check_suite_focus=true

2m 45s vs 3m 24s
That is total run time though.

The majority of that time difference comes from installation of python dependencies (where the pip python cache isn't working properly...). That's 23s all on its own on GHA.
When comparing just the `smoketest_docker.sh` runs against each other, the difference is 2m 47s (GHA) vs 2m 31s (Circle), for a delta of 16s.

It didn't seem like anything that was worth spending time on optimizing.

Kami:
@dlobue Thanks for the info.

Now that those changes have been merged into master, I assume we should remove those jobs from Circle CI config?

On a related note, how does this affect our release pipeline? As part of release pipeline, we have code in Jenkins which polls Circle CI API and only continues in case the jobs pass.

/cc @ArthurKamalov

PR #941, opened 2022-07-20
Fix config export config

ArthurKamalov:
Fix `--import-config`, `--export-config` options of the `scalyr-agent-2-config` comand which is caused by Python2,3 incompatibility.

I also noticed, that docker images don't have `scalyr-agent-2-config` compatibility script, which caused errors on docker images when running config commands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Windows:
 Bug fixes:
 * Fix a bug with ``import_vars`` functionality which didn't work correctly when the same variable name prefix was used (e.g. ``SCALYR_FOO_TEST``, ``SCALYR_FOO``).
 * Fix a bug with handling the log file of the Kubernetes Event Monitor twice, which led to duplication in the agent's status. 
+* Fix a bug in scalyr-agent-2-config ``--export-config``, ``import-config`` options caused by Python 2 and 3 code incompatibility.
 
 Docker images:
 * Upgrade various dependencies: orjson, requests, zstandard, lz4, docker.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Windows:
 Bug fixes:
 * Fix a bug with ``import_vars`` functionality which didn't work correctly when the same variable name prefix was used (e.g. ``SCALYR_FOO_TEST``, ``SCALYR_FOO``).
 * Fix a bug with handling the log file of the Kubernetes Event Monitor twice, which led to duplication in the agent's status. 
+* Fix a bug in scalyr-agent-2-config ``--export-config``, ``import-config`` options caused by Python 2 and 3 code incompatibility.
 
 Docker images:
 * Upgrade various dependencies: orjson, requests, zstandard, lz4, docker.
```
Kami, line 21: Do we also need a changelog entry here for docker image? Aka something along the lines of "Make sure ``scalyr-agent-2-config`` binary is present in the Docker image``.

tests/unit/agent_main_test.py
```diff
@@ -670,3 +673,142 @@ def test_windows_service(self):
...
+        tar = tarfile.open(self.output_file)
+        self._extract_result(tar)
+
+        # That same agent config has to be extracted.
+        assert os.path.isfile(self.extracted_config_path)
+
+        with open(self.extracted_config_path, "r") as f:
+            extracted_config = json.load(f)
+
+        assert extracted_config == original_config
```
Kami, line 736: Thanks for adding this check. It's indeed good we assert on the parsed (and also unparsed) content to ensure it doesn't contain some unnecessary `b""`` prefix (which can sometimes end up there in case bytes / string handling is not done correctly in the context of Python 2 / 3).

Kami (approved):
LGTM :+1: 

Thanks for fixing this and adding test cases.

CHANGELOG.md
ArthurKamalov: Ah sure, ~was trying to hide my oversight~ forgot about it.

PR #942, opened 2022-07-21
Instrumentation logging improvements

Kami:
This pull request includes two small improvements:

1. Make sure Kubernetes Events monitor only logs message on initialize when monitor is disabled via environment variable or config option. I also added rate limit for that log call to avoid duplicate messages under some scenarios.

2. Allow user to change logging interval which is used to log instrumentation related statistics to agent.log using INFO log level. This can come handy in various troubleshooting scenario, but most users should leave this value as is on production deployments (since low value could add a lot of overhead).

   Since code where those messages are logged doesn't have direct access to global Configuration object instance (technically, I could access it via ``monitor._global_config``, but I'm not 100% sure this is always set and it could also cause cyclic dependency issues), I needed to a different pattern for updating that module level constant - I set / update it as part of the code where the config file is re-read and things are reloaded, etc. in case changes are detected.

   Logging can also be disabled by setting that value to ``0`` (which will be the case for non Kubernetes Explorer deployments where that functionality is not used).

### DIFF
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -1222,6 +1222,10 @@ class Configuration(object):
         return self.__get_config().get_float("config_change_check_interval")
 
     @property
+    def instrumentation_stats_log_interval(self):
+        return self.__get_config().get_int("instrumentation_stats_log_interval")
+
+    @property
     def overall_stats_log_interval(self):
         return self.__get_config().get_float("overall_stats_log_interval")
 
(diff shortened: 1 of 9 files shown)


scalyr_agent/configuration.py
```diff
@@ -3161,6 +3165,21 @@ def __verify_main_config_and_apply_defaults(
             config, "disable_leak_config_reload", None, description, apply_defaults
         )
 
+        # How often to long instrumentation related stats using INFO log level to agent.log
```
ArthurKamalov, line 3168: s/long/log/g ?

ArthurKamalov (approved):
LGTM

Kami:
Re my comment above - I decided to implement simpler approach where we access global config via ScalyrMonitor object - 5a51f9bedc67b4a3d4037ea30fdac9721a4a8a1b.

I'll keep an eye on it to make sure it doesn't introduce any cyclic dependencies or surprised.

scalyr_agent/configuration.py
Kami: Fixed, thanks.

PR #943, opened 2022-07-22
Introduce clean up routine for RateMetricFunction, fix log function argument order

Kami:
This pull request includes two changes:

1. Fix invalid ``log.info()`` function call arguments order
2. Introduce clean up routine for the ``RateMetricFunction`` function which periodically cleans up old / stale values.

We need periodic clean up routine to ensure internal data structure which stores metric values doesn't grow too large and that stale / unused values are correctly removed (a lot of stored metrics will become stale and unused over time since they will refer to resources which are now gone - imagine metric which refer to old containers, old nodes, etc.)

NOTE: In the future I may refactor some of that functionality and instead of having a global singleton instance of ``RateMetricFunction`` move it to an approach where we have an instance per ``ScalyrMonitor`` (I think this should work well, reduce some overhead / spread it across monitor threads and simplify some of the thread safety issues. A lot of that functionality is already scoped to a monitor and should work well as long as we never need cross monitor metrics calculations in the future).

### DIFF
--- a/scalyr_agent/metrics/functions.py
+++ b/scalyr_agent/metrics/functions.py
@@ -115,15 +115,20 @@ class RateMetricFunction(MetricFunction):
     # metric name with a short hash of the monitor FQDN (<monitor module>.<monitor class name>) +
     # monitor instance id. We use a short hash and not fully qualified monitor name to reduce memory
     # usage a bit.
-    # NOTE: Those values are not large, but we should probably still implement some kind of watch dog
-    # job where we periodically purge out entries for values which are older than MAX_RATE_TIMESTAMP_DELTA_SECONDS
-    # (since those won't be used for calculation anyway).
+    # TODO: Perhaps we could store values ordered by timestamp to make clean up faster.
     RATE_CALCULATION_METRIC_VALUES = defaultdict(lambda: (None, None))
 
     # If the time delta between previous metric collection timestamp value and current metric collection
     # timestamp value is longer than this amount of seconds (11 minutes by default), we will ignore
     # previous value and not calculate the rate with old / stale metric value.
     MAX_RATE_TIMESTAMP_DELTA_SECONDS = 11 * 60
+
+    # Periodically when a clean up routine runs we delete any values with timestamp older than
+    # this threshold.
+    DELETE_OLD_VALUES_THRESHOLD_SECONDS = 30 * 60
+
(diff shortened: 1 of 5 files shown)


scalyr_agent/metrics/functions.py
```diff
@@ -300,6 +307,25 @@ def calculate(
...
+            monitor._global_config
+            and monitor._global_config.metric_functions_cleanup_interval
+            or 0
+        )
+
+        if metric_functions_cleanup_interval > 0 and (
+            cls.LAST_CLEANUP_RUNTIME_TS <= (now_s - metric_functions_cleanup_interval)
+        ):
+            cls._remove_old_entries(monitor=monitor)
+
```
ArthurKamalov, line 328: Do you think we should have option to disable that cleanup by making interval zero? Or maybe at least log a warning message on zero?

scalyr_agent/metrics/functions.py
```diff
@@ -407,6 +433,42 @@ def should_calculate(cls, monitor, metric_name, extra_fields=None):
...
         return False
 
+    @classmethod
+    def _remove_old_entries(cls, monitor):
+        # type: (ScalyrMonitor) -> None
+        """
+        This function removes old entries for values with the timestamps which are older than the
+        defined threshold.
+
+        TODO: Thread safety?
```
ArthurKamalov, line 443: Yeah.. I guess we could guard it by some singleton lock but I'am not sure how it may affect performance..

ArthurKamalov (approved):

scalyr_agent/metrics/functions.py
Kami: For now I avoided it since ``calculate()`` is called on every single ``emit_value()`` function call and it could add massive overhead.

I plan to look more into thread safety and performance in the future, perhaps by moving some of the abstractions directly on ``ScalyrMonitor`` to mitigate thread safety and performance issues (since ``RateMetricFunction()`` class instance would then be monitor specific and monitors already run in a separate dedicated thread).

scalyr_agent/metrics/functions.py
Kami: Yeah, if the interval is set to ``0``, it will be disabled. It's mostly there for testing purposes. We don't really plan to expose this functionality to the end user (config option is mostly there for us for testing and similar).

PR #944, opened 2022-07-22
CT-528

jmakar-scalyr:
Changes for [CT-528](https://scalyr.myjetbrains.com/youtrack/issue/CT-528)

- Created a new windows_event_log_monitor api class called NewJsonApi
  - NewJsonApi is a subclass of NewApi
    - The new class use is triggered by a new monitor config field called json; ref lines 144-150
    - Renamed NewApi.{__bookmarks,_bookmark_lock} to {_bookmarks,_bookmark_lock} for accessibility in NewJsonApi
  - Context: OldApi, NewApi (and NewJsonApi) are used by WindowEventLogMonitor to query the win32 event log api
    - NewJsonApi uses the same approach as NewApi but formats events differently
      - Specifically formats events as xml via the win32 api then converts it to json using xmltodict
      - Additionally applies a couple transformations to the json
        - _convert_json_array_to_object: Converts json arrays to objects to accommodate Dataset limitation
        - _strip_xmltodict_prefixes: Remove xmltodict added field prefixes if safe to do so
    - NewJsonApi (from WindowEventLogMonitor) sets a DummyFormatter on the logging handler to ensure the generated log record is 100% json
      - The DummyFormatter is to avoid unmarshalling and remarshalling the json within the formatter
      - Preferred the log record to be 100% json to avoid having to create a new server side parser definition (ie can just rely on dottedJson)

Sample output available on [qatesting](https://app-qatesting.scalyr.com/y/j1Xoo1uEzX6)


scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -17,11 +17,14 @@
 from __future__ import unicode_literals
 from __future__ import absolute_import
 
+import datetime
 import os
+import json
```
Kami, line 22: Minor performance thing - we may want to use ``orjson`` library here (it's already bundled with the Windows binary).

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -625,18 +636,92 @@ def log_event(self, event):
...
+        # Ref: json.dumps(xmltodict.parse('<entry>foo<text>bar</text></entry>'))
+        #        => '{"entry": {"text": "bar", "#text": "foo"}}'
+        _strip_xmltodict_prefixes(event_json)
+
+        # Populate the record here with fields that would normally be added by the log formatter,
+        # this avoids having to unmarshal and remarshal later in the log formatter.
+        # Refer to the use of DummyFormatter in WindowEventLogMonitor.open_metric_log().
+        event_json["timestamp"] = datetime.datetime.utcnow().isoformat(" ")
+        event_json["name"] = self._logger.name
+        self._logger.emit_value("unused", json.dumps(event_json))
```
Kami, line 686: As discussed on Slack, if we want this to work out of the box we will also need to do the following, right?

1. Define new built-in parser on the server side which parses ``message`` as json (or utilize existing json parser, may still be better to have a new one to have more flexibility in the future)
2. Define ``parser: json`` attribute for messages in those log file to ensure they are parsed as JSON on the server side.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -625,18 +636,92 @@ def log_event(self, event):
...
+            channel = values[win32evtlog.EvtSystemChannel][0]
+            if channel not in self._bookmarks:
+                self._bookmarks[channel] = win32evtlog.EvtCreateBookmark(None)
+            win32evtlog.EvtUpdateBookmark(self._bookmarks[channel], event)
         finally:
-            self.__bookmark_lock.release()
+            self._bookmark_lock.release()
+
+
+def _convert_json_array_to_object(obj):
```
Kami, line 698: Would you also mind sharing some example payloads in this PR so we can see if additional processing is needed (in case nested arrays make it hard to query specific values)?

Kami:
I had a look at the example payload on qatesting and per discussion on Slack the other day, we likely want to discuss this format with @dariocalia and @harrisonScalyr (and likely also involve early adopters, to see if it meets their needs).

This deconstructed array will likely make it very hard to query those events so we should discuss  possible alternatives (perhaps deconstructing each ``EventData`` array item into new DataSet ``Event`` object, carrying over / repeating parent Windows Event Log ``Event`` attributes - this won't be 1:1 mapping with WEL format, but it may make it more useful to the end users due to making it easier to query for those events).

docs/monitors/windows_event_log_monitor.md
```diff
@@ -86,3 +86,4 @@ and System sources:
 | `remote_user`                | Optional (defaults to ``None``). The username to use for authentication on the remote server.  This option is only valid on Windows Vista and above. | 
 | `remote_password`            | Optional (defaults to ``None``). The password to use for authentication on the remote server.  This option is only valid on Windows Vista and above. | 
 | `remote_domain`              | Optional (defaults to ``None``). The domain to which the remote user account belongs.  This option is only valid on Windows Vista and above. | 
+| `json`                       | Optional (defaults to ``False``). Format events as json? Supports inclusion of all event fields. This option is only valid on Windows Vista and above. |
```
Kami, line 89: We should also figure out / decide if ``true`` should be the default value.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -625,18 +636,92 @@ def log_event(self, event):
...
+            event, win32evtlog.EvtRenderEventValues, Context=self._render_context
+        )
+
+        metadata = win32evtlog.EvtOpenPublisherMetadata(
+            values[win32evtlog.EvtSystemProviderName][0]
+        )
+
+        event_json = xmltodict.parse(
+            win32evtlog.EvtFormatMessage(
+                metadata, event, win32evtlog.EvtFormatMessageXml
```
Kami, line 669: @dariocalia mentioned JSON format may also be available by the win32 API, but that doesn't appear to be the case - https://docs.microsoft.com/en-us/windows/win32/api/winevt/ne-winevt-evt_format_message_flags

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -625,18 +636,92 @@ def log_event(self, event):
...
+
+        # If safe, strip off the @ / # prefixes xmltodict uses to prevent conflicts.
+        # Ref: json.dumps(xmltodict.parse('<entry>foo<text>bar</text></entry>'))
+        #        => '{"entry": {"text": "bar", "#text": "foo"}}'
+        _strip_xmltodict_prefixes(event_json)
+
+        # Populate the record here with fields that would normally be added by the log formatter,
+        # this avoids having to unmarshal and remarshal later in the log formatter.
+        # Refer to the use of DummyFormatter in WindowEventLogMonitor.open_metric_log().
+        event_json["timestamp"] = datetime.datetime.utcnow().isoformat(" ")
```
Kami, line 684: Can we actually use timestamp from the WEL event itself (which may be different than the current time, right?

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Ah thank you, will fix

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Right, the parser attribute is set on line 880 (in WindowEventLogMonitor.__get_api).
It's not set in the NewJsonApi because the WindowEventLogMonitor.log_config attrib is not accessible from it.

I opted to use dottedJson because it's one of the builtin parsers, open to defining a new parser if you think it's worth it.

jmakar-scalyr:
Some sample data (just grabbed entries with different Event.System.EventID)

[sample.txt](https://github.com/scalyr/scalyr-agent-2/files/9183061/sample.txt)

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Sure, added sample.txt in a comment.

I haven't look too hard but so far the EventData has a pretty consistent structure of arrays of Text generally with a Name attribute.  There was one occasion where the EventData object itself had an Name attribute as well, both those cases are shown [here](https://scalyr.myjetbrains.com/youtrack/issue/CT-528#focus=Comments-4-56476.0-0)

docs/monitors/windows_event_log_monitor.md
jmakar-scalyr: From a conservation early on with Dario, the intention is this would be false initially to be backwards-compatible for  queries/dashboards customers have built but ultimately it would become the default (or the NewJsonApi impl. would replace the NewApi impl.)

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Yep, I looked pretty hard & didn't find an approach to get json directly from the win32 api.

It's reassuring to know that Elastic does an xml-to-json conversion as well: https://github.com/elastic/beats/blob/main/winlogbeat/eventlog/wineventlog.go#L323

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Will fix

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Hm it seems orjson is not supported with python 2.7: https://github.com/ijl/orjson/issues/35#issuecomment-545578561

Sticking with the standard library json module for now for the sake of the python 2.7 unittests.

ArthurKamalov: I guess, you also can use our wrapper json functions. They select json library itself and they have to use standard json lib as a fallback if `orjson` isn't fount.  https://github.com/scalyr/scalyr-agent-2/blob/ab6d933907884a9ccc61aa76e0857ec4d691a67b/scalyr_agent/util.py#L357-L386

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
Kami: Thanks, I missed that line :)

We can discuss during the sync up if it makes sense to define new server side parser or not (may help with the queries and make it easier to distinguish between differently formatted messages from WEL log file).

agent_build/requirement-files/testing-requirements.txt
```diff
@@ -40,3 +40,5 @@ futures==3.3.0; python_version <= '2.7'
 more-itertools==8.5.0; python_version >= '3.0'
 more-itertools==5.0.0; python_version <= '2.7'
 distro==1.6.0
+xmltodict==0.13.0; python_version >= '3.0'
+xmltodict==0.12.0; python_version <= '2.7'
```
Kami, line 44: pre-built Windows Python binaries are Python 3 only so we can remove this dependency for <= 2.7 and update unit tests so they only run under Python 3 (aka skip tests which rely on this dependency under Python 2).

scalyr_agent/third_party/README.md
```diff
@@ -35,8 +35,8 @@ The following libraries are included:
   * [idna](#idna)
   * [chardet](#chardet)
   * [urllib3](#urllib3)
-  * [python-dateutil][#dateutil]
-  * [repoze.lru][#repoze.lru]
+  * [python-dateutil](#dateutil)
```
Kami, line 38: whoops, thanks for the fix.

agent_build/requirement-files/testing-requirements.txt
jmakar-scalyr: Will fix

jmakar-scalyr:
@dariocalia @Kami I discussed the current approach with folks more familiar with the use case & they're happy with it.  They also let me know that we should already be able to search fields recursively using wildcards (ie `Event.EventData.* contains 'Negotiate'`)

We found that `Event.* contains ...` recursively searches ok but that `Event.EventData.* contains ..` didn't, Joel will file a bug with the EventDB team.

So assuming Tomaz & Arthur approve, this PR should be ready to merge.  (Incidentally the code coverage check failed because I added `@skip(sys.platform != "Windows")` to the tests to avoid requiring xmltodict for Python 2.7)

ArthurKamalov (approved):
LGTM I wonder if we can add that case in our Windows AMI test which runs in AWS instance, but now sure about how long it may take.

Kami (approved):

PR #946, opened 2022-07-25
Include full module name in the agent log messages

Kami:
This pull request updates ``AgentLogFormatter`` to include full module name (previously only the file name was included).

This allows us to disthinghuish between different modules which use the same name (but belong to a different package).

Before:

```bash
2022-07-24 13:15:19.697Z WARNING [monitor:linux_system_metrics()] [tcollector.py:148] iostat.py: b'Traceback (most recent call last):'
```

After:

```bash
2022-07-24 13:15:19.697Z WARNING [monitor:linux_system_metrics()] [scalyr_agent.third_party.tcollector.tcollector:148] iostat.py: b'Traceback (most recent call last):'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Scalyr Agent 2 Changes By Release
 =================================
 
-## 2.1.32 "Occao" - July 20, 2022
+## 2.1.32 "Occao" - July 20, 2022 - In development
 
 <!---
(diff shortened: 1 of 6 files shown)


ArthurKamalov (approved):

Kami:
While working on it I noticed there are a couple of more small edge cases which I will fix in a separate PR.

PR #950, opened 2022-07-29
CT-574 Migrate Unittests from CircleCI to GHA

dlobue:
This PR gets unittests working in GHA. There are some differences from the CircleCI implementation due to my attempts to increase code-reuse, and to try the naive approach in order to simplify things. I also increased the number of python versions we run tests on because, why not?

Changes that should get special attention:
- No python 2.7 test run
- No python 3.5 test run
- Capturing coverage on all test runs
- Combined the default tox testenv and tox windows testenv together
- Running tests against python 3.6 through python 3.10 on ubuntu & osx (only python 3.10 on windows)
- Removed pinning of virtualenv and pip
- Using a GH action to parse the junit report and generate report summaries as comments, on the workflow summary, and on the check summary. It also adds annotations to the check summary of test failures. I can tune these summaries, so please let me know if you like them, want less, or if there's something missing.

CT-574


.github/workflows/unittests.yml
```diff
@@ -0,0 +1,83 @@
...
+          path: test-results/junit*.xml
+        # Use always() to always run this step to publish test results when there are test failures
+        if: ${{ always() }}
+
+      - name: Publish Test Results
+        uses: EnricoMi/publish-unit-test-result-action@v2
+        if: ${{ always() }}
+        with:
+          junit_files: "test-results/junit*.xml"
+          comment_title: Test Results - Python ${{ matrix.python-version }} - ${{ matrix.platform }}
```
Kami, line 129: I think having comment per job is a bit too spammy - is there a way to consolidate results of all the jobs in a single comment?

.github/workflows/unittests.yml
```diff
@@ -0,0 +1,129 @@
...
+          - ubuntu
+          - macos
+        include:
+          - platform: windows
+            python-version: "3.10"
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Setup Python
```
Kami, line 88: To speed up the build, I assume we also want to add step where we cache the pip dependencies.

Kami:
Since OS X is only used for development I think it's fine to only run unit tests there under a single Python version (e.g. same one as we use for Windows unit tests).

Kami:
> Removed pinning of virtualenv and pip

We try to pin everything / as much as possible to try to be things reproducible as much as possible and to avoid surprises (breaking changes in pip and virtualenv have bitten us many times in the past already).

Kami:
> No python 2.7 test run

We should still add those tests back until we officially deprecate Python 2.7 (since those tests still catch some Python 2.7 compatibility related issues).

dlobue:
> > No python 2.7 test run
> 
> We should still add those tests back until we officially deprecate Python 2.7 (since those tests still catch some Python 2.7 compatibility related issues).

I was hoping we'd be ready to roll out frozen binaries by the time I was ready to deprecate circleci. Added them in.

dlobue:
> Since OS X is only used for development I think it's fine to only run unit tests there under a single Python version (e.g. same one as we use for Windows unit tests).

Alrighty. Done.

.github/workflows/unittests.yml
dlobue: Yes: https://github.com/scalyr/scalyr-agent-2/pull/950#issuecomment-1208734836
I don't know if it will properly handle linking back to the original test properly in this configuration. I guess we'll just have to wait and see....

.github/workflows/unittests.yml
dlobue: Not really needed ATM. It pretty much is doing that already for the requirements I'm installing from the commandline. That pip install step is sub-10sec.
Right now the docker image tests take more time than these unittests. We can revisit caching if it becomes problematic.

dlobue:
> > Removed pinning of virtualenv and pip
> 
> We try to pin everything / as much as possible to try to be things reproducible as much as possible and to avoid surprises (breaking changes in pip and virtualenv have bitten us many times in the past already).

Okay, well we need to update the virtualenv and pip version more often than every 2 years. The old version was breaking the windows + python 3.10 test suite.

If we're going to go all in on pinning, we might want to look into [pip-tools](https://pypi.org/project/pip-tools/), [pipenv](https://pipenv.pypa.io/en/latest/) or [poetry](https://python-poetry.org/).

.github/workflows/unittests.yml
Kami: With pip dependency caching enabled, that cache should also be used by tox, so it should also speed tox target run

Kami:
> I was hoping we'd be ready to roll out frozen binaries by the time I was ready to deprecate circleci. Added them in.

that was optimistic :P

Kami:
Btw, what we should do with required github status checks? Right now we have bunch of Circle CI checks marked as required for PR merging.

Should we leave this as is or we should update checks to use new GHA jobs?

.github/workflows/unittests.yml
```diff
@@ -0,0 +1,133 @@
...
+        with:
+          path: reports
+
+      - name: Prepare Coverage Data for codecov.io
+        run: |
+          coverage combine reports/**/.coverage
+          coverage xml -i -o coverage.xml
+
+      - name: Upload Coverage to Codecov.io
+        uses: codecov/codecov-action@v3
```
Kami, line 55: For security we should probably pin all the actions to specific git commit sha

dlobue:
> Btw, what we should do with required github status checks? Right now we have bunch of Circle CI checks marked as required for PR merging.
> 
> Should we leave this as is or we should update checks to use new GHA jobs?

@Kami 
My plan was to leave the required status checks as-is for the time being. I'll switch the required status checks once all test suites have been ported. This will give the ported tests some time to bake and give us confidence in their reliability (or identify their problems).

.github/workflows/unittests.yml
dlobue: Pinned all third-party actions.

Kami (approved):
LGTM 👍

Kami:
I'll go ahead and merge it since I'm curious how test reporting via comment works in "real life" (since this PR has a lot of spam due to developing and testing this functionality).

Kami:
I was curious how much (if anything) does caching Python dependencies speed up the tox run - 8f36004efd8a08b9fb2c602bbfeef380f9c2d313.

And yeah, speed up is tiny in the range of 10-20s (eventually it may make sense to cache created .tox/ virtualenv sub-directories, but handling that correctly has many more edge cases and it's harder to do), but probably doesn't hurt to keep it.

PR #951, opened 2022-07-31
Add check for dead essential monitors

ArthurKamalov:
This PR introduces new option `stop_agent_if_fails` for the monitor configuration. If monitor with such option (which has to be set as `True`) fails, then whole agent has to fail too. That has bring additional confidence that agent won't run without "essential" monitor, for example kubernetes monitor.

Agent checks health of its monitors in the main loop with its own pace (once in 5 sec for now). To achieve that, I had to refactor that main loop, because before that, the rate of the loop was[ hardcoded to the interval ](https://github.com/scalyr/scalyr-agent-2/blob/483a144321721e90722e87c75a53f98c14d1c342/scalyr_agent/agent_main.py#L1227)of the "config change check "(default - 30 sec), so it's hard to add new checks that have to run with smaller intervals. I extracted all code which is related to the "config change check" to a separate [generator function](https://github.com/scalyr/scalyr-agent-2/blob/0b32d3393b741977ede8d0d22f2de71ec0948a8b/scalyr_agent/agent_main.py#L1072), and now that generator is just called from the main loop when it's time to check the config. Since that check now is performed in "asynchronous" manner, another checks can be added with their own intervals.  
That's "config change check" code is a bit long and messy and it's hard to "extract" it without bringing chaos to PR's diff, so I decided not to touch original [__run](https://github.com/scalyr/scalyr-agent-2/blob/0b32d3393b741977ede8d0d22f2de71ec0948a8b/scalyr_agent/agent_main.py#L1589) method and create new method [__new_run](https://github.com/scalyr/scalyr-agent-2/blob/0b32d3393b741977ede8d0d22f2de71ec0948a8b/scalyr_agent/agent_main.py#L1440) which has everything that original `__run` has but without "config change check" code. If everything is OK, before the final merge, I just remove the original `__run` method and replace it with `__new_run`.  The change itself may seem very debatable, but I believe that it has to improve readability of the agent_main code and also has to make it easier to add new similar checks to the agent without having to be in sync with the "config change check". Please let me know if there are any problems with reviewing that.

I also added new end to end test that verifies new logic https://github.com/scalyr/scalyr-agent-2/blob/0b32d3393b741977ede8d0d22f2de71ec0948a8b/tests/end_to_end_tests/test_from_source/test_from_source.py#L220.

I found it as a good starting point towards a new final pytest based testing framework for smoke/end2end/whatever tests. The idea is to have one directory for all pytest-based end-to-end tests. I named it `tests/end_to_end_tests`. For now there are just tests called `test_from_source` and they have to replace `standalone` smoke tests and other poorly organised tests with the same purpose.  For now, as I said, those tests fully replicates the [standalone smoke test](https://github.com/scalyr/scalyr-agent-2/blob/master/.circleci/smoketest_standalone.sh), but there's still some work required to get rid of [those test](https://github.com/scalyr/scalyr-agent-2/tree/master/tests/smoke_tests).  Old tests won't be touched and will be removed only with `python2` agent packages, but I want to make one single place for all end-to-end tests when frozen binaries are released.


.github/actions/perform-deployment/action.yml
```diff
@@ -7,6 +7,14 @@ inputs:
     description: "Name of the deployment"
     required: true
 
+  cache-key-runner-part:
+    description: | 
+      Part of a cache key that represents system and architecture of the current runner. This is needed to provide 
+      prevent cache collisions for the sema step which runs on different environments.
```
Kami, line 13: sema step? I assume that's a typo :)

agent_build/tools/common.py
```diff
@@ -45,6 +46,16 @@ def init_logging():
     )
 
 
+# Make shlex.join for Python 3.7
```
Kami, line 49: Should we just run those tests under Python 3.8 (same version we use for Windows binaries and Docker images)?

scalyr_agent/configuration.py
```diff
@@ -3905,6 +3905,15 @@ def __verify_monitor_entry_and_set_defaults(
             monitor_entry, "log_path", module_name + ".log", description
         )
 
+        # Verify option which tells that agent has to fail if that monitor stops.
+        self.__verify_or_set_optional_bool(
+            monitor_entry,
+            field="stop_agent_if_fails",
```
Kami, line 3911: Perhaps ``stop_agent_on_failure`` is a better name

tests/credentials_template.json
```diff
@@ -0,0 +1,12 @@
+// This file can be used in local development by specifying needed credentials and other options for the tests.
+// Copy it with new name "credentials.json" and tests will look in it."
+
+// NOTE: also remove those invalid comments.
+{
+  "scalyr_api_key": "",
```
Kami, line 6: I assume you used JSON and not YAML since I just removed pyyaml dependency the other day since it was unused? :)

tests/end_to_end_tests/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+from _pytest.runner import pytest_runtest_protocol as orig_pytest_runtest_protocol
+
+from agent_build.tools.constants import SOURCE_ROOT
+
+IN_CICD = os.environ.get("AGENT_BUILD_IN_CICD", False)
+
+
+def pytest_runtest_protocol(item, nextitem):
+    """
+    Wrap existing pytest protocol to print special grouping directive that
```
Kami, line 36: 👍

tests/end_to_end_tests/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+    makes logs of each test case collapsable on GitHub Actions.
```
Kami, line 37: One option would be to explicitly check for GHA here, but I think it's also fine since it's unlikely we will switch CI system in the near term.

Kami:
Do you think we could add Kubernetes end to end tests specifically for this issue? Or would it be to hard to prepare / mock the environment to replicate that issue?

scalyr_agent/agent_main.py
```diff
@@ -1069,7 +1069,527 @@ def __get_log_files_initial_positions(self, only_new_files=False):
...
+                    if (
+                        time.time()
+                        >= last_config_change_check_time
+                        + self.__config.config_change_check_interval
+                    ):
+                        next(check_config_for_change_gen)
+                        last_config_change_check_time = time.time()
+
+                    # Check for essential monitors every 5 sec.
+                    if time.time() >= last_essential_monitors_check_time + 5:
```
Kami, line 1533: Is there a chance that the code above could take 5 seconds or longer and we would always end up checking those essential monitors and code busy waiting / 100% cpu usage?

I don't think so, but wanted to double check.

scalyr_agent/agent_main.py
```diff
@@ -1069,7 +1069,527 @@ def __get_log_files_initial_positions(self, only_new_files=False):
...
+            self.__controller.emit_init_log(log, self.__config.debug_init)
+
+            self.__start_or_stop_unsafe_debugging()
+
+            raw_scalyr_server = self.__config.raw_scalyr_server
+            self.__print_force_https_message(
+                self.__config.scalyr_server, raw_scalyr_server
+            )
+
+            last_config_change_check_time = (
```
Kami, line 1507: This double assignment with parenthesis got me a bit confused at first, but I assume black forces parenthesis so it's fine :)

scalyr_agent/configuration.py
Kami: I assume we will define k8s monitor as essential as part of the Kubernetes deployment config, right?

Also, I wonder if simply ``essential`` may be a better config option name (although, on the other hand, I kinda prefer explicit nature of the ``stop_agent_on_failure`` option name).

Kami:
I do agree that the main run loop is quite large and messy - eventually we should try to refactor into into multiple smaller functions, but that's not a blocker here.

scalyr_agent/configuration.py
ArthurKamalov: > I assume we will define k8s monitor as essential as part of the Kubernetes deployment config, right?
correct,  I just have next coming PR where I extend pytest ent-to-end tests with docker end k8s tests, and I want to include that there.

> Also, I wonder if simply essential may be a better config option name (although, on the other hand, I kinda prefer explicit nature of the stop_agent_on_failure option name).

yes, I am also on the fence here, and I also thought that explicit naming would be easier for customers understanding.

tests/credentials_template.json
ArthurKamalov: I guess I just didn't see any problem with using json this time :) I even didn't touch existing similar YAML config in our other tests, since I really want to move everything in new end to end tests and them gone.

scalyr_agent/agent_main.py
ArthurKamalov: It seems that it shouldn't, since we call new `time.time()` on each check, so we don't give any chance for the usage of the outdated timestamp.

ArthurKamalov:
> Do you think we could add Kubernetes end to end tests specifically for this issue? Or would it be to hard to prepare / mock the environment to replicate that issue?

As I mentioned in one of the comments, I working on next portion of tests for k8s and docker and I want to put that kind of test there. But yeah I am still trying to find how to simulate that kind of failures on K8s/minikube

scalyr_agent/agent_main.py
ArthurKamalov: Yeah, I guess it's better to do a reformatting only after the review :)

agent_build/tools/common.py
ArthurKamalov: I actually started thinking that besides frozen binaries, we may want to provide a "last stand" tarball with source code, as we do it now, so it would be better to at least comply with currently supported version of the python. WDYT?

ArthurKamalov:
Added needed changes and also modified duplicate job skipping in end-to-end tests workflow. Now, since we also have another jobs in that workflow (smoketests migrated by Dominic, new tests from this pr), we may want not skipping them as we do with k8s, openmetrics e2e tests, so the workflow triggering behaviour should be:

- on non-master branch: all tests excluding  k8s e2e tests.
- on master branch: all tests including k8s e2e tests.
- on PR against master: all tests including k8s e2e tests. Since there may be double triggering if we push commit on branch which has a PR, then the `skip-duplicates` job has to cancel the 'push' event because it does not have e2e tests (which we want to see during the PR)

scalyr_agent/scalyr_monitor.py
```diff
@@ -212,6 +212,13 @@ def __init__(
 
         StoppableThread.__init__(self, name="metric thread")
 
+    @property
+    def stop_agent_on_failure(self):
```
Kami, line 216: Should we also have a method of setting this programmatically via code?

Imagine a scenario where some monitors manager / other codes decides, based on some criteria that monitor should be treated as essential.

I know we don't really have good patterns establishes for that yet - e.g. hoes does "code vs config" precedence works, but we do have some use cases like that already in the wild, notably Kubernetes Explorer and related monitors - those monitors should only be be treated as essential in case a specific environment is set.

I guess technically this could be achieved by overriding this method on this ScalyrMonitor class, so it's probably fine.

tests/end_to_end_tests/pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+log_cli = True
+log_level = INFO
```
Kami, line 3: I assume this can be overridden on per pytest run invocation basis in case it's needed?

tests/end_to_end_tests/tools.py
```diff
@@ -0,0 +1,92 @@
...
+        output = self._check_output_command(
+            ["status", "-v", "--format", "json"]
+        ).decode()
+        return json.loads(output)
+
+    @property
+    def is_running_and_healthy(self) -> bool:
+        try:
+            self._check_output_command(["status"]).decode()
+        except subprocess.CalledProcessError as e:
```
Kami, line 85: Should we also check for non-zero status code and return False in case non zero is returned?

tests/end_to_end_tests/verify.py
```diff
@@ -0,0 +1,309 @@
...
+            to_fail = True
+
+            # There is an issue with dns resolution on GitHub actions side, so we skip some of the error messages.
+            connection_error_mgs = '[error="client/connectionFailed"] Failed to connect to "https://agent.scalyr.com" due to errno=-3.'
+
+            if connection_error_mgs in messages:
+                # If the traceback that follows after error message contains particular error message,
+                # then we are ok with that.
+                errors_to_ignore = [
+                    "socket.gaierror: [Errno -3] Try again",
```
Kami, line 80: I believe this code is duplicated in another file, right?

If so, it would be good if we could refactor some of it into some common module / similar.

tests/end_to_end_tests/verify.py
```diff
@@ -0,0 +1,309 @@
...
+        # Ensure CA validation is not disabled with default install
+        assert "sslverifyoff" not in agent_log_content
+        assert (
+            "Server certificate validation has been disabled" not in agent_log_content
+        )
+
+    check_agent_log_for_errors(content=agent_log_content)
+
+    log.info("Wait for agent log requests stats.")
+    while not check_requests_stats_in_agent_log(content=get_agent_log_content()):
```
Kami, line 235: It seems like there is no timeout for this while loop. Or did I miss something?

tests/end_to_end_tests/verify.py
```diff
@@ -0,0 +1,309 @@
...
+        time.sleep(1)
+
+    log.info(
+        "Verify that previously written test log file content has been uploaded to server."
+    )
+    while True:
+        resp = ScalyrQueryRequest(
```
Kami, line 242: Same here, I also don't see any timeout for this while loop, but maybe I missed it.

tests/end_to_end_tests/verify.py
ArthurKamalov: Yes, I just got it from the tests/package_tests, but I plan to remove and replace them with these tests. I don't want to get any dependencies from here, since it has many OOP abstractions that only add extra complexity and I really not happy about them now...

scalyr_agent/scalyr_monitor.py
ArthurKamalov: I guess we can at least introduce instance attribute so we can manipulate it during initialization.

tests/end_to_end_tests/pytest.ini
ArthurKamalov: Yes, CLI arguments such as `--log-format`, `--log-level` has to override this config options. I haven't checked, but I believe that those ini files also have to respect the scope where they are located, so this particular file shouldn't have any effect outside of the tests/end-to-end-tests folder.

tests/end_to_end_tests/tools.py
ArthurKamalov: subprocess `check_call` and `check_output` already assume non-zero exit code if they raise an error.

Agent `status` command returns with non-zero exit code when agent is not running.

tests/end_to_end_tests/verify.py
ArthurKamalov: Ah yes, since all those helper functions are meant to be used inside `pytest` test cases, we just put timeouts to cases themselves. I just forgot to do that, thanks for reminder :)

tests/end_to_end_tests/tools.py
ArthurKamalov: PS. I also made it just `is_running` because I wanted to add the health check here, but then I started thinking that it has to be out of the scope of the method.

tests/end_to_end_tests/verify.py
ArthurKamalov: https://github.com/scalyr/scalyr-agent-2/pull/951#discussion_r935285122

agent_build/tools/common.py
Kami: If it's not tons of compatibility code I think it should be fine.

tests/end_to_end_tests/conftest.py
ArthurKamalov: yes, I think we can change that whenever we want, for now if should be any big difference, since we are already putting out existing "IN_CI_CD" flag variable.

scalyr_agent/agent_main.py
```diff
@@ -1069,7 +1069,527 @@ def __get_log_files_initial_positions(self, only_new_files=False):
...
+            )
+            if stopped_monitor.stop_agent_on_failure:
+                raise Exception(
+                    "Monitor '{}' with short hash '{}' is not running, stopping the agent because it is configured "
+                    "not to run without this monitor.".format(
+                        monitor_status.monitor_name, monitor_status.monitor_short_hash
+                    )
+                )
+
     def __run(self, controller):
```
Kami, line 1589: So you will remove this method just before merging and rename ``__new_run`` to ``__run``?

scalyr_agent/scalyr_monitor.py
```diff
@@ -121,6 +121,13 @@ def __init__(
         # The logger instance that this monitor should use to report all information and metric values.
         self._logger = logger
         self.monitor_name = monitor_config["module"]
+
+        # Flag that indicates that the agent can not run without this monitor and
+        # has to fail if monitor fails.
+        # The default value (False) of this config option has to be already set for all monitors,
+        # but also add it here for extra safety.
+        self.stop_agent_on_failure = monitor_config.get("stop_agent_on_failure", False)
```
Kami, line 129: I would actually still prefer this to be a getter method since it gives us more flexibility in overriding that logic in sub classes (I know that technically we can also use ``@property``, but ``get_stop_agent_on_failure()`` is a bit more explicit.

tests/end_to_end_tests/verify.py
Kami: So to clarify, this will be refactored in a future PR?

Kami:
@ArthurKamalov How do we plan to handle this?

> I assume we will define k8s monitor as essential as part of the Kubernetes deployment config, right?

I think that probably for now, we should leave it as is, merge this PR, deploy it on our environments for a while (with that monitor config option set to ``True``) and then if everything looks fine, update default value for this monitor config option as part of configs which are included with Docker images for Docker + Kubernetes deployments.

WDYT?

scalyr_agent/scalyr_monitor.py
Kami: EDIT: I had a look and it appears existing base ``ScalyrMonitor`` uses both patterns - ``get_()`` method and ``@property`` decorator.

So as far as consistency goes, both methods would probably work, but getter method is probably preferred since it's much more widely uses than ``@property`` (e.g. ``get_user_agent_fragment()``, ``get_calculate_rate_metric_names()``, etc.).

Kami (approved):
I've added a couple of more comments and questions, but mostly LGTM 👍

.github/workflows/end_to_end_tests.yml
```diff
@@ -519,3 +538,43 @@ jobs:
...
+
+      - name: Perform the deployment of the test environment
+        uses: ./.github/actions/perform-deployment
+        with:
+          deployment-name: "test_environment"
+
+      - name: Run tests
+        shell: bash
+        run: |
+          python3 -m pytest  tests/end_to_end_tests/test_from_source/test_from_source.py \
```
Kami, line 575: It would be good to also track and upload coverage to codecov.io for this job as well.

It's fine to implement that in a separate PR.

scalyr_agent/agent_main.py
ArthurKamalov: Yes, removed, thanks.

tests/end_to_end_tests/test_from_source/test_from_source.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    test_monitors_path = pl.Path(__file__).parent / "fixtures"
+
+    default_config.update(
+        {
+            "additional_monitor_module_paths": str(test_monitors_path),
+            "monitors": [
+                {
+                    "module": "monitors.essential_failing_monitor",
+                    "stop_agent_on_failure": True,
```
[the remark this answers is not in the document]
ArthurKamalov: Done. Thanks for pointing on that, because I found bug in how I raise error in the dummy monitor.

ArthurKamalov:
> I think that probably for now, we should leave it as is, merge this PR, deploy it on our environments for a while (with that monitor config option set to True) and then if everything looks fine, update default value for this monitor config option as part of configs which are included with Docker images for Docker + Kubernetes deployments.
WDYT?

sounds good to me.

.github/workflows/end_to_end_tests.yml
ArthurKamalov: Yes,  maybe create another action for this too.

tests/end_to_end_tests/test_from_source/test_from_source.py
```diff
@@ -261,20 +262,109 @@ def test_with_failing_essential_monitor(
...
+            "monitors": [
+                {
+                    "module": "monitors.essential_failing_monitor",
+                    "id": "not_essential",
+                    "stop_agent_on_failure": False,
+                },
+                {
+                    "module": "monitors.essential_failing_monitor",
+                    "id": "not_essential_default",
+                },
```
Kami, line 317: 👍

Kami (approved):

tests/unit/agent_main_test.py
```diff
@@ -579,21 +603,116 @@ def test_no_log_messages_printed_on_config_parse_on_status_stop(self):
...
-        with open(tmp_path, "wb") as fp:
-            fp.write(json.dumps(config_data).encode("utf-8"))
+        # 2. MSI install (only intermediate_certs_path check should be skipped)
+        with mock.patch(
+            "scalyr_agent.__scalyr__.INSTALL_TYPE", __scalyr__.PACKAGE_INSTALL
+        ):
 
-        return tmp_path
+            config_data = {
+                "api_key": "0v7dW1EIPpglMcSjGywUdgY9xTNWQP/kas6qHEmiUG5w-",
```
Kami, line 621: I assume this is not a valid api key? If it is, please make sure it's revoked :)

tests/unit/configuration_test.py
```diff
@@ -2520,6 +2520,49 @@ def test__verify_required_attributes(self):
...
+    def test_monitor_stop_agent_on_failure_option(self):
+        """
+        Test the 'stop_agent_on_failure' option.
+        By the default it has to be False for all monitor configs.
+        """
+
+        self._write_file_with_separator_conversion(
+            """
+            {
+                "api_key": "0v7dW1EIPpglMcSjGywUdgY9xTNWQP/kas6qHEmiUG5w-",
```
Kami, line 2532: Same here - re api key.

PR #952, opened 2022-08-01
Update bundled CA certs and add additional LetsEncrypt root CA cert

Kami:
This pull request updates CA certifications we bundle to verify the API endpoint certificate. It adds new LetsEncrypt root certificate which was not part of the bundled certs yet - https://letsencrypt.org/certificates/.

Some of our environments use LetsEncrypt issued certificates so we should bundle this root to avoid surprises and issues.

While at it, I also renamed the existing certifications to use more descriptive names - I don't think this should cause issues since the sort order is still the same.


ArthurKamalov (approved):

ArthurKamalov:
Sorry, missed this PR.

PR #953, opened 2022-08-01
Log server SSL certificate in PEM format on certificate validation failure

Kami:
This pull request updates the code to log the actual server certificate in PEM format to make troubleshooting easier when cert validation fails (so we can see which certificate is returned by the server in case of server side misconfiguration or similar).

## Note on Implementation Limitations

Sadly with primitives offered / exposed by Python, there is no way for us to retrieve server certificate from the existing established socket / connection once validation fails.

This means that in case we want to log server certificate we need to establish a new connection / socket for logging purposes only with disabled cert validation and retrieve certificate for logging purposes only from that socket. Again, this short lived socket is used purely for logging purposes and we intentionally disable certificate validation for this code path.

This approach is not 100% ideal since a new connection is created and we may miss some edge cases since on new connection, server side could return a different certificate (e.g. server side misconfiguration, being routed to a different LB, etc.), but we can't do much about it using APIs exposed by Python. So overall, won't cover 100% edge cases, but better than nothing.


scalyr_agent/connection.py
```diff
@@ -445,6 +445,63 @@ def connect(self):
...
+                        sock, do_handshake_on_connect=True, server_hostname=self.host
+                    )
+
+                else:
+                    sock = ssl.wrap_socket(sock, cert_reqs=ssl.CERT_NONE)
+
+                # NOTE: In case binary_form is False and validation wasn't perform an empty dict
+                # will be returned so we need to use binary_form to still get cert back for logging
+                # purposes - https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.getpeercert
+                der_cert = sock.getpeercert(binary_form=True)
```
ArthurKamalov, line 494: Shouldn't we guard it from possible ValueError exception, which is mentioned in docs?

scalyr_agent/connection.py
```diff
@@ -445,6 +445,63 @@ def connect(self):
...
+
+            try:
+                sock = socket.create_connection((self.host, self.port), self.__timeout)
+
+                if py_post_equal_279:
+                    ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
+                    ssl_context.options |= ssl.OP_NO_SSLv2
+                    ssl_context.options |= ssl.OP_NO_SSLv3
+                    ssl_context.options |= ssl.OP_NO_TLSv1
+                    ssl_context.options |= ssl.OP_NO_TLSv1_1
```
ArthurKamalov, line 480: Not sure if it worth doing, but I guess we can extract this to a common function, so we can exclude that ssl_context is updated in connection code, but remains outdated here, in log method.

scalyr_agent/connection.py
Kami: oops, sorry, yeah, good catch - i thought I also added ``except`` clause, but I only added ``finally``. Will fix it and also add ``except Exception``.

scalyr_agent/connection.py
Kami: I was a bit torn here -  I'm a bit worried that someone will want to make a change which only affects logging code but it will then also affect main code path aka the non logging code.

Granted we have many tests which should catch this, but still :)

ArthurKamalov: Yea, haven't thought about that, that's definitely may be the case.

ArthurKamalov (approved):

Kami:
Per discussion on Slack - I will also add a change to log server certificate on successful connection using DEBUG log level.

PR #954, opened 2022-08-03
Add new build k8s kill on monitors death

ArthurKamalov:
This PR creates new temporary build `k8s-restart-agent-on-monitor-death` which has enabled config option `stop_agent_on_failure` for the Kubernetes monitor, which has to provide additional safety if Kubernetes monitor fails.


agent_build/package_builders.py
```diff
@@ -1156,3 +1179,21 @@ class DockerApiPackageBuilder(ContainerPackageBuilder):
         base_image_deployment_step_cls=deployments.BuildAlpineDockerBaseImageStep,
     )
 )
+
+# Those builds for a temporary testing of the "stop_agent_on_failure" option for the k8s image build.
+K8S_CONTAINER_RESTART_AGENT_ON_MONITOR_DEATH_DEBIAN = K8sRestartAgentOnMonitorsDeath(
+    name="k8s-restart-agent-on-monitor-death-debian",
+    config_path=_CONFIGS_PATH / "k8s-config",
+    base_image_deployment_step_cls=deployments.BuildDebianDockerBaseImageStep,
+    additional_config_paths=[
```
Kami, line 1188: Ah, OK, you added support for additional config paths - I thought you will just copy over the contents from https://github.com/scalyr/scalyr-agent-2/tree/master/docker/k8s-config/agent.d and then modify docker.json, but this approach works as well :)

agent_build/package_builders.py
```diff
@@ -343,15 +343,24 @@ def _install_info_str(self) -> str:
...
         # Copy config
-        shutil.copytree(config_source_path, output_path)
+        shutil.copytree(base_config_source_path, output_path)
+
+        # Copy additional config files.
+        for a_config_path in additional_config_paths or []:
+            shutil.copytree(a_config_path, output_path, dirs_exist_ok=True)
 
         # Make sure config file has 640 permissions
         config_file_path = output_path / "agent.json"
```
Kami, line 366: We should probably also do the same for other config files (set permissions to 640 for all config fragments in base_config_source_path), but that's not a blocker for this PR and just an improvement over the existing behavior.

Kami (approved):
LGTM, thanks 👍

agent_build/package_builders.py
ArthurKamalov: Those files also have to be checked, because that checking code goes right after :)

https://github.com/scalyr/scalyr-agent-2/blob/54e52da290fa263adf8fd6a032ccb1ea01779837/agent_build/package_builders.py#L361-L379

PR #956, opened 2022-08-05
CT-575 Create lint GHA workflow

dlobue:
This is a direct port of the circleci lint job. I'm going to restrain myself from adding fancy frills like reports. There are too many other things that need to be ported for me to go nuts with polish on each one. I can iterate on these later. 

CT-575


ArthurKamalov:
LGTM, for a first iteration and for moving from CircleCI.

Kami (approved):

PR #958, opened 2022-08-09
CT-578 Add codespeed workflow for GHA

dlobue:
Adds codespeed workflow and micro-benchmark job to GHA.

The value of `CODESPEED_ENVIRONMENT` does not match what is used in CircleCI. I did this because the hardware configuration is different, so I thought it made sense to reflect that. I had to create the Environment database record by hand in the codespeed django admin interface.

CT-578

### DIFF
--- a/.github/workflows/codespeed.yml
+++ b/.github/workflows/codespeed.yml
@@ -63,12 +63,12 @@ jobs:
         run: tox -e micro-benchmarks
 
       - name: Process and submit results to CodeSpeed
-        # if: (github.ref == 'refs/heads/master' || github.head_ref == 'master')
+        if: (github.ref == 'refs/heads/master' || github.head_ref == 'master')
         env:
           PYTHONPATH: .
           CODESPEED_URL: "https://scalyr-agent-codespeed.herokuapp.com/"
           CODESPEED_PROJECT: "scalyr-agent-2-microbenchmarks"
-          CODESPEED_EXECUTABLE: ${{ steps.setup-python.outputs.python-version }}
+          CODESPEED_EXECUTABLE: Python ${{ steps.setup-python.outputs.python-version }}
           CODESPEED_ENVIRONMENT: GitHub Hosted Action Runner
           CODESPEED_AUTH: ${{ secrets.CODESPEED_AUTH }}
         run: python benchmarks/scripts/send_microbenchmarks_data_to_codespeed.py --data-path="benchmark_results/*.json" --debug
--- a/.github/workflows/codespeed.yml
+++ b/.github/workflows/codespeed.yml
@@ -7,6 +7,8 @@ on:
   pull_request:
     branches:
       - master
+  schedule:
+    - cron: '0 4 * * *'
 
 jobs:
   # Special job which automatically cancels old runs for the same branch, prevents runs for the
--- a/.github/workflows/codespeed.yml
+++ b/.github/workflows/codespeed.yml
@@ -48,13 +48,16 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install tox
+      - name: Install Deps
         env:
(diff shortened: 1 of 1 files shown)


ArthurKamalov:
@dlobue I am not aware of best practices here, and I may be missing a good reason behind this,  but what about moving all "code analysis" related tests (such as lint testing https://github.com/scalyr/scalyr-agent-2/pull/956) into one single workflow?

Kami:
> The value of CODESPEED_ENVIRONMENT does not match what is used in CircleCI. I did this because the hardware configuration is different, so I thought it made sense to reflect that. 

Thanks.

I will also go ahead and update database fixtures in the codespeed repo so they are in sync with what's in the database ("technically" we follows infra as code approach there as well and all the db fixtures are committed to git repo and auto deployed on heroku push).

.github/workflows/codespeed.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Process and submit results to CodeSpeed
+        if: (github.ref == 'refs/heads/master' || github.head_ref == 'master')
+        env:
+          PYTHONPATH: .
+          CODESPEED_URL: "https://scalyr-agent-codespeed.herokuapp.com/"
+          CODESPEED_PROJECT: "scalyr-agent-2-microbenchmarks"
+          CODESPEED_EXECUTABLE: Python ${{ steps.setup-python.outputs.python-version }}
+          CODESPEED_ENVIRONMENT: GitHub Hosted Action Runner
+          CODESPEED_AUTH: ${{ secrets.CODESPEED_AUTH }}
+        run: python benchmarks/scripts/send_microbenchmarks_data_to_codespeed.py --data-path="benchmark_results/*.json" --debug
```
Kami, line 74: Per comment on Slack / Zoom the other day, we have multiple sets of benchmarks tracked in CodeSpeed:

1. micro benchmarks (what you have migrated here)
2. agent process level "end to end" benchmarks (where we run complete agent process with various different configurations and record various resource utilization metrics)

Do you plan to migrate 2) as part of this PR or as part of a different one?

dlobue: @Kami I am migrating 2) as part of CT-577. I broke them into separate tasks to try and keep PR sizes down (among other reasons).

dlobue:
> > The value of CODESPEED_ENVIRONMENT does not match what is used in CircleCI. I did this because the hardware configuration is different, so I thought it made sense to reflect that.
> 
> Thanks.
> 
> I will also go ahead and update database fixtures in the codespeed repo so they are in sync with what's in the database ("technically" we follows infra as code approach there as well and all the db fixtures are committed to git repo and auto deployed on heroku push).

Okay. If you haven't done that yet, I'm happy to handle that.

dlobue:
> @dlobue I am not aware of best practices here, and I may be missing a good reason behind this, but what about moving all "code analysis" related tests (such as lint testing #956) into one single workflow?

The number one reason I didn't do this originally was to prevent merge conflicts. I plan on iterating on both the linting workflow and the codespeed workflow shortly, and it would be helpful for them to remain separate for the time being.

Kami:
I just updated data fixture in codespeed repo, but I'm still working on heroku push related failures (looks like there are some breaking changes in upstream libraries.... 🤷  )

dlobue:
> I just updated data fixture in codespeed repo, but I'm still working on heroku push related failures (looks like there are some breaking changes in upstream libraries.... 🤷 )

There's been no code change though, right? The heroku push is basically a no-op, right? Why bother?

Kami:
@dlobue heroku push also involves a deploy which also runs django command which loads a fixture....

Kami (approved):

Kami:
Confirmed data for micro benchmarks is making it to CodeSpeed instance

Screenshot 2022-08-16 at 09 35 01

PR #961, opened 2022-08-10
Remove Circle CI jobs which have been been migrated to Github Actions

Kami:
Now that unit tests, lint checks and smoke test jobs have been moved from Circle CI to Github Actions, we can remove those, now redundant Circle CI job.

Initially we thought we may want to keep both set of jobs in place until frozen binaries are in place, but frozen binaries may take a while and things are getting unwieldy with 80+ github status checkes - not to mention that ported jobs are not exactly 1:1 mappings (there are some changes in pytest targets, etc.) which means it will be a pain to keep both set of jobs in sync.

This PR removes jobs which are not a hard dependency and strictly needed for Jenkins release pipeline (only hard dependency for release pipeline are jobs which build deb, rpm and .exe artifacts + ami tests).

## TODO

- [x] Update corresponding Github status checks for master branch protection


Kami:
@dlobue While working on this change I think I may have missed some things during the review of the corresponding PR. I think we should make sure those things are also migrated:

1. For all the jobs, step which sends a notification to Slack on job failure for master branch runs (we already have existing pattern for that for Kubernetes end to end tests - just need to very incoming slack webhook secret is still active since it got revoked / removed multiple times in the past)
2. Cron trigger which runs all those migrated jobs on a daily basis - those daily runs are useful and catch various issues with upstream dependencies changing and similar.

In addition to that, I'm trying to figure out what happened / what's the plan for the following smoke test jobs which didn't seem to be migrated as part of #939 - if I missed something, please let me know:

* ``smoke-rpm-package-*``
* ``smoke-deb-package-*``
* ``smoke-standalone-*``
* ``smoke k8s``
* ``smoke-monitors-tests``

Until those are migrated, I left them in Circle CI.

Kami:
@dlobue It appears that the code coverage has dropped after removing Circle CI jobs which would indicate that perhaps not all the code coverage steps have been correctly migrated. Can you please have a look when you get a chance?

.circleci/config.yml
```diff
@@ -1504,153 +943,6 @@ jobs:
...
-            git status -- *docs/monitors/*.md */docs/monitorsDmd | (grep -q "nothing to commit" || (echo "Auto-generate monitor doc files are not up to date. Make sure you run tox -e generate-monitor-docs and commit any changed files." && exit 1))
-      - save_cache:
-          key: deps2-tox-{{ .Branch }}-lint-venv-{{ checksum "dev-requirements.txt" }}-{{ checksum "lint-requirements.txt" }}
-          paths:
-            - "~/.cache/pip"
-      - slack/status:
-          fail_only: true
-          only_for_branches: master
-
-  sonarcloud-scan:
```
ArthurKamalov, line 1570: it seems that we only keep the `codecov` for now, right?

dlobue:
@Kami 

> @dlobue While working on this change I think I may have missed some things during the review of the corresponding PR. I think we should make sure those things are also migrated:
> 
>     1. For all the jobs, step which sends a notification to Slack on job failure for master branch runs (we already have existing pattern for that for Kubernetes end to end tests - just need to very incoming slack webhook secret is still active since it got revoked / removed multiple times in the past)
> 
>     2. Cron trigger which runs all those migrated jobs on a daily basis - those daily runs are useful and catch various issues with upstream dependencies changing and similar.

I wanted to do those last originally. I'll create a PR shortly to add that functionality.

> 
> 
> In addition to that, I'm trying to figure out what happened / what's the plan for the following smoke test jobs which didn't seem to be migrated as part of #939 (if I missed something, please let me know):
> 
>     * `smoke-rpm-package-*`

https://scalyr.myjetbrains.com/youtrack/issue/CT-583/Migrate-packaging-build-test-smoke-for-rpm-deb-from-CircleCI-to-GHA

> 
>     * `smoke-deb-package-*`

https://scalyr.myjetbrains.com/youtrack/issue/CT-583/Migrate-packaging-build-test-smoke-for-rpm-deb-from-CircleCI-to-GHA

> 
>     * `smoke-standalone-*`

https://scalyr.myjetbrains.com/youtrack/issue/CT-579/Migrate-tox-standalone-smoke-tests-from-CircleCI-to-GHA

> 
>     * `smoke k8s`

You mean this one? https://github.com/scalyr/scalyr-agent-2/blob/master/.github/workflows/end_to_end_tests.yml#L400-L472

> 
>     * `smoke-monitors-tests`

https://scalyr.myjetbrains.com/youtrack/issue/CT-579/Migrate-tox-standalone-smoke-tests-from-CircleCI-to-GHA

> 
> 
> Until those are migrated, I left them in Circle CI.

dlobue:
> @dlobue It appears that the code coverage has dropped after removing Circle CI jobs which would indicate that perhaps not all the code coverage steps have been correctly migrated. Can you please have a look when you get a chance?


Will do.

.circleci/config.yml
```diff
@@ -2366,26 +1622,12 @@ workflows:
      - smoke-standalone-35-rate-limit:
          context: scalyr-agent
       # NOTE: smoke test jobs below still use old smoke tests framework
-     - smoke-docker-json:
-         context: scalyr-agent
      - smoke-docker-api-raw-logs-disabled:
          context: scalyr-agent
```
dlobue, line 1626: This one should be able to be disabled. Here's a link to a run of the GHA version: https://github.com/scalyr/scalyr-agent-2/runs/7770418121?check_suite_focus=true

.circleci/config.yml
```diff
@@ -2475,32 +1717,7 @@ workflows:
...
-           branches:
-             only:
-               - win_py30_changes
-   jobs:
-     - ami-tests-development-windows:
-         context: scalyr-agent
-         requires:
-            - build-windows-package
-     - build-windows-package
-     - build-linux-packages-and-installer-script
```
dlobue, line 2503: There is no equivalent for these yet.

dlobue:
FYI I figured out the cause of the change in coverage. I'm working on a fix, but are the problems:
- The job that converts the coverage data to xml did not have access to the source code (which is necessary for it to work)
- The coverage combine job did not have the coveragerc config file
- The coveragerc config file does not have all necessary source paths listed, so combine fails
- The codecov config does not have a path fix for code from GHA runners

That's why coverage is broken. I also found other misc problems with our configuration:
- python stdlib is being measured for coverage

Kami:
@dlobue yeah, iirc we also had some path processing code in in the coverage report files since we run coverage on various platforms and setups and paths are not always the same in the coverage report files and combine doesn't handle that so we need to process the paths to "normalize" them to be the same so the coverage report is correct.

.circleci/config.yml
Kami: Thanks - I missed this one + smoke-k8s one. Will remove it from Circle CI config.

.circleci/config.yml
Kami: Yeah, that was actually a left over build from the past which we forgot to remove it (it was only triggered for ``win_py30_changes`` branch).

We have another target which basically does the same for ``master`` branch.

Kami:
@dlobue Thanks for addressing coverage issue. I will go ahead and merge this PR (+ update status checks once it has been merged).

Kami:
@dlobue Hm, looks like coverage dropped by ~3% or so after removing ``smoke-k8s`` and ``smoke-docker-api-raw-logs-disabled`` jobs from Circle CI which has apparently already been migrated to GHA.

When you get a chance, can you please look into it and verify we indeed capture coverage from those GHA jobs? Thanks.

Kami:
I also removed micro benchmarks targets which have been migrated to GHA.

I will merge this PR into master to avoid mess with so many status checks before the release, but we should still track down code coverage discrepancy.

Kami:
Something weird is going on - now the coverage report shows -68% coverage and all the jobs which upload coverage completed.

/cc @dlobue

ArthurKamalov (approved):
LGTM

PR #964, opened 2022-08-11
Log warning message in case of unknown connection error

ArthurKamalov:
Make this exception log message as warning to calm down our tests, since they look for any ERROR log messages and this particular error is not fatal.

NOTE: Since it caches unknown error, we still have to provide exception traceback.

### DIFF
--- a/scalyr_agent/scalyr_client.py
+++ b/scalyr_agent/scalyr_client.py
@@ -660,8 +660,9 @@ class ScalyrClientSession(object):
                         error_code="client/requestFailed",
                     )
                 else:
-                    log.exception(
+                    log.warning(
                         "Failed to receive response due to exception.  Closing connection, will re-attempt",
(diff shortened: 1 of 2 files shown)


Kami (approved):
Thanks, LGTM 👍 

Please just add a changelog entry.

scalyr_agent/scalyr_client.py
```diff
@@ -660,8 +660,9 @@ def __receive_response(self, body_str, send_time):
                         error_code="client/requestFailed",
                     )
                 else:
-                    log.exception(
+                    log.warning(
                         "Failed to receive response due to exception.  Closing connection, will re-attempt",
+                        exc_info=True,
```
Kami, line 665: Yes, as you said it yourself - we definitely want to include original stack trace here to provide more context on what is going on.

PR #966, opened 2022-08-12
Experiment: Update pycoverage lib

dlobue:


Kami:
If we can upgrade it, that would be great.

I don't remember the exact reason anymore why we couldn't upgrade it before (IIRC something related to supported Python versions and underlying coverage file format changing which caused issues with something, perhaps combining + path rewritting).

PR #967, opened 2022-08-12
CT-579 Add tox standalone smoketests to GHA

dlobue:
Port over tox standalone smoketests from CircleCI.

Two differences from CircleCI version:
- instead of running the rate-limit tests on python 3.5, I'm running them on 3.10 as 3.5 is EOL and we don't officially support it anymore.
- Caching isn't enabled on the monitor smoketests yet. I want to get all tests ported over to GHA, and then do an optimization pass on everything all at once. I also kind of want to remove that docker wrapper around the monitor tests.

CT-579


.github/workflows/unittests.yml
```diff
@@ -131,6 +134,143 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        shell: bash
+    needs: pre_job
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
```
Kami, line 148: Yeah, eventually we will need to prune down the version matrix since we have like a gazillion of jobs (if not earlier, we will be able to do that once frozen binaries are in place)

Kami:
>  I also kind of want to remove that docker wrapper around the monitor tests.

@ArthurKamalov would that be possible / good idea? I don't remember anymore why we added that, but I'm sure we had a good reason for it (perhaps something to do with the dependencies and making sure all the system / Docker level dependencies are in place for various monitors)

Kami (approved):

dlobue:
> > I also kind of want to remove that docker wrapper around the monitor tests.
> 
> > @ArthurKamalov would that be possible / good idea? I don't remember anymore why we added that, but I'm sure we had a good reason for it (perhaps something to do with the dependencies and making sure all the system / Docker level dependencies are in place for various monitors)
> 
> Yes, just to provide unified environment which does not differ in CI/CD or locally. It was also an attempt to do some caching for CircleCI to reduce overall time.
> 
> @dlobue I would suggest keep those docker wrapped tests in CircleCI for now, because it may be really painful to deal with them, and just wait for new frozen binaries tests.

@ArthurKamalov Yeah, I'm not going to go down that rabbit hole right now. My purpose in bringing up the removal of the docker wrappers was to explain why I didn't want to invest time ATM on getting that docker image cache stuff working (because I might throw it away).

PR #973, opened 2022-08-16
CT-573 Fix coverage issues with smoke tests

dlobue:
This PR is to fix the issue identified in the following comment:  https://github.com/scalyr/scalyr-agent-2/pull/961#issuecomment-1216311694

The problem was the sed processing of the coverage files was creating bad paths. Instead of fixing the issue, I decided to use the same coverage reporting pattern used in the unittests workflow. Changes made:

- Add agent code path in smoke test docker image to .coveragerc and
  codecov fixes config.
- Use same coverage reporting pattern as unittests (as we know that
  workflow works).
- Remove manual processing of coverage files via sed.

CT-573

### DIFF
--- a/.github/workflows/end_to_end_tests.yml
+++ b/.github/workflows/end_to_end_tests.yml
@@ -43,6 +43,38 @@ jobs:
           do_not_skip: "${{ steps.get_events_to_not_skip.outputs.events_to_not_skip }}"
           github_token: ${{ github.token }}
 
+  coverage:
+    runs-on: ubuntu-latest
+    if: success() || failure()
+    needs:
+      - pre_job
+      - k8s-smoketest
+      - docker-smoketest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        id: setup-python
+        with:
+          python-version: 3.8
+      - name: Install pycoverage
+        run: pip install coverage==4.5.4
+      - name: Download coverage reports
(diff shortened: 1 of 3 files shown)


.github/workflows/end_to_end_tests.yml
```diff
@@ -43,6 +43,38 @@ jobs:
           do_not_skip: "${{ steps.get_events_to_not_skip.outputs.events_to_not_skip }}"
           github_token: ${{ github.token }}
 
+  coverage:
```
Kami, line 46: It probably makes sense to eventually refactor those steps which are duplicated across multiple workflows in a separate standalone re-usable GHA action.

Kami (approved):
LGTM, thanks for fixing that.

And yeah, now that we don't use sonarcloud anymore, I don't think we need do to local path fixing anymore, we can just rely on codecov functionality (IIRC, sonarcloud doesn't support that functionality and to make it work there, we needed to fix paths locally ourselves on CI).

.github/workflows/end_to_end_tests.yml
dlobue: Agreed.

dlobue:
> LGTM, thanks for fixing that.
> 
> And yeah, now that we don't use sonarcloud anymore, I don't think we need do to local path fixing anymore, we can just rely on codecov functionality (IIRC, sonarcloud doesn't support that functionality and to make it work there, we needed to fix paths locally ourselves on CI).

FYI the paths are fixed by `coverage combine` with the current config before submission to codecov.

PR #974, opened 2022-08-17
Add docker image builder

ArthurKamalov:
This PR add refactored[ docker image builders](https://github.com/scalyr/scalyr-agent-2/blob/add-refactored-docker-image-builders/agent_build_refactored/docker_image_builders.py)

The main idea remains the same: 
We build base docker image which can be cached by GHA, and the final docker image is built upon this base image.
There's only one difference - the base image now also consist of other, let's say, "child" steps. those child steps build the same base image but just for one particular platform/architecture, so the "parent" step just has to combine them into a resulting, multi-platform base image.  Since each platform image now is built in a separate step, we can run those steps in a separate GHA jobs to reduce overall build time. (Logic that does separate jobs in GHA will be added with updated build workflow in next PR).

Other things that worth mentioning:

- Refactored [build_package_new.py](https://github.com/scalyr/scalyr-agent-2/blob/add-refactored-docker-image-builders/build_package_new_refactored.py) script. As it can be seen now, it does not define command line argument for docker images because docker image builder class is a subclass of the ``Runner``, and runner allows to define CLI by using ``add_command_line_arguments`` and ``handle_command_line_arguments`` methods. (see[ here](https://github.com/scalyr/scalyr-agent-2/blob/2df74d433fbf8ee0732bf6e657b437bbac5bfb1a/agent_build_refactored/docker_image_builders.py#L492))

- Image [Dockerfiles](https://github.com/scalyr/scalyr-agent-2/tree/add-refactored-docker-image-builders/agent_build_refactored/docker/dockerfiles) also mainly remain same, but for the [base image Dockerfile](https://github.com/scalyr/scalyr-agent-2/blob/add-refactored-docker-image-builders/agent_build_refactored/docker/dockerfiles/base.Dockerfile) I extracted all shell commands to a logically separated shell scripts (see them alongside with Dockerfiles). I think this helps to keep Dockerfile more readable, because we don't have to do that 'debian', 'alpine' stage wizardry anymore. It also allows to include those new scripts to a shell lint check.

### DIFF
--- a/agent_build_refactored/docker/dockerfiles/install-python-libs.sh
+++ b/agent_build_refactored/docker/dockerfiles/install-python-libs.sh
@@ -17,7 +17,7 @@ set -e
 
 REQUIREMENTS_FILES="agent_build/requirement-files"
 
-# orjson is wheel is not available for armv7 + musl yet so we exclude it here. We can't exclude it
+# orjson wheel is not available for armv7 + musl yet so we exclude it here. We can't exclude it
 # with pip environment markers since they are not specific enough.
 if [ "$TARGETVARIANT" = "v7" ] && [ "$DISTRO_NAME" = "alpine" ]; then
   sed -i '/^orjson/d' "${REQUIREMENTS_FILES}/main-requirements.txt"
--- a/agent_build_refactored/docker/dockerfiles/install-python-libs.sh
+++ b/agent_build_refactored/docker/dockerfiles/install-python-libs.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+# Copyright 2014-2022 Scalyr Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+REQUIREMENTS_FILES="agent_build/requirement-files"
+
+# orjson is wheel is not available for armv7 + musl yet so we exclude it here. We can't exclude it
+# with pip environment markers since they are not specific enough.
+if [ "$TARGETVARIANT" = "v7" ] && [ "$DISTRO_NAME" = "alpine" ]; then
+  sed -i '/^orjson/d' "${REQUIREMENTS_FILES}/main-requirements.txt"
+fi
+
+
+# Right now we don't support lz4 on server side yet so we don't include this dependency since it doesn't offer pre built
+# wheel and it substantially slows down base image build
+sed -i '/^lz4/d' "${REQUIREMENTS_FILES}/compression-requirements.txt"
+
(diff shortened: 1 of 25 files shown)


agent_build_refactored/docker/dockerfiles/install-python-libs.sh
```diff
@@ -0,0 +1,36 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+REQUIREMENTS_FILES="agent_build/requirement-files"
+
+# orjson is wheel is not available for armv7 + musl yet so we exclude it here. We can't exclude it
```
Kami, line 20: i assume this "is" is redundant here

agent_build_refactored/docker/steps/build_base_image.py
```diff
@@ -0,0 +1,117 @@
...
+            continue
+
+        base_step_output_path = pl.Path(base_step_output_path)
+
+        # Find tarball with platform-specific image.
+        found = list(base_step_output_path.glob(f"{BASE_IMAGE_NAME_PREFIX}*.tar"))
+
+        assert (
+            len(found) == 1
+        ), "Number of result image tarball for each platform-specific builder step has to be 1."
```
Kami, line 66: Perhaps it would be good to also include glob pattern + expanded glob in the assertion error? Or that's not needed since context will already be included with assertion error?

Kami:
>  It also allows to include those new scripts to a shell lint check.

I assume you will do this in a future PR, or?

Also, how do you plan to test this + publishing to docker hub? Will you use testing docker hub account we have?

agent_build_refactored/docker/steps/build_base_image.py
ArthurKamalov: Yeah, it has to be useful to see what's actually going on there. Added.

ArthurKamalov:
> > It also allows to include those new scripts to a shell lint check.
> 
> I assume you will do this in a future PR, or?
> 
> Also, how do you plan to test this + publishing to docker hub? Will you use testing docker hub account we have?

I think it should be ok to add it now)

Yes, CI/CD is in the next PR.

agent_build_refactored/docker_image_builders.py
```diff
@@ -0,0 +1,771 @@
...
+                "BASE_IMAGE_NAME_PREFIX": self.base_image_name_prefix,
+            },
+            github_actions_settings=GitHubActionsSettings(
+                cacheable=True,
+            ),
+        )
+        self.base_distro = base_distro
+        self.supported_platforms = supported_platforms
+
+    def __post_init__(self):
```
Kami, line 182: Do we need this method for anything? I don't see it being overridden anywhere, but maybe I missed it.

ArthurKamalov: woops, sorry. It was forgotten when I tried to make ``RunnerStep`` as a ``dataclass``. Didn't worked well since this class is to big, and "dataclass-style" declaration of it is less readable

Kami (approved):

agent_build_refactored/docker_image_builders.py
ArthurKamalov: PS: The main idea was to make a steps dataclass `frozen` to ensure that checksum calculation for cache  can not be somehow mutated.

PR #975, opened 2022-08-17
CT-577 Add codespeed agent benchmark jobs to GHA

dlobue:
Port codespeed agent benchmark jobs from circleci to GHA.

FYI the job fails unless it is run on the master branch or release branch due to codespeed not accepting results for any other branch.

CT-577

### DIFF
--- a/.github/workflows/codespeed-agent-benchmarks.yaml
+++ b/.github/workflows/codespeed-agent-benchmarks.yaml
@@ -108,6 +108,8 @@ jobs:
           RUN_TIME: ${{ inputs.run_time }}
           CAPTURE_INTERVAL: ${{ inputs.capture_interval }}
           AGENT_CONFIG_FILE: ${{ inputs.agent_config }}
+          SCALYR_SERVER: https://agent.scalyr.com
+          SCALYR_API_KEY: ${{ secrets.SCALYR_API_KEY }}
           SCALYR_SERVER_ATTRIBUTES: "{\"serverHost\": \"${{ inputs.agent_server_host }}\"}"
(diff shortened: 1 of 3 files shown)


.github/workflows/codespeed-agent-benchmarks.yaml
```diff
@@ -0,0 +1,169 @@
...
+        default: false
+        required: false
+      dry_run:
+        description: "True to enable dry run which runs the benchmarks without submitting data to CodeSpeed."
+        type: boolean
+        default: false
+        required: false
+
+jobs:
+  codespeed-reusable-workflow:
```
Kami, line 65: Perhaps agent benchmark or similar would be better name for this job.

Kami: EDIT: I see you give actual jobs which call this re-usable flow better names so it's probably fine :)

Kami (approved):
LGTM 👍

.github/workflows/codespeed-agent-benchmarks.yaml
Kami: EDIT: Looks like job names are not good so I will go ahead and try to rename them / give them better name - https://github.com/scalyr/scalyr-agent-2/actions/runs/2902530969

Kami: Pushed a change to make it a bit better (https://github.com/scalyr/scalyr-agent-2/runs/7948166748?check_suite_focus=true), but it's still not great - not much we can do though with the current UI job name truncation.

PR #977, opened 2022-08-18
Add new Ci/CD workflow with images build

ArthurKamalov:
This PR adds initial CI/CD logic how we build our images (and future packages).

I've decided to do it as a separate, smaller PR to give a chance to understand better the idea that I am using here.
The idea is to generate job matrices outside of the jobs themselves with a [python script](https://github.com/scalyr/scalyr-agent-2/blob/add-ci/cd/agent_build_refactored/scripts/get_github_actions_job_matrices.py). This has to allow us to build job matrices depending on some conditions, for example we can do a "limited" set of jobs only for certain build artefacts
when we are not in the `master` branch or not in the "master-targeted"  pull request.  

**Non-master branch  run:**

image

**Master branch / PR run:**

image

As it can be seen, "limited" run does not include `alpine` based images to reduce amount of jobs during an active development. The same thing we can do with Kubernetes versions to test for k8s images and etc.

This also allows us to achieve another feature that I've partially mentioned in previous relevant PRs - generate matrix for a special job that pre-builds each platform-specific base image for our [docker images.](https://github.com/scalyr/scalyr-agent-2/actions/runs/2884036941)

image

Each platform now is built separately which has to save some time. 

Another thing that may worth mentioning is that images now are stored as GHA artefacts. That allows us to use them as single source of truth for testing and for publishing (In current build logic we build them every time).


ArthurKamalov:
PS: As it was said, this PR does not include actual testing and publishing, but I think it's important to review that particular aspect first, and see if it does not create some arguments agains it.

ArthurKamalov:
PPS: I've also made [copy](https://github.com/scalyr/scalyr-agent-2/tree/add-ci/cd/.github/actions/execute-runner) of existing GitHub Action which handles caching to not interfere with existing code and to also apply naming changes to it, but it basically remains the same.

agent_build_refactored/docker_image_builders.py
```diff
@@ -755,7 +745,6 @@ class K8sRestartAgentOnMonitorsDeathAlpine(ContainerImageBuilderAlpine):
 ALPINE_K8S_IMAGE_BUILDERS = [
     K8sAlpine,
     K8sWithOpenMetricsAlpine,
```
Kami, line 747: Since we are cleaning it up I guess we may just go ahead and delete K8sWithOpenMetricsAlpine + related code as well.

agent_build_refactored/scripts/get_github_actions_job_matrices.py
```diff
@@ -0,0 +1,140 @@
...
+# push to a "production" tag.
+elif GITHUB_EVENT_NAME == "push" and GITHUB_REF_TYPE == "tag":
+    version_file_path = SOURCE_ROOT / "VERSION"
+    current_version = version_file_path.read_text().strip()
+    if GITHUB_REF_NAME == f"v{current_version}":
+        limited_run = False
+
+# Set collection of builders that will be used during the CI/Cd workflow run.
+if limited_run:
+    used_image_builders = DEBIAN_IMAGE_BUILDERS[:]
```
Kami, line 69: So this means we will only run Alpine builders on master run? Does this means that we won't notice if PR breaks Alpine image until it's merged into master?

Kami:
I do agree with Dominic that this "custom step generation" code is a bit magical / harder to grasp (I always prefer simpler solutions) and straight up hard coding build matrix may be easier, but I know this may not provide enough flexibility.

.github/actions/execute-runner/index.js
```diff
@@ -0,0 +1,173 @@
...
+    //
+    // Save the cache directory for a particular step if it hasn't been saved yet.
+    //
+
+    const fullPath = path.join(cacheDir, name)
+    const cacheKey = `${name}-${finalCacheSuffix}`
+
+    // Skip files. Cache can be only the directory.
+    if (!fs.lstatSync(fullPath).isDirectory()) {
+        return
```
Kami, line 63: Do we want to end some logging in case we end up in this if branch?

.github/actions/execute-runner/index.js
```diff
@@ -0,0 +1,173 @@
...
+    }
+    else {
+        console.log(`Save cache for the step with key ${cacheKey}.`)
+
+        try {
+            await cache.saveCache([fullPath], cacheKey)
+        } catch (error) {
+            console.warn(
+                `Can not save cache by key ${cacheKey}.
+                It seems that seems that it has been saved somewhere else.\nOriginal message: ${error}`
```
Kami, line 78: Would this indicate a race condition - e.g. another job saves cache before this step finished?

.github/actions/execute-runner/index.js
```diff
@@ -0,0 +1,173 @@
...
+    const step_cache_names = JSON.parse(json_encoded_step_names)
+
+    const cacheHits = {}
+
+    const finalCacheSuffix = `${cacheKeyRunnerPart}-${cacheVersionSuffix}`
+
+    // Run through step names and look if the is any existing cache for them.
+    for (let name of step_cache_names) {
+        console.log(name);
+        cacheHits[name] = await checkAndGetCache(
```
Kami, line 134: Probably want to improve logging a bit here to make it easier to see what is going on - e.g. ``console.log("Processing cache steps"); console.log(`Step name...`).`

Kami: And we probably also want to log the result of ``checkAndGetCache``. Similar story for line 150 below - should add more logging to make troubleshooting easier (if / when needed).

agent_build_refactored/scripts/get_github_actions_job_matrices.py
ArthurKamalov: We should notice that on PR creation, since the it also has to initiate the "full" run.

Kami:
Minor type I noticed in the existing code in the build output:

```python
agent_build_refactored/tools/runner.py
477:            log.info(f"Result if the step '{self.id}' is found in cache, skip.")
```

if -> of

.github/workflows/agent-build-refactored.yml
```diff
@@ -0,0 +1,163 @@
...
+          driver-opts: network=host
+
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@8b122486cedac8393e77aa9734c3528886e4a1a8 # v2
+        with:
+          image: tonistiigi/binfmt:qemu-v6.2.0
+          platforms: all
+
+      - name: Run cached step.
```
Kami, line 105: Let's remove "." from the end of the step names since most steps don't use it :)

.github/workflows/agent-build-refactored.yml
```diff
@@ -0,0 +1,163 @@
...
+        uses: ./.github/actions/execute-runner
+        with:
+          runner-fqdn: ${{ matrix.step-runner-fqdn }}
+
+
+  build-images:
+    name: Build image '${{ matrix.builder-name }}'.
```
Kami, line 112: Same here, minor step naming consistency thing

.github/actions/execute-runner/index.js
```diff
@@ -0,0 +1,173 @@
...
+    const cacheKey = `${name}-${finalCacheSuffix}`
+
+    // Skip files. Cache can be only the directory.
+    if (!fs.lstatSync(fullPath).isDirectory()) {
+        return
+    }
+
+    // If there's no cache hit, then save directory to the cache now.
+    if (isHit) {
+        console.log(`Cache for the step with key ${cacheKey} has been hit. Skip saving.`)
```
Kami, line 68: Perhaps may make it a bit easier to understand if it said something along the lines of "Cache for key already exist, skip saving".

Since for a second I was bit confused what hit means in this context (I assumed it means cache for that key has been found / exists, but I was still confused for a second).

ArthurKamalov:
Created a new action which accepts JSON matrix that we can define right in the workflow file.
https://github.com/scalyr/scalyr-agent-2/blob/fdbb9184c68e8c47b584d2995ba9f812cb8c91e2/.github/workflows/agent-build-refactored.yml#L48-L67

Still not a conventional way of defining a job matrix, but now it at least a hard-coded with workflow itself. [The script](https://github.com/scalyr/scalyr-agent-2/blob/fdbb9184c68e8c47b584d2995ba9f812cb8c91e2/agent_build_refactored/scripts/get_github_actions_job_matrices.py) that generated whole matrices before, is still there, but now it just works with jobs that are given to him from the new action.

.github/actions/execute-runner/index.js
ArthurKamalov: No, that just has to be already existing value for that key. I've changed that message so it is not seem like an error.

agent_build_refactored/scripts/get_github_actions_job_matrices.py
```diff
@@ -0,0 +1,189 @@
...
+    # List of all runners that are used by this workflow run.
+    used_runners = []
+
+    # Generate a final agent image build job matrix and filter out job for non-limited run, if needed.
+    result_images_build_matrix = {"include": []}
+    for job in images_build_matrix:
+        is_basic = job.get("basic", False)
+
+        # Skip job which id not basic if run is not limited.
+        if not is_basic and limited_run:
```
Kami, line 148: I think the comment is a bit confusing and should probably say "Skip job with basic attribute for limited run".

.github/workflows/agent-build-refactored.yml
```diff
@@ -0,0 +1,166 @@
...
+      # Jobs that has to be in a non-master workflow run has to have a '"basic": true' field.
+      - name: Init job matrices
+        uses: ./.github/actions/init-job-matrices
+        id: init-matrices
+        with:
+
+          # In the 'non-master' runs we only build debian-based images.
+          images-to-build: |
+            [
+              { "name": "docker-json-debian", "basic": true },
```
Kami, line 59: I know there is explanation in the comment above which explains what "basic" means, but I was still confused a bit before checking the comment.

I'm just thinking out loud if we could come up with some more explanatory attribute name - but maybe that's not possible since I can't think of anything better either :)

I guess another way would perhaps be to change this behavior and add ``master_run_only`` (or similar) attribute and set it to true for alpine jobs and false for debian ones.

ArthurKamalov:
@Kami changed `basic` to `master_run_only `, thanks for suggesting a better naming.

Kami (approved):

PR #980, opened 2022-08-22
CT-573 remove deprecated jobs, workflows, and commands from circleci

dlobue:
Removes deprecated jobs, workflows and commands from circleci for benchmarks and smoke-tests.

CT-573


ArthurKamalov:
All looks good to me except the fact that I tried to explain in this slack message https://sentinelone.slack.com/archives/C01S3UJTBLG/p1661195575221639?thread_ts=1661195222.549139&cid=C01S3UJTBLG

Maybe we should port at least that missing rate limiting tests first?

dlobue:
> All looks good to me except the fact that I tried to explain in this slack message https://sentinelone.slack.com/archives/C01S3UJTBLG/p1661195575221639?thread_ts=1661195222.549139&cid=C01S3UJTBLG
> 
> Maybe we should port at least that missing rate limiting tests first?

They have been ported over: https://github.com/scalyr/scalyr-agent-2/blob/master/.github/workflows/unittests.yml#L140-L210
https://github.com/scalyr/scalyr-agent-2/actions/runs/2906419700
Here's the rate-limit test for this PR specifically: https://github.com/scalyr/scalyr-agent-2/runs/7959264643?check_suite_focus=true

ArthurKamalov:
Ah, so they are in unittests. Sorry, my bad.

ArthurKamalov (approved):

.circleci/config.yml
```diff
@@ -91,352 +91,6 @@ parameters:
     default: ""
 
 commands:
-  unittest_tox:
```
Kami, line 94: Previously I only removed job names from the section at the bottom where workflows are defined (aka which jobs run for which workflow), but removing it all together from this file is fine as well (if we ever need to add it back for some reason, there is always git history :)).

PR #982, opened 2022-08-23
[DTIN-2150]Add images test and publishing

ArthurKamalov:
This PR finally introduces new testing and publishing for agent images.

The main changes here are tests for `k8s` images and `docker` images which are located in [tests/end_to_end_tests/container_image_tests](https://github.com/scalyr/scalyr-agent-2/tree/add-images-test-and-publishing/tests/end_to_end_tests/container_image_tests). They are also have the same "pytest-driven" structure as other tests that we already have in the [tests/end_to_end_tests](https://github.com/scalyr/scalyr-agent-2/tree/add-images-test-and-publishing/tests/end_to_end_tests) folder.

Kubernetes image tests can be configured to use different Kubernetes version, runtime and etc, as we already do in GHA-based [e2e tests](https://github.com/scalyr/scalyr-agent-2/blob/188d97c06bff2097fe5d80999124c51da309d628/.github/workflows/end_to_end_tests.yml#L92), so if this format of testing is approved, I guess we can start transferring all checks from those e2e tests here, so have only one place where we tests images. The same also may be applicable to the "legacy" smoketests which were transferred from the CircleCI.
But this, of course, is only possible if everybody is ok with that new way of testing.

Besides that, new publishing has finally been added to the workflow.


.github/workflows/agent-build-refactored.yml
```diff
@@ -163,4 +189,190 @@ jobs:
         with:
           name: ${{ matrix.name }}-registry
           path: /tmp/result
-          retention-days: 1
\ No newline at end of file
+          retention-days: 1
+
+  test-k8s-images:
```
Kami, line 194: I assume old / existing workflow will be removed once that PR is merged and verified?

tests/end_to_end_tests/container_image_tests/docker_test/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+                image_name,
+            ]
+        )
+
+        with timeout_tracker(20):
+            while True:
+                try:
+                    get_agent_log_content(container_name=agent_container_name)
+                    return agent_container_name
+                except subprocess.CalledProcessError:
```
Kami, line 110: Is there any way for us to provide additional context (e.g. in a somewhat generic manner) when timeout has been reached - things like logs accumulated so far, etc?

Kami:
> Kubernetes image tests can be configured to use different Kubernetes version, runtime and etc, as we already do in GHA-based [e2e tests](https://github.com/scalyr/scalyr-agent-2/blob/188d97c06bff2097fe5d80999124c51da309d628/.github/workflows/end_to_end_tests.yml#L92), so if this format of testing is approved, I guess we can start transferring all checks from those e2e tests here, so have only one place where we tests images. The same also may be applicable to the "legacy" smoketests which were transferred from the CircleCI.
But this, of course, is only possible if everybody is ok with that new way of testing.

I would almost prefer to reduce the scope of this PR and leave it out of this PR and tackle it in the future, but since it's already here... :)

ArthurKamalov:
> > Kubernetes image tests can be configured to use different Kubernetes version, runtime and etc, as we already do in GHA-based [e2e tests](https://github.com/scalyr/scalyr-agent-2/blob/188d97c06bff2097fe5d80999124c51da309d628/.github/workflows/end_to_end_tests.yml#L92), so if this format of testing is approved, I guess we can start transferring all checks from those e2e tests here, so have only one place where we tests images. The same also may be applicable to the "legacy" smoketests which were transferred from the CircleCI.
> > But this, of course, is only possible if everybody is ok with that new way of testing.
> 
> I would almost prefer to reduce the scope of this PR and leave it out of this PR and tackle it in the future, but since it's already here... :)

Yes, existing tests won't be touched in this PR. Need to give time to the team to make familiar with it.

.github/workflows/agent-build-refactored.yml
ArthurKamalov: No, it will be removed when it is sure that all existing tests are moved here and also  the new testing framework itself is tested and approved by the rest of the team.

tests/end_to_end_tests/container_image_tests/docker_test/conftest.py
ArthurKamalov: Slightly changed this code. Now instead of checking the existence of the agent log, we use the `is_running` method of the existing class which controls the agent.

We now also accumulate logs in a background thread by using a new special fixture, which dumps that logs at the end of the test case (on failure too)

agent_build_refactored/scripts/get_github_actions_job_matrices.py
```diff
@@ -149,22 +170,49 @@ def main():
...
+    # Generate a final agent k8s image test job matrix and filter out job for non-master run, if needed.
+    k8s_images_test_matrix = json.loads(
+        k8s_images_test_matrix_json_file_path.read_text()
+    )
+    result_k8s_images_test_matrix = {"include": []}
+    for job in k8s_images_test_matrix:
+        is_master_run_only = job.get("master_run_only", True)
+        # If this is non-master run, skip jobs which are not supposed to be in it.
+        if is_master_run_only and not master_run:
+            continue
```
Kami, line 194: Do we want to log a message here which says that job is X skipped because it's set to only run on master and this is not a master running or something along those lines?

I know this script also returns result so we probably need to log to stderr to not pollute stdout.

tests/end_to_end_tests/container_image_tests/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+@pytest.fixture(scope="session")
+def all_image_names_to_publish(image_builder_cls, source_registry_path):
+    """
+    Registry that contains target image. It may be an externally specified one, or if omitted,
+        new local registry with image is created.
+    """
+
+    with LocalRegistryContainer(
+        name="images_registry",
+        registry_port=0,
```
Kami, line 66: What does port 0 means? Use a default port? Use a random port? Or something else?

ArthurKamalov: Random/given by OS.

PR #983, opened 2022-08-23
Windows event monitor json api to use dottedEscapedJson parser

jmakar-scalyr:
Changed the Windows event monitor NewJsonApi parser to be dottedEscapedJson to accommodate embedded double quotes.

More notes/details in [DSET-355](https://jira-ng.sentinelone.com/browse/DSET-355)


Kami:
I think it would be nice to add integration / end to end tests for that (e.g. AMI based ones), but that may take a while.

Perhaps best course of action right now is to open JIRA ticket for that and then decide on the priority.

Kami (approved):

jmakar-scalyr:
Will merge this for now & revisit later to introduce an ami test

PR #984, opened 2022-08-29
Update syslog monitor to ignore invalid utf-8 byte sequences

Kami:
This pull request updates ``SyslogRequestParser`` class in the syslog monitor to ignore errors which may error when trying to decode byte sequence into utf-8 unicode string. This class is used when TCP transport is used and when ``tcp_request_parser`` parser is set to ``default`` (aka default behavior).

Previously an exception would be thrown in case we encountered invalid utf-8 sequence, but now we simply ignore non valid utf-8 data.

It's worth noting that exactly the same thing already happens inside ``SyslogBatchedRequestParser`` class which is used when ``tcp_request_parser`` is set to ``batch``.

### DIFF
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -490,7 +490,7 @@ class SyslogUDPHandler(six.moves.socketserver.BaseRequestHandler):
     """
 
     def handle(self):
-        data = six.ensure_text(self.request[0].strip(), errors="ignore")
+        data = six.ensure_text(self.request[0].strip(), "utf-8", errors="ignore")
         self.server.syslog_handler.handle(data)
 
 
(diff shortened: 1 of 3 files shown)


scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -613,7 +613,9 @@ def process(self, data, handle_frame):
                     )
 
                     # skip invalid bytes which can appear because of the buffer overflow.
-                    frame_data = six.ensure_text(self._remaining, errors="ignore")
+                    frame_data = six.ensure_text(
+                        self._remaining, "utf-8", errors="ignore"
```
Kami, line 617: ``utf-8`` is already the default value for the ``encoding`` argument, but I explicitly added it here to make it more obvious and consistent with the rest of the code.

ArthurKamalov (approved):
LGTM

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -628,7 +630,7 @@ def process(self, data, handle_frame):
             frame_length = frame_end - self._offset
 
             frame_data = six.ensure_text(
-                self._remaining[self._offset : frame_end].strip()
+                self._remaining[self._offset : frame_end].strip(), "utf-8", "ignore"
```
Kami, line 633: That was the primary place where exception could previously get thrown.

PR #987, opened 2022-08-30
Don't skip Github Action job runs on master branch merge / push

Kami:
I noticed some of the Github action jobs were being incorrectly skipped on master branch runs (e.g. https://github.com/scalyr/scalyr-agent-2/runs/8068790013).

We always want to run jobs on master branch merge / push and only skip duplicate / redundant jobs on pull request.

This pull request fixes all the jobs to make sure we don't skip run on master branch and also fixes the conditional which were likely unintentionally changed from "or" to "and" which won't produce the desired behavior.

```dif
 -    if: ${{ needs.pre_job.outputs.should_skip != 'true' || github.ref == 'refs/heads/master' }}
 +    if: ${{ needs.pre_job.outputs.should_skip != 'true' && (github.ref == 'refs/heads/master' || github.base_ref == 'master') }}
```

We always want to run checks on master branch to catch any issues which may only manifests itself on merging and since some Jenkins jobs depend on master runs.

Technically, if we strictly enforce "branch / pr must be up to date before merging" rule, we may also be able to safely skip redundant builds on master merge / run as well.

### DIFF
--- a/.github/actions/setup-minikube-cluster/action.yml
+++ b/.github/actions/setup-minikube-cluster/action.yml
@@ -52,6 +52,11 @@ runs:
       if: ${{ inputs.k8s_version != '' }}
       shell: bash
       run: |
+        # Workaround to ensure we use kubectl version which comes with specific k8s release as part
+        # of minikube and not system level kubectl
+        which kubectl
+        rm /usr/local/bin/kubectl || true
+        ln -sf $(which minikube) /usr/local/bin/kubectl
         echo "kubectl version"
         kubectl version
         echo ""
--- a/.github/actions/setup-minikube-cluster/action.yml
+++ b/.github/actions/setup-minikube-cluster/action.yml
@@ -5,7 +5,7 @@ inputs:
   minikube_version:
     description: "Minikube version to use"
     required: false
-    default: "v1.26.0"
+    default: "v1.26.1"
   k8s_version:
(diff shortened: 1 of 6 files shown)


Kami:
>  Technically, if we strictly enforce "branch / pr must be up to date before merging" rule / status check, we may also be able to safely skip redundant builds on master merge / run as well.

@ArthurKamalov @dlobue WDYT about that?

dlobue (approved):

dlobue:
> > Technically, if we strictly enforce "branch / pr must be up to date before merging" rule / status check, we may also be able to safely skip redundant builds on master merge / run as well.
> 
> @ArthurKamalov @dlobue WDYT about that?
> 
> In the past, we didn't enforce that status check so making sure all the jobs still run on master was essential to catch any issues which would have only been caught after merge. If we do want to continue to enforce that status check, we could update skip rules to also skip redundant builds on master branch runs.
> 
> If we decide to remove that github status check, then we do need to keep this skip if change to ensure we always run jobs on master branch.

+1 to removal of that check, and to running checks post-merge on the master branch.

ArthurKamalov:
Looks good to me

Kami:
@dlobue Thanks, will remove that check.

@ArthurKamalov to confirm, you are also on board with that change?

PR #988, opened 2022-08-30
Update slack channel for GHA notifications to renamed channel

dlobue:
The cloud-tech channel was renamed a while ago. Update the GHA workflows
with the current channel name to fix notifications.


jmakar-scalyr (approved):
Looks good.

I noticed that there is also a reference to #cloud-tech in scripts/send-benchmark-completion-notification-to-slack.py  May as well update that script as well

PR #989, opened 2022-08-31
Added quoted text to windows event log monitor ami test

jmakar-scalyr:
The intention here was to add an additional test to help cover the change introduced by DSET-355 / PR #983 

However the DSET-355 change doesn't manifest until after the event has been received by the Scalyr servers (because it relies on a specific parser), so it's debatable whether this added check is worth including as it is.

Alternatively a check could be added to the ami test script that queries for the event from Scalyr and verifies it's what is expected.  There are some challenges associated with this:
- Will need to make the server hostname & api key available to the powershell script
- Will need to take care that simultaneous executions don't interfere with each other (by introducing a random value in the checked message?)

Let me know what you think

### DIFF
--- a/tests/ami/scripts/test_windows.ps1.j2
+++ b/tests/ami/scripts/test_windows.ps1.j2
@@ -221,7 +221,7 @@ if ("$event_log_output" -notmatch "Level")   { Throw "Event log monitor didnt em
 # Verify that the event we inserted exists
 if ("$event_log_output" -notmatch "ScalyrAgentTest")   { Throw "Matching event not found in event log" };
 # TODO: Find out how to make this powershell script to write event logs with non-ASCII characters.
-if ("$event_log_output" -notmatch 'Scalyr automated tests event \\\\"with quoted text\\\\" .+')   { Throw "Matching event not found in event log" };
+if ("$event_log_output" -notmatch 'Scalyr automated tests event \\"with quoted text\\" .+')   { Throw "Matching event not found in event log" };
 
 #
 # Verify agent start up line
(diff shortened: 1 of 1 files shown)


Kami:
> Alternatively a check could be added to the ami test script that queries for the event from Scalyr and verifies it's what is expected. 

Sadly right now we don't have a framework for that setup in the AMI tests, but we do have similar patterns in place for smoke and end to end tests.

We have Scalyr Query Python script (https://github.com/scalyr/scalyr-agent-2/blob/master/.circleci/docker_unified_smoke_unit/smoketest/smoketest.py) used by many smoke tests and there is also simple bash script which is used by the Kubernetes end to end tests (https://github.com/scalyr/scalyr-agent-2/blob/master/scripts/cicd/scalyr-query.sh).

We should likely be able to re-use either of those inside the AMI tests script.

Perhaps bash version may be a bit simpler (Circle CI windows runner should also support bash shell), but long term we may want to invest some time into refactoring scalyr query Python script to make it a bit more re-usable / generic so it can be re-used across different repos and integrations for end to end tests which rely on querying DataSet (to avoid copy and pasting this script across different repos and making small changes to the script itself - sadly that's the status quo today).

> Will need to take care that simultaneous executions don't interfere with each other (by introducing a random value in the checked message?)

For other tests which query DataSet API we usually use serverHost which is unique to a specific test run (it includes random string + timestamp + build run id) and test invocation. This way tests are "isolated" by server host and multiple tests can run concurrently.

Kami (approved):
We can get this merged right now and figure out how to approach end to end tests which query DataSet separately.

PR #992, opened 2022-09-06
CT-596

jmakar-scalyr:
NOV requested more sophisticated handling of EventData array fields, discussion in [CT-596](https://scalyr.myjetbrains.com/youtrack/issue/CT-596)

The changes here take care to avoid clobbering existing fields (even though such pathological cases may never occur), also added tests for the explicit scenarios I was concerned about.

### DIFF
--- a/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
+++ b/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
@@ -703,13 +703,13 @@ def _convert_json_array_to_object(x):
         # Otherwise the (stringified) index number becomes the field name.
         rv = {}
         for i in range(len(x)):
-            if isinstance(x[i], dict) and '@Name' in x[i]:
-                orig_name = name = x[i].pop('@Name')
+            if isinstance(x[i], dict) and "@Name" in x[i]:
+                orig_name = name = x[i].pop("@Name")
                 if name in rv:
                     name += str(i)
                     # Fallback to the original behavior if this pathological case occurs
                     if name in rv:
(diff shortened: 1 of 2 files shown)


scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -697,7 +697,25 @@ def log_event(self, event):
 
 def _convert_json_array_to_object(x):
     if isinstance(x, list):
-        return {str(i): _convert_json_array_to_object(x[i]) for i in range(len(x))}
+        # Special handling of @Name fields per customer request:
+        # Ie [{"@Name:":"n1", "a1":"v1", ...}, ...] becomes {"n1":{"a1":"v1", ...}, ...}
+        # Duplicate @Names are handled by appending the index number if possible.
+        # Otherwise the (stringified) index number becomes the field name.
+        rv = {}
+        for i in range(len(x)):
```
ArthurKamalov, line 705: It seems that the `x[i]` is not changing during the whole iteration, so I think we can use something like
`for i, x_e in enumerate(x):` to avoid extra dict lookups, but I guess, that's more a matter of taste rather than real issue :)

ArthurKamalov (approved):
LGTM

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Ah good point, will start using enumerate.  (I would change this code but trying to get this release out quickly)

PR #997, opened 2022-09-20
windows_event_log_monitor: NewJsonApi: Ensure formatting when RateLim…

jmakar-scalyr:
…itLogFilter is not applied

From investigation into DSET-536 I came across an issue if the scalyr_logging.RateLimiterLogFilter is not applied (eg when monitor_log_write_rate & monitor_log_max_write_burst are both -1: ref scalyr_logging.MetricLogHandler.__init__) then the LogRecord generated by the NewJsonApi will not have a message attribute (note that RateLimiterLogFilter has a formater as its first argument).  This change resolves that issue

### DIFF
--- a/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
+++ b/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
@@ -928,8 +928,13 @@ and System sources:
         return result
 
     def open_metric_log(self):
-        class DummyFormatter:
+        class DummyFormatter(scalyr_logging.MetricLogFormatter):
             def format(self, record):
+                # Ensure the LogRecord has a message attribute.
+                # If the rate limit is disabled, formatting is not applied by RateLimitLogFilter.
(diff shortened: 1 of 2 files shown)


scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -928,8 +928,13 @@ def __get_api(self, sources, events, channels):
         return result
 
     def open_metric_log(self):
-        class DummyFormatter:
+        class DummyFormatter(scalyr_logging.MetricLogFormatter):
             def format(self, record):
+                # Ensure the LogRecord has a message attribute.
+                # If the rate limit is disabled, formatting is not applied by RateLimitLogFilter.
+                # Ref: scalyr_logging.MetricLoggingHandler.__init__
+                if not hasattr(record, "message"):
```
Kami, line 936: Can we add some tests for this scenario?

Kami (approved):

PR #1005, opened 2022-10-10
Agent connection logging improvements and misc changes

Kami:
This pull request updates scalyr client code to log how long it takes to read / process the response from the socket.

This will hopefully make it slightly easier to see what is going on when we see connection reset by peer / connection timeout and similar errors.

It also includes some other changes:

* Log which version of docker-py and requests library is used so we can see if it's using system version or bundled / vendored one
* Allows user to configure Docker client library connection timeout and maximum connection pool size when using Docker container enumerator via new `k8s_docker_client_timeout` and `k8s_docker_client_max_pool_size` config option.

### DIFF
--- a/scalyr_agent/scalyr_client.py
+++ b/scalyr_agent/scalyr_client.py
@@ -633,35 +633,47 @@ class ScalyrClientSession(object):
                     response = '{ "status":"success" }'
                     status_code = 200
                 else:
+                    start_ts = time.time()
                     status_code = self.__connection.status_code()
                     response = self.__connection.response()
+                    end_ts = time.time()
 
                 bytes_received = len(response)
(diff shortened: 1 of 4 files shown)


scalyr_agent/scalyr_client.py
```diff
@@ -633,35 +633,47 @@ def __receive_response(self, body_str, send_time):
                     response = '{ "status":"success" }'
                     status_code = 200
                 else:
+                    start_ts = time.time()
                     status_code = self.__connection.status_code()
                     response = self.__connection.response()
+                    end_ts = time.time()
```
ArthurKamalov, line 639: It seems that this end timestamp ends up being unused if no exception is occurred.

ArthurKamalov (approved):
LGTM :+1:

PR #1011, opened 2022-10-17
Add hard guard rail for leader election to prevent OOM conditions

dlobue:
Add a hard guard rail that stops the k8s events monitor if leader election detects more than 1000 nodes that are eligible for leader election.

This is to prevent the agent from consumer gobs of memory and getting OOM killed. The k8s events monitor should stop and produce a very nervous-making error message, but the rest of the agent should continue to work as though nothing was wrong.

The k8s events monitor can still be used in clusters containing 1000+ nodes. The user just needs to configure the use of labels to restrict the number of nodes that are considered for leader election.

DSET-424

### DIFF
--- a/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
@@ -512,9 +512,9 @@ This monitor was released and enabled by default in Scalyr Agent version `2.0.43
         """
         # Hard guard-rails to prevent the agent from consuming large amounts of memory.
         if query_fields:
-            query_fields += '&limit=1000'
+            query_fields += "&limit=1000"
         else:
-            query_fields = '?limit=1000'
+            query_fields = "?limit=1000"
 
         response = k8s.query_api_with_retries(
             "/api/v1/nodes%s" % query_fields,
--- a/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
+++ b/scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
@@ -510,13 +510,31 @@ This monitor was released and enabled by default in Scalyr Agent version `2.0.43
         @param k8s: a KubernetesApi object for querying the k8s api
         @query_fields - optional query string appended to the node endpoint to allow for filtering
         """
+        # Hard guard-rails to prevent the agent from consuming large amounts of memory.
+        if query_fields:
+            query_fields += '&limit=1000'
+        else:
+            query_fields = '?limit=1000'
+
         response = k8s.query_api_with_retries(
             "/api/v1/nodes%s" % query_fields,
             retry_error_context="nodes%s" % query_fields,
             retry_error_limit_key="k8se_check_nodes_for_leader",
         )
-        nodes = response.get("items", [])
+        if "continue" in response.get("metadata", {}):
+            # This is a hard guard-rail that stops only the
+            # KubernetesEventsMonitor should we run into a situation where
+            # 1000+ nodes are considered for leader election.
+            # TODO: Add in URL for how to use labels to rectify this once we have the documentation written up.
+            global_log.error(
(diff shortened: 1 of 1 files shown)


scalyr_agent/builtin_monitors/kubernetes_events_monitor.py
```diff
@@ -510,13 +510,31 @@ def _check_nodes_for_leader(self, k8s, query_fields=""):
         @param k8s: a KubernetesApi object for querying the k8s api
         @query_fields - optional query string appended to the node endpoint to allow for filtering
         """
+        # Hard guard-rails to prevent the agent from consuming large amounts of memory.
+        if query_fields:
```
Kami, line 514: This sounds good as the first step.

Next step would be (as discussed) implementing pagination where we only retrieve up to X nodes at once (and only limit sets of of fields / metadata field we need for leader election) and process this data incrementally to avoid memory overhead.

Ideally we would also use streaming JSON parser (so we parse result incrementally), but this may be hard to do with the requests abstraction which is used right now.

On a related note, we should make this limit a configuration option in case it ever needs to be changed.

And ideally we would also do some analysis - if we retrieve metadata + id + other mandatory fields for X nodes, what kind of temporary memory usage increase this represents so we can come up with the approximate limit based on this data.

Kami: I guess technically a simpler solution would be to only include those X retrieved nodes in leader election process instead of all nodes.

That's probably sufficient in case this X (limit) is large enough.

Only thing I would prefer to do in case we go with this approach is to try to retrieve those X nodes at random, but that may not be possible with available pagination query parameters (so the election is not biased against first X nodes or similar).

jmakar-scalyr:
Marking this PR as draft until final decisions made.

Looking at the api reference, pagination is easy & guaranteed to be a consistent snapshot, refer to https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/, search for "GET /api/v1/nodes" and see the continue & limit parameters.

A proposal:
- Use the limit=1000 & continue parameters to paginate
- Also use fieldSelector parameter to limit returned metadata
- For every page processed, only store the oldest node & update each page
  - Since we only care about the oldest there's no need to store more than that, which caps memory usage to page size

WDYT @Kami ?

Kami:
@jmakar-scalyr What kind of ordering / sorting is used by default when using pagination? Can we ensure that order is consistent and stable (e.g. ids or creationTimestamp is used for sort order)?

In case it is, then it should probably be fine.

Also, since afaik, Kubernetes API doesn't expose actual atomic CAS (compare and swap) operation for updating node metadata, there is still possible race condition situation there (and how significant it is kinda depends on the time windows - which will be longer with more nodes and more paginated requests since we need to do more processing).

jmakar-scalyr:
Afaict it's unsorted & I didn't find an approach to sort before the api returns the results.

I guess you're leaning towards designing a new leader election algorithm?  Anyways let me know what you'd like me to look into.

Here's a quick pagination PoC if you want to play with it:

```
#!/usr/bin/env python3
# Demonstrate pagination of Kubernetes nodes using its api

import json
import requests

# Retrieved from something like `kubectl --v=10 get svc`
host = '<elided>'

# TODO These are short-lived, run the command and parse its output here
# aws eks get-token --cluster-name $cluster | jq -r .status.token
token = '<elided>'

nodes = []
cont = None
while True:
    url = f'https://{host}/api/v1/nodes?limit=1000' + (f'&continue={cont}' if cont else '')
    # TODO Use the certificate from ~/.kube/config instead of verify=False
    resp = requests.get(url, headers={'Authorization': f'Bearer {token}'}, timeout=15, verify=False)
    resp.raise_for_status()
    resp_json = resp.json()

    nodes += resp_json['items']

    metadata = resp_json['metadata']
    cont = metadata.get('continue')
    if not cont:
        break

print(json.dumps(nodes, indent=2))
```

PR #1020, opened 2022-10-25
Try adding "securityContext.allowPrivelegeEscalation: false" to the agent container

Kami:
This pull request adds ``securityContext.allowPrivelegeEscalation: false`` annotation to the agent container in the Kubernetes DaemonSet definition file to see if it works (best practice, least privilege mantra - albeit this is a bit moot here since agent is considered as a privileged process).

I don't think it would work with regular installation (iirc, tcollectors rely on setuid, but those don't run in k8s images), but we will see if tests pass or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 Scalyr Agent 2 Changes By Release
 =================================
 
+## 2.1.38 "TBD" - November 17, 2022
+<!---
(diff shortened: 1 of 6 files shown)


Kami:
End to end tests seemed to have confirmed (everything passed) that this annotation works fine, so I applied it to all the container definitions.

ArthurKamalov (approved):
LGTM

PR #1021, opened 2022-10-26
Also run some micro + process level benchmarks under Python 3.11

Kami:
This pull request updates micro + process level benchmarks to also run under new Python 3.11 to see how it impacts those benchmarks.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -366,8 +366,8 @@ deps =
     -r benchmarks/micro/requirements-compression-algorithms.txt
     # need < 9.0.0 for python 2.7
     py-cpuinfo==8.0.0
-    pytest-benchmark[histogram]==3.2.3; python_version < '3.1'
-    pytest-benchmark[histogram]==4.0.0; python_version == '3.1'
+    pytest-benchmark[histogram]==3.2.3; python_version < '3.11'
+    pytest-benchmark[histogram]==4.0.0; python_version == '3.11'
     # Depends on libsnappy-dev / snappy-devel
     # NOTE: We don't include this dependency in requirements.txt to avoid having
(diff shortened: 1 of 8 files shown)


Kami:
Removed testing changes (5352d49) after confirming everything is working as it should (https://github.com/scalyr/scalyr-agent-2/actions/runs/3454844153).

tox.ini
```diff
@@ -364,11 +364,14 @@ deps =
     -r dev-requirements.txt
     -r benchmarks/scripts/requirements.txt
     -r benchmarks/micro/requirements-compression-algorithms.txt
-    pytest-benchmark[histogram]
+    # need < 9.0.0 for python 2.7
+    py-cpuinfo==8.0.0
```
ArthurKamalov, line 368: hmm, I've tried to install various earlier versions of `py-cpuinfo` but got same error.

tox.ini
```diff
@@ -364,11 +364,14 @@ deps =
...
+    pytest-benchmark[histogram]==3.2.3; python_version < '3.1'
```
ArthurKamalov, line 369: I guess it was meant 3.11

ArthurKamalov (approved):
LGTM

tox.ini
Kami: Good catch - I believe the reason it still works is because this requirement is duplicated in two places (requirements file, tox file) so the correct version still gets installed under Python 3.11.

tox.ini
Kami: Yeah, multiple dependencies were causing problems, not just py-cpuinfo (also pygal, etc.).

PR #1022, opened 2022-10-28
DSET-424: Alternate algorithm for Kuberenetes event monitor leader election

jmakar-scalyr:
This relates to DSET-424: High memory usage in Agent leader election within k8s event monitor

Rather than paginating nodes or using a subset of them take the simpler option of expecting a deployment or daemonset to only run a single (or few) pods with the Kubernetes event monitor running.  Then the leader election is restricted to only those pods owned by the deployment/replicaset or daemonset thereby ensuring a much smaller pool.

This approach came from discussions with Mugzy, Dario & Tomaz earlier this week & we concluded this was the best solution.

In addition to the change above (mostly localized to _check_pods_for_leader) the leader_node and ignore_master options were removed.


docs/monitors/kubernetes_events_monitor.md
```diff
@@ -109,7 +89,5 @@ This monitor was released and enabled by default in Scalyr Agent version `2.0.43
...
 | `event_object_filter`    | Optional (defaults to ['CronJob', 'DaemonSet', 'Deployment', 'Job', 'Node', 'Pod', 'ReplicaSet', 'ReplicationController', 'StatefulSet', 'Endpoint']). A list of event object types to filter on. Only events whose ``involvedObject`` ``kind`` is on this list will be included.  To not perform filtering and to send all event kinds, set the environment variable ``SCALYR_K8S_EVENT_OBJECT_FILTER=null``. | 
 | `leader_check_interval`  | Optional (defaults to 60). The number of seconds to wait between checks to see if we are still the leader. | 
-| `leader_node`            | Optional (defaults to None). Force the `leader` to be the scalyr-agent that runs on this node. | 
-| `check_labels`           | Optional (defaults to False). If true, then the monitor will check for any nodes with the label `agent.config.scalyr.com/events_leader_candidate=true` and the node with this label set and that has the oldestcreation time will be the event monitor leader. |
```
Kami, line 113: Since this is technically a breaking change, let's please also add some info on it to the release notes file.

Kami:
Per the discussion the other day, I'm still missing one piece - for this change to work, we also need to update actual kubernetes / kustomize resource files (aka define Deployment for agent which only runs Kubernetes Events monitor + also update helm chart), right?

docs/monitors/kubernetes_events_monitor.md
jmakar-scalyr: Will do.  Will make a separate PR for this

PR #1024, opened 2022-11-07
DSET-840: Adjust memory_stats.usage for consistency with other tools

jmakar-scalyr:
From DSET-840:

The memory usage metric retrieved from the Docker api has the filesystem cache included, this change removes the cache value (if present) for consistency with the Kubelet api (kubectl top) / metrics-server, docker stats, etc

Ref: https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerStats

```
import docker
client = docker.APIClient(base_url='unix://var/run/docker.sock', version='auto')
client.stats(container='<container-id>', stream=False)['memory_stats']
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -9,6 +9,13 @@ incompatible changes have been made in that release.
 For a list of all the changes in a particular release, please refer to the changelog file -
 https://github.com/scalyr/scalyr-agent-2/blob/master/CHANGELOG.md.
 
+## 2.1.38 "TBD" - TBD
+
+* This release updates the Kubernetes monitor to more accurately calculate used memory.
+
+  Formerly `docker.mem.usage` was equal to `memory_stats.usage` now it it equal to `memory_stats.usage` minus
+  `memory_stats.stats.cache` which represents file system cache usage.
+
 ## 2.1.21 "Ultramarine" - Jun 1, 2021
 
 * This is the last release that still supports Python 2.6. This version of the agent will emit a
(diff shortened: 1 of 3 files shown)


Kami:
Since this changes the default behavior, we should also document this change in the release notes file + ideally also DataSet docs (iirc, this memory usage metric is shown on default built-in Kubernetes dashboards so we need to make sure this change is communicated well otherwise some people may get surprised / confused).

Kami:
> this change removes the cache value (if present) 

Do we know when this value is present and when it's not? (and what kind of factors affects the presence of that value)

jmakar-scalyr:
> Since this changes the default behavior, we should also document this change in the release notes file + ideally also DataSet docs (iirc, this memory usage metric is shown on default built-in Kubernetes dashboards so we need to make sure this change is communicated well otherwise some people may get surprised / confused).

Updated the release notes & it looks like another team is handling the DataSet docs / customer updates.

> Do we know when this value is present and when it's not? (and what kind of factors affects the presence of that value)

I was just being cautious, afaik it will always be present on Linux systems (not sure about other Unix-like systems but I don't think that's a concern)

RELEASE_NOTES.md
```diff
@@ -9,6 +9,13 @@ incompatible changes have been made in that release.
 For a list of all the changes in a particular release, please refer to the changelog file -
 https://github.com/scalyr/scalyr-agent-2/blob/master/CHANGELOG.md.
 
+## 2.1.38 "TBD" - TBD
+
+* This release updates the Kubernetes monitor to more accurately calculate used memory.
```
ArthurKamalov, line 14: I guess it would be also good to reflect this in the CHANGELOG file. It also seems that it is short enough to just have it in the CHANGELOG file, because, IIRC, we use RELEASE NOTES for more descriptive notes, but nothing wrong with keeping it here too, of course :)

ArthurKamalov (approved):
LGTM

RELEASE_NOTES.md
jmakar-scalyr: Good catch, fixed

PR #1025, opened 2022-11-09
[WIP] DSET-738 Add agents dependency packages

ArthurKamalov:


agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -0,0 +1,932 @@
...
+PYTHON_PACKAGE_NAME = "scalyr-agent-python3"
+
+# name of the dependency package with agent requirement libraries.
+AGENT_LIBS_PACKAGE_NAME = "scalyr-agent-libs"
+
+
+EMBEDDED_PYTHON_VERSION = "3.11.0"
+EMBEDDED_PYTHON_SHORT_VERSION = ".".join(EMBEDDED_PYTHON_VERSION.split(".")[:2])
+
+PYTHON_PACKAGE_SSL_VERSION = "1.1.1k"
```
Kami, line 81: Does this version here need to be the lowest common denominator for compatibility reasons? Or we can use new / latest openssl version? Also, I assume openssl is dynamically linked to OpenSSL?

agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -0,0 +1,932 @@
...
+        param_filters = {
+            "deb": "debs",
+            "rpm": "rpms"
+        }
+
+        params["filter"] = param_filters[self.PACKAGE_TYPE]
+
+        with requests.Session() as s:
+            resp = s.get(
+                url=f"https://packagecloud.io/api/v1/repos/{user_name}/{repo_name}/search.json",
```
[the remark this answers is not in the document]
Kami: EDIT: It looks like it does - https://packagecloud.io/docs/api#pagination

So we likely either need to handle this scenario (in case the filter results in more than one page of results being returned) or simply throw in case we encounter such condition (and we can tackle it later once more packages are in the repo or similar).

Initially we may also want to use larger value for ``?per_page`` query parameter.

agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -0,0 +1,932 @@
...
+            token=token,
+            user_name=user_name,
+            repo_name=repo_name
+        )
+
+        if len(packages) == 0:
+            logger.info(f"Could not find any package with name {package_name}")
+            return None
+
+        last_package = packages[-1]
```
Kami, line 557: So search.json API has guaranted and stable sort order? If so, what is based on? package id / upload timestamp or similar?

agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -0,0 +1,932 @@
...
+        else:
+            logging.error(f"Unknown command {args.command}.")
+            exit(1)
+
+
+def create_build_dependencies_step(
+        base_image: EnvironmentRunnerStep
+) -> EnvironmentRunnerStep:
+    """
+    This function creates step that installsP Python build requirements, to a given environment.
```
Kami, line 765: Small typo, extra "P"

Kami:
More general architecture / implementation question - how do we plan to handle publishing "development" packages - aka package which are built and published on each master run.

I think it's probably a good idea to handle them a bit differently from stable / production packages (e.g. use a different repo for development / unstable packages) so we can maintain that repo differently from stable / production packages (e.g. prune old dev packages more often, etc.).

IIRC, packagecloud doesn't support release "channels" (think stable, beta, alpha, dev, etc.) so a way to handle that is to simply have multiple repos (e.g. in StackStorm we had StackStorm/stable repo for stable / production packages and StackStorm/unstable for development and testing packages).

ArthurKamalov:
> More general architecture / implementation question - how do we plan to handle publishing "development" packages - aka package which are built and published on each master run.
> 
> I think it's probably a good idea to handle them a bit differently from stable / production packages (e.g. use a different repo for development / unstable packages) so we can maintain that repo differently from stable / production packages (e.g. prune old dev packages more often, etc.).
> 
> IIRC, packagecloud doesn't support release "channels" (think stable, beta, alpha, dev, etc.) so a way to handle that is to simply have multiple repos (e.g. in StackStorm we had StackStorm/stable repo for stable / production packages and StackStorm/unstable for development and testing packages).


AFAIK, yes, packagecloud does not support channels.


Yes, I think we are on the same page here. The plan is to have two repositories, internal and prod (just using current naming, but we can change it to stable/unstable or smth). 

On `master` branch pushes we publish each new build into the `internal` repo. 
On the push of the "production" tag (e.g. 2.1.50), we publish it into prod repo and internal.

agent_build_refactored/managed_packages/managed_packages_builders.py
ArthurKamalov: This has to be latest version. The plan is to add periodical job that checks crucial libraries such as OpenSSL for new versions/critical bug fixes.
 
Yes, It is dynamically linked and it is included to the Python package, so we can run the interpreter and linking to this version of OpenSSL

According to my tests, latest version of the OpenSSl does not have any compatibility issues with all distros that we plan to maintain. But it will be shown closer in one of the next PRs with AMI tests.

agent_build_refactored/managed_packages/managed_packages_builders.py
ArthurKamalov: It has to group by distro_version  (aka `Ubuntu/22.04`) and then sort by the package version inside each group. 
But as it was said in previous comment  https://github.com/scalyr/scalyr-agent-2/pull/1025#discussion_r1030641030, new do not rely on that now and do full run through all pages.

agent_build_refactored/managed_packages/steps/install_build_dependencies.sh
```diff
@@ -0,0 +1,265 @@
...
+./configure --enable-libgdbm-compat
+make -j "$(nproc)"
+make install
+popd
+popd
+
+
+mkdir /tmp/build-libffi
+pushd /tmp/build-libffi
+curl -L "https://codeload.github.com/libffi/libffi/tar.gz/v${LIBFFI_VERSION}" > libffi.tar.gz
```
Kami, line 239: I know we use TLS transport, but long term we should also figure out how to make those downloads more secure - aka verifying checksums / PGP signatures (if available). That's basically what happens when using package manager (at least metadata signature is verified).

ArthurKamalov: Yes, that would be very nice to have. Will add it.

agent_build_refactored/managed_packages/steps/download_build_dependencies/download_build_dependencies.sh
```diff
@@ -114,72 +161,63 @@ Huqba4uzzTOVO3KXwQEwEJhQVY1MssUXyjnvxRnfvYt22Y1+KOuv2Ps+DCw8J0Gs
...
-#curl -L "https://ftp.gnu.org/gnu/texinfo/texinfo-${TEXTINFO_VERSION}.tar.gz" > texinfo.tar.gz
-#curl -L "https://ftp.gnu.org/gnu/texinfo/texinfo-${TEXTINFO_VERSION}.tar.gz.sig" > texinfo.tar.gz.sig
-#
-#gpg --verify --keyring "${SOURCE_ROOT}/agent_build_refactored/managed_packages/steps/download_build_dependencies/gnu-keyring.gpg" \
-#  texinfo.tar.gz.sig \
-#  texinfo.tar.gz
+mkdir "${STEP_OUTPUT_PATH}/texinfo"
+pushd "${STEP_OUTPUT_PATH}/texinfo"
+curl -L "https://ftp.gnu.org/gnu/texinfo/texinfo-${TEXINFO_VERSION}.tar.gz" > texinfo.tar.gz
+curl -L "https://ftp.gnu.org/gnu/texinfo/texinfo-${TEXINFO_VERSION}.tar.gz.sig" > texinfo.tar.gz.sig
```
Kami, line 173: To be on the safer side, we should likely bundle signatures either directly in the script or in files somewhere in the repo.

Since if we download signature file from the same location from where we download the software artifact, it doesn't really add any additional security since that file could have been compromises / similar as well :)

agent_build_refactored/managed_packages/steps/download_build_dependencies/download_build_dependencies.sh
```diff
@@ -1,9 +1,55 @@
...
+
 set -e
 
+apt update
 
+DEBIAN_FRONTEND=noninteractive apt install -y git curl gnupg gnupg2 xz-utils
 
 
-XZ_PUBLIC_KEY="-----BEGIN PGP PUBLIC KEY BLOCK-----
+# Import XZ public key
```
Kami, line 51: Thanks for adding key verification. And yeah, as you've implemented it - if we want additional security benefits, we indeed to to bundle the actual keys which are used for signing inside the script / repo.

agent_build_refactored/managed_packages/steps/download_build_dependencies/download_build_dependencies.sh
```diff
@@ -77,27 +123,28 @@ eUy2pBqS1302sBB+nfrGFTFmJC4ypzY8uQFG98dxBozqME836KrMhiBbeLIghXgs
 OQHHt+gXyQVzvzskedUxZ4mxzZ1D/+JfJhwxGvxZFn5sAGLbAmRbXoIv0duHk/JW
 2F1vt/S03rzec8y37B3LLWnIO9LK
 =V4Re
------END PGP PUBLIC KEY BLOCK-----"
+-----END PGP PUBLIC KEY BLOCK-----" | gpg --import
 
-gpg --import <<< ${XZ_PUBLIC_KEY}
```
Kami: Just thinking out loud - do we need / want to clear gpg data directory before validating each artifact? So we ensure that only key for a specific package is in the gpg key database once verifying that artifact.

Kami (approved):
LGTM, thanks for adding artifact signature validation.

PR #1026, opened 2022-11-11
Disable Python 2 micro benchmark.

ArthurKamalov:
Disable python 2 benchmark since it has dependency issues.


Kami:
I've fixed it in #1021 as part of Python 3.11 support.

So for now, we can keep it in place and remove it as soon as frozen binaries are ready.

Kami:
This has been done / superseded by #1073.

PR #1028, opened 2022-11-16
DSET-848: windows_event_log_monitor: Support %%n placeholder replacement

jmakar-scalyr:
Adding support for this is a little complex so I took the liberty of adding a new feature flag: placeholder_render

If you would like to test can always run: `python scalyr_agent/run_monitor.py -c '{json:true,placeholder_render:true}' scalyr_agent.builtin_monitors.windows_event_log_monitor`

Some references:
- https://learn.microsoft.com/en-us/windows/win32/eventlog/event-identifiers#message-definitions
- https://learn.microsoft.com/en-us/windows/win32/eventlog/message-files
- FormatMessage of https://github.com/mhammond/pywin32/blob/main/win32/Lib/win32evtlogutil.py

### DIFF
--- a/docs/monitors/windows_event_log_monitor.md
+++ b/docs/monitors/windows_event_log_monitor.md
@@ -159,6 +159,7 @@ Log into DataSet and query [monitor = 'windows_process_metrics'](https://app.sca
 | `remote_password`            | Optional (defaults to `none`). Password to use for authentication on the remote server.  This option is only valid on Windows Vista and above. | 
 | `remote_domain`              | Optional (defaults to `none`). The domain for the remote user account. This option is only valid on Windows Vista and above. | 
 | `json`                       | Optional (defaults to `false`). Format events as json? Supports inclusion of all event fields. This option is only valid on Windows Vista and above. | 
+| `placeholder_render`         | Optional (defaults to `false`). Render %%n placeholders in event data?  This option is only valid on Windows Vista and above. | 
 
 <a name="events"></a>
 ## Event Reference
--- a/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
+++ b/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
@@ -736,11 +736,13 @@ class NewJsonApi(NewApi):
             win32evtlog.EvtRenderContextSystem
         )
 
-        self._placeholder_render = config.get('placeholder_render')
+        self._placeholder_render = config.get("placeholder_render")
         if self._placeholder_render:
             self._dll_cache = Cache(size=10)
             self._param_cache = Cache(size=1000)
-            self._langid = win32api.MAKELANGID(win32con.LANG_NEUTRAL, win32con.SUBLANG_NEUTRAL)
+            self._langid = win32api.MAKELANGID(
+                win32con.LANG_NEUTRAL, win32con.SUBLANG_NEUTRAL
+            )
 
     def log_event(self, event):
         values = win32evtlog.EvtRender(
@@ -800,17 +802,17 @@ class NewJsonApi(NewApi):
             self._bookmark_lock.release()
 
     def _param_placeholder_value(self, channel, provider, param):
-        if not re.match('^%%\d+$', param):
+        if not re.match("^%%\d+$", param):
             return param
 
-        if channel == 'Security':
-            provider = 'Security'
+        if channel == "Security":
+            provider = "Security"
 
-        param_key = '%s-%s-%s' % (channel, provider, param)
+        param_key = "%s-%s-%s" % (channel, provider, param)
         if param_key in self._param_cache:
             return self._param_cache[param_key]
 
-        dll_key = '%s-%s' % (channel, provider)
+        dll_key = "%s-%s" % (channel, provider)
         if dll_key in self._dll_cache:
             if not isinstance(self._dll_cache[dll_key], _DLL):
                 return param
@@ -844,33 +846,43 @@ class NewJsonApi(NewApi):
                 self._dll_cache[dll_key].handle,
                 int(param[2:]),
                 self._langid,
-                None
+                None,
             )
             self._param_cache[param_key] = value.strip()
         except:
             self._param_cache[param_key] = param
 
         return self._param_cache[param_key]
-    
+
     def _replace_param_placeholders(self, event):
         rv = copy.deepcopy(event)
 
         try:
-            event_data = event['Event']['EventData']['Data']
-            channel = event['Event']['System']['Channel']
-            provider = event['Event']['System']['Provider']['Name']
+            event_data = event["Event"]["EventData"]["Data"]
+            channel = event["Event"]["System"]["Channel"]
+            provider = event["Event"]["System"]["Provider"]["Name"]
         except:
             return rv
 
-        if isinstance(event_data, str) and re.match('^%%\d+$', event_data):
-            rv['Event']['EventData']['Data'] = self._param_placeholder_value(channel, provider, event_data)
-            
+        if isinstance(event_data, str) and re.match("^%%\d+$", event_data):
+            rv["Event"]["EventData"]["Data"] = self._param_placeholder_value(
+                channel, provider, event_data
+            )
+
         elif isinstance(event_data, dict):
             for key, val in event_data.items():
-                if isinstance(val, str) and re.match('^%%\d+$', val):
-                    rv['Event']['EventData']['Data'][key] = self._param_placeholder_value(channel, provider, val)
-                elif isinstance(val, dict) and 'Text' in val and re.match('^%%\d+$', val['Text']):
-                    rv['Event']['EventData']['Data'][key]['Text'] = self._param_placeholder_value(channel, provider, val['Text'])
+                if isinstance(val, str) and re.match("^%%\d+$", val):
+                    rv["Event"]["EventData"]["Data"][
+                        key
+                    ] = self._param_placeholder_value(channel, provider, val)
+                elif (
+                    isinstance(val, dict)
+                    and "Text" in val
+                    and re.match("^%%\d+$", val["Text"])
+                ):
+                    rv["Event"]["EventData"]["Data"][key][
+                        "Text"
+                    ] = self._param_placeholder_value(channel, provider, val["Text"])
 
         return rv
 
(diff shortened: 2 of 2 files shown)


docs/monitors/windows_event_log_monitor.md
```diff
@@ -159,6 +159,7 @@ Log into DataSet and query [monitor = 'windows_process_metrics'](https://app.sca
 | `remote_password`            | Optional (defaults to `none`). Password to use for authentication on the remote server.  This option is only valid on Windows Vista and above. | 
 | `remote_domain`              | Optional (defaults to `none`). The domain for the remote user account. This option is only valid on Windows Vista and above. | 
 | `json`                       | Optional (defaults to `false`). Format events as json? Supports inclusion of all event fields. This option is only valid on Windows Vista and above. | 
+| `placeholder_render`         | Optional (defaults to `false`). Render %%n placeholders in event data?  This option is only valid on Windows Vista and above. |
```
dkoshe, line 162: Do we support older than Vista ? If we do, what happens if this is set to `true` on older versions?  Do we silently ignore?

jmakar-scalyr: Yes, for context there are two Windows apis for retrieving event logs, one is pre-Vista (its use in the monitor code is encapsulated by the OldApi class) and one is Vista & onwards (NewApi & NewJsonApi classes).

Yes, if json or placeholder_render are set as true for pre-Vista it is silently ignored.

ArthurKamalov, line 162: Maybe we should also add the server equivalent, like Vista/WindowsServer2008?

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -754,6 +774,14 @@ def log_event(self, event):
         #        => '{"entry": {"text": "bar", "#text": "foo"}}'
         event_json = _strip_xmltodict_prefixes(event_json)
 
+        # FormatMessage will replace insertion and parameter strings in the event message,
+        # however it will not not replace parameter strings in the event data.
```
ArthurKamalov, line 778: duplicate "not", not sure if this can be considered as double negative :D

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -773,6 +801,91 @@ def log_event(self, event):
         finally:
             self._bookmark_lock.release()
 
+    def _param_placeholder_value(self, channel, provider, param):
+        if not re.match("^%%[0-9]+$", param):
```
ArthurKamalov, line 805: I guess I may miss something here, but it seems that every branch that starts from [there](https://github.com/scalyr/scalyr-agent-2/blob/d0d45d49327087a454d8be6900e4307dcd82a925/scalyr_agent/builtin_monitors/windows_event_log_monitor.py#L867)  already does the `^%%[0-9]+$` match and only then calls the `_param_placeholder_value ` function.  Does this mean that this check will always and up with false?

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -773,6 +801,91 @@ def log_event(self, event):
...
+        if channel == "Security":
+            provider = "Security"
+
+        param_key = "%s-%s-%s" % (channel, provider, param)
+        if param_key in self._param_cache:
+            return self._param_cache[param_key]
+
+        dll_key = "%s-%s" % (channel, provider)
+        if dll_key in self._dll_cache:
+            if not isinstance(self._dll_cache[dll_key], _DLL):
```
ArthurKamalov, line 817: Ah, if it's not a _DLL instance, then this has to be an exception. Got it. I just was wondering why we put exceptions to the cache :) 
And it seems that there's no way for a cached Exception to make it to the `FormatMessageW `

ArthurKamalov (approved):
LGTM, I wonder if we can trigger come relevant sample events in our [windows AMI tests](https://github.com/scalyr/scalyr-agent-2/blob/master/tests/ami/scripts/test_windows.ps1.j2), and see how it works, but I won't insist since it may be not so trivial to do :)

docs/monitors/windows_event_log_monitor.md
jmakar-scalyr: Actually, I would like to drop support for the older / pre-Vista api to simplify the implementation.   Will make a ticket for later discussion

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: You're right, I am just being cautious here because _param_placeholder_value expects that format for param (later it does `int(param[2:])`).  This extra check is to quickly identify a problem if something else gets passed into the function in a future change.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Yes, I should've either commented this or made a note in the PR description.  The idea is to store an exception to avoid attempting to (unsuccessfully) open a dll over and over again.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Oops, will fix, thank you!

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
ArthurKamalov: Yeah, got it.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -773,6 +801,91 @@ def log_event(self, event):
...
+            found = False
+            for dll in self._dll_cache.values():
+                if dll.path == dllpath:
+                    self._dll_cache[dll_key] = dll
+                    found = True
+                    break
+
+            if not found:
+                try:
+                    self._dll_cache[dll_key] = _DLL(channel, provider)
```
Kami, line 838: Does this mean we keep an open file handle to each loaded DLL which is referenced?

I guess that should not be a big issue since I see the cache size is limited to 10 entries.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -773,6 +801,91 @@ def log_event(self, event):
...
+        try:
+            value = win32api.FormatMessageW(
+                win32con.FORMAT_MESSAGE_FROM_HMODULE,
+                self._dll_cache[dll_key].handle,
+                int(param[2:]),
+                self._langid,
+                None,
+            )
+            self._param_cache[param_key] = value.strip()
+        except:
```
Kami, line 852: Probably a good idea to log a message here with the actual exception + context? At very least, under debug in case it doesn't warrant info / warn log level.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -823,6 +936,47 @@ def _strip_xmltodict_prefixes(x):
...
+            win32api.RegCloseKey(keyhandle)
+
+    def __init__(self, channel, provider):
+        try:
+            self.path = self.dllpath(channel, provider)
+            self.handle = win32api.LoadLibraryEx(
+                self.path, 0, win32con.LOAD_LIBRARY_AS_DATAFILE
+            )
+        except Exception as e:
+            self.handle = None
```
Kami, line 961: Same here, may also want to log the error with context.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -823,6 +936,47 @@ def _strip_xmltodict_prefixes(x):
...
+            raise e
+
+    def __del__(self):
```
Kami, line 964: IIRC, ``__del__`` is Python implementation and GC algorithm specific so it may be safer / make more sense to use a context manager or similar (although not sure, how hard would it be to re-organize the code to use context manager or similar instead).

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -724,6 +736,14 @@ def __init__(self, config, logger, channels):
             win32evtlog.EvtRenderContextSystem
         )
 
+        self._placeholder_render = config.get("placeholder_render")
+        if self._placeholder_render:
+            self._dll_cache = Cache(size=10)
```
Kami, line 741: Does it make sense to make those size parameters user configurable?

Kami:
Since it's a non-trivial functionality, it would also be good to add some unit tests for this new functionality (and as Arthur said, ideally also AMI tests albeit those may be harder to orchestrate).

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
Kami: Do we want to have some kind of TTL on the cache items here? In case some library which provides some DLL gets installed later one and future calls would not result in an exception anymore?

Or it's just really an overkill and we can just assume user will also restart the agent in such case or similar?

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
Kami: For pure observability purposes, it may make sense to periodically log under INFO (e.g. once a day or every 12 hours) size of those caches + perhaps also keys of the dll_cache.

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Yes

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Will do

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Will do

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Yeah I wasn't happy with this either, will revisit it

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: Will do

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -773,6 +801,91 @@ def log_event(self, event):
...
+
+        try:
+            value = win32api.FormatMessageW(
+                win32con.FORMAT_MESSAGE_FROM_HMODULE,
+                self._dll_cache[dll_key].handle,
+                int(param[2:]),
+                self._langid,
+                None,
+            )
+            self._param_cache[param_key] = value.strip()
```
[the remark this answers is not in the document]
jmakar-scalyr: It should, because the values being cached are from message tables but of course there are no guarantees.  Will change the param cache to have some ttl to guard against this

jmakar-scalyr:
DSET-1106 & DSET-1107 created from feedback

PR #1036, opened 2022-11-29
DSET-1018: Add file template support in syslog_monitor

jmakar-scalyr:
DSET-1018: Add file template support in syslog_monitor

There are two aspects to the changes here:
- Collection and passing to SyslogHandler.handle of additional info for each syslog message
  - Specifically proto, srcip, destport, hostname, appname
- Use of this additional info in the creation of a dynamic set of loggers using a template
  - Implemented in SyslogHandler.__handle_syslog_logs

The message_log_template (which overrides the use of message_log if not-none) defines the template for which the additional info (eg proto, srcip, etc) are used as params.  The other added options (eg check_for_unused_logs_mins, etc) support their cleanup.


jmakar-scalyr:
Temporarily converting to draft while PR checks fixed

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -42,6 +43,12 @@
 from six.moves import range
 import six.moves.socketserver
 
+try:
+    # Only available for python >= 3.6
+    import syslogmp
```
ArthurKamalov, line 48: If this lib may be used by agent which is not from our images, then it seems that it has to be bundled here as a third party library. 

luckily we are close to new linux binaries so soon we won't have to do that.

jmakar-scalyr: Awesome

tests/unit/syslog_monitor_test.py
```diff
@@ -13,8 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ------------------------------------------------------------------------
-#
-# author: Imron Alston <imron@scalyr.com>
+# author: scalyr-cloudtech@scalyr.com
```
dkoshe, line 16: I would change this to `scalyr@sentinelone.com`

tests/unit/syslog_monitor_test.py
```diff
@@ -23,7 +22,7 @@
 if False:
     from typing import List
 
-__author__ = "imron@scalyr.com"
+__author__ = "scalyr-cloudtech@scalyr.com"
```
dkoshe, line 25: I would change this to `scalyr@sentinelone.com`

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -116,6 +123,19 @@
     default="agent_syslog.log",
 )
 
+define_config_option(
+    __monitor__,
+    "message_log_template",
```
ArthurKamalov, line 128: Since we do not bundle the `syslogmp` source with agent, I think we should mention here, that this option can be used only in dockerized agent, and not agent that is installed from packages.

tests/unit/syslog_monitor_test.py
jmakar-scalyr: Will do

tests/unit/syslog_monitor_test.py
jmakar-scalyr: Will do

scalyr_agent/builtin_monitors/syslog_monitor.py
jmakar-scalyr: Will do, I didn't like introducing a dependency but implementing the parsing myself would've been too time consuming

ArthurKamalov: Yeah, I can totally understand that...
So I think it should be enough just to be sure that the `syslog_monitor` still can operate without `syslogmp` and just explicitly state here, that new config is not for packages agent.

ArthurKamalov (approved):
LGTM

scalyr_agent/builtin_monitors/syslog_monitor.py
jmakar-scalyr: Yep, verified that the monitor works fine without syslogmp if the new option message_log_template is not used

PR #1037, opened 2022-11-29
DSET-842 Add dependency packages tests

ArthurKamalov:
This PR adds end to end testing for new managed Linux packages, for now only for dependency package, but this acts like a base for the actual agent package tests when it is added.

### How tests work
This test mainly recreates existing AMI tests, but also tries to remove some complexities. 
For example, instead of sending a deployment templated script to a libcloud Node, we just build and send standalone frozen binary of the `pytest` executable, which can run Python-written tests on any linux environment without any dependencies.

When this test binary is in the target machine, it starts testing the packages that are also deployed to the machine in form of the mock repository.  Testing the packages in form of a ready repository is important to verify that dependency relations between packages are correct.

Another small change is that `python` and `agent-libs` dependency packages now have simplified package structure. 
Instead of previous structure. where package had its own subdirectory in each folder of the `/usr/` directory, now,
it just has subdirectories in `/usr/lib/scalyr-agent-2-dependencies` - for platform-dependent files and `/usr/share/scalyr-agent-2-dependencies` - for platform-independent files.  (You may also notice how [build_python.sh](https://github.com/scalyr/scalyr-agent-2/blob/master/agent_build_refactored/managed_packages/steps/build_python.sh) and [build_agent_libs.sh](https://github.com/scalyr/scalyr-agent-2/blob/master/agent_build_refactored/managed_packages/steps/build_agent_libs.sh) scripts have become much simpler because of that)  Now approach is simpler and more reliable and more encouraged by the Python itself. And the previous approach, in opposite, was too fragile and also has non-zero chance that Python, in some circumstances,  may load something from system Python, because its prefix is still `/usr` as for the system's one. 
`sys.prefix` is very important for the interpreter because it uses it as a reference point for finding all other packages and libraries.

Now the Python's root prefix is guaranteed to be `/usr/lib/<subdir>`. Due to importance of being sure that Python starts with correct prefix settings I added simple sanity test [script](https://github.com/scalyr/scalyr-agent-2/blob/add-dependenci-packages-tests/tests/end_to_end_tests/managed_packages_tests/verify_python_interpreter.py) which is also executed in each ec2 instance to verify that Python interpreter works as expected on those distributions.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -90,10 +90,10 @@ jobs:
           - ubuntu
         include:
           - os: windows-2022
-          - platform: windows
+            platform: windows
             python-version: "3.10"
           - os: macos-11
-          - platform: macos
+            platform: macos
             python-version: "3.10"
     steps:
       - name: Checkout Repository


Kami:
> This test mainly recreates existing AMI tests, but also tries to remove some complexities.
For example, instead of sending a deployment templated script to a libcloud Node, we just build and send standalone frozen binary of the pytest executable, which can run Python-written tests on any linux environment without any dependencies.

I like this new approach.

Writing test files in Python should be much nicer than bash. IIRC, bash was used in the past as a lowest common denominator, but if we can get this new approach to work with all the distros we test with, then it works for me :)

agent_build_refactored/managed_packages/steps/build_agent_libs.sh
```diff
@@ -44,60 +44,14 @@ REQUIREMENTS_FILES_PATH="${SOURCE_ROOT}/agent_build/requirement-files"
...
-cp -a "${AGENT_LIBS_BUILD_ROOT}/usr/lib/python${PYTHON_SHORT_VERSION}"  "${AGENT_LIBS_ROOT}/usr/lib/${SUBDIR_NAME}"
-rm -r "${AGENT_LIBS_BUILD_ROOT}/usr/lib/python${PYTHON_SHORT_VERSION}"
-
-mkdir -p "${AGENT_LIBS_ROOT}/usr/libexec"
-mv "${AGENT_LIBS_BUILD_ROOT}/usr/bin" "${AGENT_LIBS_ROOT}/usr/libexec/${SUBDIR_NAME}"
-
-# Check if there are any files that have been left uncopied from the build root.
-REMAINING_AGENT_FILES=$(find "${AGENT_LIBS_BUILD_ROOT}" -type f)
-test "$(echo -n "${REMAINING_AGENT_FILES}" | wc -l)" = "0" || die "There are still files that are remaining non-copied to a result agent libs directory. Files: '${REMAINING_AGENT_FILES}'"
+"/usr/lib/${SUBDIR_NAME}/bin/python3" -m pip install -v --force-reinstall --root "${STEP_OUTPUT_PATH}/agent_libs"  \
```
Kami, line 55: Yeah, this approach is indeed much simpler and better 👍

tests/end_to_end_tests/managed_packages_tests/conftest.py
```diff
@@ -0,0 +1,409 @@
...
+        package_type=package_builder.PACKAGE_TYPE,
+        packages_dir_path=packages_dir
+    )
+    repo_dir = repo_builder.output_path
+
+    return repo_dir
+
+
+@pytest.fixture(scope="session")
+def repo_url(package_builder_name, package_builder, repo_dir, tmp_path_factory):
```
Kami, line 287: I was hoping we could keep this simpler - e.g. only install .rpm/.deb packages directly from file, but yeah I do understand the reasoning with testing the complete flow and "installing packages from the package repo".

At least for the stable tests (as we do for the AMI tests today) which run periodically, we should use actual upstream stable packagecloud repo, right? (so we exercise that repo code path and ensure repo is in order)

tests/end_to_end_tests/managed_packages_tests/test_packages.py
```diff
@@ -0,0 +1,154 @@
...
+from agent_build_refactored.managed_packages.managed_packages_builders import PYTHON_PACKAGE_NAME, \
+    AGENT_LIBS_PACKAGE_NAME, AGENT_DEPENDENCY_PACKAGE_SUBDIR_NAME
+
+logger = logging.getLogger(__name__)
+
+"""
+This test module preforms end to end testing of the Linux agent package and its dependency packages. 
+Since we perform testing for multiple distributions, those tests are mainly run inside another remote machines, 
+such as ec2 instance or docker container. If needed, it can be run locally, but you have to be aware that those tests
+are changing system state and maust be aware of risks.
```
Kami, line 31: small type - "maust"

tests/end_to_end_tests/managed_packages_tests/conftest.py
ArthurKamalov: Correct, for the stable tests, as we do now, we just use url of existing repo.

Yeah, maybe I've gone too far by trying to mimic real repo for the dev testing, and it should be enough to do it on only on stable tests. :sweat_smile:

I'll simplify that part.

Kami: Thanks - I'm always for simplicity and as at least abstractions / indirections as possible to make understanding, troubleshooting, etc. easier :)

agent_build/requirement-files/testing-requirements.txt
```diff
@@ -48,3 +48,7 @@ more-itertools==8.5.0; python_version >= '3.0'
 more-itertools==5.0.0; python_version <= '2.7'
 distro==1.6.0
 xmltodict==0.13.0; python_version >= '3.0'
+
+paramiko==2.12.0; python_version >= '3.6'
```
Kami, line 52: Do we still need / want to use a separate more minimum requirements file just for the AMI tests (as we do today, iirc)?

tests/end_to_end_tests/run_in_remote_machine/cleanup_test_ec2_objects.py
```diff
@@ -0,0 +1,225 @@
+import argparse
```
Kami, line 1: IIRC, there is duplicated between this and existing AMI clean up script, right?

I assume that eventually we will be able to get rid all of the old / duplicated code?

tests/end_to_end_tests/run_in_remote_machine/ec2.py
```diff
@@ -0,0 +1,370 @@
...
+    return int(prefix_list["Version"])
+
+
+def add_current_ip_to_prefix_list(
+    client,
+    prefix_list_id: str,
+    workflow_id: str = None
+):
+    """
+    Add new CIDR entry with current public IP in to the prefix list. We also additionally store json object in the
```
Kami, line 316: Would be good to add a comment somewhere on why maintaining the custom prefix list is necessary.

agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -57,13 +57,12 @@
     files of local system Python interpreter. To achieve that, the dependency packages files are installed in their
     own 'sub-directories'
 
-    /usr/libe/scalyr-agent-2-dependencies
-    /usr/libexec/scalyr-agent-2-dependencies
-    /usr/shared/scalyr-agent-2-dependencies
-    /usr/include/scalyr-agent-2-dependencies
+    For now there are two subdirectories:
+        - /usr/libe/scalyr-agent-2-dependencies - for platform dependent files.
```
jmakar-scalyr, line 61: Small typo, /usr/libe/...

agent_build_refactored/managed_packages/steps/build_python.sh
```diff
@@ -71,92 +69,53 @@ INCLUDE_DIR="include"
...
-	--bindir="${PYTHON_INSTALL_PREFIX}/${BIN_DIR}/${SUBDIR_NAME}" \
-	--datarootdir="${PYTHON_INSTALL_PREFIX}/${DATA_ROOT_DIR}/${SUBDIR_NAME}" \
-	--includedir="${PYTHON_INSTALL_PREFIX}/${INCLUDE_DIR}/${SUBDIR_NAME}" \
-	--with-platlibdir="${LIB_DIR}/${SUBDIR_NAME}"
+	--prefix="${PACKAGE_INSTALL_PREFIX}" \
+	--exec-prefix="${PACKAGE_INSTALL_EXEC_PREFIX}" \
+	--with-ensurepip=upgrade \
+	--with-suffix="-orig"
+
+#		--enable-optimizations \
```
jmakar-scalyr, line 77: Why were --enable-optimizations & --with-lto commented out?

tests/end_to_end_tests/managed_packages_tests/conftest.py
```diff
@@ -0,0 +1,194 @@
...
+        "--builder-name",
+        dest="builder_name",
+        required=True
+    )
+
+    parser.addoption(
+        "--packages-source",
+        dest="packages_source",
+        required=False,
+        help="Directory to packages to test. If not specified, packages will be built inplace."
```
jmakar-scalyr, line 33: Package source would be a directory or url (for a repo) right?
Should change the help text to be "Directory or tarball or repo url of packages to test. ..."

If repos are always specified as urls, can eliminate the --packages-source-type option (but up to you of course)

tests/end_to_end_tests/managed_packages_tests/fixtures/centos6.repo
```diff
@@ -0,0 +1,40 @@
+[C6.10-base]
```
jmakar-scalyr, line 1: CentOS 6 EOL was a couple years ago, is this still needed?

tests/end_to_end_tests/run_in_remote_machine/cleanup_test_ec2_objects.py
```diff
@@ -0,0 +1,232 @@
...
+    entries = resp["Entries"]
+
+    entries_to_remove = {}
+
+    current_time = time.time()
+
+    # Remove old prfix list entries.
+    for entry in entries:
+        timestamp = _parse_entry_timestamp(entry)
+        if timestamp <= current_time - PREFIX_LIST_ENTRY_REMOVE_THRESHOLD:
```
jmakar-scalyr, line 114: May want to ensure timestamp is a float or int in _parse_entry_timestamp to avoid potentially comparing a timestamp as a string to current_time

tests/end_to_end_tests/run_in_remote_machine/cleanup_test_ec2_objects.py
```diff
@@ -0,0 +1,232 @@
...
+        args.region
+    )
+
+    cleanup_old_prefix_list_entries(
+        client=boto_client,
+        prefix_list_id=args.prefix_list_id,
+        workflow_id=args.workflow_id
+    )
+
+    cleanup_old_ect_test_instance(
```
jmakar-scalyr, line 228: Should ect here be ec2?

tests/end_to_end_tests/run_in_remote_machine/cleanup_test_ec2_objects.py
```diff
@@ -0,0 +1,232 @@
...
+        "--region",
+        required=True
+    )
+    parser.add_argument(
+        "--workflow-id",
+        required=False
+    )
+    args = parser.parse_args()
+
+    boto_client = boto3.client(
```
jmakar-scalyr, line 207: This can be a TODO / future work, is there a way to generalize using libcloud to avoid having to specifically make a boto3 client here?  (I haven't played with it but I figured libcloud would provide generalized classes to make it easy to switch between cloud providers)

tests/end_to_end_tests/run_in_remote_machine/ec2.py
```diff
@@ -0,0 +1,361 @@
...
+"""
+This module defines main logic that is responsible for manipulating with ec2 instances to run end to end tests
+inside them.
+"""
+
+logger = logging.getLogger(__name__)
+
+# All the instances created by this script will use this string in the name.
+INSTANCE_NAME_STRING = "automated-agent-tests-"
+assert "-tests-" in INSTANCE_NAME_STRING
```
jmakar-scalyr, line 20: Why have this assertion?

tests/end_to_end_tests/run_in_remote_machine/ec2.py
```diff
@@ -0,0 +1,361 @@
...
+    ssh_username: str
+
+
+def destroy_node_and_cleanup(driver, node):
+    """
+    Destroy the provided node and cleanup any left over EBS volumes.
+    """
+    volumes = driver.list_volumes(node=node)
+
+    assert (
```
jmakar-scalyr, line 42: :+1:

tests/end_to_end_tests/run_in_remote_machine/ec2.py
```diff
@@ -0,0 +1,361 @@
...
+            )
+            print(str(e))
+        else:
+            raise e
+
+    assert len(volumes) <= 1
+    print("Cleaning up any left-over EBS volumes for this node...")
+
+    # Give it some time for the volume to become detached from the node
+    if volumes:
```
jmakar-scalyr, line 66: I believe there is a way to check volume status, but it can be a TODO

tests/end_to_end_tests/run_in_remote_machine/ec2.py
```diff
@@ -0,0 +1,361 @@
...
+        "ec2",
+        aws_access_key_id=access_key,
+        aws_secret_access_key=secret_key,
+        region_name=region
+    )
+
+    # Add current public IP to security group's prefix list.
+    # We have to update that prefix list each time because there are to many GitHub actions public IPs, and
+    # it is not possible to whitelist all of them in the AWS prefix list.
+    add_current_ip_to_prefix_list(
```
jmakar-scalyr, line 257: This can be another TODO but it would be good to remove it when done (I'm assuming the security group will always have it & just end up collecting lots of ips otherwise)

jmakar-scalyr (approved):
I'm not that familiar with some of these aspects but I didn't see any real issues (all of the isolated comments were for minor points).  Let me know if there are areas that need more careful review.

agent_build_refactored/managed_packages/steps/build_python.sh
ArthurKamalov: Correct, just to save time during the active development. Added reminder TODO to uncomment that for stable release.

tests/end_to_end_tests/managed_packages_tests/fixtures/centos6.repo
ArthurKamalov: We may not explicitly state  that we support CentOS 6, but since new package does not rely one any system dependency except `glibc`, and minimum supported `glibc` version is `2.12`, then still running tests for CentOS 6 should to be a good marker that the package does indeed work properly on systems with such old `glibc`.

tests/end_to_end_tests/run_in_remote_machine/cleanup_test_ec2_objects.py
ArthurKamalov: Sure, done.

tests/end_to_end_tests/run_in_remote_machine/cleanup_test_ec2_objects.py
ArthurKamalov: The `libcloud` does provide generalized interface, but, IFAIK, it does not provide it for AWS prefix lists manipulation, at least I could not find it :(

tests/end_to_end_tests/run_in_remote_machine/ec2.py
ArthurKamalov: Yeah, I think that may not be needed. I just copied it from the current [ami tests,](https://github.com/scalyr/scalyr-agent-2/blob/f5d4602cec828c11795bdd840d3f0830cd925a97/tests/ami/packages_sanity_tests.py#L242) as some other ec2 related things :)

tests/end_to_end_tests/run_in_remote_machine/ec2.py
ArthurKamalov: Added volumes status check.

ArthurKamalov:
PS: I will also revert the removal of the repository mocks. See more in this comment https://github.com/scalyr/scalyr-agent-2/pull/1037#discussion_r1039537706

PR #1039, opened 2022-11-30
DSET-1106: PR 1024 feedback

jmakar-scalyr:
- Added debug logs for couple scenarios
- Periodically (every 12 hours) log cache usage stats
- Ensured safe to call `_DLL.__del__` multiple times
- Added support for entry ttl in Cache, added unit test
- Made cache sizes and ttls configurable

### DIFF
--- a/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
+++ b/scalyr_agent/builtin_monitors/windows_event_log_monitor.py
@@ -830,10 +830,10 @@ class NewJsonApi(NewApi):
             if time.time() > self._next_cache_report:
                 self._next_cache_report = time.time() + self._cache_report_interval
                 self._logger.info(
-                    "placeholder param cache size = %d" % (len(self._param_cache),)
+                    "placeholder param cache size = %d" % len(self._param_cache)
                 )
                 self._logger.info(
-                    "dll handle cache size = %d" % (len(self._dll_cache),)
+                    "dll handle cache size = %d" % len(self._dll_cache)
                 )
 
         # Populate the record here with fields that would normally be added by the log formatter,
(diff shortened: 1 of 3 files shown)


scalyr_agent/builtin_monitors/windows_event_log_monitor.py
```diff
@@ -738,8 +774,18 @@ def __init__(self, config, logger, channels):
...
+            self._dll_cache = Cache(
+                size=config.get("dll_handle_cache_size"),
+                ttl=config.get("dll_handle_cache_ttl"),
+            )
+            self._param_cache = Cache(
+                size=config.get("placeholder_param_cache_size"),
+                ttl=config.get("placeholder_param_cache_ttl"),
+            )
+
+            self._cache_report_interval = 3600 * 12
```
ArthurKamalov, line 786: Since we already provide ability to fine-tune the cache, maybe it would be also helpful to configure the interval?

tests/unit/cache_test.py
```diff
@@ -0,0 +1,77 @@
...
+        time.sleep(1)
+        self.assertTrue("a" not in cache)
+
+    def test_ttl_via_iter(self):
+        cache = Cache(1000, 1)
+
+        cache["a"] = 1
+        self.assertTrue("a" in cache)
+
+        time.sleep(1)
```
ArthurKamalov, line 58: Since it does not seems like we do rounding to seconds in the cache logic, maybe it would be enough to sleep just some fraction of a second to prove the point?

ArthurKamalov (approved):
LGTM, just added some very minor comments

scalyr_agent/builtin_monitors/windows_event_log_monitor.py
jmakar-scalyr: You're probably right, but I'll hold off for now

tests/unit/cache_test.py
jmakar-scalyr: Ooh great idea, will fix

PR #1043, opened 2022-12-06
[DSET-841] Build test and publish agent package to packagecloud

ArthurKamalov:
This pull request adds production versions of the agent package to the Packageloud.

Another big change is that current end-to-end (or so-called 'AMI' tests) have been fully rewritten in Python ([link](https://github.com/scalyr/scalyr-agent-2/blob/0896974089c106ac251e7f731f27c9d6463a7a98/tests/end_to_end_tests/managed_packages_tests/test_packages.py#L138)).


agent_build_refactored/managed_packages/convenience_install_script/install_agent_template.sh
```diff
@@ -0,0 +1,432 @@
+#!/bin/bash
```
ArthurKamalov, line 1: This is basically just a copy of [existing install script ](https://github.com/scalyr/scalyr-agent-2/blob/master/scripts/circleci/release/installScalyrAgentV2.sh) but without some legacy parts.

agent_build_refactored/managed_packages/files/init.d/scalyr-agent-2
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
ArthurKamalov, line 1: Also Just a copy of existing file.

agent_build_refactored/managed_packages/install-scriptlets/preuninstall.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
ArthurKamalov, line 1: Full copy of the https://github.com/scalyr/scalyr-agent-2/blob/master/installer/scripts/preuninstall.sh

agent_build_refactored/scripts/filter_matrix_jobs.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2014-2022 Scalyr Inc.
```
ArthurKamalov, line 1: This basically just a part of the `agent_build_refactored/scripts/get_github_actions_job_matrices.py` file, which was logically split into three files.

agent_build_refactored/scripts/get_workflow_info.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2014-2022 Scalyr Inc.
```
ArthurKamalov, line 1: This basically just a part of the `agent_build_refactored/scripts/get_github_actions_job_matrices.py` file, which was logically split into three files.

.github/actions/init-job-matrices/action.yml
```diff
@@ -45,34 +55,38 @@ runs:
...
+        echo "to_publish=$(cat /tmp/workflow_info.json | jq -c '.to_publish')" >> $GITHUB_OUTPUT
+        echo "is_production=$(cat /tmp/workflow_info.json | jq -c '.is_production')" >> $GITHUB_OUTPUT
+        echo "version=$(cat /tmp/workflow_info.json | jq -c '.version')" >> $GITHUB_OUTPUT
+        
+        mkdir -p /tmp/matrices
+        
+        cat <<EOT | python3 agent_build_refactored/scripts/filter_matrix_jobs.py --is-master-run "${IS_MASTER_RUN}" > /tmp/matrices/agent_image_build_matrix.json
         ${{ inputs.images-to-build }}
         EOT
+        echo "agent_image_build_matrix_json=$(cat /tmp/matrices/agent_image_build_matrix.json)" >> $GITHUB_OUTPUT
```
jmakar-scalyr, line 76: I think this is a litter clearer:
```
echo ${{ inputs.images-to-build }} | python3 agent_build_refactored/... > /tmp/matrices/...
echo "agent_image_build_matrix_json=$(cat /tmp/matrices/...)" >> $GITHUB_OUTPUT
```

If the temp files are not needed later could even do this:
```
echo "agent_image_build_matrix_json=$(echo ${{ inputs.image-to-build }} | python3 agent_build_refactored/...)" >> $GITHUB_OUTPUT
```
That may be less readable so I'll leave it up to you.

This overall comment applies to the other two variables below also.

.github/workflows/agent-build-refactored.yml
```diff
@@ -23,9 +25,10 @@ jobs:
   pre_job:
     runs-on: ubuntu-20.04
     outputs:
-      should_skip: ${{ steps.skip_check.outputs.should_skip }}
```
jmakar-scalyr, line 26: Out of curiosity, why was this removed / no longer necessary?

agent_build_refactored/docker_image_builders.py
```diff
@@ -259,8 +259,8 @@ def build_container_filesystem(self, output_path: pl.Path):
             shutil.rmtree(package_root_path)
         package_root_path.mkdir(parents=True)
 
-        # Build 'LFS-structured' filesystem.
-        build_linux_lfs_agent_files(
+        # Build 'HFS-structured' filesystem.
```
jmakar-scalyr, line 262: Looks like a typo here: HFS-structured => FHS-structured

By the way, FHS refers to https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard right?  What does LFS refer to?

agent_build_refactored/scripts/get_workflow_info.py
```diff
@@ -0,0 +1,153 @@
...
+GITHUB_TOKEN = os.environ.get("GITHUB_TOKEN")
+GITHUB_SHA = os.environ.get("GITHUB_SHA", "")
+
+
+def is_branch_has_pull_requests() -> bool:
+    """
+    Determine if current branch has at least one opened pull request.
+    """
+    data = subprocess.check_output([
+        "curl",
```
jmakar-scalyr, line 50: I think will also need a `-s` here to avoid showing the progress meter

agent_build_refactored/tools/runner.py
```diff
@@ -831,7 +830,7 @@ def run_in_docker(
 
             mount_args.extend([
                 "-v",
-                f"{arg.path}:{in_docker_path}:z"
```
jmakar-scalyr, line 834: Why is the :z no longer necessary?  It's no longer shared?

tests/end_to_end_tests/managed_packages_tests/test_packages.py
```diff
@@ -167,81 +189,400 @@ def test_packages(
...
+    assert "ca_cert_path: /etc/ssl/system-ca-bundle.crt" in agent_log
+    assert "HttpConnection uses native os ssl" in agent_log
+
+    agent_status = agent_commander.get_status()
+    assert "Last successful communication with Scalyr:" in agent_status
+    assert "Last successful communication with Scalyr: Never" not in agent_status
+    assert "Last copy response status:                 success" in agent_status
+    _stop_agent_and_remove_logs_and_data(agent_commander=agent_commander)
+
+    # 3. Mimic MITM attack by pointing a random domain to agent.scalyr.com and verify hostname
```
jmakar-scalyr, line 389: Nicely done!

agent_build_refactored/scripts/filter_matrix_jobs.py
```diff
@@ -0,0 +1,75 @@
...
+DEFAULT_OS = os.environ["DEFAULT_OS"]
+DEFAULT_PYTHON_VERSION = os.environ["DEFAULT_PYTHON_VERSION"]
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--is-master-run",
+        required=True
+    )
```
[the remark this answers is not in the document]
ArthurKamalov: Since we get input value for this option from JSON, then it can be `true` or `false`, in case of `false` the `store_true` option will still be true, so, on the "command caller" side, we must not provide the `--is-master-run`  at all, but that's also adds some extra verbosity.

I guess we can do that with bash's conditional substitution, (like `${value:--is-master-run}`) but in this case we still have to return empty string instead of `false` in our JSON object.

agent_build_refactored/tools/runner.py
ArthurKamalov: I, actually, haven't got any difference with or without it. I just added it in hope that it will solve user permission differences of host and guest OS's, but it didn't.

agent_build_refactored/docker_image_builders.py
ArthurKamalov: Oops, yes it does, fixed.  Thanks!

.github/workflows/agent-build-refactored.yml
ArthurKamalov: Because for this workflow I just left `opened`  event for the `pull_request` trigger type, https://github.com/scalyr/scalyr-agent-2/blob/0896974089c106ac251e7f731f27c9d6463a7a98/.github/workflows/agent-build-refactored.yml#L4-L7
 so there should not be any "double triggering" from both the `push` and `pull_request` when you push new commits to already opened pull request.

.github/actions/init-job-matrices/action.yml
ArthurKamalov: There reason is because that values are multiline string and with `echo ${{ inputs.images-to-build }}`, it just leads to bash syntactic error. And with `echo "${{ inputs.images-to-build }}"` it just removes all double quoted from the string, so it's not a valid JSON anymore.
But thanks for pointing on this again, because it seems that single quote does the trick `echo '${{ inputs.images-to-build }}'` 👍

tests/end_to_end_tests/managed_packages_tests/test_packages.py
ArthurKamalov: Thanks, but that's already existing checks and I just moved them here :sweat_smile:

PR #1055, opened 2022-12-19
DSET-1019: Log file attributes via file template

jmakar-scalyr:
As discussed in DSET-1019, the attributes are collected in order to populate the file template (message_log_template) so this change is just a matter of also adding the attributes as server attributes/fields.

This pr also includes an unrelated, minor bugfix (__syslog_attributes must be a JsonObject).


ArthurKamalov (approved):
LGTM

PR #1056, opened 2022-12-20
DSET-1296 Improve arm64 build

ArthurKamalov:
This PR speeds up building of new agent packages with for ARM by spinning up ARM EC2 instance with docker engine. Now all compute intensive compilation steps of ARM binaries can be run in this remote docker without emulation, reducing overall build time from 2-3 hours to similar to x86_64 packages.

I also had to stop using `apache-libcloud` as a tool to spawn ec2 instances since it does not provide some key features to normally clean up volumes, so now everything is done by the `boto3` library.


agent_build_refactored/tools/run_in_ec2/boto3_tools.py
```diff
@@ -0,0 +1,443 @@
+# Copyright 2014-2022 Scalyr Inc.
```
ArthurKamalov, line 1: This mainly just a refactored `tests/end_to_end_tests/run_in_remote_machine/ec2.py`, Since we now use ec2 instances during the build process too, it was moved from tests to build packages.

agent_build_refactored/tools/run_in_ec2/cleanup_ec2_objects.py
```diff
@@ -0,0 +1,236 @@
+# Copyright 2014-2022 Scalyr Inc.
```
ArthurKamalov, line 1: The same. Just moved `tests/end_to_end_tests/run_in_remote_machine/cleanup_test_ec2_objects.py` And replaced libcloud  with boto3

agent_build_refactored/tools/runner.py
```diff
@@ -15,31 +15,32 @@
```
ArthurKamalov, line 15: This diff is quite messy, but its mainly just refactored in order to be able to use remote docker engine form ec2 to run some steps in it.

.github/workflows/agent-build-refactored.yml
```diff
@@ -123,10 +140,41 @@ jobs:
...
+
+      - name: Prepare SSH
+        id: prepare-ssh
+        shell: bash
+        run: |
+          echo '${{ secrets.AWS_DEV_US_EAST_1_KEYPAIR_NAME }}' > /tmp/private_key.pem
+          chmod 600 /tmp/private_key.pem
+          eval `ssh-agent -s`
+          
+          cat "/tmp/private_key.pem" | ssh-add -
```
jmakar-scalyr, line 161: I think these lines can be simplified (no need to write the key to the filesystem):
```
eval `ssh-agent -s`
echo '${{ secrets.AWS_DEV_US_EAST_1_KEYPAIR_NAME }}' | ssh-add -
```

jmakar-scalyr: Hm maybe I'm wrong because I see the `aws_private_key_path: "/tmp/private_key.pem"` below

In that case why use both ssh-agent and store the key on the filesystem?

agent_build_refactored/managed_packages/steps/build_python.sh
```diff
@@ -95,8 +95,8 @@ cp -a /usr/local/lib/libuuid.so* "${BUILD_ROOT}${PACKAGE_INSTALL_EXEC_PREFIX}/li
 cp -a /usr/local/lib/libgdbm.so* "${BUILD_ROOT}${PACKAGE_INSTALL_EXEC_PREFIX}/lib"
 cp -a /usr/local/lib/libgdbm_compat.so* "${BUILD_ROOT}${PACKAGE_INSTALL_EXEC_PREFIX}/lib"
 cp -a /usr/local/lib64/libffi.so* "${BUILD_ROOT}${PACKAGE_INSTALL_EXEC_PREFIX}/lib"
-cp -a /usr/local/lib64/libcrypto.so* "${BUILD_ROOT}${PACKAGE_INSTALL_EXEC_PREFIX}/lib"
-cp -a /usr/local/lib64/libssl.so* "${BUILD_ROOT}${PACKAGE_INSTALL_EXEC_PREFIX}/lib"
+cp -a /usr/local/${LIBSSL_DIR}/libcrypto.so* "${BUILD_ROOT}${PACKAGE_INSTALL_EXEC_PREFIX}/lib"
+cp -a /usr/local/${LIBSSL_DIR}/libssl.so* "${BUILD_ROOT}${PACKAGE_INSTALL_EXEC_PREFIX}/lib"
```
jmakar-scalyr, line 99: I think it's better to make LIBSSL_DIR a full path but I'll leave it up to you if you want to change that now or not

agent_build_refactored/managed_packages/steps/install_gcc_7_arm64.sh
```diff
@@ -0,0 +1,38 @@
...
+# This script prepares base build environment for the ARM64 linux GLIBC binary packages, it expects to be run in
+# Centos 7 to compile against lower GLIBS (2.17).
+
+set -e
+
+# RHSCL is installed, so we can install newer tools, such as gcc-7
+yum install -y centos-release-scl
+yum install -y devtoolset-7
+
+# Remove this preinstalled packages, since we build and install those libraries from source.
```
jmakar-scalyr, line 30: Is this comment accurate?  Why build perl from source?

agent_build_refactored/tools/run_in_ec2/boto3_tools.py
```diff
@@ -0,0 +1,445 @@
...
+            private_key_path=_validate_setting("AWS_PRIVATE_KEY_PATH"),
+            private_key_name=_validate_setting("AWS_PRIVATE_KEY_NAME"),
+            region=_validate_setting("AWS_REGION"),
+            security_group=_validate_setting("AWS_SECURITY_GROUP"),
+            security_groups_prefix_list_id=_validate_setting("AWS_SECURITY_GROUPS_PREFIX_LIST_ID"),
+            ec2_objects_name_prefix=_validate_setting("AWS_OBJECTS_NAME_PREFIX")
+        )
+
+    def create_boto3_session(self):
+        import boto3
```
jmakar-scalyr, line 89: This line can be removed

agent_build_refactored/tools/run_in_ec2/boto3_tools.py
```diff
@@ -0,0 +1,445 @@
...
+            ssh_connection=ssh,
+            files=files_to_upload
+        )
+        if deployment_command:
+            ssh_run_command(
+                ssh_connection=ssh,
+                command=deployment_command
+            )
+
+        ssh.close()
```
jmakar-scalyr, line 173: The ssh.close() should occur even if there are no files_to_upload

agent_build_refactored/tools/run_in_ec2/boto3_tools.py
```diff
@@ -0,0 +1,445 @@
...
+                                f"support for 'gp2' volumes. Our system will be working on provisioning additional " \
+                                f"capacity."
+            if no_capacity_error not in message:
+                logger.exception("Unrecoverable error has occurred.")
+                raise
+
+            logger.info(f"    {e}")
+            logger.info(f"    Retry...")
+            attempts -= 1
+            time.sleep(random.randint(5, 10))
```
jmakar-scalyr, line 266: Maybe sleep longer here?  I don't expect ec2 instances to be destroyed so frequently, maybe a sleep of 1 to 3 minutes is more appropriate, wdyt?

jmakar-scalyr (approved):
Looks good.  A silly question, why use a docker engine for this?  Cost savings?  I figured it would be simpler just to spin up native arm64 ec2 instances instead.

ArthurKamalov:
> Looks good. A silly question, why use a docker engine for this? Cost savings? I figured it would be simpler just to spin up native arm64 ec2 instances instead.


Yes, the whole point is to use native arm64 ec2 instance, but Docker is useful because it allows to run any distribution we want. For example we build Python on CentOS (because it has lowest glibc), but do other things on new Ubuntu, and it's just faster to spin up ec2 instance with docker once, and then run containers with CentOS, Ubuntu, rather than create two ec2 instances with CentOS and Ubuntu.  

Using docker also allows us to save whole image of the build environment and, for example, cache it in GitHub actions, so there's no need to rebuild it if it's not changed.

agent_build_refactored/tools/run_in_ec2/boto3_tools.py
ArthurKamalov: oops. Thanks.

agent_build_refactored/managed_packages/steps/install_gcc_7_arm64.sh
ArthurKamalov: We have to build and install perl for X86_64 builds because Centos:6 has pretty old perl and some configure script do not work with it. For ARM, even if it has never version, I think it would be still better to do the same to keep consistent tools versions.

PR #1059, opened 2023-01-06
rename scalyr agent prod write token

ArthurKamalov:


Kami (approved):
LGTM, thanks (explicit names are always good) 👍 

IIRC, this token is also for the cloudtest testing account, right? If so, it may make sense to also encode that into the key name.

PR #1060, opened 2023-01-07
Fix of malformed JSON

MrKragh:
Hi There, not much to say, just making it valid JSON as my editor didn't like the malformed version.

### DIFF
--- a/config/agent.json
+++ b/config/agent.json
@@ -14,7 +14,7 @@
 
      // You can add whatever additional fields you'd like.
      // tier: "production"
-  }
+  },
 
   // Log files to upload to Scalyr. You can use '*' wildcards here.
   logs: [


Kami:
Thanks for the contribution - just to clarify, DataSet agent configs are not pure JSON, they utilize Hocon (https://myarch.com/dpbuddy-docs/hocon.html) which is closer to a subset of JavaScript (it allows comments and commas and quoted values are not mandatory).

Kami (approved):

ArthurKamalov:
Unfortunately, I have to revert this change because requires additional attention from the user because the `apt` package manager can not handle this change automatically:

```
Configuration file '/etc/scalyr-agent-2/agent.json'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** agent.json (Y/I/N/O/D/Z) [default=N] ? 
```

For now we would like to avoid this, since some of user installations may be automated, and this change may add complications.

PR #1064, opened 2023-01-10
DSET-1408 Restructure agent package

ArthurKamalov:
This changes (I hope for the last time) structure of the new agent dependency packages.  Now all files from the dependency packages are located on the `/opt/scalyr-agent-dependencies` folder, which should guarantee that we do not mess with system packages.

Another change is that Python3 dependency package now can use system OpenSSL, and, only if it can't be found, fallbacks to "embedded"  OpenSSL. It's better to try to use system's SSL first, since OS's maintainer may deliver security updates faster than us.
 
Additional changes:
*  New requirements file in `/opt/scalyr-agent-dependencies/etc/additional-requirements.txt` allows used to specify additional requirements for the agent and to install them, for example, when they use custom monitor with additional dependencies. This as achievable now, because the `scalyr-agent-libs` package, when installed, creates `venv` with all needed requirements, including additional ones.

* New single[ requirements file](https://github.com/scalyr/scalyr-agent-2/blob/e262167619e33af416d07cdcf2a99ff0d5277353/dev-requirements-new.txt) with all project requirements.


agent_build_refactored/prepare_agent_filesystem.py
```diff
@@ -391,3 +391,350 @@ def build_linux_fhs_agent_files(
         "..", "share", "scalyr-agent-2", "bin", "scalyr-agent-2-config"
     )
     scalyr_agent_config_symlink_path.symlink_to(scalyr_agent_config_symlink_target_path)
+
```
ArthurKamalov, line 394: This just a copy-paste from the build_package.py

agent_build_refactored/managed_packages/steps/download_build_dependencies/download_build_dependencies.sh
```diff
@@ -18,29 +18,21 @@
...
 #
-# This script downloads source files for all libraries that are required to build Python. Since we build Python and its
-# requirements in a legacy OS (such as Centos 6) to achieve binary compatibility, we also have to build many essential
-# build tools because tools from legacy OS's ate too old.
+# This script downloads source files for all libraries that are required to build Python.
 #
 # It expects next environment variables:
 #
 #   XZ_VERSION: version of XZ Utils to build. Python requirement. ALso required by some make and configure scripts.
-#   PERL_VERSION: version of Perl to build. Required by some make and configure scripts.
```
ArthurKamalov, line 28: I found out that even centos:6 can provide more recent build tools, so we don't have to build them, only python dependencies.

Kami:
> Another change is that Python3 dependency package now can use system OpenSSL, and, only if it can't be found, fallbacks to "embedded" OpenSSL. It's better to try to use system's SSL first, since OS's maintainer may deliver security updates faster than us.

This seems reasonable for me.

> New requirements file in /opt/scalyr-agent-dependencies/etc/additional-requirements.txt allows used to specify additional requirements for the agent and to install them, for example, when they use custom monitor with additional dependencies. This as achievable now, because the scalyr-agent-libs package, when installed, creates venv with all needed requirements, including additional ones.

How will this affect upgrades and such? Or on upgrade, we simply skip and ignore anything under `/opt/scalyr-agent-dependencies/etc`?

agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -81,10 +74,26 @@
 EMBEDDED_PYTHON_VERSION = "3.11.0"
 EMBEDDED_PYTHON_SHORT_VERSION = ".".join(EMBEDDED_PYTHON_VERSION.split(".")[:2])
 
-PYTHON_PACKAGE_SSL_VERSION = "1.1.1k"
+PYTHON_PACKAGE_SSL_1_1_1_VERSION = "1.1.1k"
```
Kami, line 77: Is this the latest 1.1 version we can use?

agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -972,269 +1082,477 @@ def create_build_dependencies_step(
...
+    steps = {}
+
+    for architecture in SUPPORTED_ARCHITECTURES:
+        run_in_remote_docker = architecture != Architecture.X86_64
+
+        build_python = ArtifactRunnerStep(
+            name=f"build_python_{openssl_version_type}_{architecture.value}",
+            script_path="agent_build_refactored/managed_packages/steps/build_python.sh",
+            base=INSTALL_BUILD_ENVIRONMENT_STEPS[architecture],
+            required_steps={
```
Kami, line 1208: Since this value ``required_steps`` is a dictionary which is unordered by default, how do we achieve right order of those steps? 

Parent class (``RunnerStep``) description also says ":param required_steps: List of other steps that has to be executed in order to run this step." which is a bit confusing.

agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -972,269 +1082,477 @@ def create_build_dependencies_step(
...
+                "dev-requirements-new.txt",
+            ],
+            base=INSTALL_BUILD_ENVIRONMENT_STEPS[architecture],
+            required_steps={
+                "BUILD_PYTHON_DEPENDENCIES": BUILD_PYTHON_DEPENDENCIES_STEPS[architecture],
+                "BUILD_OPENSSL": BUILD_OPENSSL_STEPS[DEFAULT_OPENSSL_VERSION][architecture],
+                "BUILD_PYTHON": BUILD_PYTHON_STEPS[DEFAULT_OPENSSL_VERSION][architecture],
+            },
+            environment_variables={
+                "RUST_VERSION": "1.63.0",
```
Kami, line 1320: Would it make sense to put this value in some module level constant to make it easier to see it at a glance with other versions + update it?

agent_build_refactored/managed_packages/scalyr_agent_libs/agent-libs-config
```diff
@@ -0,0 +1,75 @@
...
+
+    # This is pre-created venv that is shipped with package.
+    venv_source_dir = DEPENDENCIES_PACKAGE_ROOT / "venv"
+
+    venv_dst_dir = pl.Path("/var/opt/scalyr-agent-2-dependencies/venv")
+    if venv_dst_dir.exists():
+        shutil.rmtree(venv_dst_dir)
+
+    # Copy venv that is shipped with the package.
+    shutil.copytree(
```
Kami, line 42: If I understand it correctly, we utilize the same virtualenv for core agent dependencies + user dependencies?

Would it make sense to utilize different environment for those (e.g. to make updating easier or similar)?

agent_build_refactored/managed_packages/scalyr_agent_python3/agent-python3-config
```diff
@@ -0,0 +1,193 @@
+#!/opt/scalyr-agent-2-dependencies/bin/python3
```
Kami, line 1: Just to confirm, we also run all the lint checks on the ``agent_build_refactored/`` directory, right?

agent_build_refactored/managed_packages/scalyr_agent_python3/agent-python3-config
```diff
@@ -0,0 +1,193 @@
...
+    """
+    Determine current system OpenSSL version, if presented.
+    :param version_type: version of the OpenSSL to use.
+    :return:
+    """
+    result = subprocess.run(
+        [
+            sys.executable,
+            "-c",
+            "import ssl; print(ssl.OPENSSL_VERSION);"
```
Kami, line 138: So we actually determine which OpenSSL is available on the system by checking against which OpenSSL version Python interpreter is compiled / linked against?

Does Python interpreter always dynamically links to OpenSSL library so this will always work as expected, or?

agent_build_refactored/managed_packages/steps/build_python_dependencies.sh
```diff
@@ -0,0 +1,125 @@
...
+
+# shellcheck disable=SC1090
+source ~/.bashrc
+
+
+DESTDIR_ROOT=/tmp/root
+
+mkdir /tmp/build-xz
+pushd /tmp/build-xz
+tar -xvf "${DOWNLOAD_BUILD_DEPENDENCIES}/xz/xz.tar.gz"
```
Kami, line 44: And here

agent_build_refactored/tools/constants.py
```diff
@@ -77,6 +95,10 @@ class Architecture(enum.Enum):
     ARM = "arm"
     ARMV7 = "armv7"
     ARMV8 = "armv8"
+    RISCV = "riscv64"
```
Kami, line 98: Just curious, do we also need to support / build for those architectures?

agent_build_refactored/tools/run_in_ec2/boto3_tools.py
```diff
@@ -129,46 +129,51 @@ def create_and_deploy_ec2_instance(
         root_volume_size=root_volume_size
     )
 
-    # Try to establish SSH connection.
```
Kami, line 132: Do we have any kind of period clean up process (like we had with libcloud) to make sure we periodically clean up any stale / stray instances?

agent_build_refactored/tools/runner.py
```diff
@@ -988,10 +992,18 @@ def get_remote_docker_host_for_step(step: RunnerStep) -> Optional[str]:
...
 
             # Get EC2 AMI image according to step's architecture.
+
             ec2_image = DOCKER_EC2_BUILDERS.get(step.architecture)
             # If image is not found then just run in local docker engine.
-            if ec2_image is None:
-                return None
+            if ec2_image is None and step.architecture != Architecture.X86_64:
+                ec2_image = EC2DistroImage(
+                    image_id="ami-09d56f8956ab235b3",
```
Kami, line 1000: Should probably make it a module level constant with some comment on what image it is, etc.

Kami:
Just a more general comment.

Since we sadly need to "reinvent" a lot of package manager functionality ourselves (downloading, building, signature verification, updates, etc.), I think we should try to make updating and managing those dependencies as easy as possible - e.g. make sure all the versions are defined in a single file as a top level constant or similar (right now it looks like versions are split across multiple files and scripts) and ideally also have some workflows which check for vulnerabilities / latest versions of those dependencies + some README / info on how to manage + update those (+ which files need to be modified for that).

agent_build_refactored/managed_packages/scalyr_agent_libs/agent-libs-config
ArthurKamalov: I've tried to make it clearer in this docstring

agent_build_refactored/managed_packages/managed_packages_builders.py
ArthurKamalov: There's, actually, a new version 1.1.1s :) The new job that scan's for openssl updates is still waits to be added.

agent_build_refactored/managed_packages/managed_packages_builders.py
ArthurKamalov: Sorry, that docstring was left outdated. The `required_steps` is used by step in order to access result artifacts of the `required_steps`, order here is irrelevant.

agent_build_refactored/managed_packages/scalyr_agent_libs/agent-libs-config
ArthurKamalov: I've tried to make it clearer in this docstring in this commit https://github.com/scalyr/scalyr-agent-2/blob/d48298a45a322219275048948038afdb0968664f/agent_build_refactored/managed_packages/managed_packages_builders.py#L73

agent_build_refactored/managed_packages/scalyr_agent_python3/agent-python3-config
ArthurKamalov: strange, it is not added to tox, but  it still was checked by the lint job. 
Anyway, added it to tox.

agent_build_refactored/managed_packages/scalyr_agent_python3/agent-python3-config
ArthurKamalov: Yes, by default, its C bindings for `_ssl` and `_hashlib` modules dynamically link to system's OpenSLL, but we also now. ship with the package another variants of those bindings that are parched to be "hardocded" to use OpenSSL
shared libraries that are also shipped with package in the `/opt/scalyr-agent-dependencies/lib/openssl` library.

I've also tried to add this to this docstring 
https://github.com/scalyr/scalyr-agent-2/blob/d48298a45a322219275048948038afdb0968664f/agent_build_refactored/managed_packages/managed_packages_builders.py#L50

agent_build_refactored/managed_packages/steps/build_openssl/build_openssl_1_1_1.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# This script builds from source the OpenSSL 1.1.1+ library.
+#
+
+
+set -e
+
+mkdir /tmp/build-openssl_1_1_1
+pushd /tmp/build-openssl_1_1_1
+tar -xvf "${DOWNLOAD_BUILD_DEPENDENCIES}/openssl_1_1_1/openssl.tar.gz"
```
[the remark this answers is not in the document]
ArthurKamalov: Sorry, maybe I can not get it properly, but the signature is verified in the separate step:
https://github.com/scalyr/scalyr-agent-2/blob/c9b4ad4ed02b7018ceda739f12d3395a5c0a2656/agent_build_refactored/managed_packages/steps/download_build_dependencies/download_build_dependencies.sh#L888-L895

agent_build_refactored/managed_packages/steps/build_openssl/build_openssl_3.sh
```diff
@@ -0,0 +1,41 @@
...
+# This script builds from source the OpenSSL 3+ library.
+#
+
+
+set -e
+
+
+mkdir /tmp/build-openssl_3
+pushd /tmp/build-openssl_3
+tar -xvf "${DOWNLOAD_BUILD_DEPENDENCIES}/openssl_3/openssl.tar.gz"
```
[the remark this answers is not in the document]
ArthurKamalov: As in previous comment, this is done in a separate step: https://github.com/scalyr/scalyr-agent-2/blob/c9b4ad4ed02b7018ceda739f12d3395a5c0a2656/agent_build_refactored/managed_packages/steps/download_build_dependencies/download_build_dependencies.sh#L1016-L1026

agent_build_refactored/managed_packages/steps/build_python.sh
```diff
@@ -45,72 +32,36 @@ set -e
 # shellcheck disable=SC1090
 source ~/.bashrc
 
+tar -xzvf "${BUILD_PYTHON_DEPENDENCIES}/common.tar.gz" -C /
```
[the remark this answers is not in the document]
ArthurKamalov: https://github.com/scalyr/scalyr-agent-2/blob/c9b4ad4ed02b7018ceda739f12d3395a5c0a2656/agent_build_refactored/managed_packages/steps/download_build_dependencies/download_build_dependencies.sh#L1141-L1150

agent_build_refactored/managed_packages/steps/build_python_dependencies.sh
ArthurKamalov: Done in a separate step.

agent_build_refactored/tools/runner.py
ArthurKamalov: It's also a leftover from my attempts to compile Python for other platform, removed it for now.

ArthurKamalov:
> Since we sadly need to "reinvent" a lot of package manager functionality ourselves (downloading, building, signature verification, updates, etc.), I think we should try to make updating and managing those dependencies as easy as possible - e.g. make sure all the versions are defined in a single file as a top level constant or similar (right now it looks like versions are split across multiple files and scripts) and ideally also have some workflows which check for vulnerabilities / latest versions of those dependencies + some README / info on how to manage + update those (+ which files need to be modified for that).

Yep, that's among thangs that I am planning to do next.

agent_build_refactored/managed_packages/managed_packages_builders.py
Kami: Sorry, I wanted to use if 1.1.1k is the latest version we can use due to the compatibility reasons or we can use latest available version out there (s or k)?

agent_build_refactored/managed_packages/scalyr_agent_python3/agent-python3-config
Kami: Thanks for adding this additional explanation.

I assume we have some post build / CI checks that ensure that the compiled Python binary was linked against the correct OpenSSL version?

agent_build_refactored/managed_packages/steps/build_openssl/build_openssl_3.sh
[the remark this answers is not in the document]
Kami: Thanks. I just wasn't completely sure if those steps are related / if this one runs before the download one which does check the signatures.

agent_build_refactored/tools/run_in_ec2/boto3_tools.py
Kami: Thanks for clarification. I assume we also run / could run this job on a cron trigger so we don't have stray instances running a run in case previous build was interrupted and in case no new build is started for a while which would clean up old / stray instances.

agent_build_refactored/managed_packages/build_dependencies_versions.py
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This module defines version for Python interpreter that is used in our dependency packages and versions
+of some of its dependencies.
+"""
+
+EMBEDDED_PYTHON_VERSION = "3.11.0"
```
Kami, line 20: Thanks 👍

Kami (approved):
LGTM 👍

agent_build_refactored/managed_packages/managed_packages_builders.py
ArthurKamalov: As I understood from my observations, all patch versions of 1.1.1 (aka 1.1.1[a-z]) have to be binary compatible. For the Python it also does not matter what patch version is used, it just has to be 1.1.1+, and it can resolve everything automatically.

Kami: Ack, thanks. In this case it would likely make sense for us to try to use the latest available version.

agent_build_refactored/managed_packages/scalyr_agent_python3/agent-python3-config
ArthurKamalov: ah, thats a good idea.
I added that checks for some distributions. 
https://github.com/scalyr/scalyr-agent-2/blob/83b963f6d04596fbdac4c25344c7efc65d8c7bad/tests/end_to_end_tests/managed_packages_tests/test_packages.py#L484-L492

I think adding checks for all tested distributions is a bad idea since their versions will eventually change which will be causing additional flakiness of the tests.

full commit: https://github.com/scalyr/scalyr-agent-2/commit/e18d38e4af333d1abd3b1ec5c706e4bc607354cd

PR #1072, opened 2023-01-25
[DSET-1580] Fix custom monitors dependency issues

ArthurKamalov:
This PR adds more information about how to install requirements from custom monitors.

It also changes the method of how Python package links to its  shared objects. Previously it was done by patching the binaries and hardcoding shared object. This was generally OK, but I've figured out that it may have some side-effects and it seems that the `patchelf` tool that is used to patch binaries is also not 100% free from bugs. 

The new approach is the old approach ,and it just to override the `LD_LIBRARY_PATH`. This may be not so elegant as patching, since we have to create shell script "wrappers" for our binaries, that have to override the `LD_LIBRARY_PATH`, but it has to be more robust, since the `LD_LIBRARY_PATH` is the first thing that is checked during linking, so nothing else
from the outside can be linked by mistake.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,7 +13,7 @@ https://github.com/scalyr/scalyr-agent-2/blob/master/CHANGELOG.md.
 
 * This release significantly changes the Agent's `deb` and `rpm` packages. Agent now is shipped with its own, independent 
   version of the Python interpreter and does not rely on system Python. The upgrade process to the new packages has to be 
-  seamless for the majority of cases, but you use custom monitors that require additional libraries, you will need to
+  seamless for the majority of cases, but if you use custom monitors that require additional libraries, you will need to
   specify those requirements in new agent's configuration.
   
   After upgrading to a new version, add required libraries to the file `/opt/scalyr-agent-2-dependencies/etc/additional-requirements.txt` 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -9,6 +9,16 @@ incompatible changes have been made in that release.
 For a list of all the changes in a particular release, please refer to the changelog file -
 https://github.com/scalyr/scalyr-agent-2/blob/master/CHANGELOG.md.
 
+## 2.2.1 "TBD"
+
+* This release significantly changes the Agent's `deb` and `rpm` packages. Agent now is shipped with its own, independent 
+  version of the Python interpreter and does not rely on system Python. The upgrade process to the new packages has to be 
+  seamless for the majority of cases, but you use custom monitors that require additional libraries, you will need to
+  specify those requirements in new agent's configuration.
+  
+  After upgrading to a new version, add required libraries to the file `/opt/scalyr-agent-2-dependencies/etc/additional-requirements.txt` 
+  and run command `/opt/scalyr-agent-2-dependencies/bin/agent-libs-config initialize` to re-initialize requirements.
+
 ## 2.1.38 "Zaotune" - Dec 1, 2022
 
 * This release changes the Kubernetes monitor leader election algorithm to use the pods in the owning ReplicaSet or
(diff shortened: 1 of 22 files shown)


agent_build_refactored/managed_packages/scalyr_agent_python3/internal/agent-python3-config.py
```diff
@@ -0,0 +1,223 @@
+# Copyright 2014-2023 Scalyr Inc.
```
ArthurKamalov, line 1: This is not a new file, it's basically just a copy of the previous version of the  `agent_build_refactored/managed_packages/scalyr_agent_python3/agent-python3-config` file. The problem is that this python script can not be executed directly anymore, and it has to be executed from the "wrapper" shell script.
But since it is a "visible" executable and can be used by a user, the wrapper has to have the same name, creating this huge diff.

agent_build_refactored/managed_packages/scalyr_agent_python3/agent-python3-config
```diff
@@ -13,182 +13,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""
```
ArthurKamalov, line 16: The removed content is moved to the `agent_build_refactored/managed_packages/scalyr_agent_python3/internal/agent-python3-config.py` file.

This is now have to be a shell script, not a Python script, but the name has to be the same.

agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -421,6 +423,8 @@ def _build_agent_package(
                 "--deb-no-default-config-files",
                 "--no-deb-auto-config-files",
                 "--config-files", "/etc/scalyr-agent-2/agent.json",
+                "--config-files", "/etc/scalyr-agent-2/agent.d",
```
ArthurKamalov, line 426: I guess the `agent.d` also has to be a config, but it isn't even in the current versions of the agent.

Kami (approved):
LGTM, thanks for addressing that 👍 

I assume you will open separate PR for including docker dependency out of the box?

RELEASE_NOTES.md
```diff
@@ -9,6 +9,16 @@ incompatible changes have been made in that release.
...
 https://github.com/scalyr/scalyr-agent-2/blob/master/CHANGELOG.md.
 
+## 2.2.1 "TBD"
+
+* This release significantly changes the Agent's `deb` and `rpm` packages. Agent now is shipped with its own, independent 
+  version of the Python interpreter and does not rely on system Python. The upgrade process to the new packages has to be 
+  seamless for the majority of cases, but if you use custom monitors that require additional libraries, you will need to
+  specify those requirements in new agent's configuration.
+  
+  After upgrading to a new version, add required libraries to the file `/opt/scalyr-agent-2-dependencies/etc/additional-requirements.txt`
```
Kami, line 19: We probably also need some docs on dataset.com/help/ on that. So we should probably also start working on that (open ticket, etc.) and then probably also involve @peteryork in the future.

ArthurKamalov: Yeah, I've already tried to find the most appropriate place for it but not sure where we can put that. I guess the closes thing has to be this one https://app.scalyr.com/help/scalyr-agent#plugins

ArthurKamalov:
> I assume you will open separate PR for including docker dependency out of the box?

https://github.com/scalyr/scalyr-agent-2/commit/0762b10ee476d33260df050999e330fabf794e48#diff-ac35fa4a779402a8735f63358c098a5e5b5ee12e67b3d7e3ca22a7e6b0921c2c

I believe this has to be already added.

ArthurKamalov:
I added testing for the process of installation of the additional libraries.  Also added more testing of what version of openssl agent uses. Before that we only tested openssl version from the result of installation, but now we also look into agent's logs and make sure that the correct openssl version is picket at the runtime.

https://github.com/scalyr/scalyr-agent-2/pull/1072/commits/e49c8d2e5687f758c71f71d744acdd23ef57b5f3

RELEASE_NOTES.md
ArthurKamalov: PS: But, I guess, since we kinda claim that those tested distros are officially supported, maybe we should take an effort and do this for all of them. WDYT?

ArthurKamalov: Sorry, this is a wrong thread.

PR #1074, opened 2023-01-27
Add package upgrade test and other

ArthurKamalov:
This PR adds tests that exercise upgrade from the stable(released) version of the agent. This is achieved by downloading stable version of the package from the prod repo and putting it to a "temporary" test repository, which we create during testing, alongside with the current package. 

It also reorganises GHA CI/CD workflows that build new packages. Now new Linux packages are built in a separate "reusable" workflow and it is called by the main workflow.  Eventually we should try to achieve the state, where all types of artifacts are build in their own reusable workflows but those workflows, in turn, are called from the single "main" workflow.


agent_build_refactored/managed_packages/scalyr_agent_python3/internal/agent-python3-config.py
```diff
@@ -175,19 +174,7 @@ def get_current_openssl_version(version_type: str) -> Tuple[bool, str]:
     if result.returncode != 0:
         return False, "Not found"
 
-    version_output = result.stdout.decode()
-
-    if version_type == OPENSSL_1_1_1_VERSION:
```
ArthurKamalov, line 180: I removed this check, so it's now completely up to Python to decide whether the available OpenSSL is valid or not.
Trying to "guess" the output of the openssl version may be a bad idea, for example on  other "unsupported" distributions, which may have valid OpenSSL, but, for some reason have different version format. In this case this check will be just an obstacle which won't let to install the agent on an absolute valid distro.

tests/end_to_end_tests/managed_packages_tests/conftest.py
```diff
@@ -171,37 +172,55 @@ def build(
         )
 
         if package_type == "deb":
-            # Create deb repository using 'reprepro'.
```
ArthurKamalov, line 174: We had to switch to `aptly` from `reprepro` because the latter does not support multiple version of the same package, which we need now in order to perform the upgrade.

VERSION
```diff
@@ -1 +1 @@
-2.1.40
+2.2.1
```
ArthurKamalov, line 1: For now I just bumped the minor version, but need to decide what version we should use next.

agent_build_refactored/managed_packages/steps/build_python.sh
```diff
@@ -53,13 +53,12 @@ pushd build
 	--with-readline=edit \
 	--prefix="${INSTALL_PREFIX}" \
 	--exec-prefix="${INSTALL_PREFIX}" \
-	--with-ensurepip=upgrade
```
Kami, line 56: Just curious why that change has been made.

tests/end_to_end_tests/run_in_remote_machine/__init__.py
```diff
@@ -16,156 +16,194 @@
...
 import subprocess
-from typing import List, Dict
+from typing import List, Dict, Union
 
 from agent_build_refactored.tools.constants import Architecture
 from agent_build_refactored.tools.run_in_ec2.constants import EC2DistroImage
 
 logger = logging.getLogger(__name__)
 
+_EMBEDDED_OPENSSL_VERSION_NUMBER = 269488447
```
Kami, line 30: Should we add comment in the place where we download + build OpenSSL that this constant in the test needs to be updated if / when OpenSSL version is upgraded there?

tests/end_to_end_tests/managed_packages_tests/test_packages.py
```diff
@@ -237,18 +239,15 @@ def test_packages(
...
+    if isinstance(target_distro.expected_openssl, list):
+        expected_openssl_versions = target_distro.expected_openssl
+    else:
+        expected_openssl_versions = [target_distro.expected_openssl]
 
-    # On Ubuntu 14.04 there's no "modern" version OpenSSl, but it still has to use 1.1.1 that is shipped with
-    # the package.
-    elif distro_name == "ubuntu1404":
-        assert "OpenSSL version=OpenSSL 1.1.1" in starting_agent_message
+    assert openssl_version in expected_openssl_versions
```
Kami, line 250: So we do a partial "contains" check here.

I assume this should be sufficient as long as the input string is not some loose partial sub-string which could still result in this assertion passing, but  possibly not doing what we want (e.g. if someone accidentally used just ``OpenSSL version`` or similar for the ``expected_openssl`` parameter).

tests/end_to_end_tests/run_in_remote_machine/__init__.py
```diff
@@ -16,156 +16,194 @@
...
+
+
+@dataclasses.dataclass
+class TargetDistro:
+    name: str
+    docker_image: str
+    ec2_images: Dict[Architecture, EC2DistroImage]
+    # Expected version (in int representation) of the OpenSSL library that has to be picked by the agent.
+    expected_openssl: Union[int, List[int]] = 3
```
Kami, line 39: Would be good if you could include an example here for the expected format.

tests/end_to_end_tests/managed_packages_tests/test_packages.py
ArthurKamalov: This has to use integer version value from [ssl module ](https://docs.python.org/3/library/ssl.html#ssl.OPENSSL_VERSION_NUMBER).  It have to get expected values from the agent starting message from the `agent.log` file. This message now has new field [version_number](https://github.com/scalyr/scalyr-agent-2/blob/3daaede149eac0ec0171e68d96d0c3980ef66dbb/scalyr_agent/util.py#L2171) which also shows numeric representation of the OpenSSL. For more info search for `OPENSSL_VERSION_NUMBER` in https://www.openssl.org/docs/man3.0/man3/OPENSSL_VERSION_NUMBER.html

ArthurKamalov: PS: This "contains" check does not a partial comparison. It does 1 to 1 match of integer versions of OpenSSL.

Kami (approved):

PR #1078, opened 2023-01-31
DSET-1628: message_log_template & attributes not applied without checkpoint

jmakar-scalyr:
- Pass the global config from SyslogMonitor (via SyslogServer) to SyslogHandler
- Then match current log path with potential log config from the global config
  - This is needed to know which parser & attributes to apply when creating the logger

### DIFF
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -25,6 +25,7 @@ __author__ = "scalyr@sentinelone.com"
 import errno
 from fnmatch import fnmatch
 import glob
+import json
 import logging
 import logging.handlers
 import os
@@ -1619,6 +1620,21 @@ class SyslogHandler(object):
                             max_bytes=self.__max_log_size,
                             backup_count=self.__max_log_rotations,
                             flush_delay=self.__flush_delay,
+                        )
+
+                        global_log.log(
+                            scalyr_logging.DEBUG_LEVEL_1,
+                            "__syslog_attributes=%s log_config_attribs=%s extra=%s"
+                            % (
+                                json.dumps(
+                                    self.__syslog_attributes.to_dict(),
+                                    separators=(",", ":"),
+                                ),
+                                json.dumps(
+                                    log_config_attribs.to_dict(), separators=(",", ":")
+                                ),
+                                json.dumps(extra, separators=(",", ":")),
+                            ),
                         )
 
                         attribs = self.__syslog_attributes.copy()
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -1204,6 +1204,7 @@ class SyslogHandler(object):
         logger,
         line_reporter,
         config,
+        global_config,
         server_host,
         log_path,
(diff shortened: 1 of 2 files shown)


Kami:
Thanks for fixing this.

So far LGTM, will have a look again when tests are ready.

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1591,33 +1594,51 @@ def __handle_syslog_logs(self, data, extra):
...
-                        filename=path,
-                        max_bytes=self.__max_log_size,
-                        backup_count=self.__max_log_rotations,
-                        flush_delay=self.__flush_delay,
-                    )
+                    # Match the new log path with a global log config (generally globbed).
+                    # Note that in log configs, .path is required but .attributes may be empty.
+                    log_config_attribs = None
+                    for log_config in self.__global_log_configs:
+                        if fnmatch(path, log_config["path"]):
```
Kami, line 1601: IIRC, we will only end up in this code path in log file for a specific path doesn't exist yet so it shouldn't add much overhead since it won't be called for every single syslog message / log line, right?

Kami: Also, We should probably try to add more debug logs to this function to make troubleshooting easier.

As long as we add it to the code paths which only run for a new log file (and not for every single syslog message), it shouldn't add much overhead and should be fine to do it.

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1591,33 +1594,51 @@ def __handle_syslog_logs(self, data, extra):
...
+                            max_bytes=self.__max_log_size,
+                            backup_count=self.__max_log_rotations,
+                            flush_delay=self.__flush_delay,
+                        )
 
-                    attribs = self.__syslog_attributes.copy()
-                    attribs.update(
-                        {k.lower(): v for k, v in extra.items() if v is not None}
-                    )
+                        attribs = self.__syslog_attributes.copy()
```
Kami, line 1614: Does this dictionary only contain simple values? In case it doesn't, we may need to use ``copy.deepcopy()`` to avoid surprises.

Kami: For testing reasons, it's probably also a good idea to refactor some of the functionality here in separate standalone utility functions.

scalyr_agent/builtin_monitors/syslog_monitor.py
jmakar-scalyr: Yes that's right.  Sure, added a couple debug logs.

scalyr_agent/builtin_monitors/syslog_monitor.py
jmakar-scalyr: Yup they're scalar values

Kami: I think it could also aid with troubleshooting if add some debug log statements here as well (actual log config object, etc.).

jmakar-scalyr: Sure, will do

Kami (approved):

PR #1080, opened 2023-02-02
Fix syslog micro benchmark, also run micro benchmarks on non-master branches

Kami:
It looks like #1078 inadvertedly broke syslog micro benchmark.

This PR fixes it.

In the past we intentionally only ran benchmark jobs for master branch and not for every PR branch since that would be too slow, but maybe it make sense to also run microbenchmarks on all the PRs now that we have trimmed down the Python version matrix, but only submit results to CodeSpeed on master branch.

### DIFF
--- a/.github/workflows/codespeed-agent-benchmarks.yaml
+++ b/.github/workflows/codespeed-agent-benchmarks.yaml
@@ -150,7 +150,8 @@ jobs:
           CAPTURE_AGENT_STATUS_METRICS: ${{ inputs.capture_agent_status_metrics }}
           DRY_RUN: ${{ inputs.dry_run }}
           PYTHONPATH: .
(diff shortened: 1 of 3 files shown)


Kami:
Per PR description, I decided to update micro benchmarks workflow so it also runs for all the PR branches (but we still only submit results to CodeSpeed on master branch).

This should allow us to catch such failures earlier (before merging to master).

Kami:
I will only be able to confirm if all the GHA workflow related logic works correctly once this has been merged into master.

ArthurKamalov (approved):

Kami:
Confirmed that it works correctly:

1) On PR runs, we run the checks, but we don't submit data to CodeSpeed - https://github.com/scalyr/scalyr-agent-2/actions/runs/4073309769
2) On master branch run, results are also submitted to CodeSpeed - https://github.com/scalyr/scalyr-agent-2/actions/runs/4073778099

PR #1081, opened 2023-02-02
[DSET-1648] Fix "skip duplicate" GHA workflow conditionals, move re-usable workflows to a sub-directory, use github.ref_name context variable

Kami:
This pull request fixes "skip duplicate" workflow conditionals to ensure we cancel / skip redundant builds for pull request / feature branches (as as such, spare some CPU cycles and speed up the CI).

Code incorrectly checked for ``github.base_ref`` which contains name of the target / destination branch (which is almost always ``master`` for PRs). This means that jobs would never be skipped for PR branches.

In reality, we want to check for the current branch name (``github.ref``) to make sure we always run jobs on master branch (but skip redundant jobs on PR branches).

References:

* https://docs.github.com/en/actions/learn-github-actions/contexts

> The base_ref or target branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.

> The head_ref or source branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.


Kami:
EDIT: Looks like gha now also supports ``github.ref_name`` value so I will update workflows to use that instead since it's more straight forward and less confusing - https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables.

> The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub. For example, feature-branch-1.

Kami:
Per discussion with @ArthurKamalov, I will also move all the re-usable workflows which are called from other "top level" workflows (and are never triggered directly) to a sub-directory to make it easier to see what is going on.

Kami:
Will verify that everything works correctly once the PR has been merged into master.

ArthurKamalov (approved):
LGTM

PR #1085, opened 2023-02-06
DSET-1409: Add tests for syslog monitor changes (specificially message_log_template option)

jmakar-scalyr:
DSET-1409: Add tests for syslog monitor changes (specificially message_log_template option)

- SyslogTemplateTest class supports testing behavior in syslog_monitor.SyslogHandler.__parse_syslog (via mocks around use of LogWatcher)
  - Also took care not to write any log files (unclear where and how the tests could be run), ie AutoFlushingRotatingFileMock, AutoFlushingRotatingFileHandlerMock
- SyslogParserTest class supports testing syslog_monitor.SyslogHandler._parse_syslog (essentially a wrapper for syslogmp.parse)
  - Skipped tests highlight syslogmp-based parsing, to be discussed elsewhere


tests/unit/syslog_template_test.py
```diff
@@ -0,0 +1,1037 @@
...
+                    "destport": self.__class__.tcp_port + 1,
+                    "hostname": "localhost",
+                    "appname": "demo",
+                },
+                "parser": "second-parser",
+            },
+        )
+
+    # On Debian based distros `localhost 127.0.1.1` is added to /etc/hosts
+    @skip("Requires aliasing the loopback interface")
```
ArthurKamalov, line 405: Should we do this skip conditional to make it work on Linux?

ArthurKamalov (approved):
LGTM, Just have one comment about skipping one of the tests cases.

tests/unit/syslog_template_test.py
```diff
@@ -0,0 +1,1037 @@
+# Copyright 2023 Scalyr Inc.
```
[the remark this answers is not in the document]
jmakar-scalyr: Oh good idea, I'll move all the syslog monitor test files there

tests/unit/syslog_template_test.py
jmakar-scalyr: We could, I'm not sure if non-Debian distros do the same thing or not.

I think I'll just leave it as plain skip for now & if there are problems with that feature in the future will work something out to ensure it's tested.

PR #1089, opened 2023-02-09
[DTIN-2240] Remove dependency packages

ArthurKamalov:
This PR returns back agent linux packages that do not have any dependency packages.  

After approval of the team, I removed all dependency packages so now there's only one agent package without any dependencies. 

The majority of changes are concentrated in few files and other changes are just cosmetic, such as directory renaming and etc.


agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -327,13 +328,232 @@ def get_agent_libs_build_command_args(cls) -> List[str]:
             # fmt: on
         ]
 
-    def _build_agent_package(
-        self,
-        agent_libs_package_version: str,
-    ) -> str:
+    def _prepare_package_python_and_libraries_files(self, package_root: pl.Path):
```
ArthurKamalov, line 331: Previously we created files for dependency packages in a separate steps by using those scripts: 
`agent_build_refactored/managed_packages/steps/build_agent_libs_package_root.sh`, `agent_build_refactored/managed_packages/steps/build_python_package_root.sh`. Now we do not create separate dependency packages, but still need to prepare those files, so everything just moved from those script to this function.

.github/workflows/agent-build-new.yml
```diff
@@ -62,3 +76,30 @@ jobs:
...
+
+
+      - name: Prepare artifacts
+        run: |
+          mkdir -p /tmp/result_artifacts
+          cp -a /tmp/all_artifacts/linux-packages-*/. /tmp/result_artifacts
+          cp -a /tmp/all_artifacts/tarball-noarch/. /tmp/result_artifacts
+
+      - name: Save result artifact
+        uses: actions/upload-artifact@v3
```
jmakar-scalyr, line 101: This is new to me, neat!  I'm guessing that artifacts like this are only accessible via actions/download-artifact ?

One thing that might be handy (to be done later) is an brief README that describes the overall build process

RELEASE_NOTES.md
```diff
@@ -16,8 +16,8 @@ https://github.com/scalyr/scalyr-agent-2/blob/master/CHANGELOG.md.
   seamless for the majority of cases, but if you use custom monitors that require additional libraries, you will need to
   specify those requirements in new agent's configuration.
   
-  After upgrading to a new version, add required libraries to the file `/opt/scalyr-agent-2-dependencies/etc/additional-requirements.txt` 
-  and run command `/opt/scalyr-agent-2-dependencies/bin/agent-libs-config initialize` to re-initialize requirements.
+  After upgrading to a new version, add required libraries to the file `/opt/scalyr-agent-2/etc/additional-requirements.txt`
```
jmakar-scalyr, line 19: These are developer notes right?  Maybe it would be clearer to say "When creating a new version"

agent_build_refactored/managed_packages/files/bin/python3
```diff
@@ -26,8 +26,11 @@
...
-DEPENDENCIES_LIBS="${DEPENDENCIES_ROOT}/lib"
+PACKAGE_OPT_DIR="/opt/scalyr-agent-2"
+PACKAGE_LIBS_DIR="${PACKAGE_OPT_DIR}/lib"
+PYTHON_DIR="${PACKAGE_OPT_DIR}/python3"
+PYTHON_LIB_DIR="${PYTHON_DIR}/lib"
 
-export LD_LIBRARY_PATH="${DEPENDENCIES_LIBS}:${DEPENDENCIES_LIBS}/openssl/current:${LD_LIBRARY_PATH}"
-exec "${DEPENDENCIES_ROOT}/bin/python3-original" "$@"
+
+export LD_LIBRARY_PATH="${PYTHON_LIB_DIR}:${PACKAGE_LIBS_DIR}/openssl/current:${LD_LIBRARY_PATH}"
```
jmakar-scalyr, line 35: Why is this needed?  I understand it's using a bundled version of openssl but couldn't the Scalyr agent package have openssl as a dependency to avoid this?

agent_build_refactored/managed_packages/install-scriptlets/preuninstall.sh
```diff
@@ -43,4 +43,10 @@ fi
 # Always remove the .pyc files and __pycache__ directories
 find /usr/share/scalyr-agent-2 -type f -name '*.py[co]' -delete -o -type d -name __pycache__ -exec rm -r {} \;
 
+# Remove dynamically generated venv.
+rm -r "/var/opt/scalyr-agent-2/venv"
+
+# Collect garbage after Python.
+find /opt/scalyr-agent-2 -type f -name '*.py[co]' -delete -o -type d -name __pycache__ -exec rm -r {} \;
```
jmakar-scalyr, line 50: Will the `*.py[co]` delete be necessary with Python 3?  (I though Python 3 puts everything in `__pycache__`)

agent_build_refactored/managed_packages/managed_packages_builders.py
```diff
@@ -60,70 +62,62 @@
...
+            library if it is presented on system. First it tries to find OpenSSL version 3+.
         - 2: If first step is not successful and system does not have appropriate OpenSSL 3, then
             we replace previous  '_ssl.cpython*.so' and '_hashlib.cpython.so' C bindings with the same bindings, but
             that are compiled and linked against the OpenSSL 1.1.1+, and will look for OpenSSL 1.1.1+. Even though the
             _ssh and _hashlib modules may resolve multiple variants of the OpenSSL libraries, they are able to resolve
             them only for the specific major version they were compiled for. So we have to have multiple variants of the
             ssl C bindings that are compiled and linked against both OpenSSL 1.1.1+ and 3.0+.
 
-        -3: If OpenSSL 1.1.1+ is also not presented in a system, then we fallback to the 'embedded' OpenSSL library that
+        - 3: If OpenSSL 1.1.1+ is also not presented in a system, then we fallback to the 'embedded' OpenSSL library that
```
jmakar-scalyr, line 87: So a version of OpenSSL is embedded is support older systems that don't have a more recent version installed?

One thing that's a little bit of a concern is that OpenSSL generally has a frequent release cycle because of discovered vulnerabilities & embedding it would mean the release cycle for the Scalyr agent should try to match it.
But then again the agent doesn't host services so maybe it's not much of a concern.

agent_build_refactored/managed_packages/steps/build_dev_requirements.sh
```diff
@@ -36,12 +36,24 @@ ldconfig
 cd ~
 export PATH="/usr/local/bin:/root/.cargo/bin:${PATH}"
 export PKG_CONFIG_PATH="/usr/local/lib64/pkgconfig:/usr/local/lib/pkgconfig:${PKG_CONFIG_PATH}"
-curl --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain "${RUST_VERSION}"
+
```
jmakar-scalyr, line 39: Why switch from rustup?  (Out of curiosity)

jmakar-scalyr (approved):
Looks good, just had a few random questions (can answer later if busy)

RELEASE_NOTES.md
ArthurKamalov: That's serves more like an additional reference where user or developer can find more info about some particular change if it's big enough to be in the CHANGELOG itself. That's because, (as I can see it), the CHANGELOG has to follow is special format where each "change" must not have a line breaks, so a really long text would be ugly and very inconvenient to edit, so it's just better to move it. to another file.

And no, that's more a user note, about how to install user's custom agent dependencies after the upgrade to a new version.

agent_build_refactored/managed_packages/steps/prepare_python_environment.sh
```diff
@@ -0,0 +1,43 @@
...
+cp -a "${BUILD_OPENSSL_1_1_1}/." /
+cp -a "${BUILD_DEV_REQUIREMENTS}/root/." /
+
+echo "${PYTHON_INSTALL_PREFIX}/lib" >> /etc/ld.so.conf.d/python3.conf
+ldconfig
+
+ln -s "${PYTHON_INSTALL_PREFIX}/bin/python3" /usr/bin/python3
+
+# shellcheck disable=SC1090
+source ~/.bashrc
```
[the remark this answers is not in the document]
ArthurKamalov: Sometimes there may be stuff that has to be activated first in order to make it work in the current bash script. For example, in order to prepare the environment to build our embedded Python interpreter ([link](https://github.com/scalyr/scalyr-agent-2/blob/3a02ca9d48f6678a6045479164679ac1f29fceba/agent_build_refactored/managed_packages/steps/install_build_environment/install_gcc_centos_7.sh#L30)) , we install a special package `devtoolset-9`  that provides more recent version of the `gcc`, but this can be used only after "enabling" it.  Unfortunately, just putting this into the `.bashrc`  is not enough in our case, because our "step" scripts are executed in docker by using `docker run ... bash ...` command, and bash, in this case, does not automatically invoke the `.bashrc`.

agent_build_refactored/managed_packages/steps/build_dev_requirements.sh
ArthurKamalov: Newer versions of Rust, including rustup, have already dropped the support of the glibc < "2.17", and we still support glibc "2.12", so newest rustup just can not start :(

That may sound as a good signal for us to also drop so old glibc versions, but I think this release is already has enough changes and it would be better to do it later.

agent_build_refactored/managed_packages/managed_packages_builders.py
ArthurKamalov: > So a version of OpenSSL is embedded is support older systems that don't have a more recent version installed?

Correct.

Yeah, that's why we try to find the system's OpenSSL first (because it has be maintained by the distribution in a more timely manner) and only fallback to the embedded one when there are no other options. In this case yeah, we'd better track the OpenSSL update as frequent as possible. I also have a draft for a script that has to track version of OpenSSL and if there's a new version, it creates PR with update.

agent_build_refactored/managed_packages/install-scriptlets/preuninstall.sh
ArthurKamalov: Ah, I just reused existing code snipped. Good to know, Will remove that. Thanks.

agent_build_refactored/managed_packages/files/bin/python3
ArthurKamalov: > but couldn't the Scalyr agent package have openssl as a dependency to avoid this?

I think it just won'y help us so much. If we specify the dependency just  like `libssl` without any versioning, then we still have to verify if the version provided by the package manager is appropriate so we can fallback to the embedded
openssl if not.

If we specify the dependency `libssl` with version, aka `libssl=>1.1.1`, then we'll just leave behind everyone who can not satisfy that requirement, because now package manager won't simple allow to install this package, so they don't even have a chance to fall back to the embedded openssl.

.github/workflows/agent-build-new.yml
ArthurKamalov: I added a comment to clarify what is going on here. In short, we just pack all build artifacts into one single artifact, so when doing a release, our release pipieline does not have to download all artifacts one by one and get all of them at once.

ArthurKamalov:
I managed how to simplify the logic of initializing of the OpenSSL when I was trying to answer on your comment, thanks for that :smile: I reflected that in the updated docstring of the package builder https://github.com/scalyr/scalyr-agent-2/blob/72142aa4fd2dc5bad8f410f793f7a5b187d0edf2/agent_build_refactored/managed_packages/managed_packages_builders.py#L73-L99

Besides that, everything else has to be also addressed.

PR #1098, opened 2023-03-01
Fix logging bug in the Docker monitor

Kami:
## Description

This PR ensures we have a rate limit in place for the ``No timestamp found on line:`` log message in the Docker monitor.

## Background, Context

Agent on one of my servers some how ended up in a bad state where ``No timestamp found on line:`` log message would get logged very often. This would cause the agent to exhaust all the CPU cycles just by logging this message.

I don't know the exact root cause for that edge case (it could be something related to Docker daemon upgrade or similar), but it doesn't matter that much since we need a rate limit (aka this fix) regardless.

Here is the message which appeared in the logs:

> 21:33:12.017​nuc.tomaz.me​/var/log/scalyr-agent-2/agent.log​2023-02-25 20:33:12.017Z ERROR [monitor:docker_monitor] [scalyr_agent.builtin_monitors.docker_monitor:1637] No timestamp found on line: 'Error grabbing logs: invalid character '\x00' looking for beginning of value

This message does potentially indicate another issue which we should handle and log differently, but that can be handled in a future PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,12 @@ Scalyr Agent 2 Changes By Release
 Packaged by Arthur Kamalov <arthurk@sentinelone.com> on Jan 19, 2023 00:00 -0800
 --->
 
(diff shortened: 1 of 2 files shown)


ArthurKamalov (approved):
LGTM

PR #1100, opened 2023-03-02
Fix post-uninstall warnings

ArthurKamalov:
This PR, to reduce noise, removes some warnings that are produced during the package uninstallation. 

Other changes: 
* It also disables check of the new version of the pip, so the update suggestion message won't shown every time.

* Also corrected venv removal during the `pre-uninstall` script execution.

* Retries added to the package status tests to reduce overall flakiness.

### DIFF
--- a/agent_build_refactored/managed_packages/files/bin/agent-libs-config
+++ b/agent_build_refactored/managed_packages/files/bin/agent-libs-config
@@ -52,6 +52,7 @@ def initialize():
             "pip",
             "install",
             "--quiet",
+            # TODO: Do a periodical check of the available new version of the pip in the CI/CD
             "--disable-pip-version-check",
             # Core agent requirements are already installed in this "pre-created" venv, but
             # we still add their requirements file to be sure that pip will throw an error if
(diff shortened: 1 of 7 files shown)


agent_build_refactored/managed_packages/files/bin/agent-libs-config
```diff
@@ -52,6 +52,7 @@ def initialize():
             "pip",
             "install",
             "--quiet",
+            "--disable-pip-version-check",
```
tomaz-s1, line 55: 👍 

Probably not a bad idea for us to have some check in CI/CD which periodically reminds us to upgrade pip in the venv or similar.

tests/end_to_end_tests/tools.py
```diff
@@ -89,6 +89,28 @@ def start(self, no_fork: bool = False, env=None):
...
+        time.sleep(0.2)
+
+        attempts = 0
+
+        while True:
+            try:
+                self.get_status_json()
+            except subprocess.CalledProcessError as e:
+                if attempts >= 10:
+                    raise e
```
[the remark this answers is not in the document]
ArthurKamalov: Sure, added more log messages, but also increased initial wait so it has more time before it throws its first warning.

tomaz-s1 (approved):
LGTM 👍

agent_build_refactored/managed_packages/files/bin/agent-libs-config
ArthurKamalov: yeah, especially with the fact that we cache our builds.

PR #1106, opened 2023-03-21
DSET-2112 Support SSL connections to PostgreSQL

dlobue:
Support SSL connections to PostgreSQL. Enable by setting `use_ssl` to `True` in the `postgres_monitor` config.

DSET-2112

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Packaged by Arthur Kamalov <arthurk@sentinelone.com> on Jan 19, 2023 00:00 -0800
 
 Other:
 * Linux `deb` and `rpm` packages are now shipped with its own Python interpreter, so agent does not have to rely on system Python interpreter anymore. Please refer to the RELEASE_NOTES document, for more information.
+* The PostgreSQL monitor now supports connecting to servers over SSL. Configure the monitor with the setting `use_ssl` set to `True` to enable this.
 
 Bug fixes:
 * Fix bug in the Docker monitor which would, under some edge cases, cause a specific log messages to be logged very often and as such, exhausting the CPU cycles available to the Docker monitor.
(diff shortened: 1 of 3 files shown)


tomaz-s1:
How are the certificate and hostname checks handled? Aka against which CA bundle (if any) do we verify server certificate? And is the hostname check performed (to verify that hostname / ip inside the cert matches the hostname / ip we are connecting to)?

It would also be good to document this in the code and inside the docs (to avoid MITM and similar issues, and also give info on how to, with big warning at your own risk, turn ca validation off).

dlobue:
> How are the certificate and hostname checks handled? Aka against which CA bundle (if any) do we verify server certificate? And is the hostname check performed (to verify that hostname / ip inside the cert matches the hostname / ip we are connecting to)?
> 
> It would also be good to document this in the code and inside the docs (to avoid MITM and similar issues, and also give info on how to, with big warning at your own risk, turn ca validation off).

We actually don't have any options in this regard as we are using such an old version of pg8000 (the library was bundled into the agent repo 7 years ago). We can turn on SSL. That's it.

In the most recent version of pg8000 supports customization of the SSL context.

The default for python SSL contexts (as I'm sure you know) is `verify_mode` is set to `CERT_REQUIRED`, `check_hostname` is set to True , and the certificate bundle used is the default system bundle.

So anything with an SSL cert that's been signed by one of the major SSL providers will work, provided it matches the hostname the user is connecting to. Self-signed certs won't work, but I'm not particularly concerned about that use-case ATM.

This seems like a sane and secure default to me. (Please educate me if I'm wrong!)

tomaz-s1:
@dlobue In this case, let's just add a note on current behavior on cert and hostname validation to docs, besides that, LGTM.

And we can update the library and the code and frozen binaries are only way to release the agent (so we don't need to deal with supporting Python 2.7 anymore).

dlobue:
@tomaz-s1 I updated the `use_ssl` option description to include the note on verification behavior as you asked, and then regenerated the help docs for the monitor.

ArthurKamalov:
LGTM

tomaz-s1 (approved):

PR #1107, opened 2023-03-22
Add AIO package version.

ArthurKamalov:


tomaz-s1:
Does this (substantially) slow down the build? If yes, is there a way for us to speed it up - e.g. by copying the existing artifact and renaming it, or is this not possible since we would also need to update embedded metadata (+ as such package signature, etc.)?

ArthurKamalov:
> Does this (substantially) slow down the build? If yes, is there a way for us to speed it up - e.g. by copying the existing artifact and renaming it, or is this not possible since we would also need to update the embedded metadata (+ as such package signature, etc.)?

The `scalyr-agent-2` packages are not that same as `scalyr-agent-2-aio`.  They are basically the same packages that we were building in CircleCI, but migrated to GHA.  They are built differently, without any "embedded" Python or OpenSSL, and still depend on system Python and OpenSSL. 

That shouldn't slow down the build since those packages are built in parallel job and they should be faster to build, since they are just 2-3 MB.

ArthurKamalov:
PS. Forgot to mention, that now our "convenience script" also has new option "--use-aio-package", that installs aio package instead of "traditional/" ones.

tomaz-s1:
> The scalyr-agent-2 packages are not that same as scalyr-agent-2-aio. They are basically the same packages that we were building in CircleCI, but migrated to GHA. They are built differently, without any "embedded" Python or OpenSSL, and still depend on system Python and OpenSSL.

Thanks for clarifying.

But the plan is for them to eventually be, right? After we reach some to be decided threshold / milestone for AIO packages...

ArthurKamalov:
> > The scalyr-agent-2 packages are not that same as scalyr-agent-2-aio. They are basically the same packages that we were building in CircleCI, but migrated to GHA. They are built differently, without any "embedded" Python or OpenSSL, and still depend on system Python and OpenSSL.
> 
> Thanks for clarifying.
> 
> But the plan is for them to eventually be, right? After we reach some to be decided threshold / milestone for AIO packages...

Yes, I think, so.  I also mentioned that in this docstring https://github.com/scalyr/scalyr-agent-2/blob/f98d02888551388693a05aa6d1abf430b89dbd4b/agent_build_refactored/managed_packages/managed_packages_builders.py#L70-L71.  We will be able to do with it whatever we want in future, e.g. 
- make it the same as `scalyr-agent-2-aio` for a "seamless" migration of the consumers from `scalyr-agent-2`.
- just stop publishing it, and keep only `scalyr-agent-2-aio`
- even keep publishing it as Python-dependent package (if that's really needed for some reason).

tomaz-s1 (approved):

PR #1125, opened 2023-04-25
DSET-3651 Add permissions for argo-rollout to scalyr-agent ClusterRole

dlobue:
Add permissions to query argo rollouts to scalyr-agent ClusterRole.

DSET-3651


tomaz-s1 (approved):
LGTM.

Let's please just add a changelog entry.

end to end test would also be nice, but I know this would be a bit involved to set up.

PR #1146, opened 2023-06-21
Try running Python 2.7 unit tests in a container to work around removal

tomaz-s1:
This PR is attempt to fix unit and smoke tests to get them running under Python 2.7 again.

In fact I can't get it to work in an hour or two, I will likely just remove those tests - we have already spend tons of time on it in the past and it's been a massive head ache trying to get tests to still work with Python 2.7 which has been EOL for 3+ years and removed from most platforms and dependencies we use.

## Background, Context

Github recently removed support for Python 2.7 from their cloud runners and setup-python action (https://github.com/actions/setup-python/issues/672#issuecomment-1589120020) so those tests started failing.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -142,61 +142,6 @@ jobs:
           steps: ${{ toJson(steps) }}
           channel: '#eng-dataset-cloud-tech'
 
-  # NOTE: Github removed 2.7 support from their runners so we need to use container workaround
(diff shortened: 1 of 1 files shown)


tomaz-s1:
It looks like the easy approach is sadly not working.

During to 2.7 being EOL anywhere, it's not worth for us to spend tons of more time on it and also add tons of more complexity and workarounds we will need to maintain.

So I will go ahead and remove 2.7 from the build.

PR #1147, opened 2023-06-22
Update workflow trigger to make sure Docker image build is also triggered on push to a tag

tomaz-s1:
This pull request updates Docker Image Build workflow trigger to make sure it's also triggered on push to a tag (which happens as part of the release process).


tomaz-s1:
I also added `workflow_dispatch` trigger so we can also trigger that workflow manually if / when needed.

ArthurKamalov (approved):

PR #1152, opened 2023-07-05
 Add secrets scanning workflow #96 

tomaz-s1:
This pull request adds a workflow which uses trufflehog to automatically scan files and commits for secrets on push, PR and as part of a daily cron trigger.


ArthurKamalov (approved):
LGTM

PR #1160, opened 2023-07-18
Sequence diagrams for agent start and stop

alesnovak-s1:
Sequence diagrams for agent start and stop


ArthurKamalov (commented):
Looks good to me overall.
I am not fully sure how those diagrams are created, but if they are created manually, there is a risk that this diagram may become outdated over time, due to a high chance that a developer will simply forget to reflect his changes also in diagrams. Not sure if that's possible, but it would be good to have some automated update mechanism for that. 
@tomaz-s1 WDYT?

tomaz-s1:
Yeah, I assume Arthur just means embedding mermaid diagram inside a code as a docstring (e.g. top level module docstring) and then generate markdown from that.

I think that's fine and doing anything more than that (aka programmatically generating complete diagram from the code flows) is not really feasible and worth the time investment.

PR #1168, opened 2023-07-27
Also run secrets scan workflow on the release branch

tomaz-s1:
This pull request updates secret scanner workflow to also run on the `release` branch - as part of the push, pr and cron trigger.

### DIFF
--- a/.github/workflows/secrets-scanner.yaml
+++ b/.github/workflows/secrets-scanner.yaml
@@ -3,9 +3,11 @@ on:
   push:
     branches:
       - master
+      - release
   pull_request:
     branches:
       - master
+      - release
   schedule:
     - cron: '0 4 * * *'
 
@@ -13,7 +15,8 @@ permissions:
   contents: read
 
 jobs:
-  TruffleHog:
+  trufflehog_push_pull:
+    name: TruffleHog (push, pull)
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/secrets-scanner.yaml
```diff
@@ -51,18 +54,92 @@ jobs:
...
+        if: ${{ failure() && (github.ref_name == 'master' || github.ref_name == 'release') }}
+        uses: act10ns/slack@ed1309ab9862e57e9e583e51c7889486b9a00b0f # v2.0.0
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        with:
+          status: ${{ job.status }}
+          steps: ${{ toJson(steps) }}
+          channel: '#eng-dataset-o11y'
+
+  trufflehog_cron_release_branch:
```
alesnovak-s1, line 109: trufflehog_cron_release_branch are identical, except for the ref parameter.
What about using strategy.matrix here?

alesnovak-s1 (changes requested):

.github/workflows/secrets-scanner.yaml
tomaz-s1: Thanks for the review.

That could work, but I wanted to use this workflow as-is (copy & paste) since we use it in many other places and right now we don't have public open source org level Github repository with re-usable actions and workflows where we could store and reference from workflows like that.

tomaz-s1: But maybe that's also an opportunity to start such repo with public and re-usable actions.

I would be happy to do that, but we also need to follow the official process and get all the necessary approvals for that.

alesnovak-s1 (approved):

PR #1173, opened 2023-08-09
Packages updated to fix vulnerabilities.

alesnovak-s1:
https://sentinelone.atlassian.net/browse/DTIN-2974

requests==2.31.0 - in order to get certifi==2023.7.22
docker==6.1.3 - version compatible with the new requests
pysnmp==4.4.12 - to in order to fix the pycrypto vulnerability

All effective changes:
```
2,3c2,3
< charset-normalizer==2.1.1
< docker==6.0.0
---
> charset-normalizer==3.2.0
> docker==6.1.3
4a5,6
> lz4==4.0.2
> orjson==3.8.0
10c12
< pycrypto==2.6.1
---
> pycryptodomex==3.18.0
13c15
< pysnmp==4.3.0
---
> pysnmp==4.4.12
17c19
< requests==2.28.1
---
> requests==2.31.0
21c23
< urllib3==1.26.16
---
> urllib3==2.0.4
22a25
> zstandard==0.19.0
```

### DIFF
--- a/dev-requirements-new.txt
+++ b/dev-requirements-new.txt
@@ -13,7 +13,10 @@ repoze.lru==0.7
 six==1.14.0
 
 # Required for redis monitor.
-redis==4.6.0
+redis==2.10.5; python_version == '3.6'
+redis==4.6.0; python_version >= '3.7'
+
+
 
 # Required for mysql monitor
 PyMySQL==0.9.3
--- a/dev-requirements-new.txt
+++ b/dev-requirements-new.txt
@@ -13,8 +13,7 @@ repoze.lru==0.7
 six==1.14.0
 
 # Required for redis monitor.
-redis==2.10.5; python_version == '3.6'
-redis==4.6.0; python_version >= '3.7'
+redis==2.10.5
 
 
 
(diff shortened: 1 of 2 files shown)


jmakar-scalyr (approved):
Looks good, thanks for taking care of this

tomaz-s1:
@alesnovak-s1 As you are probably aware, this will only affect dependencies inside the Docker Images, but not the dependencies we vendor with the artifacts (rpm, deb and also directory we include inside docker image) - https://github.com/scalyr/scalyr-agent-2/tree/master/scalyr_agent/third_party

In the past, we didn't want / couldn't update vendored / bundled dependencies since we still support Python 2.7 and we couldn't update them.

One thing we need to do is to verify that the actual Docker images use dependencies from this requirements file and not the one bundled inside `scalyr_agent/` directory - we need to verify this with Docker image. I forget exactly the priority since the agent does `PYTHONPATH` manipulator to ensure dependencies `scalyr_agent/third_party` are actually being used.

tomaz-s1:
My comment above can be partially ignored.

I confirm that recent Docker images don't bundle / vendor dependencies from `third_party_*` directory anymore (except tcollector) so we should be fine here.

> root@25babb35531c:/# ls -la /usr/share/scalyr-agent-2/py/scalyr_agent/third_party*
total 16
drwxr-xr-x 3 root root 4096 Aug 14 14:18 .
drwxr-xr-x 1 root root 4096 Aug 14 14:50 ..
drwxr-xr-x 4 root root 4096 Aug 14 14:18 tcollector

PR #1182, opened 2023-08-22
Try to fix Docker image tag publishing to make sure "v" prefix is not present

tomaz-s1:
This is an attempt to fix the Docker Image tag publish to make sure docker image tags don't include ``v`` prefix.

Not tested, may not work.

### DIFF
--- a/.github/workflows/reusable-agent-build-container-images.yml
+++ b/.github/workflows/reusable-agent-build-container-images.yml
@@ -859,18 +859,22 @@ jobs:
         run: |
           if [ "${{ github.ref_type }}" = "tag" ]; then
             AGENT_VERSION="$(cat VERSION | tr -d '\n')"
+            TAG="${{ github.ref_name }}"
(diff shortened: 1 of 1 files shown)


alesnovak-s1 (approved):

alesnovak-s1:
@tomaz-s1 
I've tested the changes in a separate branch with a mocked workflow.
Removing  "exit 1" failsafe lines and will be merging after all the tests pass.

tomaz-s1:
@alesnovak-s1 Great, thanks.

I assume before / after it's merged we also want to test end to end it with our ``test4scalyr`` dev / testing Docker Hub account?

alesnovak-s1:
> @alesnovak-s1 Great, thanks.
> 
> I assume before / after it's merged we also want to test end to end it with our `test4scalyr` dev / testing Docker Hub account?

I will test it before merging the next PR, which is related to this one.

PR #1184, opened 2023-08-23
Push to test4scalyr on pr commit

alesnovak-s1:
We want every commit to a PR on branch release or tagged with docker_push_test to push docker images to the test4scalyr account.
It will allow us to check which images will be pushed when the release tag is created.

Decision tree:

* PR commit => branch == release or label == docker_push_test => push image:VERSION to TEST
* PR tag => tag == VERSION => push image:VERSION to PROD
* Branch commit => branch == master => push image:sSHA to PROD

https://sentinelone.atlassian.net/browse/DTIN-3069

The documentation update is not the part of this issue. It will be recreated when the release process is refactored:

https://sentinelone.atlassian.net/browse/DTIN-2930


tomaz-s1:
Great, this will make directly testing Docker Images from PR changes much easier 👍 

Is there any way to see how this change is working? I checked https://hub.docker.com/r/test4scalyr/scalyr-k8s-agent/tags, but I couldn't find a docker tag related to this PR (maybe I missed it). Or is it because we use version for the tag?

.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -857,21 +856,18 @@ jobs:
...
-            fi
-          else
-              echo "publish=true" >> "${GITHUB_OUTPUT}"
+          AGENT_VERSION="$(cat VERSION | tr -d '\n')"
+
+          if [ "${{ github.ref_type }}" = "tag" ] && [ "${{ github.ref_name }}" != "v${AGENT_VERSION}" ]; then
+            echo "tags=empty" >> "${GITHUB_OUTPUT}"
+          elif ${{ github.ref_name }} == "master"; then
               echo "tags=${{ github.sha }}" >> "${GITHUB_OUTPUT}"
+          else
```
tomaz-s1, line 865: I think that it would be better to use something like that for tags for images which were created as part of PR builds: `pr-<num>-<commit sha>`. 

For example: `pr-1184-bc0610268a25ae80a4f8e8088814571fa4a30401`.

In case this results in too many tags, we could omit the commit sha (last part) from the tag.

And I imagine we may eventually also want up some clean up process / workflow which deletes PR tags which are older than X (e.g. 90) days.

alesnovak-s1: Regarding the naming - good point, I like that and will change it.

Ad repository cleaning - https://sentinelone.atlassian.net/browse/DTIN-3073

alesnovak-s1: > I think that it would be better to use something like that for tags for images which were created as part of PR builds: `pr-<num>-<commit sha>`.
> 
> For example: `pr-1184-bc0610268a25ae80a4f8e8088814571fa4a30401`.
> 
> In case this results in too many tags, we could omit the commit sha (last part) from the tag.
> 
> And I imagine we may eventually also want up some clean up process / workflow which deletes PR tags which are older than X (e.g. 90) days.

Changes commited.

tomaz-s1:
@alesnovak-s1 It appears to be working now - https://hub.docker.com/r/test4scalyr/scalyr-k8s-agent/tags?page=1&name=pr-.

Having said that,  see `latest` tag was updated as well and not just PR specific tags. Is this desired / expected? Or it's a bug in the workflow?

alesnovak-s1:
> 

I've commited t

> @alesnovak-s1 It appears to be working now - https://hub.docker.com/r/test4scalyr/scalyr-k8s-agent/tags?page=1&name=pr- 🎉
> 
> Having said that, see `latest` tag was updated as well and not just PR specific tags. Is this desired / expected? Or it's a bug in the workflow?

I've made some fixes in last commits, including this one.
Now only the pr related tags are pushed.

VERSION and latest are pushed from the release branch only

tomaz-s1:
Thanks, I had another look at Docker Hub and it looks fine now 👍

tomaz-s1 (approved):
I added a question on the release branch conditional, besides that, LGTM.

.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -857,29 +856,37 @@ jobs:
...
-            fi
-          else
-              echo "publish=true" >> "${GITHUB_OUTPUT}"
+              echo "tags=$AGENT_VERSION,latest" >> "${GITHUB_OUTPUT}"
+            else
+              echo "tags=empty" >> "${GITHUB_OUTPUT}"
+            fi        
+          elif [ "${{ github.ref_name }}" = "master" ]; then
               echo "tags=${{ github.sha }}" >> "${GITHUB_OUTPUT}"
+          elif [ "${{ github.head_ref }}" = "release" ]; then
```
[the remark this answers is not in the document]
alesnovak-s1: On release branch, we push pr-X tags and VERSION tag as well.
The VERSION tag is added here to enable us to check what exactly is going to be pushed to the scalyr dockerhub.

PR #1185, opened 2023-08-28
Docs about AIO clarified

alesnovak-s1:


tomaz-s1 (approved):
Thanks a lot of clarifying this.

This makes it much more obvious now that those changes only applied to the AIO packages :)

PR #1186, opened 2023-08-28
Manual merge from scalyr-agent-2-s1-internal

alesnovak-s1:


.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -75,12 +75,14 @@ jobs:
   # Build cacheable requirements of the images.
   build_image_requirements:
     name: requirements-${{ matrix.builder_name }}-${{ matrix.architecture }}
-    runs-on: ubuntu-22.04
+    runs-on: self-hosted
```
tomaz-s1, line 78: For security reasons, we should only use Github cloud and not self hosted runners.

.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -98,7 +100,7 @@ jobs:
           platforms: all
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@4c0219f9ac95b02789c1075625400b2acbff50b1 # v2.0.0
```
tomaz-s1, line 101: As a best practice and for security reasons, we tend to pin all the non-official actions to specific commit sha (granted, with the official large organization github actions like Docker, potential security risk is smaller, but still).

PR #1197, opened 2023-09-24
Parametrizing actions and fixing for self-hosted

alesnovak-s1:


.github/workflows/agent-build-new.yml
```diff
@@ -39,14 +39,13 @@ jobs:
...
 
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4
 
       - id: define_constants
         run: |
           echo "python_version=3.8" >> "${GITHUB_OUTPUT}"
-          echo "cache_version=6" >> "${GITHUB_OUTPUT}"
+          echo "cache_version=$GITHUB_REF_NAME" >> "${GITHUB_OUTPUT}"
```
tomaz-s1, line 47: This means we won't be able to reuse cache across branches since this variable contains branch name, right?

alesnovak-s1: Yes. What I found was that some tests were passing only because of the cache (from a different branch perhaps?) and hiding and error. So I wanted to restrict it.

tomaz-s1: Do we know if it will have significant impact on test run time?

But otherwise I think that's fine and similar approach we use in other places (re-build everything from scratch in master branch).

.github/workflows/agent-build-new.yml
```diff
@@ -187,11 +193,9 @@ jobs:
...
           AWS_ACCESS_KEY: ${{ secrets.CT_AWS_DEV_EC2_ACCESS_KEY }}
           AWS_SECRET_KEY: ${{ secrets.CT_AWS_DEV_EC2_SECRET_KEY }}
-          AWS_PRIVATE_KEY_NAME: ${{ needs.pre_job.outputs.aws_private_key_name }}
-          AWS_PRIVATE_KEY_PATH: "/tmp/private_key.pem"
+          AWS_PRIVATE_KEY_NAME: ${{ secrets.CT_AWS_DEV_EC2_PRIVATE_KEY_NAME }}
+          AWS_PRIVATE_KEY: ${{ secrets.CT_AWS_DEV_EC2_PRIVATE_KEY }}
           AWS_REGION: ${{ needs.pre_job.outputs.aws_region }}
           CICD_WORKFLOW: ${{ needs.pre_job.outputs.cicd_workflow }}
         run: |
-          echo "${{ secrets.CT_AWS_DEV_EC2_PRIVATE_KEY }}" > "${AWS_PRIVATE_KEY_PATH}"
```
tomaz-s1, line 195: I assume this is not needed anymore since the Python script only needs access to the AWS credentials available in the environment variables and not also access to private key, right?

.github/workflows/codespeed-benchmarks.yml
```diff
@@ -17,6 +17,7 @@ jobs:
   pre_job:
     name: Skip Duplicate Jobs Pre Job
     runs-on: ubuntu-latest
+    if: ${{ ! vars.UNIT_TESTS_DISABLED }}
```
tomaz-s1, line 20: Sadly benchmarks (at least the publishing of results to CodeSpeed instance) has been broken for a while. Maybe we can just disable the steps which publish results to CodeSpeed, but leave everything else in place (so we still exercise micro benchmarks and other benchmarking related code).

.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -15,17 +15,24 @@ on:
...
 
-      aws_private_key_name:
-        description: "Name of a private key that is used to connect to AWS EC2 instances that will be created by this workflow."
-        type: string
-
       cicd_workflow:
         description: "Additional string that has to differentiate AWS object from another workflows."
         type: string
+      
+      builder_name:
```
tomaz-s1, line 22: Can you please add some description on what this parameter represents and valid / possible values?

.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -15,17 +15,24 @@ on:
...
+        type: string
+
+      base_image:
+        type: string
+
+      architectures:
```
tomaz-s1, line 28: If the is a fixed list of values we support for this parameter, we should perhaps utilize `options` attribute (https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/).

.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -74,33 +85,52 @@ jobs:
...
         with:
           python_version: ${{ inputs.python_version }}
 
+      - name: Login to DockerHub
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3
+        with:
+          username: ${{ secrets.PUBLISH_REGISTRY_USERNAME }}
+          password: ${{ secrets.PUBLISH_REGISTRY_PASSWORD }}
+
+      - name: Login to an internal build registry
```
tomaz-s1, line 108: Should we only run this step in case some variable (e.g. runner) indicates we are building it for internal registry / using on prem runner?

.github/workflows/codespeed-benchmarks.yml
alesnovak-s1: The benchmark work for me. Disabling Unit tests is meant for the internal repo only.

tomaz-s1: Thanks for clarifying. In this case it would be good to document in that readme file (when disabling unit tests makes sense or is desired).

.github/workflows/reusable-agent-build-container-images.yml
alesnovak-s1: I am sending there a json representing a list.
There's no structure for that in GHA inputs unfortunatelly.

https://github.com/scalyr/scalyr-agent-2/blob/parametrize_gha_2/.github/workflows/README.md

agent_build_refactored/utils/aws/ec2.py
```diff
@@ -98,15 +96,45 @@ def _create_ssh_container(
...
         )
 
+        # The AWS_PRIVATE_KEY must be saved inside the running container using an echo command with a pipe.
+        # Using a volume mount is not possible when docker is installed via snap and leaves a key file in the filesystem.
         cmd_args.extend(
-            additional_cmd_args
+            [     
+                "/bin/bash",
+                "-c",
+                f"mkdir /tmp/mounts; echo \"{os.environ['AWS_PRIVATE_KEY']}\" > {self._ssh_client_container_in_docker_private_key_path}; chmod 600 {self._ssh_client_container_in_docker_private_key_path}; cat {self._ssh_client_container_in_docker_private_key_path}; " +
```
tomaz-s1, line 105: Is it a good idea to cat private key file at the end? This could cause a secret leak in the logs, right?

I guess if the private key material (file content) is defined as a secret in Github Actions / similar, this should not be an issue, but still probably better to remove this step (I assume it was there only for troubleshooting / debugging or a similar reason).

agent_build_refactored/utils/aws/ec2.py
```diff
@@ -98,15 +96,45 @@ def _create_ssh_container(
...
+
+        # Since the docker is run in the background, we need to wait for the bash command to be really excuted.
+        cmd = [
+                    "docker",
+                    "exec",
+                    "-i",
+                    container_name,
+                    "bash",
+                    "-c",
+                    f"while [ ! -f {self._ssh_client_container_in_docker_private_key_path} ]; do sleep 1; echo Waiting for the key file ...; done"
```
tomaz-s1, line 125: Do we need some kind of timeout or similar? So we don't wait for ever in case key file is not written for whatever reason (I imagine that in case key file is not written / available in some short amount of time, this likely indicates developer error or similar and it doesn't make sense to try to wait any longer)

.github/workflows/reusable-agent-build-container-images.yml
alesnovak-s1: For workflow_call you can use only boolean, number and string.

tomaz-s1: Ah, OK, so the `choice` type  is not supported for `workflow_call` trigger. Good to know, thanks.

For reference:

* https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputsinput_idtype
* https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputsinput_idtype

.github/workflows/agent-build-new.yml
alesnovak-s1: The build and test time is still up to 20 minutes.

agent_build_refactored/utils/aws/ec2.py
alesnovak-s1: You're right! It's *-ed, but still a very bad practice. Will fix it.

.github/workflows/codespeed-benchmarks.yml
alesnovak-s1: Modified the description in README.md. Btw, I didn't mention the internal repo on purpose.

tomaz-s1 (approved):
I had a couple of small questions and comments, but overall, LGTM.

.github/workflows/reusable-agent-build-container-images.yml
tomaz-s1: EDIT: I see build is failing due to missing username and password, so we likely indeed only want to run this step in case `INTERNAL_BUILD_REGISTRY_USERNAME` and `INTERNAL_BUILD_REGISTRY_PASSWORD` is defined / not empty.

alesnovak-s1: Yeah, I wanted to let if fail and continue in case there are no credentials for it. But it will be cleaner to do it the way you suggested.

build_package_new.py
```diff
@@ -1,237 +1,241 @@
...
+    image_parser = subparsers.add_parser("image")
+
+    image_parser.add_argument(
+        "builder_name",
+        choices=ALL_CONTAINERISED_AGENT_BUILDERS.keys(),
+        help="Name of the builder.",
+    )
+
+    image_parser.add_argument(
+        "--base-image", required=True, help="Name of the image to build"
```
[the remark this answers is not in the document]
alesnovak-s1: Oh, that's a copy paste problem there :-D

PR #1201, opened 2023-10-03
Fixing test_packages.py test - upgrade flask version

alesnovak-s1:
Fixing test_packages.py test cause by a new version of werkzeug incompatible with flask version used in the test.
Upgraded to the version used in the project


scalyr_agent/connection.py
```diff
@@ -589,7 +589,7 @@ def log_server_ssl_certificate(py_post_equal_279, log_level=logging.WARN):
                     log_server_ssl_certificate(
                         py_post_equal_279=False, log_level=logging.WARN
                     )
-                    raise e
+                    raise Exception(f"Error wrapping a ssl socket. ca_file: {self.__ca_file}") from e
```
tomaz-s1, line 592: Do you happen to have an example how a message from this new exception now looks like?

I just want to make sure it still includes all the relevant context (which it likely should since it's using "from" notation, but wanted to double check)

tomaz-s1 (approved):
LGTM with one small comment / question on the new exception message field value.

scalyr_agent/connection.py
alesnovak-s1: I don't have the example, but it prints the new exception and the original one as well.

PR #1205, opened 2023-10-12
Fixing container being included after failed request to K8s API

alesnovak-s1:


czerwingithub (approved):

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2262,13 +2262,23 @@ def _get_containers(
...
+                    pod = None
+                    try:
+                        pod = k8s_cache.pod(
+                            pod_namespace,
+                            pod_name,
+                            current_time,
+                            allow_expired=False,
+                            ignore_k8s_api_exception=False,
+                        )
+                    except k8s_utils.K8sApiException as e:
```
jmakar-scalyr, line 2275: It might be good to use the http status code from the K8sApiException to help decide what to do

For example looking at the notes from DTIN-3156, a 404 means the pod doesn't exist so it should always be excluded

jmakar-scalyr (approved):
Looks good, just a had a comment on how to better handle this (no need to do that now if this fix is urgent)

scalyr_agent/builtin_monitors/kubernetes_monitor.py
alesnovak-s1: That makes sense.
404 might also mean the pod existed and then was garbage collected before the agent had a chance to process its logs.
But in that case, we wouldn't know if it was annotated to be included anyway.
So the safe action is to exclude it and log a warning.

alesnovak-s1: 404 => Pod not found, Log Warning, excluded
401 => Unauthorized, Log Error, excluded

PR #1206, opened 2023-10-12
Added Centos 6 Vault Repo mirror

alesnovak-s1:
Added vault.centos.org repo mirror to have a better chance of packages being available.


jmakar-scalyr (changes requested):
I think it's better to replace the existing files (CentOS-Base.repo, etc) rather than add new ones, especially since https://vault.epel.cloud is offline

Also CentOS 6 end-of-life was in Nov 2020, can we just remove CentOS 6 support?
Is there a DataSet query to we can run that shows the number (if any) Scalyr agents still running on CentOS 6?  WDYT @czerwingithub ?

czerwingithub (approved):

czerwingithub:
> I think it's better to replace the existing files (CentOS-Base.repo, etc) rather than add new ones, especially since https://vault.epel.cloud is offline
> 
> Also CentOS 6 end-of-life was in Nov 2020, can we just remove CentOS 6 support? Is there a DataSet query to we can run that shows the number (if any) Scalyr agents still running on CentOS 6? WDYT @czerwingithub ?

Sorry, I shouldn't have approved with Joey's outstanding questions.  Yeah, there should be a way to figure this out with a query.

alesnovak-s1:
> > I think it's better to replace the existing files (CentOS-Base.repo, etc) rather than add new ones, especially since https://vault.epel.cloud is offline
> > Also CentOS 6 end-of-life was in Nov 2020, can we just remove CentOS 6 support? Is there a DataSet query to we can run that shows the number (if any) Scalyr agents still running on CentOS 6? WDYT @czerwingithub ?
> 
> Sorry, I shouldn't have approved with Joey's outstanding questions. Yeah, there should be a way to figure this out with a query.

My intention is to have two mirrors - vault.centos.org and vault.epel.cloud. 
From what I read on forums, both of them are not very realiable. At this moment, vault.epel.cloud is online again.

alesnovak-s1:
> > > I think it's better to replace the existing files (CentOS-Base.repo, etc) rather than add new ones, especially since https://vault.epel.cloud is offline
> > > Also CentOS 6 end-of-life was in Nov 2020, can we just remove CentOS 6 support? Is there a DataSet query to we can run that shows the number (if any) Scalyr agents still running on CentOS 6? WDYT @czerwingithub ?
> > 
> > 
> > Sorry, I shouldn't have approved with Joey's outstanding questions. Yeah, there should be a way to figure this out with a query.
> 
> My intention is to have two mirrors - vault.centos.org and vault.epel.cloud. From what I read on forums, both of them are not very realiable. At this moment, vault.epel.cloud is online again. But this is not the correct way to do it apparently. I will fix it.

Looking at the docs, there's no straightforward way to have a fallback repo configured and it's not worth spending more time on it.
Switching to vault.centos.org, because it seems to be more stable recently.

PR #1210, opened 2023-10-29
Fixing the threading issue in the syslog monitor

alesnovak-s1:
Requests are read in a thread pool first to release the socket fd as soon as possible to ensure higher throughput.
The data is handled in a separate thread pool.

https://sentinelone.atlassian.net/browse/DTIN-3231

### DIFF
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -1092,7 +1092,7 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
 
         try:
             DONE = "DONE"
-            work_queue = queue.SimpleQueue()
+            work_queue = queue.Queue()
             def worker(queue):
                 data = queue.get(block=True)
                 log_to_file("SyslogTCPHandler.handle - worker got data data=" + str(data))
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -610,9 +610,6 @@ class SyslogUDPHandler(six.moves.socketserver.BaseRequestHandler):
             self.server.syslog_handler.handle, data, extra
         )
 
-        # self.server.syslog_handler.handle(data, extra)
-
-
 
 class SocketNotReadyException(Exception):
     def __init__(self, cause):
@@ -1038,13 +1035,9 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
                 traceback.format_exc(),
             )
 
-    def handle(self):
-        syslog_request = SyslogRequest(
-            socket=self.request,
-            max_buffer_size=self.server.tcp_buffer_size
-        )
+    def _syslog_request_parser(self):
         if self.request_parser == "default":
-            syslog_parser = SyslogRequestParser(
+            return SyslogRequestParser(
                 socket_client_address=self.client_address,
                 socket_server_address=self.server.server_address,
                 max_buffer_size=self.server.tcp_buffer_size,
@@ -1052,7 +1045,7 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
                 message_size_can_exceed_tcp_buffer=self.server.message_size_can_exceed_tcp_buffer
             )
         elif self.request_parser == "batch":
-            syslog_parser = SyslogBatchedRequestParser(
+            return SyslogBatchedRequestParser(
                 socket_client_address=self.client_address,
                 socket_server_address=self.server.server_address,
                 max_buffer_size=self.server.tcp_buffer_size,
@@ -1061,7 +1054,7 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
                 message_delimiter=self.message_delimiter
             )
         elif self.request_parser == "raw":
-            syslog_parser = SyslogRawRequestParser(
+            return SyslogRawRequestParser(
                 socket_client_address=self.client_address,
                 socket_server_address=self.server.server_address,
                 max_buffer_size=self.server.tcp_buffer_size,
@@ -1069,6 +1062,14 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
             )
         else:
             raise ValueError("Invalid request parser: %s" % (self.request_parser))
+
+    def handle(self):
+        syslog_request = SyslogRequest(
+            socket=self.request,
+            max_buffer_size=self.server.tcp_buffer_size
+        )
+        syslog_parser = self._syslog_request_parser()
+        
         global_log.log(
             scalyr_logging.DEBUG_LEVEL_1,
             "SyslogTCPHandler.handle - created syslog_parser. Thread: %d",
@@ -1076,6 +1077,8 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
         )
 
         try:
+            # Worker is responsible for processing the data read from one request.
+            # Using the queue ensures the data is processed in the order it was read.
             DONE = "DONE"
             work_queue = queue.Queue()
             def worker(queue):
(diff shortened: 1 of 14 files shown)


scalyr_agent/builtin_monitors/thread_pool.py
```diff
@@ -0,0 +1,59 @@
+import threading
```
tomaz-s1, line 1: Let's please add ASF license header to this file.

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -18,6 +18,10 @@
 from __future__ import absolute_import
```
tomaz-s1, line 18: It would be good to add a docstring to this module to explain to the developers how request processing works and that we use a thread pool for processing requests and reading data from the socket.

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -570,77 +576,113 @@ class SyslogUDPHandler(six.moves.socketserver.BaseRequestHandler):
...
     def handle(self):
         data = six.ensure_text(self.request[0].strip(), "utf-8", errors="ignore")
         extra = {
             "proto": "udp",
             "srcip": self.client_address[0],
             "destport": self.server.server_address[1],
         }
-        self.server.syslog_handler.handle(data, extra)
 
+        self.__request_processing_executor.submit(
```
tomaz-s1, line 602: Since `handle` is now called inside a thread, do we need any kind of locking around this method?

It appears we already have a lock in https://github.com/scalyr/scalyr-agent-2/blob/888447321333b7f817739cb06806a461188b3ba7/scalyr_agent/builtin_monitors/syslog_monitor.py#L1583 for file creation related reasons and all `handle()` method does is pass data to the logger (https://github.com/scalyr/scalyr-agent-2/blob/888447321333b7f817739cb06806a461188b3ba7/scalyr_agent/builtin_monitors/syslog_monitor.py#L1738).

I believe we already have a lock in our custom logger methods, but it would be good to confirm there are no race conditions - starting a monitor and sending a lot of data to in in parallel manner and verifying that syslog monitor log file is correct aka that data is not interleaved.

scalyr_agent/builtin_monitors/thread_pool.py
```diff
@@ -0,0 +1,59 @@
...
+from concurrent.futures import ThreadPoolExecutor
+
+from scalyr_agent import scalyr_logging
+
+global_log = scalyr_logging.getLogger(__name__)
+
+class ExecutorMixIn:
+    def __init__(self):
+        # Using 4 threads for processing requests, because of GIL and CPU bound tasks higher number would not help process the data faster.
+        self._request_processing_executor = ThreadPoolExecutorFactory.get_singleton("request_processing_executor", max_workers=4)
```
tomaz-s1, line 11: Do we want to make `max_workers` configurable?

Perhaps it's a good idea in case someone hits a race condition or similar so we can provide them with a workaround to set this value to `1`.

scalyr_agent/monitors_manager.py
```diff
@@ -208,6 +210,10 @@ def stop_manager(self, wait_on_join=True, join_timeout=5):
 
         start_time = time.time()
 
+        log.info("Stopping shared thread pools")
+        ThreadPoolExecutorFactory.shutdown(wait=False, cancel_futures=True)
```
tomaz-s1, line 214: To me it seems like that this should live on the syslog monitor `stop()` method and not inside the MonitorsManager itself.

alesnovak-s1: The thread pools are shared across the syslog monitors. That's why it needs to be here.

tomaz-s1: Ah, ok, makes sense, thanks for clarifying.

Could this cause any issues since each monitor is running in a separate thread already?

alesnovak-s1: That's a good question.
ThreadPool.submit (this method is thread safe) adds a callable item with parameters to a queue and starts a new thread and adds it to an internal pool if num_threads < self._max_workers.
What I was worried about is creating and starting a long running nested thread - possibly blocking joining of the parent thread. But that's not the case in python.
It would be a better design not to call processing_executor.submit from inside thread from a request_reading_executor pool, but abstractions around BaseServer don't allow it. 
Ideally we should use an async server like tornado to handle these requests, but that's for another PR :)

scalyr_agent/builtin_monitors/thread_pool.py
alesnovak-s1: This was a laziness on my part, I will make it configurable :)

scalyr_agent/builtin_monitors/syslog_monitor.py
tomaz-s1: Thanks for the clarification :)

scalyr_agent/monitors_manager.py
alesnovak-s1: > To me it seems like that this should live on the syslog monitor `stop()` method and not inside the MonitorsManager itself.

Looking more into the code of syslog monitors, I've realized you're right.
Since SyslogMonitor is in charge of starting and stopping all Syslog Servers, it makes sense for it to be responsible for creation and shutting down of the pools.

alesnovak-s1: In the end it really needs to be in the Monitor Manager, I understood the Syslog Monitor wrong. It starts servers for one monitor configuration only.

alesnovak-s1:
@tomaz-s1 
I have added configuration, changed tests to follow the new api of some of the classes (Syslog Request and Parser).
And now to ensure the request data is processed in the correct order, the worker submitted to the processing executor gets his data from a Queue (one queue and one worker per request). 
Please review again.

tomaz-s1:
@alesnovak-s1 Thanks.

Have we done any end to end testing with agent running multiple monitor instances and producer sending data in parallel (e.g. using https://github.com/scalyr/syslog-collector/tree/main/data-and-load-generator tool)?

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -570,77 +583,110 @@ class SyslogUDPHandler(six.moves.socketserver.BaseRequestHandler):
...
-            return None
-        except NON_BLOCKING_SOCKET_DATA_NOT_READY_EXCEPTIONS:
-            return None
+                raise SocketClosed()
+        except (socket.timeout, *NON_BLOCKING_SOCKET_DATA_NOT_READY_EXCEPTIONS) as e:
+            raise SocketNotReadyException(e)
         except socket.error as e:
             if e.errno == errno.EAGAIN:
-                return None
+                raise SocketNotReadyException(e)
```
tomaz-s1, line 643: Just curious why this change is needed? In the past, we had bug related to this code path throwing, but this could have changed with the multi threading model.

tomaz-s1 (approved):
LGTM, but I would feel more comfortable if we performed more end to end testing with multiple monitor instances and highly parallel producer.

scalyr_agent/builtin_monitors/syslog_monitor.py
alesnovak-s1: Before the method returned None for both cases - when data on socket were not ready and when the socket was closed.
And you had to guess what happened. 
The exceptions are path of the code flow and are expected and catched right away, avoiding any propagating.

alesnovak-s1: And I don't like returning None when something unexpected happened. It's quite error-prone.

alesnovak-s1:
> @alesnovak-s1 Thanks.
> 
> Have we done any end to end testing with agent running multiple monitor instances and producer sending data in parallel (e.g. using https://github.com/scalyr/syslog-collector/tree/main/data-and-load-generator tool)?

Yes, I have a config with multiple Syslog (~15) monitors configured with both udp and tcp ports.
I was using loggen and syslog-loadgen and sending tons of data to all the ports in parallel, several processes per port.

PR #1216, opened 2023-11-09
allow_http_monitors config implementation

alesnovak-s1:
Implementation of https://sentinelone.atlassian.net/browse/DTIN-3192.
For FIPS compliant environments we want to have a config (env) flag that will force HTTPS configuration on all relevant monitors.

I've used the same patten as allow_http does.

### DIFF
--- a/scalyr_agent/builtin_monitors/apache_monitor.py
+++ b/scalyr_agent/builtin_monitors/apache_monitor.py
@@ -49,7 +49,7 @@ define_config_option(
     "number, at which the Apache status module is served. The URL should end in `/?auto` to "
     "indicate that the machine-readable version of the page should be returned.",
     default="http://localhost/server-status/?auto",
-    insecure_http_url=True
+    allow_http=False
 )
 define_config_option(
     __monitor__,
--- a/scalyr_agent/builtin_monitors/apache_monitor.py
+++ b/scalyr_agent/builtin_monitors/apache_monitor.py
@@ -49,6 +49,7 @@ define_config_option(
     "number, at which the Apache status module is served. The URL should end in `/?auto` to "
     "indicate that the machine-readable version of the page should be returned.",
     default="http://localhost/server-status/?auto",
+    insecure_http_url=True
 )
 define_config_option(
     __monitor__,
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -523,12 +523,6 @@ class Configuration(object):
                 monitor_config[name] = ensure_https_url(value)
         return monitor_config
 
-        # from scalyr_agent.builtin_monitors import unsecure_http_config_fields
-        # for field in unsecure_http_config_fields:
-        #     if field in monitor_config:
-        #         monitor_config[field] = ensure_https_url(monitor_config[field])
-
-
     def __verify_workers(self):
         """
         Verify all worker config entries from the "workers" list in config.
--- a/scalyr_agent/configuration.py
+++ b/scalyr_agent/configuration.py
@@ -515,6 +515,7 @@ class Configuration(object):
 
     @staticmethod
     def force_http_monitor_config(monitor_config):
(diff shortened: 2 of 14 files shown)


tomaz-s1:
IIRC, we have multiple ways (for backward compatibility and various historical reasons) for sending HTTP(s) requests.

There is a custom httplib client (which is used by the main agent code) and in some places (monitors) we utilize requests library.

Having said that, it appears that shouldn't matter here since this PR is only about forcing https URL, but not actually about also forcing cert + hostname validation, or?

scalyr_agent/builtin_monitors/apache_monitor.py
```diff
@@ -49,6 +49,7 @@
     "number, at which the Apache status module is served. The URL should end in `/?auto` to "
     "indicate that the machine-readable version of the page should be returned.",
     default="http://localhost/server-status/?auto",
+    insecure_http_url=True
```
jmakar-s1, line 52: For consistency could we rename this as `allow_http`?

I know that could be confused with the unrelated allow_http for scalyr server but I think it's clearer to have the monitor config parameter with a similar name

jmakar-s1:
> IIRC, we have multiple ways (for backward compatibility and various historical reasons) for sending HTTP(s) requests.
> 
> There is a custom httplib client (which is used by the main agent code) and in some places (monitors) we utilize requests library.
> 
> Having said that, it appears that shouldn't matter here since this PR is only about forcing https URL, but not actually about also forcing cert + hostname validation, or?

Yes, the request was to disallow http

jmakar-s1 (approved):
Looks good but had a couple requested changes

alesnovak-s1:
> IIRC, we have multiple ways (for backward compatibility and various historical reasons) for sending HTTP(s) requests.
> 
> There is a custom httplib client (which is used by the main agent code) and in some places (monitors) we utilize requests library.
> 
> Having said that, it appears that shouldn't matter here since this PR is only about forcing https URL, but not actually about also forcing cert + hostname validation, or?

It's only a safe-guard for the configuration as allow_http. And I wanted to make the fix as simple as possible, because I still believe user should be responsible for what he puts inside the configuration.
All the same, it would be cleaner to force this on a lower level. But to achieve that, we would need to change httplib client, patch requests library, docker client and others probably. Or am I mistaken?

scalyr_agent/configuration.py
```diff
@@ -499,6 +505,25 @@ def parse(self):
...
+
+    @staticmethod
+    def force_http_monitor_config(monitor_config):
+        # Loads a monitor module, checks for options marked as insecure_http_url and forces https scheme there.
+        __import__(monitor_config["module"])
+        monitor_info = MonitorInformation.get_monitor_info(monitor_config["module"])
+        for name, value in monitor_config.iteritems():
+            monitor_config_option = monitor_info.config_option(name)
+            if monitor_config_option and monitor_config_option.insecure_http_url:
+                monitor_config[name] = ensure_https_url(value)
```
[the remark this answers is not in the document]
alesnovak-s1: This applies ensure_http_url to only those options marked as insecure_http_url. 
I was thinking about checking based on pattern like "url", but that could have some unexpected effects for new monitors. "url" could be a part of option that does not contain an actuall url, i.e. normalize_url: bool

scalyr_agent/builtin_monitors/apache_monitor.py
alesnovak-s1: That sounds good. I wasn't sure about how to name it and insecure_http_url is definitelly confusing.
I will make allow_http=True by default and  allow_http=False will mark the option for https enforcing. 
Do you agree?

jmakar-s1: Sounds good

scalyr_agent/configuration.py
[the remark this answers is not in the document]
jmakar-s1: Ah got it, my mistake

PR #1218, opened 2023-11-10
Fixing support for python2

alesnovak-s1:
Implementing a threading fallback for python2 and syslog monitor server. The fallback uses ThreadingMixin and works the same way as before the change to ThreadPool, which broke the python2 compatibility in the first place.

Added some necessary test fixes and loggings.

Some of the tests were using python3 instead of python2 or testing against the stable version downloaded from the repo instead of the working one. 

I've changed current version to a working snapshot version 2.2.7.pre.8.1 and will continue using snapshot version between releases  to avoid conflicting builds.

https://sentinelone.atlassian.net/browse/DTIN-3358


VERSION
```diff
@@ -1 +1 @@
-2.2.7
\ No newline at end of file
+2.2.7.pre8.1
```
jmakar-s1, line 1: I'm not sure if there is any processing done on this value that expects a `x.y.z` format, to be on the safe side it might be better to change this to something like `2.2.7-pre8-1`

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -56,7 +54,10 @@
 from six.moves import range
 import six.moves.socketserver
 
-from scalyr_agent.builtin_monitors.thread_pool import ExecutorMixIn
+if six.PY2:
+    from scalyr_agent.builtin_monitors.thread_pool_dummy import ExecutorMixIn
+else:
+    from scalyr_agent.builtin_monitors.thread_pool import ExecutorMixIn
```
jmakar-s1, line 60: There are aspects of thread_pool that aren't compatible with python 2?  I see the thread_pool_dummy doesn't do anything, it would be better if an equivalent approach could be implemented for python 2

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1077,25 +1083,30 @@ def handle(self):
...
+                DONE = "DONE"
+                work_queue = queue.Queue()
+                def worker(queue, is_shutdown):
                     data = queue.get(block=True)
+                    while data != DONE:
+                        if is_shutdown():
+                            global_log.info("ThreadPool shutting down, skipping further request processing.")
+                            break
+                        self.__request_data_process(syslog_parser, data)
+                        data = queue.get(block=True)
```
jmakar-s1, line 1098: I would consider rearranging this to avoid the special value & handle queue cleanup during shutdown:

```
def worker(queue, is_shutdown):
    while True:
        if is_shutdown():
            global_log.info("ThreadPool shutting down, skipping further request processing.")
            break
        try:
            data = queue.get(block=False, timeout=1.0)
        except Empty:
            continue
        self.__request_data_process(syslog_parser, data)
        queue.task_done()

    while not queue.empty():
        data = queue.get()
        queue.task_done()
```

Then the producer calls `queue.join()` to ensure everything from the queue has been consumed

tests/image_builder/monitors/base/Dockerfile
```diff
@@ -4,6 +4,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt update -y
 RUN apt install -y mysql-server
 RUN apt install -y postgresql
+RUN apt install -y git
```
jmakar-s1, line 7: Better to do this to avoid creating many unnecessary layers:
```
RUN apt update -y && apt install -y git mysql-server postgresql
```

jmakar-s1 (approved):
Looks fine, minor suggestions

scalyr_agent/builtin_monitors/syslog_monitor.py
alesnovak-s1: In python2 there's ThreadPool with a different API.
I was considering implementing the python2 solution around it, but then decided no to touch a code that already works, especially the python2 part which will be deprecated soon.
The ExecutorMixin from dummy package inherits ThreadingMixin - this is the way it was implemented before, it starts a new thread for every incoming request.

VERSION
alesnovak-s1: Good point. I will change it and see if anything fails :)

PR #1219, opened 2023-11-15
Thread pool per monitor, so monitors don't block each other

alesnovak-s1:
This PR fixes the issue of agent not stopping gracefully and producing logs only for one syslog monitor under load:
https://sentinelone.atlassian.net/browse/DTIN-3231

*  processing_threads set to 4 and reading_threads  to auto (that was the intention from the start, I just mixed it up)
* Thread pool is not shared anymore, but one per syslog server. When share it the threads were not being evenly distributed across monitors and one monitor under load caused other monitors to stop producing any logs. 
* Removed ThreadPoolExecutorFactory
* queue.task_done call added to avoid any future regression bugs when expecting the consistency of the queue object
* Extended the time to wait for the agent to stop gracefully before killing it.

### DIFF
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -1070,9 +1070,10 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
     def __worker(data_processor, data, last_processing_future):
         # Hardcoded timeout total time to stop polluting Thread Pool in case of a bug or an unexpected error
         TIMEOUT_TOTAL_TIME = 10
-        CANCEL_FURTHER_PROCESSING_MSG = "Cancelling further processing."
-        TIMEOUT_ERROR_MSG = "Timeout error while waiting for previous message being parsed." + " " + CANCEL_FURTHER_PROCESSING_MSG
-        CANCELLED_ERROR_MSG = "Future of the revious message parser cancelled." + " " + CANCEL_FURTHER_PROCESSING_MSG
+        CANCEL_FURTHER_PROCESSING_MSG = "Cancelling further processing for this request."
+        TIMEOUT_ERROR_MSG = "Timeout error while waiting for previous message being parsed. If the server is not shutting down, this may be a bug or an unexpected error." + " " + CANCEL_FURTHER_PROCESSING_MSG
+        CANCELLED_ERROR_MSG = "Future of the revious message parser cancelled. The server is probably shutting down. " + " " + CANCEL_FURTHER_PROCESSING_MSG
+        PREVIOUS_CANCELEED_ERROR_MSG = "Previous message parser encountered an error. " + " " + CANCEL_FURTHER_PROCESSING_MSG
         GENERIC_EXCEPTION_MSG = "Exception from the previous message handler. " + " " + CANCEL_FURTHER_PROCESSING_MSG
 
         try:
@@ -1080,11 +1081,15 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
                 try:
                     last_processing_future.result(timeout=TIMEOUT_TOTAL_TIME)
                 except concurrent.futures.TimeoutError as e:
-                    global_log.error(TIMEOUT_ERROR_MSG, exc_info=e)
+                    global_log.warn(TIMEOUT_ERROR_MSG)
                     raise SyslogTCPHandler.PreviousMessageHandlingError(TIMEOUT_ERROR_MSG)
                 except concurrent.futures.CancelledError as e:
-                    global_log.error(CANCELLED_ERROR_MSG, exc_info=e)
+                    global_log.warn(CANCELLED_ERROR_MSG)
+                    global_log.info(f"Thread: {threading.current_thread().name} __worker {e}")
                     raise SyslogTCPHandler.PreviousMessageHandlingError(CANCELLED_ERROR_MSG)
+                except SyslogTCPHandler.PreviousMessageHandlingError as e:
+                    global_log.debug(PREVIOUS_CANCELEED_ERROR_MSG, exc_info=e)
+                    raise e
                 except Exception as e:
                     global_log.error(GENERIC_EXCEPTION_MSG, exc_info=e)
                     raise SyslogTCPHandler.PreviousMessageHandlingError(GENERIC_EXCEPTION_MSG)
@@ -1107,7 +1112,7 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
 
         try:
             if self.__request_processing_executor:
-                # Using last_future to ensure the data from one requests is processes in the same order as it was received
+                # Using last_future to ensure the data from one requests is processed in the same order as it was received
                 last_future = None
                 for data in self.__request_stream_read(syslog_request, self.server.is_running):
                     last_future = self.__request_processing_executor.submit(self.__worker, syslog_parser.process, data, last_future)
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -1007,7 +1007,7 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
             six.moves.socketserver.BaseRequestHandler.__init__(self, *args, **kwargs)
 
     @staticmethod
-    def __request_stream_read(syslog_request, server_is_funning_fn):
+    def __request_stream_read(syslog_request, server_is_running_fn):
         count = 1
         while not syslog_request.is_closed:
             check_running = False
@@ -1027,7 +1027,7 @@ class SyslogTCPHandler(six.moves.socketserver.BaseRequestHandler):
             except SocketClosed:
                 continue
 
-            if check_running and not server_is_funning_fn():
+            if check_running and not server_is_running_fn():
                 return
 
             count += 1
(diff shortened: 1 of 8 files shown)


tomaz-s1:
Just wondering - is there any way we could add integration / end to end (regression) test which would catch this issue?

I know it can't be a directly unit test since we need to start the actual agent with a specific configuration, so it would require more work.

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1095,6 +1095,7 @@ def worker(queue, is_shutdown):
                             global_log.info("ThreadPool shutting down, skipping further request processing.")
```
tomaz-s1, line 1095: I assume this means any data which is still in queue will be lost, right?

I think that's probably fine, but I wonder if it would make sense to update the shutdown logic to work something like that (if that's not already the case):

1. Stop syslog server which is accepting data (so producers / lb in front / similar can route data to another instance or similar (only work in case of TCP receiver of course)
2. Try to process what we have in queue for up to GRACE_PERIOD seconds
3. After grace period has been reached, exit regardless

IIRC, we already have a similar graceful shutdown process in place with the agent itself, but I don't remember how exactly it works and interacts with the monitors manager itself.

Maybe I'm also over complicating it a bit since likely a lot of producers only support UDP where that doesn't matter anyway...

tomaz-s1:
I had some general question / comment on monitor graceful shutdown procedure, besides that, LGTM.

alesnovak-s1:
> Just wondering - is there any way we could add integration / end to end (regression) test which would catch this issue?
> 
> I know it can't be a directly unit test since we need to start the actual agent with a specific configuration, so it would require more work.

That's a good point!
Unit test should be possible as well, only thing I need is to start several Syslog[TCP|UDP]Server's on random free ports and start sending requests there.
Integration / E2E will be a bit more complicated, but would make sense as well. To make it more readable and less complex I might just create a bash test for it.

scalyr_agent/builtin_monitors/syslog_monitor.py
alesnovak-s1: Monitor manager calls the stop method of each monitor and then, if configured so, waits for a monitors to stop.

I agree  it's a good idea to let the syslog monitor process the requests already received to avoid loosing any data. I will add it there. It's a good point with the LB, I can already see the Support Ticket describing a strange behavior, lost data and us spending 1 day figuring out, what happened :)

alesnovak-s1:
@tomaz-s1 
Final version commited, please re=-review.
The processing changed from using one queue and one worker for every request to one worker per data read. Each worker has a reference to a future of the last worker from the same request submitted to the thread pool to ensure the processing keeps the same order.
In case of request parts count > processing threads, we might be effectively processing single threaded, but that's optimal, because of GIL. 10s deadlock might happen only in a case of bug.

tomaz-s1:
@alesnovak-s1 Thanks. @jmakar-s1 can you please also review when you get a chance? Thanks.

scalyr_agent/platform_posix.py
```diff
@@ -720,8 +720,8 @@ def stop_agent_service(self, quiet):
 
         # Try killing the daemon process
         try:
-            # Do 5 seconds worth of TERM signals.  If that doesn't work, KILL it.
-            term_attempts = 50
+            # Do 50 seconds worth of TERM signals.  If that doesn't work, KILL it.
+            term_attempts = 500
```
tomaz-s1, line 724: Could this increase in shutdown delay affect other setups / configurations when running syslog monitor? Perhaps Kubernetes one where we have health checks and shut down grace period defined?

jmakar-s1:
> @alesnovak-s1 Thanks. @jmakar-s1 can you please also review when you get a chance? Thanks.

Will do, but likely won't get a chance for a couple days

scalyr_agent/platform_posix.py
alesnovak-s1: That's a good point. Also, this change shouldn't be part of this PR. Reverting.

tomaz-s1:
> In case of request parts count > processing threads,

I'm not sure I fully understand that statement.

Wouldn't that scenario simply mean that in case parts count > processing threads, we would still be processing with `processing threads` number of threads, but other requests will wait until a thread frees up?

scalyr_agent/builtin_monitors/thread_pool.py
```diff
@@ -20,23 +20,38 @@
...
+        processing_threads = 8
         # Let the ThreadPoolExecutor decide how many threads to use based the numbre of logical CPUs
-        processing_threads = None
+        reading_threads = None
 
         if global_config:
             reading_threads = global_config.syslog_socket_thread_count
             processing_threads = global_config.syslog_processing_thread_count
+            self._shutdown_grace_period = global_config.syslog_monitors_shutdown_grace_period
+        else:
```
tomaz-s1, line 41: I assume we end in this else branch inside tests?

alesnovak-s1:
> > In case of request parts count > processing threads, we might be effectively processing single threaded, but that's optimal, because of GIL. 10s deadlock might happen only in a case of bug.
> 
> I'm not sure I fully understand that statement.
> 
> Wouldn't that scenario simply mean that in case parts count > processing threads, we would still be processing with `processing threads` number of threads, but other requests will wait until a thread frees up?

Consider this scenario:
We have N processing threads
We receive only 1 request and read data from it in >=N parts.
Then the processing threads will be like: TN waiting for TN-1 .... waiting for T1 (processing)
The pool is full with threads waiting for the previous ones and only the first one is being processed.

But that will happen very rarely and it will have an insignificant impact on performance.

alesnovak-s1:
@tomaz-s1 @jmakar-s1 
This change was implemented within the limits of python Server / Handler abstractions to avoid  changing the logic request reading / waiting for data, because I expect it to be time-proven, since the syslog protocol is so fluid.
Should we need to change it further in future, we might consider an async based solution like tornado.

scalyr_agent/builtin_monitors/thread_pool.py
alesnovak-s1: No, inside the IF branch, there's a MockConfig being passed.
I've created a test to run it without global_config.

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1019,7 +1027,7 @@ def __request_stream_read(syslog_request, server_is_funning_fn):
             except SocketClosed:
                 continue
 
-            if check_running and not server_is_funning_fn:
+            if check_running and not server_is_funning_fn():
```
jmakar-s1, line 1030: s/funning/running/

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1069,6 +1066,36 @@ def _syslog_request_parser(self):
         else:
             raise ValueError("Invalid request parser: %s" % (self.request_parser))
 
+    @staticmethod
+    def __worker(data_processor, data, last_processing_future):
+        # Hardcoded timeout total time to stop polluting Thread Pool in case of a bug or an unexpected error
```
jmakar-s1, line 1071: s/polluting/populating/

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1069,6 +1066,36 @@ def _syslog_request_parser(self):
...
+        TIMEOUT_TOTAL_TIME = 10
+        CANCEL_FURTHER_PROCESSING_MSG = "Cancelling further processing for this request."
+        TIMEOUT_ERROR_MSG = "Timeout error while waiting for previous message being parsed. If the server is not shutting down, this may be a bug or an unexpected error." + " " + CANCEL_FURTHER_PROCESSING_MSG
+        CANCELLED_ERROR_MSG = "Future of the revious message parser cancelled. The server is probably shutting down. " + " " + CANCEL_FURTHER_PROCESSING_MSG
```
jmakar-s1, line 1075: s/revious/previous/

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1069,6 +1066,36 @@ def _syslog_request_parser(self):
...
+        PREVIOUS_CANCELEED_ERROR_MSG = "Previous message parser encountered an error. " + " " + CANCEL_FURTHER_PROCESSING_MSG
```
jmakar-s1, line 1076: s/CANCELEED/CANCELLED/

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -1084,28 +1111,25 @@ def handle(self):
...
-                    self.__request_data_process(syslog_parser, data)
+                try:
+                    for data in self.__request_stream_read(syslog_request, self.server.is_running):
+                        syslog_parser.process(data)
+                except Exception as e:
+                    global_log.warning(
+                        "Error processing request: %s\n\t%s",
+                        six.text_type(e),
+                        traceback.format_exc(),
+                    )
```
jmakar-s1, line 1132: The try/except should probably be inside the for loop (one bad iteration shouldn't trash the whole batch)

scalyr_agent/builtin_monitors/thread_pool.py
```diff
@@ -20,23 +20,38 @@
...
 
 # A MixIn class used for adding a thread poll processing to a BaseServer (i.e. SyslogTCPServer, SyslogUDPServer)
 class ExecutorMixIn:
     def __init__(self, global_config):
-        # Using 4 threads for processing requests, because of GIL and CPU bound tasks higher number would not help process the data faster.
-        reading_threads = 4
+        # Using 8 threads for processing requests, because of GIL and CPU bound tasks higher number would not help process the data faster.
+        # The reason for multiple threads is to avoid deadlock in an unexpected situation before waiting for previous data
+        # on a single request being processed fails and it's not caught do to a bug.
+        processing_threads = 8
```
jmakar-s1, line 33: This shouldn't be hardcoded, it makes more sense for it to be some fraction/percentage of the number of available cores

tests/unit/builtin_monitors/syslog_monitor_test.py
```diff
@@ -156,6 +159,236 @@ def assertNoException(self, func):
...
+                server.server_close()
+
+                for _ in range(100):
+                    time.sleep(0.1)
+                    if len(server.syslog_handler.logged_data) == CONNECTIONS * MESSAGES_PER_CONNECTION:
+                        break
+
+                assert len(server.syslog_handler.logged_data) == CONNECTIONS * MESSAGES_PER_CONNECTION
+
+
```
jmakar-s1, line 285: Could you look into these test failures?

jmakar-s1 (approved):
Some minor things to clean up but look ok to me.
(I personally would have preferred to fix the queue/thread approach but if using futures is easier/cleaner then that's works too)

scalyr_agent/builtin_monitors/thread_pool.py
alesnovak-s1: The reason for hardcoding it is that it wouldn't help to have more threads since these tasks are CPU bound and because of GIL it wouldn't make it faster. I have set more threads for redundancy only. And it's also configurable.

scalyr_agent/builtin_monitors/syslog_monitor.py
alesnovak-s1: I wasn't sure about that. But it wouldn't hurt to try to process even partial data.

scalyr_agent/builtin_monitors/thread_pool.py
```diff
@@ -63,24 +78,25 @@ def process_request(self, request, client_address):
...
-                global_log.error("Failed shutting down the ThreadPoolExecutor %s", executor, exc_info=e)
+        self.__wait_for_tasks_to_complete(self._request_reading_executor, self._request_processing_executor, timeout=self._shutdown_grace_period)
+
+        try:
+            self._request_reading_executor.shutdown(wait=False, cancel_futures=True)
+            self._request_processing_executor.shutdown(wait=False, cancel_futures=True)
+        except TypeError:
+            # Older versions of python do not support cancel_futures
+            self._request_reading_executor.shutdown(wait=False)
+            self._request_processing_executor.shutdown(wait=False)
```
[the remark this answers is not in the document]
alesnovak-s1: An elegant solution, thanks.

PR #1223, opened 2023-12-07
DTIN-3315

alesnovak-s1:
https://sentinelone.atlassian.net/browse/DTIN-3315

CRI logs - partial lines support.

Partial and final lines are marked by P and F tags respectively.
Note that the timestamp of the matching message parts do not need to match and both stdout and stderr are mixed together.

Sample:

```
2016-10-06T00:17:09.669794202Z stdout F The content of the stdout log entry 1
2016-10-06T00:17:09.669794202Z stderr F The content of the stderr log entry 1
2016-10-06T00:17:09.669794202Z stdout P First line of stdout log entry 2
2016-10-06T00:17:09.669794202Z stderr F The content of the stderr log entry 2
2016-10-06T00:17:09.669794202Z stdout F Second line of the stdout log entry 2
```

Resulting in:
stdout:
```
The content of the stdout log entry 1
First line of stdout log entry 2 Second line of the stdout log entry 2
```

Resulting in:
stderr:
```
The content of the stderr log entry 1
The content of the stderr log entry 2
```

### DIFF
--- a/.github/workflows/reusable-agent-build-container-images.yml
+++ b/.github/workflows/reusable-agent-build-container-images.yml
@@ -349,24 +349,29 @@ jobs:
           fi
 
       # Here we build the dummy container which continuously prints data to stdout and stderr
-      - name: Build Dummy App Docker Image
+      - name: Build Dummy App Docker Images
         run: |
           docker build -f docker/Dockerfile.docker_monitor_testing_config -t std-printer scripts/
+          docker build -f docker/Dockerfile.long_message_printer -t long-message-printer scripts/
           docker image ls
 
           # Needed for containerd or cri-o runtime
           if [ "${{ matrix.k8s_version.runtime }}" = "containerd" ] || [ "${{ matrix.k8s_version.runtime }}" = "cri-o" ]; then
-            docker tag std-printer:latest docker.io/library/std-printer:latest
-            minikube image load docker.io/library/std-printer:latest
+            for IMAGE in std-printer long-message-printer; do
+              docker tag ${IMAGE}:latest docker.io/library/${IMAGE}:latest
+              minikube image load docker.io/library/${IMAGE}:latest
+            done
           fi
 
       # Create pod for our mock std printer container which logs will be ingested by the agent
-      - name: Create mock pod
+      - name: Create mock pods
         run: |
           minikube image ls
           kubectl apply -f tests/e2e/k8s_k8s_monitor/std_printer_deployment.yaml
+          kubectl apply -f tests/e2e/k8s_k8s_monitor/long_message_printer_deployment.yaml
           
           kubectl wait --for=condition=ready pod -l app=std-printer
+          kubectl wait --for=condition=ready pod -l app=long-message-printer
           kubectl get pods -A
 
           export APP_POD_NAME=$(kubectl get pod --namespace=default --selector=app=std-printer -o jsonpath="{.items[0].metadata.name}")
@@ -385,6 +390,7 @@ jobs:
           # We need to use a different file for different K8s versions due to promotion (removal of
           # old v1beta alias) from v1beta to v1 in v1.25.0. v1 has been available since v1.21.0
           kubectl apply -f tests/e2e/k8s_events_monitor/cronjob_v1.yaml
+
           kubectl get cronjob -A
 
       - name: Create scalyr-agent-2 daemonset
@@ -428,6 +434,7 @@ jobs:
           # annotation set as part of the deployment YAML.
           # After a while, we change that dynamically using kubectl and verify that this change has
           # been correctly picked up by the agent.
+          echo Waiting 20s for some data to be ingested
           sleep 20
 
           echo "Initial pod ingested data checks"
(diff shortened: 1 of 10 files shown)


scalyr_agent/log_processing.py
```diff
@@ -88,45 +88,150 @@
...
+    @classmethod
+    def from_raw_line(cls, raw_line):
+        decoded_raw_line = raw_line.decode("utf-8", "replace")
+        timestamp, stream, tags, message, raw_timestamp = cls.__parse_cri_log(
+            decoded_raw_line
+        )
+
+        log_line = cls(timestamp, stream, tags, message, raw_timestamp)
+
+        if not log_line.is_partial and not log_line.is_full:
```
jmakar-s1, line 121: Should also verify that is_partial & is_full are not both set

scalyr_agent/log_processing.py
```diff
@@ -88,45 +88,150 @@
...
-    # skip over the ' ' delimiter and parse the stream type
-    line = line[index + 1 :]
-    index = line.find(" ")
-    if index < 0:
-        return None, None, None, None
+        # parse the timestamp
+        timestamp = rfc3339_to_nanoseconds_since_epoch(parts[0])
+        if timestamp is None:
+            raise CRIParseError("Could not parse timestamp", line)
+        raw_timestamp = timestamp
```
jmakar-s1, line 141: A minor thing, it might be a little clearer to rename raw_timestamp to unixtime or epochtime

scalyr_agent/log_processing.py
```diff
@@ -88,45 +88,150 @@
...
+        return any(
+            self.__buffer_completed(buffer, current_time)
+            for buffer in self.__buffers.values()
+        )
+
+    def pop_completed(self, current_time):
+        for stream, buffer in self.__buffers.items():
+            if self.__buffer_completed(buffer, current_time):
+                del self.__buffers[stream]
+                return buffer
```
jmakar-s1, line 234: Something to consider, could take a different approach where append_log_line returns the buffer if it was just completed (otherwise it returns None)

This would save the trouble of requiring explicit, separate check

jmakar-s1 (approved):
LGTM, just had minor suggestions

.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -463,6 +470,29 @@ jobs:
...
+          echo 'Looking for mixed streams'
+          RETRY_ATTEMPTS=1 ./scripts/cicd/scalyr-query.sh '$serverHost="'${SCALYR_AGENT_POD_NAME}'" app="std-printer" parser="test-parser-1" stream="stdout" "stderr"' && echo -e $MIXED_STREAMS_ERR_MSG && exit 1 || echo -e $MIXED_STREAMS_SUCCESS_MSG
+          RETRY_ATTEMPTS=1 ./scripts/cicd/scalyr-query.sh '$serverHost="'${SCALYR_AGENT_POD_NAME}'" app="std-printer" parser="test-parser-1" stream="stderr" "stdout"' && echo -e $MIXED_STREAMS_ERR_MSG && exit 1 || echo -e $MIXED_STREAMS_SUCCESS_MSG
+          
+          echo "Wait for BEGIN_LONG_MESSAGE .* END_LONG_MESSAGE lines. They should not be splitted."
+          MINIMUM_RESULTS=10 ./scripts/cicd/scalyr-query.sh '$serverHost="'${SCALYR_AGENT_POD_NAME}'" app="long-message-printer" parser="test-parser-2" stream="stdout" "stdout_BEGIN_LONG_MESSAGE"'
+          MINIMUM_RESULTS=10 ./scripts/cicd/scalyr-query.sh '$serverHost="'${SCALYR_AGENT_POD_NAME}'" app="long-message-printer" parser="test-parser-2" stream="stderr" "stderr_BEGIN_LONG_MESSAGE"'
+          echo "Looking for partial lines."
```
tomaz-s1, line 487: Thanks for adding those new end to end tests.

scalyr_agent/log_processing.py
```diff
@@ -88,45 +88,150 @@
...
 
-    # skip over the ' ' delimiter and parse the stream type
-    line = line[index + 1 :]
-    index = line.find(" ")
-    if index < 0:
-        return None, None, None, None
+        # parse the timestamp
+        timestamp = rfc3339_to_nanoseconds_since_epoch(parts[0])
+        if timestamp is None:
+            raise CRIParseError("Could not parse timestamp", line)
```
tomaz-s1, line 140: Do we know when this could happen? Invalid timestamp line or broken timestamp logic? In case of the latter, would it, in some cases, make sense to default to current timestamp in situations like that?

scalyr_agent/log_processing.py
```diff
@@ -88,45 +88,150 @@
...
+    @property
+    def is_partial(self):
+        return self.__is_partial
+
+    def payload_length(self):
+        return sum(map(len, self.__lines))
+
+    def build_cri_result(self, include_raw_timestamp_field):
+        def strip_new_line(s):
+            if s and s[-1] == "\n":
```
tomaz-s1, line 190: Could the lines also contains carriage retention character `\r` (e.g. on Window host)? And if so, do we need / want to strip it?

scalyr_agent/log_processing.py
```diff
@@ -88,45 +88,150 @@
...
 
-    # parse the timestamp
-    timestamp = rfc3339_to_nanoseconds_since_epoch(line[:index])
-    if timestamp is None:
-        return None, None, None, None
+    @staticmethod
+    def __parse_cri_log(line):
+        """
+        Parse a log line that uses the K8S Container Runtime Interface (CRI) format
+        https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/kubelet-cri-logging.md
```
tomaz-s1, line 130: It appears that link doesn't work anymore, but this one does - https://github.com/kubernetes/design-proposals-archive/blob/main/node/kubelet-cri-logging.md

scalyr_agent/log_processing.py
```diff
@@ -88,45 +88,150 @@
...
+        if stream not in self.__buffers:
+            self.__buffers[stream] = CRILogLineBuffer(current_time)
+
+        self.__buffers[stream].append(log_line)
+
+    def __buffer_completed(self, buffer, current_time):
+        return (
+                not buffer.is_partial
+                or current_time - buffer.first_line_time > self.__line_completion_wait_time
+                or buffer.payload_length() >= self.__max_line_size
```
tomaz-s1, line 221: Do we today, some how indicate is a line has been stripped due to `max_line_size` config option? E.g. by appending `...` to the end of the line or similar.

I know we do that on the server side, but not sure of about the client / agent side.

scalyr_agent/log_processing.py
alesnovak-s1: I am not sure if I understand. They will be always set to True or False.
This line is to check the validity of the raw_line.

scalyr_agent/log_processing.py
alesnovak-s1: I like this approach.

scalyr_agent/log_processing.py
alesnovak-s1: Invalid timestamp format. In this case I've copied the code that was used before and changed the handling to use the exception. My reasoning was that it's been proved to be working in a desired way in production and I didn't want to change it.

scalyr_agent/log_processing.py
alesnovak-s1: Thanks!

scalyr_agent/log_processing.py
alesnovak-s1: The line is not stripped but split, the rest is sent in the next message.
Do you think it would make sense to indicate that somehow?
i.e.:
message_part 1 ....
... message_part 2

How is this handled on the server side?

tomaz-s1: On the server, we just truncate the field value and append `...` to the values which are over the server side account level defined limit (so we don't split it across multiple lines).

In case we split it and not truncate it, I think it's fine to keep that logic as-is.

tests/unit/log_processing_test.py
```diff
@@ -270,6 +340,179 @@ def test_halt_with_matcher(self):
...
+            {"path": self.__path, "parse_format": "cri"}
+        )
+        self.scan_for_new_bytes()
+
+        lines = [
+            b"2023-12-20T16:18:41.129027195+00:00 stdout F stdout: line 1\n",
+            b"2023-12-20T16:18:41.129027195+00:00 stdout P BEGIN PARTIAL_MESSAGE\n",
+            b"2023-12-20T16:18:41.129027195+00:00 stdout P -MIDDLE OF PARTIAL_MESSAGE\n",
+            b"2023-12-20T16:18:41.129069776+00:00 stderr F stderr: line 1\n",
+            b"2023-12-20T16:18:41.129069776+00:00 stderr P BEGIN PARTIAL_MESSAGE\n",
```
tomaz-s1, line 423: In case partial message is split across multiple lines, we use timestamp of the first partial message, right?

scalyr_agent/log_processing.py
```diff
@@ -88,45 +88,150 @@
...
+
+        result = LogLine(lines_string.encode("utf-8"))
+        result.timestamp = self.__timestamp
+        result.attrs = {"stream": self.__stream}
+        if include_raw_timestamp_field:
+            result.attrs["raw_timestamp"] = self.__raw_timestamp
+
+        return result
+
+class CRIStreamBuffers(object):
```
tomaz-s1, line 204: Since it's a new code, it may be a good idea to add type annotations (at least via comments, in case code still needs to be Python 2 compatible).

tomaz-s1 (approved):
Small comment / questions on carriage return characters, besides that, LGTM.

scalyr_agent/log_processing.py
alesnovak-s1: Yes, I will add it there.

tests/unit/log_processing_test.py
alesnovak-s1: Yes.

scalyr_agent/log_processing.py
jmakar-s1: I meant it should also check that both is_partial & is_full are not both true, ie:
```
if (not log_line.is_partial and not log_line.is_full) or (log_line.is_partial and log_file.is_full):
   ...
```

alesnovak-s1: is_partial and is_full will both be always set int the consturctor:
```
self.is_partial = "P" in tags if tags else False
self.is_full = "F" in tags if tags else False
```

alesnovak-s1:
@jmakar-s1 @tomaz-s1 
Thank you for the review!

PR #1225, opened 2023-12-15
Optimizing and stabilization of K8s related GHAs

alesnovak-s1:
* Making the K8s related GHAs a bit faster and more stable
* Adding CRI-O to the Kubernetes Monitor and OpenMetrics Monitor tests

### DIFF
--- a/.github/workflows/reusable-agent-build-container-images.yml
+++ b/.github/workflows/reusable-agent-build-container-images.yml
@@ -278,6 +278,8 @@ jobs:
           - { "version": "v1.24.14", "driver": "docker", "runtime": "containerd", "run_cronjob_checks": true }
           - { "version": "v1.25.10", "driver": "docker", "runtime": "containerd", "run_cronjob_checks": true }
           - { "version": "v1.26.5", "driver": "docker", "runtime": "containerd", "run_cronjob_checks": true }
+          # CRI-O
+          - { "version": "v1.27.4", "driver": "docker", "runtime": "cri-o", "run_cronjob_checks": true }
 
     steps:
       - name: Checkout Repository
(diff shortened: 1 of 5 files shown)


.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -474,6 +477,10 @@ jobs:
         run: |
           export RETRY_ATTEMPTS="14"
           export SLEEP_DELAY="10"
+          
+          # Wait for at least one hello-printer job to be scheduled and completed. 
+          RETRIES=1; 
+          until echo -n "Waiting for first job to be completed - retry $RETRIES: " && kubectl wait --for=condition=complete job -l app=hello-printer || [ $RETRIES -gt 60 ]; do sleep 1; RETRIES=$(($RETRIES+1)); done
```
tomaz-s1, line 483: So it can take up to 1 minute for one pod to become scheduled and ready?

tomaz-s1 (approved):
LGTM, thanks for adding those improvements.

.github/workflows/reusable-agent-build-container-images.yml
jmakar-s1, line 483: If the retry limit is hit this won't cause the job to fail.  Instead of a loop couldn't the `kubectl wait` just have 60 sec timeout?

jmakar-s1 (approved):
Thanks for these changes!

.github/workflows/reusable-agent-build-container-images.yml
alesnovak-s1: We are waiting for a cronjob to start a first job.
The smallest scheduled cronjob time is "every minute".

alesnovak-s1: Good point. I will add a check and make it fail.
I can't use kubectl wait here, because I am waiting for a job to complete and that job might not even be created yet.
Unfortunatelly there's not better way to do this.

alesnovak-s1:
Thank you @jmakar-s1 , @tomaz-s1 for the review.

PR #1229, opened 2024-01-04
Open Telemetry Transport

ansoniS1:
WIP

### DIFF
--- a/scalyr_agent/client_auth.py
+++ b/scalyr_agent/client_auth.py
@@ -1,0 +1,79 @@
+import datetime
+import json
+import urllib
+from base64 import b64encode
+
+from scalyr_agent.connection import ConnectionFactory
+from urllib.parse import urlparse
+
+from scalyr_agent.third_party import requests
+
+
+class ClientAuth(object):
+    def __init__(self, configuration, headers):
+        self.configuration = configuration
+        self.headers = headers
+        if self.configuration.auth == "oauth2":
+            self.auth = OAuth2(self.configuration, self.headers)
+        elif self.configuration.auth == "bearer":
+            self.auth = BearerToken(self.configuration, self.headers)
+        else:
+            self.auth = NoAuth(self.configuration, self.headers)
+
+    def authenticate(self):
+        return self.auth.authenticate()
+
+class NoAuth(object):
+    def __init__(self, configuration, headers):
+        self.headers = headers
+
+    def authenticate(self):
+        # Add a header just so for traceability
+        self.headers["x-no-auth"]="true"
+        return True
+
+# Simple Authorization Header as Bearer Token using `api_key` configuration
+class BearerToken(object):
+    def __init__(self, configuration, headers):
+        headers.set("Authorization", "Bearer " + configuration.api_key)
+
+    def authenticate(self):
+        return True
+
+# Implement https://datatracker.ietf.org/doc/html/rfc6749#section-4.4 flow
+class OAuth2(object):
+    def __init__(self, configuration, headers):
+        self.headers = headers # Headers modified for external requests
+        self.client_id = configuration.client_id
+        self.client_secret = configuration.client_secret
+        self.token_url = configuration.token_url
+        scopes = " ".join(configuration.scopes)
+        # Payload Body for the token exchange
+        self.auth_request = "grant_type=client_credentials&scope=" + urllib.parse.quote_plus(scopes)
+        # Our token to use in requests
+        self.token = None
+        # When the token expires
+        self.expiry_time = datetime.datetime.now()
+        self.verify_ssl = configuration.verify_server_certificate
+        # Authentication Headers for the token exchange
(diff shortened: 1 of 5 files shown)


scalyr_agent/client_auth.py
```diff
@@ -0,0 +1,78 @@
...
+# Simple Authorization Header as Bearer Token using `api_key` configuration
+class BearerToken(object):
+    def __init__(self, configuration, headers):
+        headers.set("Authorization", "Bearer " + configuration.api_key)
+
+    def authenticate(self):
+        return True
+
+# Implement https://datatracker.ietf.org/doc/html/rfc6749#section-4.4 flow
+class OAuth2(object):
```
alesnovak-s1, line 43: Isn't there a python library for OAuth2 to avoid having an in-house implementation?

scalyr_agent/configuration.py
```diff
@@ -1497,6 +1499,41 @@ def api_key(self):
...
+        """Returns the configuration value for 'server_url'."""
+        return self.__get_config().get_string("server_url")
+
+    @property
+    def auth(self):
+        """Returns the configuration value for 'auth'."""
+        return self.__get_config().get_string("auth")
+
+    @property
+    def client_id(self):
```
alesnovak-s1, line 1513: Please move all oauth2 configuration keys under a new oauth2 key:
oauth2:
->> client_id:
->> client_secret:
  ....

scalyr_agent/copying_manager/worker.py
```diff
@@ -354,6 +355,7 @@ def __init__(self, configuration, worker_config_entry, session_id, is_daemon=Fal
 
         # The current pending AddEventsTask.  We will retry the contained AddEventsRequest several times.
         self.__pending_add_events_task = None
+        self.__receive_response_status = {}
```
alesnovak-s1, line 358: Where is this private field used?

alesnovak-s1:
Cover the code with unit tests and implement an integration test that will use the OTLP endpoint.

alesnovak-s1 (changes requested):
I've made few comments and will continue going through the otlp code.

scalyr_agent/client_auth.py
ansoni: Python2 requirement paints us into some really weird corners like having to be on older versions of code that opens up security risks. This seems like the path of least resistance. I originally used the OpenTelemetry Python SDK, but had to back out of that because of no Python2 support.

scalyr_agent/copying_manager/worker.py
ansoni: https://github.com/scalyr/scalyr-agent-2/blob/d9c9779fb46bb52b4bebd0b02cae37af83c966b6/scalyr_agent/copying_manager/worker.py#L494

Trying to resolve this error (it doesn't):

```
2024-01-26 17:17:51.891Z ERROR [core] [scalyr_agent.scalyr_logging:707] Failed while attempting to scan and transmit logs :stack_trace:
    File "/Users/anthonyj/_Scalyr/scalyr-agent-2/scalyr_agent/copying_manager/worker.py", line 500, in run
      "Repeatedly failed to parse response due to exception.  Dropping events",
  AttributeError: 'AddEventsTask' object has no attribute '_CopyingManagerWorkerSession__receive_response_status'
```

I defined the attribute in hopes that this was just a sequencing issue in the way the `scalyr_client` works, but this didn't fix it. Happens when we start to hit publishing delays. Easiest way for me to hit it is to Debug Agent;Pause Execution for X minutes;Resume Agent

Since this is experimental, I'm comfortable shipping this bug to the customer so they can provide feedback on the feature.

alesnovak-s1: I meant the second one on the line 358. I believe that's a mistake.

scalyr_agent/otlp_client.py
```diff
@@ -0,0 +1,463 @@
...
+            sl.log_records.append(log_record)
+        rl.scope_logs.append(sl)
+        return LogsData(resource_logs=[rl]).SerializeToString()
+
+    def __event_to_log_record(self, event):
+        return LogRecord(time_unix_nano=event.timestamp,
+                         body=PB2AnyValue(string_value=event.message),
+                         attributes=_encode_attributes(event.attrs))
+
+    def __get_user_agent(
```
alesnovak-s1, line 217: This whole method is copy pasted from scalyr_client.py.

alesnovak-s1 (changes requested):

scalyr_agent/otlp_client.py
ansoni: yes. This was intentional to avoid touching `scalyr_client.py`. I went ahead and moved this method and the wrap method to `util.py`

scalyr_agent/configuration.py
ansoni: I went with `oauth_` prefix instead which goes with the current pattern used for kubernetes keys (`k8s_`)

PR #1230, opened 2024-01-08
Python2 fixes

alesnovak-s1:
Removing unused queue import, which is not supported in python2. 
concurrent.futures imported only for PY3

### DIFF
--- a/scalyr_agent/builtin_monitors/syslog_monitor.py
+++ b/scalyr_agent/builtin_monitors/syslog_monitor.py
@@ -25,8 +25,6 @@ from __future__ import absolute_import
 
 from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


tomaz-s1 (approved):

tomaz-s1:
Do we have some kind of CI check which catches issues like that?

alesnovak-s1:
> Do we have some kind of CI check which catches issues like that?

Not really. The only place I know of where the python2 is used for testing is one of the E2E tests running on amazonlinux2.
Unit tests run on python3 as well.
I was considering adding python2 to the unit tests matrix, but I think since it will be deprecated soon, it's not worth the effort.

What do you think?

PR #1238, opened 2024-01-28
Support sending events to multiple Dataset accounts via K8s annotations

alesnovak-s1:
Support for multiple account per container.
Doc - https://github.com/scalyr/scalyr-agent-2/blob/DTIN-3339_many_workers_per_log_file/docs/monitors/kubernetes_monitor.md#configuring-multiple-accounts-per-container

https://sentinelone.atlassian.net/browse/DTIN-3339


.circleci/smoketest_docker.sh
```diff
@@ -145,6 +145,12 @@ echo "Uploader container ID == ${uploader_hostname}"
 echo "Using smoketest.py script from ${SMOKE_TESTS_SCRIPT_BRANCH} branch and URL ${SMOKE_TESTS_SCRIPT_URL}"
 
 function print_debugging_info_on_exit() {
+    echo ""
```
jmakar-s1, line 148: If there isn't one already, could you make a ticket to move everything from CircleCI to Github Actions

docs/monitors/kubernetes_monitor.md
```diff
@@ -30,6 +30,26 @@ fields behave differently when configured via k8s annotations:
 * lineGroupers (not supported at all)
 * path (the path is always fixed for k8s container logs)
 
+### Configuring multiple accounts per container
+
+Containers can be assigned one or more Scalyr API keys by specifying the `log.config.scalyr.com/{container_name}.teams.{team_number}.secret` annotation.  
+The `container_name` is the name of the container as specified in the pod spec, and the `team_number` is a unique number for each team.  
+The `secret` is the name of the secret holding the Scalyr API key for the team.  
+A default fallback api key can be specified by using the `log.config.scalyr.com/teams.{team_number}.secret` annotation.
+The monitor first looks into the container specific annotation, and if not found, it falls back to the default annotation or the global configuration api key.
```
jmakar-s1, line 39: Was this extra default a requirement?  To me it makes more sense to just use the global write key as the default.

If this is retained, why use `log.config.scalyr.com/teams.{team_number}.secret` as the name?  Why not just `log.config.scalyr.com/teams.secret`?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -3409,9 +3451,76 @@ def __get_log_config_for_container(self, cid, info, k8s_cache, base_attributes):
...
+        # Based on the pod annotations in a format {container_name}.{team}.secret={secret_name}
+        # we might want to add api_keys parameter
+        if container_annotations or all_annotations:
+            api_keys = self.__container_api_keys_from_annotations(k8s_cache, pod_namespace, pod_name, container_annotations, all_annotations)
+
+            if api_keys:
+                results = [
+                    {**result, "api_key": api_key}
+                    for api_key in list(filter(lambda api_key: api_key is not None, api_keys))
+                ]
```
jmakar-s1, line 3465: This is a little subtle so I would consider adding a comment here about this, something like "Multiple matching api keys will result in multiple log configs"

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -3409,9 +3451,76 @@ def __get_log_config_for_container(self, cid, info, k8s_cache, base_attributes):
...
+            ]
+
+        api_keys = [
+            get_secret_api_key(
+                fetch_secret(
+                    secret_name
+                )
+            )
+            for secret_name in secret_names
+        ]
```
jmakar-s1, line 3518: I think it's better to filter out api_key = None here rather than in line 3464

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2627,10 +2659,14 @@ def __init__(
         self.__controlled_warmer = controlled_warmer
 
         # give this an initial empty value
-        self.raw_logs = []
+        self.raw_logs = [] # type: List[DockerLog]
+        self.docker_logs = [] # type: List[DockerLog]
```
jmakar-s1, line 2663: I may have missed it but what was the purpose of having a separate docker_logs field?

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -72,6 +74,62 @@
...
+                    self.__workers_by_id[worker.worker_id] = worker
+                    worker.start_sessions_and_block_until_copying()
+
+        return self.__workers_by_api_key[api_key]
+
+    def values(self):
+        return self.__workers_by_id.values()
+
+    def items(self):
+        return self.__workers_by_id.items()
```
jmakar-s1, line 114: With the GIL I think this is ok but as I understand the GIL only applies to Cython-based implementations.  To be on the safe side it might be a good idea to protect these accesses with the create_worker_lock (probably should rename it to worker_lock as well)

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -312,6 +370,74 @@ def generate_status(self, warn_on_rate_limit=False):
...
+    def set(self, path, worker_id, value):
+        # type: (six.text_type, six.text_type, object) -> None
+        self.__paths[path][worker_id] = value
+
+    def get_path(self, path):
+        return self.__paths[path].items()
+
+    def get(self, path, worker_id, default=None):
+        # type: (six.text_type, six.text_type, Optional[object]) -> Optional[object]
+        return self.__paths[path].get(worker_id, default)
```
jmakar-s1, line 394: May want `self.__paths.get(path, {}).get(worker_id, default)`

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -312,6 +370,74 @@ def generate_status(self, warn_on_rate_limit=False):
...
+        """
+        return [
+            (path, worker_id)
+            for worker_id, _ in self.get_path(path)
+            if worker_id not in worker_ids
+        ]
+
+    def contains(self, path, worker_id):
+        # type: (six.text_type, six.text_type) -> bool
+        return worker_id in self.__paths[path]
```
jmakar-s1, line 409: Same may want `self.__paths.get(path)` here

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -312,6 +370,74 @@ def generate_status(self, warn_on_rate_limit=False):
...
+
+    def copy(self):
+        # type: () -> PathWorkerIdDict
+        result = PathWorkerIdDict()
+        result.__paths=self.__paths.copy()
```
jmakar-s1, line 414: Is the shallow copy intentional?  Or want a deepcopy() instead?

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -312,6 +370,74 @@ def generate_status(self, warn_on_rate_limit=False):
         return result
 
 
+class PathWorkerIdDict(object):
```
jmakar-s1, line 373: This is a good abstraction for handling a log file that is managed multiple times!

scripts/cicd/verify-multiaccount-records-have-been-ingested.sh
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env bash
```
jmakar-s1, line 1: Just /bin/bash is fine, env is appropriate for python / node / etc that generally are custom-installed

jmakar-s1 (approved):
Looks good, just had some misc comments

docs/monitors/kubernetes_monitor.md
alesnovak-s1: Yes, this was a requirement. And {team_number} is there to support multiple default keys. 
The way I understand it is that you can have default keys specifically for a single pod. But I don't know the use case.

.circleci/smoketest_docker.sh
alesnovak-s1: Good Idea. I will. CircleCI is not used anymore, just the name remains in the directory structure. Agent Build Refactored, dev-requirements-new.txt and so on will be good candidates for renaming to include in this ticket :)

scalyr_agent/builtin_monitors/kubernetes_monitor.py
alesnovak-s1: That's a good point. The only difference I can see is that raw_logs are already being monitored and docker_logs are those monitored and ready to be monitored.
Since it's a public field I didn't want to mess with it and kept the same logic, just used DockerLog instead of Dict make it more clear.
This monitor as a whole could be refactored to make it much less complex.

scalyr_agent/builtin_monitors/docker_monitor.py
```diff
@@ -1483,6 +1483,7 @@ def __create_docker_logger(self, log, last_request):
             name,
             stream,
             log["log_config"]["path"],
+            log["log_config"],
```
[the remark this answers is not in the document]
alesnovak-s1: I agree this seems redundant, but I want to make it clear where the path comes from. In this monitor class a lot of field are retrieved from a JsonObject and Dict somewhere underneath and in the end it make the code less readable and harder to reason about.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
alesnovak-s1: Definitely!

scalyr_agent/copying_manager/copying_manager.py
alesnovak-s1: Good idea.

scalyr_agent/copying_manager/copying_manager.py
alesnovak-s1: I have it covered by self.__paths being a defaultdict(dict).
Otherwise I would need to be using the default get value in complement_keys, set and pop as well. And this might make it a bit error prone.

scalyr_agent/copying_manager/copying_manager.py
alesnovak-s1: Custom deepcopy will be safer for future changes.

scalyr_agent/copying_manager/copying_manager.py
alesnovak-s1: Thanks! I was getting lost without that :)

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -72,6 +73,64 @@
...
+        return self.__workers_by_id[worker_id]
+
+    def get_or_create_worker(self, api_key, global_config):
+        # type: (str, Configuration) -> CopyingManagerWorker
+        """
+        Add worker.
+        """
+        if not api_key in self.__workers_by_api_key:
+            with self.__worker_lock:
+                if not api_key in self.__workers_by_api_key:
```
tomaz-s1, line 101: If conditional seems to be duplicated (outside and inside the lock)? Can we just remove the outside conditional and only have one (with the lock)?

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1614,6 +1647,40 @@ def update_cache(self, run_state):
...
+
+        result = self._secrets_cache.lookup(
+            local_state.k8s,
+            current_time,
+            namespace,
+            name,
+            kind="Secret",
+            allow_expired=allow_expired
+        )
+        global_log.info("Secret %s/%s: %s" % (namespace, name, result.data))
```
tomaz-s1, line 1681: `result.data` will contain the actual secret value, right?

If that's indeed the case, we probably don't want to log it at all.

alesnovak-s1: Definitely, I've removed that.

scalyr_agent/copying_manager/copying_manager.py
alesnovak-s1: We need both to avoid race condition in case the worker creation started between lines 99 and 100 in another thread.

PR #1250, opened 2024-03-19
Short lived pods included by SCALYR_K8S_INCLUDE_ALL_CONTAINERS

alesnovak-s1:


scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2326,6 +2329,11 @@ def _get_containers(
 
                         k8s_info["pod_info"] = pod
                         k8s_info["pod_uid"] = pod.uid
+                    else:
+                        # Pod not found by the API, including based on SCALYR_K8S_INCLUDE_ALL_CONTAINERS
+                        if not k8s_include_by_default:
```
tomaz-s1, line 2334: Do we want to log something here, e.g. Pod foo is included because `SCALYR_K8S_INCLUDE_ALL_CONTAINERS=true`.

I guess that if we do we probably want to use rate limited info / debug to avoid the overhead since I could imagine we could end up in this branch often.

tests/unit/builtin_monitors/kubernetes_monitor_test.py
```diff
@@ -77,6 +77,17 @@
 from six.moves import range
 
 
+class StringMatcher():
```
tomaz-s1, line 80: Do we need to subclass this from `str`?

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2281,14 +2281,17 @@ def _get_containers(
...
                         # Exclude the pod
                         if e.status_code == 404:
-                            global_log.warning(
-                                "Pod %s/%s not found in K8s API. Excluding pod from collection."
-                                % (pod_namespace, pod_name),
-                                exc_info=e,
+                            if k8s_include_by_default:
+                                include_message = "Including pod based on SCALYR_K8S_INCLUDE_ALL_CONTAINERS=true."
+                            else:
+                                include_message = "Excluding pod based on SCALYR_K8S_INCLUDE_ALL_CONTAINERS=false."
```
tomaz-s1, line 2291: Just to make sure I understand it correctly, specific (short lived) container will be included if it was already in cache, even if we had `exclude: true` annotation on the pod / container, right (since it may already disappear from the API)?

If that is indeed the case, it may be a good idea to document this somewhere.

And if that is not the case, we should have an end to end test case which exercises that (pod / container which has `exclude: true` annotation and verifying that those logs are not included).

I also wonder if there could be some yet undetermined negative consequences of this approach - e.g. it may result in more pod / container logs to be included under some edge cases and not just short lived ones.

Perhaps it would be a good idea to deploy it to some larger testing environment with `SCALYR_K8S_INCLUDE_ALL_CONTAINERS=true` and some pods having set `exlcude: true` and observe how it behaves.

tests/unit/builtin_monitors/kubernetes_monitor_test.py
alesnovak-s1: In this case I would rather expect implementing an interface like Matcher in hamcrest in Java.
But with python's ducktyping approach, you just need a class implementing __eq__ :)

scalyr_agent/builtin_monitors/kubernetes_monitor.py
alesnovak-s1: Yes, the requirement was to include it even though it might have been annotated by "exclude: true".
The idea behind it is to be on the safe side and rather log something unnecessarily than lose something you need.
Long term testing it is a good idea. In this case I am reverting to a previous state of things and need to release it soon, but my plan is to deploy and test any change for at least a week to a testing or staging env. I will include some of the already release, including this one.
And I will write the doc.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
alesnovak-s1: It's logged above:
` if k8s_include_by_default:
	include_message = "Including pod based on SCALYR_K8S_INCLUDE_ALL_CONTAINERS=true."
else:
	include_message = "Excluding pod based on SCALYR_K8S_INCLUDE_ALL_CONTAINERS=false."
`

I know it's far from obvious and not very readable, but I didn't want to break the code structure or refactor it.

I've added a limit key based on pod name.

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -2281,14 +2281,19 @@ def _get_containers(
...
+                            global_log.info(
+                                "Pod %s/%s not found in K8s API. %s"
+                                % (pod_namespace, pod_name, include_message),
+                                limit_once_per_x_secs=300,
+                                limit_key=pod_name,
```
tomaz-s1, line 2296: Limit keys needs to to be globally unique so to avoid collisions / conflicts, it would be better to call it something along the lines of `k8s-api-pod-not-found-{pod_name}`.

Having said that, we should also confirm with the logger limit implementation to make sure those entries eventually get purged otherwise we could have problems with memory usage since this is a high cardinality value. If they don't, we may need to use a different key.

tomaz-s1 (approved):

scalyr_agent/builtin_monitors/kubernetes_monitor.py
alesnovak-s1: Removing the logging key after discussion with @tomaz-s1 , because of the logging framework not implementing eviction of old timestamp records.

PR #1256, opened 2024-04-12
Multi Dataset Account ChR

alesnovak-s1:


scalyr_agent/monitor_utils/k8s.py
```diff
@@ -1320,6 +1401,9 @@ def __init__(
         self._pod_processor = PodProcessor(self._controllers)
         self._pods_cache = _K8sCache(self._pod_processor, "Pod")
 
+        # create the namespace cache
+        self._namespace_cache = _K8sCache(NamespaceProcessor(self._controllers), "Namespace")
```
tomaz-s1, line 1405: Any idea how also caching the namespace information affects the agent memory consumption on larger clusters?

I imagine that overhead is not so big since there will usually be much less namespaces than Pods (which we already cache).

scripts/cicd/verify-multiaccount-records-have-been-ingested.py
```diff
@@ -0,0 +1,298 @@
+#!/usr/bin/env -S python3 -u
```
tomaz-s1, line 1: That's for moving this logic to a Python script. It's much more readable, easier to maintain and we can also run lint, style and other static analysis checks on it...

scripts/cicd/verify-multiaccount-records-have-been-ingested.py
```diff
@@ -0,0 +1,298 @@
...
+    for account_name in ENV_ACCOUNT_NAME_MAPPING.values():
+        assert_env_non_empty(account_name)
+
+    assert_env_non_empty("SERVER_HOST")
+
+
+def query(token: str, server_host: str, time_start: str, retries: int):
+    def post_request():
+        return requests.post(
+            "https://app.scalyr.com/api/query",
```
tomaz-s1, line 67: Should we also set `timeout` parameter for the requests method (e.g. 20 seconds or something reasonable)

scalyr_agent/builtin_monitors/kubernetes_monitor.py
```diff
@@ -3044,26 +3048,29 @@ def check_containers(self, run_state):
...
                             scalyr_logging.DEBUG_LEVEL_1,
                             "Starting loggers for container '%s'" % info["name"],
                         )
                         starting[cid] = info
-                    elif cid in prev_digests:
+                    elif cid in prev_digests or cid in prev_digests_namespaces:
                         # container was running and it exists in the previous digest dict, so see if
                         # it has changed
-                        if pod and prev_digests[cid] != pod.digest:
+                        if pod and prev_digests[cid] != pod.digest or namespace and prev_digests_namespaces[cid] != namespace.digest:
```
tomaz-s1, line 3062: May want to add parenthesis around the whole `or` part of the if clause.

docs/monitors/kubernetes_monitor.md
```diff
@@ -32,23 +32,221 @@ fields behave differently when configured via k8s annotations:
...
-        log.config.scalyr.com/cont2.teams.1.secret: "scalyr-api-key-team-3"
-        log.config.scalyr.com/cont2.teams.2.secret: "scalyr-api-key-team-4"
-
-In this example
-* `cont1` would use the api key from `scalyr-api-key-team-3` secret
-* `cont2` would be sending the data to two teams accessible by api keys stored in `scalyr-api-key-team-3` and `scalyr-api-key-team-4` secrets.
-* `cont3` would use the default api key from `scalyr-api-key-team-2` secret
+Besides the default API key stored in the scalyr/scalyr-api-key secret, the user can specify API keys for namespaces, pods and containers using annotations.
+
+#### Namespace level API keys
```
tomaz-s1, line 37: To me it's pretty clear how annotation precedence works from this description, but some people visualize better with images so a perhaps a simple chart on annotation priority / precedence with an example may be in order (`container -> pod -> namespace -> default agent key`) :)

scalyr_agent/builtin_monitors/kubernetes_monitor.py
alesnovak-s1: Yes, that will make it more readable.

scalyr_agent/monitor_utils/k8s.py
alesnovak-s1: That was my thinking. The NamespaceInfo object holds only necessary parts of metadata about the object and gets purged when expired.

docs/monitors/kubernetes_monitor.md
alesnovak-s1: Good idea.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -877,6 +884,80 @@ def process_object(self, k8s, obj, query_options=None):
         raise NotImplementedError("process_object not implemented for _K8sProcessor")
 
 
+class NamespaceInfo():
+
+    @staticmethod
+    def __flatten_dict(dict_to_flatten):
```
jmakar-s1, line 890: Usually flattening a dict is recursive, is this only for the md5 calc below?  Probably better to use json.dumps or even just str.

scalyr_agent/monitor_utils/k8s.py
```diff
@@ -877,6 +884,80 @@ def process_object(self, k8s, obj, query_options=None):
...
+        flattened = []
+        for key in sorted(dict_to_flatten.keys()):
+            flattened.append(key)
+            flattened.append(six.text_type(dict_to_flatten[key]))
+
+        return "".join(flattened)
+    def __init__(self, name, uid, labels, annotations, spec):
+        self.name = name
+        self.namespace = name
```
jmakar-s1, line 899: Why have two attributes for name(space)?

jmakar-s1 (approved):
LGTM

scalyr_agent/monitor_utils/k8s.py
alesnovak-s1: Both attributes are needed elsewhere in our k8s library, which expects namespaced objects. This was the easiest way to ensure compatibility and no to break things.

scalyr_agent/monitor_utils/k8s.py
alesnovak-s1: It's for the md5 below only. Switching to json.dumps.

PR #1270, opened 2024-05-15
Syslog memory usage fix

alesnovak-s1:
After several attempts to make syslog monitor threading robust and memory efficient, I decided to implement a simple solution of bounded ThreadingMixin.
A better candidate might seem to be a ThreadPoolExecutor, but it proved to be very memory hungry.

The reason for using threads in the first place is:
* UDP - avoid loosing too many sockets
* TCP - avoid blocking on unfinished and not yet closed requests

The ideal solution would be to get right of our rigid in-house implementation based on BaseServer and etither
* use socket.poll() on all open sockets
or
* use async library for hansling sockets, i.e. tornado. 
But such a big refactoring is not a priority at the moment.


.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+  schedule:
+    - cron: '0 4 * * *'
```
jmakar-s1, line 11: This doesn't really need to be run on a schedule

scripts/cicd/test_syslog_monitor_memory_usage/run_test.sh
```diff
@@ -0,0 +1,105 @@
...
+  echo "{
+      \"module\":                    \"scalyr_agent.builtin_monitors.syslog_monitor\",
+      \"protocols\":                 \"udp:$(($N+1500))\",
+      \"accept_remote_connections\": \"true\",
+      \"message_log\": \"udp_$N.log\"
+    }," >> agent.json
+done
+
+echo "]" >> agent.json
+
```
jmakar-s1, line 58: Using a here doc might be easier for writing the agent.json file

scripts/cicd/test_syslog_monitor_memory_usage/run_test.sh
```diff
@@ -0,0 +1,105 @@
...
+for N in `seq $UDP_SERVERS`;
+do
+	loggen --rate $(($RATE/$UDP_SERVERS)) --interval $INTERVAL --dgram --active-connections=5 -Q localhost $(($N+1500)) &
+	pids+=($!)
+done
+
+PID=`cat ~/scalyr-agent-dev/log/agent.pid`
+
+while ps -p $pids > /dev/null; do
+  MEM_USED=`cat /proc/$PID/status | grep VmRSS | awk '{print $2}'`
```
jmakar-s1, line 89: Can just use `awk '/^VmRSS/ {print $2}' /proc/$PID/status`
I would doublecheck that the units are always in kB, I'm not sure if that's always true

scalyr_agent/builtin_monitors/syslog_monitor.py
```diff
@@ -636,6 +624,7 @@ def __init__(self, socket, max_buffer_size):
         self._socket = socket
         self._max_buffer_size = max_buffer_size
         self._socket.setblocking(True)
+        self._socket.settimeout(0.1)
```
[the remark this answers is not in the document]
alesnovak-s1: Line 638:
```
except socket.timeout as e:
    raise SocketNotReadyException(e)
```

Line 1008:
```
   except SocketNotReadyException as e:
                time.sleep(0.01)
                check_running = True
```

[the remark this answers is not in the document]
alesnovak-s1: I know far from elegant or efficient, but I just kept the solution that was there before instead of big refactorings.

scripts/cicd/test_syslog_monitor_memory_usage/run_test.sh
alesnovak-s1: Thanks, I will use that.
Luckily it's always in Kb.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/proc/task_mmu.c?id=39a8804455fb23f09157341d3ba7db6d7ae6ee76#n22

scripts/cicd/test_syslog_monitor_memory_usage/run_test.sh
alesnovak-s1: Good idea!

scalyr_agent/builtin_monitors/syslog_monitor.py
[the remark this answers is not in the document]
jmakar-s1: Ok, just wanted to be sure it was handled

jmakar-s1 (approved):

PR #1291, opened 2024-08-14
Fix GHA Workflows

alesnovak-s1:
* https://sentinelone.atlassian.net/browse/DTIN-4453 - Fix GHA workflows - Linux package building and testing
* https://sentinelone.atlassian.net/browse/DTIN-4686 - Fix GHA workflows - Unit tests
* https://sentinelone.atlassian.net/browse/DTIN-4698 - FIX GHA worfklows - Windows installer package
* https://sentinelone.atlassian.net/browse/DTIN-4697 - FIX GHA worfklows - Split Agent Build Refactored
* https://sentinelone.atlassian.net/browse/DTIN-4408 - Scalyr Agent - Fix K8s monitor tests
* https://sentinelone.atlassian.net/browse/DTIN-4704 - FIX GHA worfklows - Retry actions for Linux packages and Container Build
* https://sentinelone.atlassian.net/browse/DTIN-4685 - FIX GHA worfklows - K8s tests


.github/workflows/build-linux-packages.yml
```diff
@@ -0,0 +1,94 @@
...
+      - master
+    tags:
+      - v*.*.*
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+
+  schedule:
+    - cron: '0 4 * * *'
```
jmakar-s1, line 15: Is it necessary to do this on a schedule?

.github/workflows/build-linux-packages.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: Checkout repository
+        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4
+
+      - id: define_constants
+        run: |
+          echo "python_version=3.8" >> "${GITHUB_OUTPUT}"
+          echo "cache_version=$GITHUB_REF_NAME" >> "${GITHUB_OUTPUT}"
+          echo "aws_region=us-east-1" >> "${GITHUB_OUTPUT}"
+          echo "cicd_workflow=${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}" >> "${GITHUB_OUTPUT}"
```
jmakar-s1, line 49: I think you only need one of github.run_id or github.run_number not both, ref: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs

.github/workflows/build-linux-packages.yml
```diff
@@ -0,0 +1,94 @@
...
+    with:
+      python_version: ${{ needs.pre_job.outputs.python_version }}
+      cache_version: ${{ needs.pre_job.outputs.cache_version }}
+      aws_region: ${{ needs.pre_job.outputs.aws_region }}
+      cicd_workflow: ${{ needs.pre_job.outputs.cicd_workflow }}
+    secrets:
+      CT_AWS_DEV_EC2_PRIVATE_KEY: ${{ secrets.CT_AWS_DEV_EC2_PRIVATE_KEY }}
+      CT_AWS_DEV_EC2_PRIVATE_KEY_NAME: ${{ secrets.CT_AWS_DEV_EC2_PRIVATE_KEY_NAME }}
+      CT_AWS_DEV_EC2_ACCESS_KEY: ${{ secrets.CT_AWS_DEV_EC2_ACCESS_KEY }}
+      CT_AWS_DEV_EC2_SECRET_KEY: ${{ secrets.CT_AWS_DEV_EC2_SECRET_KEY }}
```
jmakar-s1, line 65: Can't the packages be built within a GitHub action instead of using a dedicated EC2?  Maybe this can be done as a separate ticket

.github/workflows/build-tarball.yml
```diff
@@ -0,0 +1,55 @@
...
+      - master
+    tags:
+      - v*.*.*
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+
+  schedule:
+    - cron: '0 4 * * *'
```
jmakar-s1, line 15: Does this need to be on a schedule?

.github/workflows/build-windows-package.yml
```diff
@@ -1,10 +1,18 @@
...
+      - master
+    tags:
+      - v*.*.*
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch:
+
+  schedule:
+    - cron: '0 4 * * *'
```
jmakar-s1, line 15: Needed?

.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -304,8 +303,16 @@ jobs:
...
+          - { "version": "v1.29.3", "driver": "docker", "runtime": "cri-o", "run_cronjob_checks": true }
+          - { "version": "v1.30.0", "driver": "docker", "runtime": "cri-o", "run_cronjob_checks": true }
 
     steps:
+#      - name: Setup tmate session
+#        uses: mxschmitt/action-tmate@v3
+#        with:
+#          detached: true
+#          timeout-minutes: 5
+#          limit-access-to-actor: true
```
jmakar-s1, line 315: Should be removed

Dockerfile_test_py310
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:20.04
+
+RUN apt update
+RUN ln -fs /usr/share/zoneinfo/Europe/Prague /etc/localtime
```
jmakar-s1, line 4: Should remove this, can stick to UTC (which I believe is the default)

Dockerfile_test_py310
```diff
@@ -0,0 +1,21 @@
...
+RUN apt install software-properties-common build-essential -y
+RUN add-apt-repository ppa:deadsnakes/ppa
+RUN apt update
+RUN apt install python3.10 python3.10-venv python3.10-dev curl -y
+
+RUN curl -sS https://bootstrap.pypa.io/get-pip.py | python3.10
```
jmakar-s1, line 10: There is no ubuntu pip package?  python3.10-pip?

Dockerfile_test_py310
jmakar-s1, line 1: Actually is this a test file that was committed by accident?

build_package.py
```diff
@@ -381,7 +381,15 @@ def build_win32_installer_package(variant, version):
         ]
     )
 
-    subprocess.check_call(command)
+    try:
+        subprocess.run(command, check=True, capture_output=True)
+    except subprocess.CalledProcessError as e:
+        print("Error building frozen binary: {}".format(e))
+        print("stdout: {}".format(e.output.decode("utf-8")))
```
jmakar-s1, line 388: Shouldn't stdout be printed line by line?

scripts/cicd/gha/retry_workflow.sh
jmakar-s1, line 1: Rather than do this, I think it's better to do this as a step in the workflows (untested code from stackoverflow):
```
rerun-failed-jobs:
    runs-on: ubuntu-latest
    needs: [ job1, job2]
    if: failure()
    steps:
      - name: Rerun failed jobs in the current workflow
        env:
          GH_TOKEN: ${{ github.token }}
        run: gh run rerun ${{ github.run_id }} --failed
```

jmakar-s1 (approved):
LGTM, minor comments here and there.  Only significant change would be to consider replacing retry_workflow.sh with a GHA workflow step

.github/workflows/build-linux-packages.yml
alesnovak-s1: Yes, there's a lot of dependencies and other moving parts and we need to know when it starts failing, which happens from time to time.

.github/workflows/build-linux-packages.yml
alesnovak-s1: I agree. I was splitting old workflows and copied it the way it was to be sure nothing breaks. You never know what depends on this one. In this case it's probably ok to change it, but I prefer to keep it that way and possibly fix it later.

.github/workflows/reusable-agent-build-container-images.yml
```diff
@@ -760,7 +763,13 @@ jobs:
           # 3. Install dummy java app container with jmx exporter side
           kubectl apply -f tests/e2e/k8s_om_monitor/java_app_deployment.yaml
 
-          sleep 20
+          sleep 2
```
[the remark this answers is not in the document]
alesnovak-s1: Yes, sometimes kubectl wait is executed before minikube is finished with registering the pod as a resource and then it fails.

Dockerfile_test_py310
alesnovak-s1: Good point with the UTC, I will change it. But the default zone needs to be linked before installing build-essential to avoid apt prompting.

PR #1293, opened 2024-09-09
Addressing lint issues

alesnovak-s1:
* https://sentinelone.atlassian.net/browse/DTIN-4696 - FIX GHA worfklows - Lint Tests
* typing import fix in log_processing.py
* tmate debugging step removed


scalyr_agent/log_processing.py
```diff
@@ -27,6 +27,9 @@
 
 import sys
 
+if False:
+    from typing import Union, Optional
+
```
jmakar-s1, line 32: Is this needed?

jmakar-s1 (approved):
LGTM

scalyr_agent/log_processing.py
alesnovak-s1: Yes, lint complains about this import missing.

PR #1302, opened 2024-10-16
Fix - Race condition when trying to add a log file that's been scheduled for removal

alesnovak-s1:
* Fixes https://sentinelone.atlassian.net/browse/DTIN-4631
* __lock has been change to reentrant RLock. 

Based on the comments the lock was initially meant to lock status fields used in generate_status.
Now it ensures synchronized access to fields needed for asynchonous workflow of log files. 
This is already causing a timeout when retrieving status from the agent (https://sentinelone.atlassian.net/browse/DTIN-4769) and will be handled in another PR.
	

**The Bug:**

When a log file (path, worker_id) is scheduled for removal, the removal is done in two separate steps:
*  **__remove_logs_scheduled_for_deletion**
	* Instructs all matchers from **__logs_pending_removal** to finish
	* Clears **__logs_pending_removal**
* **__purge_finished_log_matchers**
	* Removes all finished matchers from **__logs_pending_removal**, **__dynamic_matchers**, **__dynamic_paths**
	
The race condition can be reached when a path is scheduled for removal and the matchers are still finishing.
When **add_log_config** is called it considers log path to be active and skips adding it. The log file is later removed.

In log we can see sequence like this one:
```
Log ( log_path='xy', worker_id='default' ) for monitor 'scalyr_agent.builtin_monitors.kubernetes_monitor' is pending removal

Tried to add new log file 'xy' for monitor 'scalyr_agent.builtin_monitors.kubernetes_monitor', but it is already being monitored by 'scalyr_agent.builtin_monitors.kubernetes_monitor'

Removing log file ( log_path='xy', worker_id='default' ) for 'scheduled-deletion'
```

### DIFF
--- a/scalyr_agent/copying_manager/copying_manager.py
+++ b/scalyr_agent/copying_manager/copying_manager.py
@@ -27,7 +27,6 @@ import signal
 import errno
 import fnmatch
 import json
-from collections import defaultdict
 
 WORKER_SESSION_PROCESS_MONITOR_ID_PREFIX = "agent_worker_"
 
--- a/scalyr_agent/copying_manager/copying_manager.py
+++ b/scalyr_agent/copying_manager/copying_manager.py
@@ -392,6 +392,10 @@ class PathWorkerIdDict(object):
 
     def get(self, path, worker_id, default=None):
         # type: (six.text_type, six.text_type, Optional[object]) -> Optional[object]
+
+        if path not in self.__paths:
+            return default
+
         return self.__paths[path].get(worker_id, default)
 
     def complement_keys(self, path, worker_ids):
@@ -435,7 +439,13 @@ class PathWorkerIdDict(object):
 
     def pop(self, path, worker_id, default=None):
         # type: (six.text_type, six.text_type, Optional[object]) -> Optional[object]
-        return self.__paths[path].pop(worker_id, default)
+
+        value = self.__paths[path].pop(worker_id, default)
+
+        if not self.__paths[path]:
+            self.__paths.pop(path)
+
+        return value
 
     def __len__(self):
         return len(list(self.keys()))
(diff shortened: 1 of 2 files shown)


scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -388,10 +388,20 @@ def set(self, path, worker_id, value):
         self.__paths[path][worker_id] = value
 
     def get_path(self, path):
+
+        # We don't want to be creating an empty default dict by get.
+        if path not in self.__paths:
+            return {}.items()
+
         return self.__paths[path].items()
```
jmakar-s1, line 396: Up to you but I think this is ok: `return self.__paths.get(path, {}).items()`

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -435,7 +445,13 @@ def keys(self):
 
     def pop(self, path, worker_id, default=None):
         # type: (six.text_type, six.text_type, Optional[object]) -> Optional[object]
-        return self.__paths[path].pop(worker_id, default)
+
+        value = self.__paths[path].pop(worker_id, default)
```
jmakar-s1, line 449: Don't we need to be careful that path exists in self.__paths?
Maybe better to be `value = self.__paths.get(path, {}).pop(worker_id, default)` or something similar

scalyr_agent/copying_manager/copying_manager.py
```diff
@@ -435,7 +445,13 @@ def keys(self):
...
+
+        if not self.__paths[path]:
```
jmakar-s1, line 451: Similarly this could be `if path in self.__paths and not self.__paths[path]:`

jmakar-s1 (approved):
Looks good.  One consideration for future work, to make the use of the lock clearer could use `with self.__lock:` to avoid having to explicitly release it

scalyr_agent/copying_manager/copying_manager.py
alesnovak-s1: Yes, definitely!

scalyr_agent/copying_manager/copying_manager.py
alesnovak-s1: self.__path is a defaultdict. But this way it's not consistent and a bit confusing. I will ditch the defaultdict, use this way and modify a set method as well so it's obvious what's expected to happen.

scalyr_agent/copying_manager/copying_manager.py
alesnovak-s1: It will work because of defaultdict, but it's confusing, I will change it. Thanks!

PR #1364, opened 2026-04-08
Small fixes for the build process

stevenc-s1:
- Change how we determine if WIX is installed on the windows builder due to changes on the Wix side
- Use different mirror for gnu tools. The current one is having availability issues.

### DIFF
--- a/build_package.py
+++ b/build_package.py
@@ -206,7 +206,7 @@ def build_win32_installer_package(variant, version):
 
     @return: The file name of the built package.
     """
-    if os.getenv("WIX") is None:
+    if not shutil.which('candle') or not shutil.which('light'):
         print(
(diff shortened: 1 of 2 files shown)


build_package.py
```diff
@@ -206,7 +206,7 @@ def build_win32_installer_package(variant, version):
 
     @return: The file name of the built package.
     """
-    if os.getenv("WIX") is None:
+    if not shutil.which('candle') or not shutil.which('light'):
```
tristanmccarty-s1, line 210: (nit) A comment noting why 'candle' and 'light' should be present to indicate WIX is installed could be helpful here

tristanmccarty-s1 (approved):

PR #1397, opened 2026-09-17
feat: [DTIN-7750] Add Auth token to Headers, IGW route option

tristanmccarty-s1:
This PR adds the Auth token to Headers and a new configuration to support aiming the agent at newer ingestion server instead. This is the smaller part of support for migration towards a new ingestion target.


scalyr_agent/config_main.py
```diff
@@ -670,6 +670,7 @@ def upgrade_windows_install(
             )
 
         # Determine if a newer version is available
+        # NOTE: intentionally config.scalyr_server to hit the legacy version check API
```
stevenc-s1: Sorry, I'm not quite sure what you are trying to say here. Can you restate it?

stevenc-s1 (approved):

